PR #3, opened 2020-08-11
Create the CODEOWNERS file

pjanotti:


.github/CODEOWNERS
```diff
@@ -0,0 +1,15 @@
...
+#
+# Learn about membership in OpenTelemetry community:
+#  https://github.com/open-telemetry/community/blob/master/community-membership.md
+# 
+#
+# Learn about CODEOWNERS file format: 
+#  https://help.github.com/en/articles/about-code-owners
+#
+
+* @macrogreg @MikeGoldsmith @pjanotti
```
cijothomas, line 15: https://github.com/open-telemetry/opentelemetry-dotnet/blob/master/.github/CODEOWNERS do you WANT to list names or just do what is done here?

pjanotti: Good point. We will have a few folks that are not approvers on .NET. I will add the team and other folks manually.

.github/CODEOWNERS
```diff
@@ -0,0 +1,15 @@
...
+#
+# Learn about membership in OpenTelemetry community:
+#  https://github.com/open-telemetry/community/blob/master/community-membership.md
+# 
+#
+# Learn about CODEOWNERS file format: 
+#  https://help.github.com/en/articles/about-code-owners
+#
+
+* @open-telemetry/dotnet-approvers @macrogreg
```
macrogreg, line 15: What is the reasoning for making all of the .NET SDK approves here?
I think some of the folks do not have the right context.

cijothomas: I think this must be a temporary thing until separate Github group is created. It should be a separate group from sdk repo with one or two folks in each other group.

pjanotti: Yeah, my idea was to have a broader list on CODEOWNERS and let the rights on the repo control who could merge PRs. However, per doc [here]():

pjanotti: My ignorance in this regard: the teams were already created, see https://github.com/orgs/open-telemetry/teams?query=dotnet-instrumentation

Only maintainers have the right to merge PRs: at this moment that will be me, @macrogreg and @cijothomas.

pjanotti:
Updated per last SIG meeting conversation. After this is merged I will update the GH teams.

cc @nrcventura, @macrogreg, @lucaspimentel @Mikhail-msft, @colin-higgins, @kevingosse, @MikeGoldsmith, @tonyredondo, @zacharycmontoya

PR #5, opened 2020-08-12
Doc on branch naming conventions and remotes

macrogreg:


BRANCHES AND REMOTES.md
```diff
@@ -0,0 +1,156 @@
+﻿# Branch name guidance
```
cijothomas, line 1: Do we need any special guidance other than what is followed by Otel in general.
Eg: https://github.com/open-telemetry/opentelemetry-dotnet/blob/master/CONTRIBUTING.md#pull-requests
https://github.com/open-telemetry/opentelemetry-python/blob/master/CONTRIBUTING.md

macrogreg: @cijothomas Yes, I think so for now and here is why (if there is a better way to achieve this, then I'd love to learn):

People who fork from DD before have potentially a full change history and them may want to preserve it. Same with DD itself. However, as per OTel guidance, the initial commit is a squash. (Otherwise, every single person who ever contributed to DD in the past years would need to sign CLA).

So people cannot create a new Repo that is a fork of OTel. I came up with a way to merge code back and forth if the Repo already exists as a fork of DD, but going forward it wants to take merges from OTel instead, without loosing the existing history. To save others fro needing to rediscover this, I documented and created this PR.

If there is a better way to do this, I'd love to learn. If there is not, I think that people would benefit from this documentation.

As a side point, this method creates the need to have branches in the OTel repo (rather than just forked repos). So I set up a naming contention described in this PR.

macrogreg: That sounds good to me. We can modify the doc accordingly. Just felt this was useful info. I'll ping you to discuss briefly. :)

pjanotti:
@macrogreg per the comments above and other discussions let's close this one and I will start a new one that we can add any specific instructions that are not covered on the standard OTel repo management.

cc @open-telemetry/dotnet-instrumentation-approvers

pjanotti:
I'm going to close this one per the reasons stated above.

PR #7, opened 2020-08-20
Clean up auxiliary Datadog files, leave all Tracer sources and tests in-tact.

macrogreg:


macrogreg:
Closing. Will redo cleaner.

PR #9, opened 2020-08-20
Forking DataDog Tracer to seed OTel Auto-Instrumentation-Agent.

macrogreg:
This is a fork of the DataDog .NET Tracer (aka "in-process agent", aka "auto-instrumentation agent").   
It acts as the initial seed for the official Open Telemetry .NET Tracer.

We are doing this by merging the current master of _https://github.com/DataDog/dd-trace-dotnet_ into an empty master of _https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation_.   
In addition, I cleaned up all auxiliary files not directly related to the product (e.g., demos, tools, vendor-specific integration tests and benchmarks, we well as others). We can add them in later from the original repo if the community requires.   
   
Everything builds, all unit tests are passing.


pjanotti:
@macrogreg it was intentional that `samples-lib` is still present?

macrogreg:
> it was intentional that samples-lib is still present?

I was not sure if it was needed and did not invest into investigating :)

nrcventura:
I know that this is just the initial PR and we're not ready to change many of the "branding" pieces yet. But there are some other things in here that I don't know if we should change sooner rather than later (or even change at all). Things like the Code of Conduct and strong naming key.

pjanotti:
agreed @nrcventura we should change some of those soon: I don't expect those to conflict much even for people doing cherry-picking between different repos and it is easy: choose to keep the one on the repo.

PR #18, opened 2020-09-24
First prototype for the DynamicActivityBinding library.

macrogreg:
**This PR must _not_ be squash-merged!**

This is the initial prototype of the `OpenTelemetry.DynamicActivityBinding` library.
The library provides a shim around APIs in the System.Diagnostics.DiagnosticSource assembly (`DS`) that allows to use such APIs without statically referencing `DS`. This is done via dynamic assembly loading (see https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/13), Reflection and compiled Expressions.

This early prototype includes:
* An early of the dynamic assembly loading logic.
* A copy of `DS` taken from the dotnet runtime repo, specifically https://github.com/dotnet/runtime/tree/master/src/libraries/System.Diagnostics.DiagnosticSource (as of today). All `DS`-files were copies including their respective Git history. (**To preserve that history, this PR must _not_ be squash-merged!**) `DS` is currently excluded from the build.

Issue #13, opened 2020-08-31
Full support for .NET Activities & deprecate proprietary Datadog.Trace library 


macrogreg:
Today we use a proprietary library (Datadog.Trace) to start / stop spans and to perform related tasks. This library is used for both, our auto-instrumentation, as well as for customer-created spans (where required).

.NET has an API to represent spans (Activity). It is a part of all .NET Core versions that we support, and is distributed as a NuGet for the .NET Framework. We are going to transition towards using the .NET API instead of Datadog’s proprietary library. 

Activity is also the underlying data exchange mechanism for OTel tracing. Using it will ensure that out auto-instrumentation can be mixed and matched with both, OTel API based telemetry, and vendor-agnostic .NET API based telemetry.

This document outlines the plan and discusses the risks and benefits.

# Integrations that will be enabled once we start using activities

It would be nice to compile a list of such libraries for reference purposes. Here is a preliminary / incomplete list.
@cijothomas , @lmolkova , all - did you happen to have a pointer to a more complete list?

**_Update_: Created https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/21 specifically to list/discuss libraries/components that support Activity or DS.**

~~A number of client libraries will "light up" without the need to develop integrations. This includes all libraries that use OTel APIs and/or .NET Activities.~~

- ~~ASP.NET WebForms		(4.5+)~~
- ~~ASP.NET MVC			(4+)~~
- ~~ASP.NET WebAPI		(4.5+)~~
- ~~ASP.NET Core			(1.1+)~~
- ~~HttpClient			(4.5+), and .NET Core (1.1+)~~
- ~~HttpWebRequest~~
- ~~SqlClient .NET Core		(1.0+), NuGet 4.3.0~~
- ~~Microsoft.Data.SqlClient		1.1.0)~~
- ~~Azure EventHubs Client SDK	(1.1.0)~~
- ~~Azure ServiceBus Client SDK	(3.0.0)~~

# Engineering plan

**Phase I** (_completed_):

1. Proof of concept: https://github.com/DataDog/dd-trace-dotnet/tree/gregp/hack-a-dog-202007-diagnosticsource-activity/src/Datadog.Trace.ClrProfiler.Managed/ActivityBasedTracing 

**Phase II**:

2. Solve the System.Diagnostics.DiagnosticSource.dll versioning problem.
[...]

PR #20, opened 2020-10-09
Add ActivityStub skeleton

pjanotti:
Bringing the skeleton for the `ActivityStub` described on issue #13, see also https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/18#issuecomment-705157637.

This first PR is just the skeleton for the stub and as such doesn't bring yet dynamic loader and related tests. The API surface is not 100% match with S.D.DS.Activity but gets us ready to start implementing the actual assembly loading. Extending the API to the desired surface is relatively mechanical after the dynamic loading of S.D.DS or the fallback to its "vendored" sources is done.

Some notes:

1. I went through the trouble of making it compatible with the current StyleCop settings for the whole repo, this may not be desirable when we bring the "vendored" sources from dotnet/runtime but for now it keeps the sources consistent.
2. For not the types for `TraceId` and `SpanId` are replaced with `string`: we likely need a wrapper for these types.
3. I didn't bring the methods that match methods from `ActivitySource` my initial thought is that we should also have a separate stub for `ActivitySource` instead of having all of those under `ActivityStub`.
4. For now supplemental data handling was not included either.

Missing `Activity` equivalent API surface:

| Property | S.D.DS.Activity version introduction |
| --------- | ---------------------------------------- |
| ActivityContext | 5.0.0.0 |
| DisplayName | 5.0.0.0 |
| Events | 5.0.0.0 |
| IsAllDataRequested | 5.0.0.0 |
| Links | 5.0.0.0 |
| ParentId | 4.0.2.1 |
| Recorced | 4.0.4.0 |
| Source | 5.0.0.0 |
| StartTimeUtc | 4.0.2.1 |
| TraceStateString | 4.0.4.0 |

| Method | S.D.DS.Activity version introduction |
| --------- | ---------------------------------------- |
| AddEvent | 5.0.0.0 |
| SetEndTime | 4.0.2.1 |
| SetIdFormat | 4.0.4.0 |
| SetParentId (overload w/ 3 parameters) | 4.0.4.0 |
| SetStartTime | 4.0.2.1 |

### DIFF
--- a/src/OpenTelemetry.DynamicActivityBinding/public/ActivityStub.cs
+++ b/src/OpenTelemetry.DynamicActivityBinding/public/ActivityStub.cs
@@ -43,7 +43,7 @@ namespace OpenTelemetry.DynamicActivityBinding
                 {
                     // TODO: is the no-op stub enough to represent "no current activity"? no need to differentiate
                     // "no current activity" from "no-op activity"?
-                    return NoOpSingeltons.ActivityStub;
+                    return NoOpSingletons.ActivityStub;
                 }
 
                 throw new NotImplementedException(MissingCapability);
@@ -92,7 +92,7 @@ namespace OpenTelemetry.DynamicActivityBinding
             {
                 if (ActivityInstance == null)
                 {
-                    return NoOpSingeltons.KvpEnumerable;
+                    return NoOpSingletons.KvpEnumerable;
                 }
 
                 throw new NotImplementedException(MissingCapability);
@@ -112,7 +112,7 @@ namespace OpenTelemetry.DynamicActivityBinding
             {
                 if (ActivityInstance == null)
                 {
-                    return NoOpSingeltons.TimeSpan;
+                    return NoOpSingletons.TimeSpan;
                 }
 
                 throw new NotImplementedException(MissingCapability);
@@ -174,7 +174,7 @@ namespace OpenTelemetry.DynamicActivityBinding
             {
                 if (ActivityInstance == null)
                 {
-                    return NoOpSingeltons.ActivityIdFormat;
+                    return NoOpSingletons.ActivityIdFormat;
                 }
 
                 throw new NotImplementedException(MissingCapability);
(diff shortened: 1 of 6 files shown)


src/OpenTelemetry.DynamicActivityBinding/public/ActivityStub.cs
```diff
@@ -0,0 +1,580 @@
...
+        /// Provides functionality of <c>System.Diagnostics.DiagnosticSource.Activity.Current</c>.
+        /// Introduction: System.Diagnostics.DiagnosticSource, Version=4.0.2.1
+        /// </remarks>
+        public static ActivityStub Current
+        {
+            get
+            {
+                if (!DynamicLoader.EnsureInitialized())
+                {
+                    // TODO: is the no-op stub enough to represent "no current activity"? no need to differentiate
```
nrcventura, line 44: Is this `TODO` something we want to start discussing in this PR, or is it a reminder for a future discussion?

src/OpenTelemetry.DynamicActivityBinding/public/ActivityStub.cs
```diff
@@ -0,0 +1,580 @@
...
+        {
+            if (ActivityInstance == null)
+            {
+                return;
+            }
+
+            throw new NotImplementedException(MissingCapability);
+        }
+
+        private static class NoOpSingeltons
```
nrcventura, line 570: Minor typo. It should probably be `NoOpSingletons`.

src/OpenTelemetry.DynamicActivityBinding/public/ActivityStub.cs
pjanotti: I would like to follow it up as a separate discussion.

nrcventura (approved):

pjanotti:
Trying to close and re-open to fix the CLA...

PR #22, opened 2020-10-20
Catchup to upstream prior to directory reorg

pjanotti:
Catches up to upstream prior to directp\ory reorg.

PR #25, opened 2020-10-28
Catchup to upstream prior to directory reorg

pjanotti:
Same as #22 to check EasyCLA


pjanotti:
EasyCLA gets in some state that it stays waiting forever: this seems related to the way that I pulled the changes...

PR #32, opened 2020-12-08
Add Zipkin as an option for exporting traces

pjanotti:
Adds a simple Zipkin exporter as a non-default option to export traces. 

This is a very simple implementation that mirrors code already existent on the repo, eg.: `Api.cs`. The goal is to be able to have a somewhat standard destination that we can start to use to experiment with changes.

In order to use it you have to set the following env vars.:

- OTEL_EXPORTER: set it to `zipkin` the default is the DD agent (`DatadogAgent`).
- DD_AGENT_TRACE_URL: set it to the proper Zipkin endpoint including the path: eg.: `http://localhost:9411/api/v2/spans`

These sets above are enough to get traces to a Zipkin docker running locally, eg.: `docker run -it --rm -p 9411:9411 openzipkin/zipkin`

### DIFF
--- a/src/Datadog.Trace/Agent/ZipkinApi.cs
+++ b/src/Datadog.Trace/Agent/ZipkinApi.cs
@@ -23,7 +23,7 @@ namespace Datadog.Trace.Agent
 
         public void SetBaseEndpoint(Uri baseEndpoint)
         {
-            // TODO: This doesn't make sense for this exporter.
+            // TODO: Remove this method when it is also removed from the interface.
             _tracesEndpoint = new Uri(baseEndpoint, _tracesEndpoint.PathAndQuery);
         }
 
@@ -47,11 +47,6 @@ namespace Datadog.Trace.Agent
                 request.Method = "POST";
                 request.ContentType = "application/json";
 
-                foreach (var header in _settings.HeaderTags)
-                {
-                    request.Headers.Add(header.Key, header.Value);
-                }
-
                 using (var requestStream = await request.GetRequestStreamAsync().ConfigureAwait(false))
                 {
                     var serializer = new ZipkinSerializer();
@@ -65,7 +60,10 @@ namespace Datadog.Trace.Agent
                     using var httpWebResponse = (HttpWebResponse)await request.GetResponseAsync().ConfigureAwait(false);
 
                     // Zipkin specifies only "Accepted" as valid response, the code is more tolerant here.
-                    if (httpWebResponse.StatusCode >= HttpStatusCode.OK && httpWebResponse.StatusCode <= HttpStatusCode.Accepted)
+                    // Following a criteria equivalent to HttpResponseMessage.EnsureSuccessStatusCode as
+                    // done by the OpenTelemetry .NET SDK for their Zipkin exporter:
+                    // See https://github.com/open-telemetry/opentelemetry-dotnet/blob/8cda9ef394a1b075fd156d73dace48e48f5b3c9b/src/OpenTelemetry.Exporter.Zipkin/ZipkinExporter.cs#L86
+                    if (httpWebResponse.StatusCode >= HttpStatusCode.OK && httpWebResponse.StatusCode < HttpStatusCode.MultipleChoices)
                     {
                         return true;
                     }
--- a/src/Datadog.Trace/Agent/ZipkinApi.cs
+++ b/src/Datadog.Trace/Agent/ZipkinApi.cs
@@ -1,0 +1,103 @@
+using System;
+using System.Net;
+using System.Threading.Tasks;
+using Datadog.Trace.Configuration;
+using Datadog.Trace.Logging;
+
+namespace Datadog.Trace.Agent
+{
+    internal class ZipkinApi : IApi
+    {
+        private static readonly Vendors.Serilog.ILogger Log = DatadogLogging.For<ZipkinApi>();
+
+        private readonly TracerSettings _settings;
+        private Uri _tracesEndpoint;
+
+        public ZipkinApi(TracerSettings settings)
+        {
+            Log.Debug("Creating new Zipkin Api");
+
+            _settings = settings ?? throw new ArgumentNullException(nameof(settings));
+            _tracesEndpoint = _settings.AgentUri; // User needs to include the proper path.
+        }
+
+        public void SetBaseEndpoint(Uri baseEndpoint)
+        {
+            // TODO: This doesn't make sense for this exporter.
+            _tracesEndpoint = new Uri(baseEndpoint, _tracesEndpoint.PathAndQuery);
+        }
+
+        public async Task<bool> SendTracesAsync(Span[][] traces)
+        {
+            if (traces == null || traces.Length == 0)
+            {
+                // Nothing to send, no ping for Zipkin.
+                return true;
+            }
+
+            // retry up to 5 times with exponential back-off
+            var retryLimit = 5;
+            var retryCount = 1;
+            var sleepDuration = 100; // in milliseconds
+
+            while (true)
+            {
+                // TODO: Initially same code for Fx and Core.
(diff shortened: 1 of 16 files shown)


src/Datadog.Trace/Agent/ZipkinApi.cs
```diff
@@ -0,0 +1,103 @@
...
+        private readonly TracerSettings _settings;
+        private Uri _tracesEndpoint;
+
+        public ZipkinApi(TracerSettings settings)
+        {
+            Log.Debug("Creating new Zipkin Api");
+
+            _settings = settings ?? throw new ArgumentNullException(nameof(settings));
+            _tracesEndpoint = _settings.AgentUri; // User needs to include the proper path.
+        }
```
pjanotti, line 22: The `Api` implementation has some logging here about the framework it seems that we should perhaps move it to some higher level so it gets logged independently of the exporter.

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -211,6 +211,14 @@ public static class ConfigurationKeys
         /// </summary>
         public const string DiagnosticSourceEnabled = "DD_DIAGNOSTIC_SOURCE_ENABLED";
 
+        /// <summary>
+        /// Configuration key for the exporter to be used the Tracer uses it to encode and
+        /// dispatch traces.
+        /// Default is <c>"DatadogAgent"</c>.
+        /// </summary>
+        /// <seealso cref="TracerSettings.Exporter"/>
+        public const string Exporter = "OTEL_EXPORTER";
```
pjanotti, line 220: Using `OTEL_` prefix to separate the configurations.

src/Datadog.Trace/Tagging/ITags.cs
```diff
@@ -1,7 +1,13 @@
+using System.Collections.Generic;
+
 namespace Datadog.Trace.Tagging
 {
     internal interface ITags
     {
+        List<KeyValuePair<string, double>> Metrics { get; }
+
+        List<KeyValuePair<string, string>> Tags { get; }
```
pjanotti, line 9: I opted to add this and simply expose this member from the abstract class but perhaps we can refine this, right now the naming is a bit awkward: `span.Tags.Tags`. That said I think that is acceptable for this change.

src/Datadog.Trace/Tracer.cs
```diff
@@ -118,7 +118,26 @@ internal Tracer(TracerSettings settings, IAgentWriter agentWriter, ISampler samp
...
+            {
+                _agentWriter = agentWriter;
+            }
+            else
+            {
+                IApi api = null;
+                switch (Settings.Exporter)
+                {
+                    case ExporterType.Zipkin:
+                        api = new ZipkinApi(Settings);
```
pjanotti, line 131: I'm passing the settings to have access to the agent URL: if there is an agreement I think this should be changed to the sources so each exporter (or other components) can do their own configuration reading.

src/Datadog.Trace/Tagging/ITags.cs
nrcventura: Would it read better if it was something like `span.Tags.All` or `span.Tags.AllTags`?

src/Datadog.Trace/Tracer.cs
nrcventura: Your suggestion makes sense, and matches the pattern used within the OTel SDK.

elucus:
@pjanotti thanks for getting this going

src/Datadog.Trace/Agent/ZipkinApi.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        public void SetBaseEndpoint(Uri baseEndpoint)
+        {
+            // TODO: This doesn't make sense for this exporter.
```
zacharycmontoya, line 26: This is slated to be removed in the datadog branch, so a TODO to remove this is sufficient

src/Datadog.Trace/Agent/ZipkinApi.cs
```diff
@@ -0,0 +1,103 @@
...
+            var sleepDuration = 100; // in milliseconds
+
+            while (true)
+            {
+                // TODO: Initially same code for Fx and Core.
+                var request = WebRequest.CreateHttp(_tracesEndpoint);
+                request.Method = "POST";
+                request.ContentType = "application/json";
+
+                foreach (var header in _settings.HeaderTags)
```
zacharycmontoya, line 50: I'm surprised to see this here and I think this is not meant to be used here. We use `_settings.HeaderTags` only to decorate web server spans based on the request headers. We don't send these headers to our agent

src/Datadog.Trace/Tracer.cs
zacharycmontoya: Yeah that's a good suggestion 👍🏼

src/Datadog.Trace/Tagging/ITags.cs
zacharycmontoya: For more context, we created this ITags interface and classes to optimize the common case performance where no modifications are made by the user to automatic instrumentation tags. That's also why the this `ITags` interface unintuitively has a member for both tags and metrics. This change is definitely good, though the naming is still awkward because of the original naming choices

src/Datadog.Trace/Agent/ZipkinSerializer.cs
```diff
@@ -0,0 +1,129 @@
...
+                    return new Dictionary<string, string>() { { "serviceName", actualServiceName } };
+                }
+            }
+
+            public IDictionary<string, string> Tags
+            {
+                get => _tags;
+            }
+
+            private static IDictionary<string, string> BuildTags(Span span, TracerSettings settings)
```
zacharycmontoya, line 108: nit: Do you still need the settings parameter?

zacharycmontoya (approved):
Overall LGTM

src/Datadog.Trace/Agent/ZipkinApi.cs
pjanotti: Thanks for noticing that. I completely confused this with some other exporter (OTel collector?) that has a similarly named configuration for the function above. I will remove it.

src/Datadog.Trace/Agent/ZipkinApi.cs
```diff
@@ -0,0 +1,103 @@
...
+                }
+
+                Exception requestException = null;
+                HttpStatusCode requestStatusCode = 0;
+                try
+                {
+                    using var httpWebResponse = (HttpWebResponse)await request.GetResponseAsync().ConfigureAwait(false);
+
+                    // Zipkin specifies only "Accepted" as valid response, the code is more tolerant here.
+                    if (httpWebResponse.StatusCode >= HttpStatusCode.OK && httpWebResponse.StatusCode <= HttpStatusCode.Accepted)
```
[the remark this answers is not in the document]
pjanotti: It seems reasonable to follow the SDK behavior, no need to diverge at this moment.

src/Datadog.Trace/Tracer.cs
pjanotti: I think we have an agreement here. Let's tackle that in a separate review.

pjanotti:
Follow-ups to track from the reviews:

- Move what looks like "initialization logging" from Api implementation to a location that can be shared by all exporters. See https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/32/files#r538752573
- Remove SetBaseEndpoint when the it is also removed from DD upstream.  Seehttps://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/32/files#r541277550
- Rename ITags.Tags per conversation https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/32/files#r538754494

PR #41, opened 2021-01-28
WIP: Disable CI jobs that require Arm64 pool

pjanotti:
Azure Pipelines CI is not running due to a pool that exists for upstream but not on this repo, trying to disable the jobs using this pool.


pjanotti:
It doesn't work: Az Pipelines still check for the resource. I will close this PR and later will get back to this.

PR #45, opened 2021-02-11
Removing Arm64 pool dependent jobs from CI

pjanotti:
Per conversation on SIG trying to restore CI by removing jobs that depend on Arm64 pool.


pjanotti:
I'm getting errors like `There is not enough space on the disk.` from CI before diving further into those I would like to merge this change and catch up to the latest from upstream then clean-up any remaining failures.

pjanotti:
Related: previous attempt to disable jobs using the Arm64 pool https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/41

tonyredondo (approved):

nrcventura (approved):

zacharycmontoya:
> I'm getting errors like `There is not enough space on the disk.` from CI before diving further into those I would like to merge this change and catch up to the latest from upstream then clean-up any remaining failures.

With the number of integration tests we've added and the fact that we *copied* the profiler assets into each application output, we finally got to the point where this happens reliably. This was addressed in a recent PR https://github.com/DataDog/dd-trace-dotnet/pull/1171

zacharycmontoya (approved):

PR #48, opened 2021-02-12
Remove dd_agent from Azure Pipelines

### LINKED ISSUE #44: CI is not running due to lack of required pool
CI configuration requires a pool, `Arm64`, that doesn't exist for the OTel repo. Until this time I wasn't able to locate someone with rights to try to create a similar pool for OTel. In lack of that, we may have to disable tests using this pool for the time being.

pellared:
## Why

Fixes #44

## What

- Remove usage of Datadog Agent to make the CI passing.
- There was a problem with building one project. The project is ignored in `unit-tests.yml`. Created #49 and #52 to fix it.

## Other Findings

I believe only changes in `unit-test.yml` do something at the moment. The `benchmark.yml` and `runner.yml` are most probably not configured at all. Not sure if we should enable these pipelines as well.

## Notes for Reviewers

As I am a newcomer, please assume I know nothing.

1. Is there something more that should be removed?

2. Is there something similar to Datadog Agent in OpenTelemetry that we could use in the CI pipeline to have some kind of monitoring if everything is stable?

I have not maintained the branch in the way described under https://github.com/kubernetes/community/blob/master/contributors/guide/github-workflow.md . I intend to "Squash merge" this PR. If you see a problem with it I can always create a new PR with a cleaner history/commits.


.azure-pipelines/benchmarks.yml
```diff
@@ -7,20 +7,8 @@ trigger:
 variables:
   buildConfiguration: release
   dotnetCoreSdkVersion: 5.0.100
-  ddApiKey: $(DD_API_KEY)
```
pjanotti, line 10: Is it possible to just make it conditional on `DD_API_KEY` not being empty? My attempt to conditionally disable the `Arm64` pool failed but want to be sure that it also doesn't work here.

.azure-pipelines/unit-tests.yml
```diff
@@ -87,6 +70,7 @@ jobs:
         test/**/*.Tests.csproj
         test/benchmarks/**/*.csproj
         !src/Datadog.Trace.Tools.Runner/*.csproj
+        !test/Vendored.System.Diagnostics.DiagnosticSource.Tests/Vendored.System.Diagnostics.DiagnosticSource.Tests.csproj
```
pjanotti, line 73: This is a left-over from when I was pushing this stuff to the `main` branch. Now I changed my mind and think that it will be better to remove this project from `main` and continue the work related to it on the branch that @macrogreg  worked on at the end of last year.

@macrogreg If you agree let's remove the vendored S.D.DS from `main`.

@pellared please keep this as it is on this PR and wait on an agreement if we are removing this from `main`.

.azure-pipelines/benchmarks.yml
pellared: What would be the benefit such approach? I see value of doing it if we would like to enable it (provide the API key variable value). Otherwise it feels for me like unnecessary waste for anyone reading the code.

Of course I can do that on Monday if there are any other reasons or you insist.

Moreover, take into consideration that merging it even now will at least make unit tests green in Azure DevOps and would give more confidence in other PRs ;)

pjanotti: By using a conditional the files can be identical between both repos. In general, keeping the diff smaller between the repos is a benefit, a modest one, in this specific case, but worth keeping in mind.

I don't think it is a must but desirable. I wouldn't block this change on this specific issue.

pellared: I can give it a try with conditional and make a second PR on Monday 🙂

pellared: I have not found any possibility to conditionally use a container/service https://docs.microsoft.com/en-us/azure/devops/pipelines/process/resources?view=azure-devops&tabs=schema#resources-containers 

I tried to disable the `HEALTHCHECK` of the dd_agent but it is also not possible:
- https://docs.microsoft.com/en-us/azure/devops/pipelines/process/service-containers?view=azure-devops&tabs=yaml#healthcheck
- https://developercommunity.visualstudio.com/content/problem/1164638/cant-change-service-container-healthcheck-timings.html

Here is a PR that disables Datadog Agent with minimal changes: TBD

.azure-pipelines/unit-tests.yml
pellared: @pjanotti @macrogreg 
PR which removes the vedored stuff: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/52

pellared:
I have split this PR into 3 more self-contained ones:
- Remove Vendored.System.Diagnostics.DiagnosticSource.Tests #52
- Disable Datadog Agent in Azure Pipelines #51 
- Update .NET version #53

.azure-pipelines/benchmarks.yml
pjanotti: Thanks for trying @pellared. Unfortunately, it seems that the options in Az Pipelines don't allow an easy disable for this. Let's go with your current solution.

pjanotti (approved):

PR #50, opened 2021-02-12
Fix tracer solution on macos

dszmigielski:
There is a problem when opening Datadog.Traces solution in Rider on macOS - if the project targets dotnet framework Rider goes into infinite project synchronisation loop (https://youtrack.jetbrains.com/issue/RIDER-45888). It is said to be resolved on the new Rider version, but for me it still doesn't work. I have added conditional TargetFramework for the projects that had this problem, which solves the issue on mac.


pjanotti (approved):

nrcventura (commented):
These projects look like they are mostly build tooling, sample apps, or test apps. Some of these may be tied to a specific framework version for a reason. 
- Do we think that changing the target framework for these projects will still cause these projects to solve the same "problem" that they were designed for?
- Should we just expect that development will require a Windows development environment?

pellared:
> These projects look like they are mostly build tooling, sample apps, or test apps. Some of these may be tied to a specific framework version for a reason.
> 
> * Do we think that changing the target framework for these projects will still cause these projects to solve the same "problem" that they were designed for?
> * Should we just expect that development will require a Windows development environment?

Imagine will have on

> These projects look like they are mostly build tooling, sample apps, or test apps. Some of these may be tied to a specific framework version for a reason.
> 
> * Do we think that changing the target framework for these projects will still cause these projects to solve the same "problem" that they were designed for?
> * Should we just expect that development will require a Windows development environment?

I think we should have a possibility to open a solution in both MacOS and Linux. Why?

1. Possibility to debug OS specifig bug
1. Being more inclusive. Easier contribution e.g. if someone wants to make a bugfix
1. Possibility to use Development Container and maybe even GitHub Codespaces in future.

Are you aware of any projects that are only .NET Framework compatible? We may play with conditionals in sln file to exclude them https://stackoverflow.com/questions/58258842/how-to-specify-build-configuration-for-referenced-projects-when-using-dotnet-bu

pellared:
Check https://github.com/dszmigielski/opentelemetry-dotnet-instrumentation/pull/1/files

pellared (approved):

PR #51, opened 2021-02-15
Disable Datadog Agent in Azure Pipelines

### LINKED ISSUE #44: CI is not running due to lack of required pool

pellared:
## Why

Fixes #44
Done according to comments from https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/48.

## What

- Remove usage of Datadog Agent to make the CI passing.

## Other Findings

Thr unit tests should start passing when https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/52 is merged as well.

I believe only changes in `unit-test.yml` do something at the moment. The `benchmark.yml` and `runner.yml` are most probably not configured at all. Not sure if we should enable these pipelines as well.

### DIFF
--- a/.azure-pipelines/benchmarks.yml
+++ b/.azure-pipelines/benchmarks.yml
@@ -33,9 +33,10 @@ stages:
     workspace:
       clean: all
 
-    # Enable the Datadog Agent service for this job
-    services:
-      # dd_agent: dd_agent # disable Datadog Agent as we do not have to test it in OpenTelemetry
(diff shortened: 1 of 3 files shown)


pjanotti:
Let's go with #44 instead.

PR #56, opened 2021-02-17
PoC: Refine integration tests

pellared:
## What

PoC on how we can spin up the docker containers which are needed for a given integration test suite.
Presentation: https://docs.google.com/presentation/d/1_EXZCm9A3nII5dVZ6uoGRtgWYwD4eVm0JKRwNbPx6JU/edit?usp=sharing

## Why

1. Possibility just to execute tests using dotnet test or from your IDE 
1. Possibility to debug tests from IDE
1. Should use only the resources needed for particular tests → no need to spin up all containers upfront

## Issues

1. I had to vendor https://github.com/devjoes/DockerComposeFixture. It misses SNK and there were some other issues that I solved after a couple of hours. I think it would be better to use https://github.com/dotnet/Docker.DotNet instead. I believe it would also remove the requirement to install Docker CE and Docker Compose in the containers.
1. For some reason the IntegrationTests are not working on Linux in Azure Pipelines... Reason unknown.

## Alternatives

Instead of https://github.com/devjoes/DockerComposeFixture (or writing something similar ourselves), it might be better to use:
- https://github.com/HofmeisterAn/dotnet-testcontainers (looks great but I am concerned how long the project would be maintained)
- https://github.com/dotnet/Docker.DotNet

## Final thoughts

We could use this approach to reduce the amount of resources used by integration tests and **maybe** it would make them more stable. It might also encourage to run the tests locally.

However, the approach of using "Docker from Docker" may be too complex to maintain/troubleshoot. Stuff like that may stop working properly if Docker Engine would change and somebody would have to fix it. Maybe there is other stuff in the approach of testing that would give more benefits even in less effort.

AFAIK there is a plan in OTel to transition to GitHub Actions. Maybe we should revisit CI in general?


pjanotti:
I like the idea @pellared. A key question is if they make the tests more reliable or not. It definitely makes it much easier to run single integration tests from an IDE. Could you do a quick demo on the next SIG meeting?

pellared:
> I like the idea @pellared. A key question is if they make the tests more reliable or not. It definitely makes it much easier to run single integration tests from an IDE. Could you do a quick demo at the next SIG meeting?

OK. I will do my best.
https://docs.google.com/presentation/d/1_EXZCm9A3nII5dVZ6uoGRtgWYwD4eVm0JKRwNbPx6JU/edit?usp=sharing

pjanotti:
Closing since it was a PoC. Robert's presentation https://youtu.be/NXsl_Xxhu2Y?t=1758 (cc @pellared)

pellared:
Probably the PoC was not working b/c of parallel test execution.
See: https://github.com/DataDog/dd-trace-dotnet/pull/1264#discussion_r589605719

PR #59, opened 2021-02-22
Fix Datadog.Trace.Minimal.sln for VS Code

pellared:
## Why

It is a good practice to be able to use VS Code for open-source projects. At least because this is the only free IDE that works on Linux and AFAIK the only tool that can be used to develop inside a container (https://code.visualstudio.com/docs/remote/containers).
Here is the reason why it currently does not work: https://github.com/OmniSharp/omnisharp-vscode/issues/1783

## What

Fixes the project files by removing .NET Framework targets from non-Windows systems.
Thanks to it, it is possible to use ` Datadog.Trace.Minimal.sln` for development in VS Code.

## Not covered

There would be more stuff to do to make `Datadog.Trace.sln`  with VS Code. At least change some `.csproj` files  from the old format to the new .NET Core `csproj` format. These changes would require more work and testing.
This might be helpful: https://www.hanselman.com/blog/upgrading-an-existing-net-project-files-to-the-lean-new-csproj-format-from-net-core

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -259,8 +259,11 @@ paket-files/
 # Python Tools for Visual Studio (PTVS)
 __pycache__/
 *.pyc
+
 # vscode
 .vscode/
+.devcontainer/
+
 # Benchmark results
 BenchmarkDotNet.Artifacts/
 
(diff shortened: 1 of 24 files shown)


pellared:
Closing per #60

PR #60, opened 2021-02-22
Add docs and configs for VS Code

pellared:
## Why

It is a good practice to be able to use VS Code for open-source projects. At least because this is the only free IDE that works on Linux and AFAIK the only tool that can be used to develop inside a container (https://code.visualstudio.com/docs/remote/containers).

## What

Add docs and configs for VS Code to make it working with `Datadog.Trace.Minimal.sln`

## Not covered

There would be more stuff to do to support `Datadog.Trace.sln` with VS Code. 

Probably, some `.csproj` files from the old format to the new .NET Core `csproj` format. These changes would require more work and testing.
This might be helpful: https://www.hanselman.com/blog/upgrading-an-existing-net-project-files-to-the-lean-new-csproj-format-from-net-core

Moreover, more work would be needed to document (and maybe add some configs, scripts) for native (C++) development on MacOS and Linux. Maybe @tonyredondo  could help here and would like to have it as well.

## Other stuff

Probably, if we change `.devcontainer.example` to `.devcontainer` and `.vscode.example` to `.vscode` then it might work in GitHub Codespaces:  https://docs.github.com/en/github/developing-online-with-codespaces/configuring-codespaces-for-your-project . I have chosen to add the `.example` suffix, because it is safer if someone already has these folders or prefer a different configuration.

If we want to share these settings we can do it and change the `# vccode` section in `.gitignore` to something like:

```
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
*.code-workspace
```

### DIFF
--- a/.devcontainer.example/Dockerfile
+++ b/.devcontainer.example/Dockerfile
@@ -1,0 +1,52 @@
+FROM mcr.microsoft.com/dotnet/sdk:5.0
+
+# [Option] Install zsh
+ARG INSTALL_ZSH="true"
+# [Option] Upgrade OS packages to their latest versions
+ARG UPGRADE_PACKAGES="false"
+# [Option] Enable non-root Docker access in container
+ARG ENABLE_NONROOT_DOCKER="true"
+# [Option] Use the OSS Moby Engine instead of the licensed Docker Engine
+ARG USE_MOBY="true"
+
+# Install needed packages and setup non-root user. Use a separate RUN statement to add your
+# own dependencies. A user of "automatic" attempts to reuse an user ID if one already exists.
+ARG USERNAME=automatic
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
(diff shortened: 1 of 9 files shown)


zacharycmontoya (commented):
These changes look good to me. I'm not really sure what the workflow looks like so I'll defer approval to another one of your team members or someone else that can exercise this. If this enables building and running, then I'm all for it 👍🏼

zacharycmontoya (approved):
Got everything working locally, this is great! Dev container FTW

tonyredondo:
I’ll check this PR tomorrow, didn’t have the time today.

pjanotti (approved):
LGTM but let's wait for Tony to give it a try on mac.

tonyredondo (approved):
Works great! :shipit:

Issue #42, opened 2021-02-04
Allow for OTel span data model without code divergence from Datadog


macrogreg:
Hey folks,

Based on the discussion in the meetings, I want to make a concrete proposal.
Please treat it as the first stub; we can modify it in any way we want. I just want to help moving the discussion from theoretical to practical, so that we can move forward. :)

Challenge: 
The OTel tracer must ensure that span tags use OTel conventions for both, tag-names and tag-content-format. For OTel it's a must-have, for Datadog it is optional. The OTel currently depends on Datadog's contributions. Until this changes we need to make code internations easy. This is a proposal on how we can achieve this.

Requirements:
- OTel data model conventions must be met.
- Code bases must not diverge as a result of this particular effort. This implies that:
   - Changes to the OTel Tracer repo must be merged into the Datadog repo.
   - Datadog will only do it, if it has zero performance impact and low complexity effort.

The following solution applies only to new-style CallTarget instrumentations, and to `DiagnosticSource`-based instrumentations. We will not modify older CallSite instrumentations.
We will delegate the logic of setting tags on spans to a tag-creator/span-initializer class, which will be a singleton _per instrumentation_.  The logic of extracting data from the instrumented code will be shared, only the tag configuration will be delegated. 

Consider the [`HttpClient` instrumentation as an example](https://github.com/DataDog/dd-trace-dotnet/blob/488ccfbb1e2a70dba921676b74fbd7f7ea4daed6/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/HttpClientHandler/HttpClientHandlerIntegration.cs#L36). The meat of the logic is [here](https://github.com/DataDog/dd-trace-dotnet/blob/488ccfbb1e2a70dba921676b74fbd7f7ea4daed6/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/SocketsHttpHandler/SocketsHttpHandlerCommon.cs#L17).

Note the line
```cs
Scope scope = ScopeFactory.CreateOutboundHttpScope(Tracer.Instance, requestMessage.Method.Method, requestMessage.RequestUri, IntegrationId, out HttpTags tags);
```
It does 2 things: (1) Create a span and (2) set tags on that span. We want to separate these 2 concerns as follows:

I. Create an instrumentation-specific iface:

```cs
interface ISpanInitializerForSocketsHttpHandler
{
[...]

PR #63, opened 2021-02-26
Add OTel semantic conventions for Outbound HTTP (span name and attributes)

pellared:
## Why

Attempt to implement @macrogreg idea described under #42

Extension points for trace/span conventions are needed so that we can support more than one model (e.g. OTel, DataDog, NewRelic, Splunk). We want to have it in a way that would maximize the possibility of having a common codebase. At the same time, we want to minimize potential conflicts if something "vendor" specific would need to be done on a fork.

## What

1. Extension points for a semantic convention that is chosen at runtime via the `OTEL_CONVENTION` environmental variable. Currently defaults to `Datadog`. 
1. Add OpenTelemetry Outbound HTTP semantic convention for span's operation name and attributes. Reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md

I tried to make make the least changes in the existing code so that it is easy to cherry-pick this change by Datadog and other forks. All the new part is in dedicated classes to minimize potential conflicts.

## TODO for this PR

1. Manual E2E testing
1. Double-checking with OTel specs and .NET SDK

## Next PRs 

1. Refactor test and use https://github.com/fluentassertions/fluentassertions
1. Address https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63#issuecomment-790050197
1. Add support for setting OTel status
1. Add support for adding OTel events (exceptions)
1. Try to address https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63#discussion_r585679841

## Questions

1. Do you want to use GitHub issues for making this ⬆️  stuff more transparent?
2. Any missing parts? E.g. do we have some docs where the environmental variables are described?
3. Do you want more descriptive comments in the code that describes the design decisions? Feel free to point me where such comments would be placed.


src/Datadog.Trace.ClrProfiler.Managed/Conventions/Convention.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        public IOutboundHttpConvention OutboundHttpScopeConvention;
+
+        private static Convention singleton;
+        private static volatile bool initialized;
+        private static object syncRoot = new object();
+
+        public static Convention Get(TracerSettings settings)
+        {
+            if (!initialized)
```
dszmigielski, line 15: We could use Lazy for singleton, the sync object would not be needed.

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
```diff
@@ -101,33 +99,25 @@ public static Scope CreateOutboundHttpScope(Tracer tracer, string httpMethod, Ur
...
-                span.Type = SpanTypes.Http;
-                span.ResourceName = $"{httpMethod} {resourceUrl}";
-
-                tags.HttpMethod = httpMethod?.ToUpperInvariant();
-                tags.HttpUrl = httpUrl;
-                tags.InstrumentationName = IntegrationRegistry.GetName(integrationId);
-
-                tags.SetAnalyticsSampleRate(integrationId, tracer.Settings, enabledWithGlobalSetting: false);
+                var scope = tracer.StartActiveWithTags(OperationName, tags: tags, serviceName: serviceName, spanId: spanId);
+                Convention.Get(tracer.Settings).OutboundHttpScopeConvention.Apply(new OutboundHttpArgs
```
pjanotti, line 105: Instead of invoking it as a `Get` method, it seems better to set a `Convention` property on the `Tracer`.

src/Datadog.Trace.ClrProfiler.Managed/Conventions/OutboundHttpArgs.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+using Datadog.Trace.Configuration;
+using Datadog.Trace.Tagging;
+
+namespace Datadog.Trace.ClrProfiler.Conventions
+{
+    internal struct OutboundHttpArgs
```
pjanotti, line 7: I like the idea of having a struct for the parameters.

pjanotti:
@macrogreg @nrcventura @zacharycmontoya PTAL: it is in the initial stages and still small so easy to take a look and help guide @pellared here.

src/Datadog.Trace.ClrProfiler.Managed/Conventions/Convention.cs
pellared: No longer valid: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63/commits/c81a57ae52909d614a7bd7b58b83efb99ec2385d

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
pellared: Addressed here: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63/commits/c81a57ae52909d614a7bd7b58b83efb99ec2385d

pellared:
Refactored and improved a little. Description of the PR is updated.

src/Datadog.Trace/Tracer.cs
```diff
@@ -117,6 +118,13 @@ internal Tracer(TracerSettings settings, IAgentWriter agentWriter, ISampler samp
                 _agentWriter = new AgentWriter(api, Statsd, queueSize: Settings.TraceQueueSize);
             }
 
+            switch (Settings.Convention)
+            {
+                default:
+                    OutboundHttpConvention = new DatadogOutboundHttpConvention(this);
```
pellared, line 124: the bi-directional reference is ugly...

dszmigielski (commented):
LGTM

src/Datadog.Trace/Tagging/HttpTags.cs
```diff
@@ -4,15 +4,25 @@ namespace Datadog.Trace.Tagging
...
     internal class HttpTags : InstrumentationTags, IHasStatusCode
     {
-        protected static readonly IProperty<string>[] HttpTagsProperties =
-            InstrumentationTagsProperties.Concat(
-                new Property<HttpTags, string>(Trace.Tags.HttpStatusCode, t => t.HttpStatusCode, (t, v) => t.HttpStatusCode = v),
-                new Property<HttpTags, string>(HttpClientHandlerTypeKey, t => t.HttpClientHandlerType, (t, v) => t.HttpClientHandlerType = v),
-                new Property<HttpTags, string>(Trace.Tags.HttpMethod, t => t.HttpMethod, (t, v) => t.HttpMethod = v),
-                new Property<HttpTags, string>(Trace.Tags.HttpUrl, t => t.HttpUrl, (t, v) => t.HttpUrl = v),
-                new Property<HttpTags, string>(Trace.Tags.InstrumentationName, t => t.InstrumentationName, (t, v) => t.InstrumentationName = v));
+        private readonly HttpTagsKeysMapping _convension;
```
nrcventura, line 7: `_convension` should be `_convention`

src/Datadog.Trace/Tagging/HttpTags.cs
```diff
@@ -4,15 +4,25 @@ namespace Datadog.Trace.Tagging
...
 
-        private const string HttpClientHandlerTypeKey = "http-client-handler-type";
+        public HttpTags(HttpTagsKeysMapping convension)
```
nrcventura, line 9: `convension` should be `convention`

nrcventura (commented):
These changes look good to me. I only noticed a couple of minor spelling errors.

src/Datadog.Trace/Tagging/HttpTags.cs
pellared: Addressed https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63/commits/115d77b78c97553d89e6122bd86071e8d7328945

src/Datadog.Trace/Tagging/HttpTags.cs
pellared: Addressed https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63/commits/115d77b78c97553d89e6122bd86071e8d7328945

src/Datadog.Trace/Tracer.cs
```diff
@@ -253,6 +261,8 @@ public static Tracer Instance
 
         internal IDogStatsd Statsd { get; private set; }
 
+        internal IOutboundHttpConvention OutboundHttpConvention { get; }
```
pjanotti, line 264: I assume that in the end tracer has a single property for the semantic conventions instead of one for outbound HTTP a separate one for DB, etc. Right now it is only the first "target", correct?

test/Datadog.Trace.Tests/Tagging/TagsListTests.cs
```diff
@@ -81,7 +81,7 @@ public void Serialization()
 
         private void ValidateProperties<T>(Type type, string methodName, Func<T> valueGenerator)
         {
-            var instance = (ITags)Activator.CreateInstance(type);
+            var instance = (ITags)Activator.CreateInstance(type, true);
```
pjanotti, line 84:
```suggestion
            var instance = (ITags)Activator.CreateInstance(type, nonPublic: true);
```

src/Datadog.Trace/Tracer.cs
pjanotti: In other words, what are your plans for it to grow?

src/Datadog.Trace/Conventions/IOutboundHttpConvention.cs
```diff
@@ -0,0 +1,12 @@
+using Datadog.Trace.Tagging;
+
+namespace Datadog.Trace.Conventions
+{
+    /// <summary>
+    /// Convention used to when defining operation names, span tags, statuses for outbound HTTP requests.
+    /// </summary>
+    internal interface IOutboundHttpConvention
+    {
+        Scope CreateScope(OutboundHttpArgs args, out HttpTags tags);
```
pjanotti, line 10: Depending on how this grows to cover other operations the name may need to change to something more specific. eg.: `tracer.Convention.CreateOutboundHttpScope(...)`

pjanotti (commented):
I like the direction that it is going @pellared, one key question about how this grows to cover the other instrumentations.

src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -337,6 +340,14 @@ public TracerSettings(IConfigurationSource source)
         /// </summary>
         public ExporterType Exporter { get; set; }
 
+        /// <summary>
+        /// Gets or sets the name of the convention to be used.
```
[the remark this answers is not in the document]
pellared: TBH it is a very good suggestion 👍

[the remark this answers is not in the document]
pellared: Addressed https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63/commits/115d77b78c97553d89e6122bd86071e8d7328945

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -280,6 +280,14 @@ public static class ConfigurationKeys
         /// <seealso cref="TracerSettings.Exporter"/>
         public const string Exporter = "OTEL_EXPORTER";
 
+        /// <summary>
+        /// Configuration key for the convention to be used.
```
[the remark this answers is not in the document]
pellared: Addressed https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63/commits/115d77b78c97553d89e6122bd86071e8d7328945

test/Datadog.Trace.Tests/Tagging/TagsListTests.cs
pellared: Thx. Addressed https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63/commits/fca477cdb1b86f129cfd2812a8cdbda8c292ba8a

src/Datadog.Trace/Tracer.cs
pellared: **Option 1**. Have multiple fields
**Option 2**. Have a facade with multiple sematic conventions implementations

I would go with Option 1 and when we start having 3 interfaces then we should probably refactor it to Option 2.
I just do not want to create a single gigantic interface.

src/Datadog.Trace/Conventions/IOutboundHttpConvention.cs
pellared: Or a facade will have such names.

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
```diff
@@ -0,0 +1,46 @@
...
+            StatusCode = Trace.Tags.HttpStatusCode,
+            HandlerType = "http-client-handler-type",
+            InstrumentationName = Trace.Tags.InstrumentationName,
+        };
+
+        private readonly Tracer _tracer;
+
+        public OtelOutboundHttpConvention(Tracer tracer)
+        {
+            _tracer = tracer;
```
zacharycmontoya, line 23: As you commented earlier, this reference is kind of ugly. The conventions are meant to differentiate spans right? Perhaps, we can have a Tracer create the scope first (not inside this method) and this method is only responsible for setting the tags or modifying the span, but that can all be done without the Tracer.

pellared: I did the way you suggest initially, but modifying the span's `OperationName` and setting the `HttpTag`'s tag mapping after the object is created looked even worse to my eyes.

This is the problematic code:
```
tags = new HttpTags(tagKeys);  // sets the tags mapping conventions
string serviceName = _tracer.Settings.GetServiceName(_tracer, "http-client"); // I do not know what it is yet :) 
var scope = _tracer.StartActiveWithTags("http.request", tags: tags, serviceName: serviceName, spanId: args.SpanId); // this sets some stuff 
```

I am not sure yet if we want to have a different ServiceName, OperationName for OTel and Datadog.

Maybe I will be able to refactor it if I simply know more about OTel and be more familiar with this repo 😉

src/Datadog.Trace/Tagging/HttpTags.cs
```diff
@@ -26,6 +36,36 @@ internal class HttpTags : InstrumentationTags, IHasStatusCode
 
         public string HttpStatusCode { get; set; }
 
-        protected override IProperty<string>[] GetAdditionalTags() => HttpTagsProperties;
+        protected override IProperty<string>[] GetAdditionalTags()
```
zacharycmontoya, line 39: I'm not sure if I should be providing optimization feedback at this time, but this `GetAdditionalTags()` will be called each time serialization is done, so we lose performance by dynamically building it up each time versus fetching a static readonly property. What if you created your own `OTelHttpTags` implementation that only implemented the OTel conventions so this class remains untouched?

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
zacharycmontoya: No I don't think this is a blocker. Maybe later we can revisit this to make it cleaner, but anyways we will need to make a runtime decision based on how the Tracer is configured, and this accomplishes that 👍🏼

src/Datadog.Trace/Tagging/HttpTags.cs
zacharycmontoya: Hmm, actually this suggestion causes several cascading changes because we expect to have an `out HttpTags` in the actual instrumentation classes. Let me revisit that thought

pellared: What about creating an issue for it? It would be easier to compare code and performance against something already working.

src/Datadog.Trace/Tracer.cs
pjanotti: SGTM

src/Datadog.Trace/Tagging/HttpTags.cs
zacharycmontoya: Before deferring this to an issue, can you investigate a new approach where you create an `OTelHttpTags` class that subclasses the existing `HttpTags`? I'd like to see if it's possible to solve this in this PR because Datadog will be sensitive to taking code changes that reduce performance and this might solve it before it becomes a problem. If this approach doesn't work, then yes we can address this concern in a separate GitHub issue

pellared: @zacharycmontoya will do tomorrow 😉

zacharycmontoya: I imagine you can override `GetAdditionalTags` so you have a pre-built list similar to `HttpTagsProperties` but with the customized mapping. This would be less flexible than what you have here, but I think it's worth the trade-off because we would only compute it once.

pellared: Addressed: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63/commits/5ac1fa56a482960563d290033e299f382b09c621

pellared:
@zacharycmontoya I noticed that the `ScopeFactory` does not assign the `http.request.headers.host` tag. Is it OK for you? If you want it I can add it to this PR.

Reference:
```
public const string HttpRequestHeadersHost = "http.request.headers.host";
```

test/Datadog.Trace.ClrProfiler.Managed.Tests/OtelScopeFactoryTests.cs
```diff
@@ -0,0 +1,214 @@
...
+
+        public struct Result
+        {
+            public string HttpMethodProperty;
+            public string HttpMethodTag;
+            public string HttpUrlProperty;
+            public string HttpUrlTag;
+            public string HttpTargetTag;
+
+            public override string ToString()
```
pellared, line 52: I would prefer to use https://github.com/fluentassertions/fluentassertions

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
```diff
@@ -0,0 +1,43 @@
...
+            tags.SetTag("http.scheme", requestUri.Scheme);
+            tags.SetTag("http.host", requestUri.Authority);
+            tags.SetTag("http.target", $"{requestUri.PathAndQuery}{requestUri.Fragment}");
+
+            string resourceUrl = requestUri != null ? UriHelpers.CleanUri(requestUri, removeScheme: true, tryRemoveIds: true) : null;
+            scope.Span.ResourceName = $"{httpMethod} {resourceUrl}";
+
+            var integrationId = args.IntegrationInfo;
+            tags.InstrumentationName = IntegrationRegistry.GetName(integrationId);
+            tags.SetAnalyticsSampleRate(integrationId, _tracer.Settings, enabledWithGlobalSetting: false);
```
pellared, line 39: Double-check:
Do you want to keep these? They are not required by OTel but I do not think it hurts to have this stuff.

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
```diff
@@ -0,0 +1,43 @@
...
+            tags.SetTag("http.scheme", requestUri.Scheme);
+            tags.SetTag("http.host", requestUri.Authority);
+            tags.SetTag("http.target", $"{requestUri.PathAndQuery}{requestUri.Fragment}");
+
+            string resourceUrl = UriHelpers.CleanUri(requestUri, removeScheme: true, tryRemoveIds: true);
+            scope.Span.ResourceName = $"{httpMethod} {resourceUrl}";
+
+            var integrationId = args.IntegrationInfo;
+            tags.InstrumentationName = IntegrationRegistry.GetName(integrationId);
+            tags.SetAnalyticsSampleRate(integrationId, _tracer.Settings, enabledWithGlobalSetting: false);
```
pellared, line 39: Double-check:
Do you want to keep these? They are not required by OTel but I do not think it hurts to have this stuff.

src/Datadog.Trace/Conventions/DatadogOutboundHttpConvention.cs
```diff
@@ -0,0 +1,37 @@
...
+            tags = new DatadogHttpTags();
+            var requestUri = args.RequestUri;
+            var httpMethod = args.HttpMethod;
+
+            string serviceName = _tracer.Settings.GetServiceName(_tracer, "http-client");
+            var scope = _tracer.StartActiveWithTags("http.request", tags: tags, serviceName: serviceName, spanId: args.SpanId);
+            scope.Span.Type = SpanTypes.Http;
+
+            tags.HttpMethod = httpMethod;
+            tags.HttpUrl = UriHelpers.CleanUri(requestUri, removeScheme: false, tryRemoveIds: false);
```
pellared, line 27: I removed the null-checks for `requestUri` and `UpercaseInvariant` for `httpMethod`.
Both looked unnecessary to me.
For application development, I tend to be a defensive programmer, but here (for auto instrumentation) I think that performance matters and we should not do any unnecessary stuff.

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
```diff
@@ -0,0 +1,43 @@
...
+            string operationName = $"HTTP {httpMethod}";
+            string serviceName = _tracer.Settings.GetServiceName(_tracer, "http-client");
+            var scope = _tracer.StartActiveWithTags(operationName, tags: tags, serviceName: serviceName, spanId: args.SpanId);
+            scope.Span.Type = SpanTypes.Http;
+
+            tags.HttpMethod = httpMethod;
+            tags.HttpUrl = $"{requestUri.Scheme}{Uri.SchemeDelimiter}{requestUri.Authority}{requestUri.PathAndQuery}{requestUri.Fragment}";
+            tags.SetTag("http.scheme", requestUri.Scheme);
+            tags.SetTag("http.host", requestUri.Authority);
+            tags.SetTag("http.target", $"{requestUri.PathAndQuery}{requestUri.Fragment}");
```
pellared, line 32: @zacharycmontoya  should I make it as `AdditionalTags` instead or would it be premature optimization?

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
zacharycmontoya: This is only used for Datadog so it should probably be removed

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
zacharycmontoya: Either optimizing now or later is fine with me. If you think it would be a good example to demonstrate OTel-specific tagging, then that would be a good reason to do that now

src/Datadog.Trace/Tagging/HttpTagsKeysMapping.cs
```diff
@@ -0,0 +1,11 @@
+namespace Datadog.Trace.Tagging
+{
+    internal struct HttpTagsKeysMapping
```
zacharycmontoya, line 3: After the `OtelHttpTags` refactor, is this used anywhere?

zacharycmontoya:
> @zacharycmontoya I noticed that the `ScopeFactory` does not assign the `http.request.headers.host` tag. Is it OK for you? If you want it I can add it to this PR.
> 
> Reference:
> 
> ```
> public const string HttpRequestHeadersHost = "http.request.headers.host";
> ```

Yeah that sounds good to me. I'm guessing we didn't have that because the project started with a OpenTracing conventions and that wasn't part of the spec

zacharycmontoya:
> image
> 
> Zipkin exporter misses the `AdditionalTags`

Yes, the ZipkinSerializer needs to also traverse `span.GetAdditionalTags();`, which is a separate data store than `span.Tags`

src/Datadog.Trace/Tagging/HttpTagsKeysMapping.cs
pellared: no. nice catch. thanks 👍 totally missed it when I was reviewing my own PR 😉

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
pellared: @pjanotti @nrcventura can you confirm, please?

test/Datadog.Trace.ClrProfiler.Managed.Tests/OtelScopeFactoryTests.cs
pellared: I can do it in the next PR

pellared:
> > image
> > Zipkin exporter misses the `AdditionalTags`
> 
> Yes, the ZipkinSerializer is failing. Currently we're not surfacing the `TagsList.GetAdditionalTags` except inside `TagsList.WriteTags`...Let me think on this one...

@zacharycmontoya : AFAIK @RassK is going to take a look at it

pellared:
> > @zacharycmontoya I noticed that the `ScopeFactory` does not assign the `http.request.headers.host` tag. Is it OK for you? If you want it I can add it to this PR.
> > Reference:
> > ```
> > public const string HttpRequestHeadersHost = "http.request.headers.host";
> > ```
> 
> Yeah that sounds good to me. I'm guessing we didn't have that because the project started with a OpenTracing conventions and that wasn't part of the spec

I will address it in the next PR, just to have this one focusing on how an OTel semantic convention can be applied to other places.

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
pjanotti: Keep `tag.InstrumentationName` (it follow an OT convention)

My vote is to remove `tags.SetAnalyticsSampleRate`: it is a Datadog convention.

pellared: @pjanotti What about `ResourceName`? I think it is consumed only by `SpanMessagePackFormatter` and  `DatadogLogger`. Currently, I will remove it, but I can bring it back.

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
pellared: Done now.

src/Datadog.Trace/Tagging/HttpTagsKeysMapping.cs
pellared: Removed

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
```diff
@@ -0,0 +1,36 @@
...
+            string serviceName = _tracer.Settings.GetServiceName(_tracer, "http-client");
+            var scope = _tracer.StartActiveWithTags(operationName, tags: tags, serviceName: serviceName, spanId: args.SpanId);
+            scope.Span.Type = SpanTypes.Http;
+
+            var uri = args.RequestUri;
+            otelTags.HttpMethod = args.HttpMethod;
+            otelTags.HttpUrl = string.Concat(uri.Scheme, Uri.SchemeDelimiter, uri.Authority, uri.PathAndQuery, uri.Fragment);
+            otelTags.HttpScheme = uri.Scheme;
+            otelTags.HttpHost = uri.Authority;
+            otelTags.HttpTarget = uri.PathAndQuery + uri.Fragment;
```
pellared, line 31: TBH I am not sure if we need (and want) these.
Reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
```diff
@@ -0,0 +1,36 @@
...
+            tags = otelTags;
+
+            string operationName = "HTTP " + args.HttpMethod;
+            string serviceName = _tracer.Settings.GetServiceName(_tracer, "http-client");
+            var scope = _tracer.StartActiveWithTags(operationName, tags: tags, serviceName: serviceName, spanId: args.SpanId);
+            scope.Span.Type = SpanTypes.Http;
+
+            var uri = args.RequestUri;
+            otelTags.HttpMethod = args.HttpMethod;
+            otelTags.HttpUrl = string.Concat(uri.Scheme, Uri.SchemeDelimiter, uri.Authority, uri.PathAndQuery, uri.Fragment);
```
pellared, line 28: The OTel. .NET SDK uses `OrginalString`. However, this can cause leakage of auth data.
See: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs#L115
I feel like this is a little vuln in .NET SDK. Do you agree?

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
```diff
@@ -101,33 +99,21 @@ public static Scope CreateOutboundHttpScope(Tracer tracer, string httpMethod, Ur
...
-
-                tags.SetAnalyticsSampleRate(integrationId, tracer.Settings, enabledWithGlobalSetting: false);
+                var args = new OutboundHttpArgs
+                {
+                    SpanId = spanId,
+                    HttpMethod = httpMethod,
+                    IntegrationInfo = integrationId,
+                    RequestUri = requestUri,
+                };
+                var scope = tracer.OutboundHttpConvention.CreateScope(args, out tags);
```
tonyredondo, line 109: I always prefer to create a readonly struct in this case and if I have a lot of fields passing it by ref, something like [this](https://sharplab.io/#v2:EYLgHgbALANALiAlgGwD4AEBMBGAsAKHQGYACLEgYRIG8CT6zT0pKAnAUwEM52BlAYwD2AB3YVOyZO1YAKAJQ06DZQDdOrEuoDmAZxIBeEgDt2AdxIB5AK5xggq0YAmACThxhAQVa7F+Zf4ZeYU4jAElHAxJsTCIoAFYYJQDlV3cAWXY4AAtBCMMAIgBxAFEAFXzEv2TlACV2AEcrdh04AFVWREiTc3bEGXyst2EQAHoR/LlK6oBfAG4k/woObj4hURltHTl5qvpphYXEIx5WIwkybAAGNi4eARF2GWtbeydUz289TYVaXf90ADsmk+ADp3hlsrkdtUFvs/gsxiMSAAeHRWAC26PUAE8AHwIsYkLxaDHsY56Kw6dgRYDYlH8XH5ULPOwOFxDCiCIwqMlwRBc/LIkYMkEEpFCtGYnH4v5HE5nZAkFqsKz8OCWGyst5DYk6Ba/arEEhWZBcrQAfhIQRC4R2C2UiJRkqxrDx9oYjrqjWacAA5HpnKVSgAFEjozI5Ryiv4OwkSjEut0xhhG9BXEjgiNQgju+iO1EJ6W5kiehpNFr+kitGqhaPVfMjZ1F5P0I29Ehe8ttDrQhhw5QLVMsJa3dgAOTM91E4kk0nkvmqag0my6Zg7XEcXOQ2JZr3Z7l1MmLymisQSx4YRTKFQv9G6VY6/UG7lG4zkxe2xZHKwnpinjw4AAzYFdE/P5+wYQ5jmkBULmub8eF/f8ZCA9dOE3Ixt13Nl3l1ECtgXZJAXwsEhghSNez2A54T+fMm1dGVYyRYlSXJY0qRpOlkQZJlsO1dxOW5Xl+SMQVhVxOsPTjRtCwYqD5XOZYMO3JU4BVNU0OUndNT3XDPn1YsjSUrc6RNM1LWtMJHDtFsSzjeik3rQlOx9StAxDMMsyjYsGwcxiiNIYzMLpNNrkzSFrJzWy6NkxzklLb0Kz0atax86S/MMwKNxMh9Ohclpeko/xMs0ky+P3D5dBkMyjAtJVgismA4JIZ9hHI3ImvbDhEu7RAflvK0GvCSIdCGyLbJSMivMiVr2vG6pajLH120Mbqu0K4sIKo/ACGmIA=).

That's a personal preference non a blocker.

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
tonyredondo: I'm missing something for sure, but, why not use `Tracer.Instance` instead? If the configuration is changed at runtime we cannot pick the new configuration. Also removes this .ctor.

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
tonyredondo: Agree we shouldn't use the `OriginalString` approach if this can contain auth data.

src/Datadog.Trace/Tracer.cs
tonyredondo: hmmm now I see the reason of the .ctor with the tracer object.

It seems weird to me storing data from integrations (`IOutboundHttpConvention`
 in this case) inside the tracer, this means that if we add new integrations that support conventions we need to modify the tracer as well?

Consider to expose only the `ConventionType` from the settings and each integration (that supports conventions) reads this value from the current `Tracer.Instance` and configure itself. 

I prefer decouple the integrations from the tracer.

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
tonyredondo: Well these are not `Required` by the specs, maybe we can think later if we should have an optional flag setting to expose those keys, but by default only use the required one for performance reasons???

tonyredondo (commented):
Overall LGTM, I left some comments.

src/Datadog.Trace/Tracer.cs
pellared: Relying on `Tracer.Instace` would make the code less maintainable/readable/testable.

I think that this proposal: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63/files#r585368841 together with https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63/files#r585374238 will address your main concerns.

The `Tracer` is the root object (aggregate) so it is normal that something "inside" would need to be changed when adding new stuff. Creating a facade would at least do it not "directly".

Maybe extracting some of the Tracer's functionalities to create spans would make it a little cleaner, but TBH I am not sure if it is worth the effort.

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
pellared: Done https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63/commits/317936c36374fdb9fe3691d0cd3273778ff5beb8

tonyredondo: Well, in big structures copying the whole struct can be more expensive than passing the struct by ref. But I agree in this case we don't need it, I did the test and the JIT is optimizing both calls in the same way: 

image

src/Datadog.Trace/Tracer.cs
tonyredondo: Good thanks for clarifying.

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
tonyredondo: Good thanks for clarifying.

tonyredondo (approved):
LGTM

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
pellared: I asked a question on #otel-specification Slack channel: https://cloud-native.slack.com/archives/C01N7PP1THC/p1614854143044600

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
zacharycmontoya: Sounds good to me

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
pjanotti: Good find @pellared definitely do not use `OriginalString`.

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
pjanotti: @pellared please, create an issue to add an option for the extra tags.

zacharycmontoya (approved):

pjanotti (approved):
LGTM, small things for follow-ups.

src/Datadog.Trace/Conventions/OtelHttpTags.cs
```diff
@@ -0,0 +1,27 @@
...
+using Datadog.Trace.Tagging;
+
+namespace Datadog.Trace.Conventions
+{
+    internal class OtelHttpTags : HttpTags
+    {
+        protected static readonly IProperty<string>[] HttpTagsProperties =
+            InstrumentationTagsProperties.Concat(
+                new Property<HttpTags, string>(Trace.Tags.HttpStatusCode, t => t.HttpStatusCode, (t, v) => t.HttpStatusCode = v),
+                new Property<HttpTags, string>("http-client-handler-type", t => t.HttpClientHandlerType, (t, v) => t.HttpClientHandlerType = v),
```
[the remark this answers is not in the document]
pellared: Nice idea. Done

pellared (commented):
@pjanotti 
I think this PR is ready for being merged.

src/Datadog.Trace/Conventions/OtelOutboundHttpConvention.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            _tracer = tracer;
+        }
+
+        public Scope CreateScope(OutboundHttpArgs args, out HttpTags tags)
+        {
+            var otelTags = new OtelHttpTags();
+            tags = otelTags;
+
+            string operationName = "HTTP " + args.HttpMethod;
```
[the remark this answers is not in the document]
pjanotti: Yes, we always should profile this kind of decision.

nrcventura (approved):

PR #66, opened 2021-03-03
Change default log location

dszmigielski:
Change of default log location for opentelemetry distribution, so it doesn't overlap with datadog.


pellared (approved):
Approving.
It would be good if someone tests the Windows Installer manually.
There are still some stuff which should be renamed (e.g. in `.github/ISSUE_TEMPLATE/bug_report.md` and some comments in `src/Datadog.Trace.ClrProfiler.Native/environment_variables.h`) but we can have a separate PR for it.

src/Datadog.Trace.ClrProfiler.Native/pal.h
```diff
@@ -58,7 +58,7 @@ inline WSTRING DatadogLogFilePath() {
   return ToWSTRING(program_data +
                    R"(\Datadog .NET Tracer\logs\dotnet-tracer-native.log)");
```
pellared, line 59: I thonk this should be renamed as well

dszmigielski: Fixed

pjanotti (approved):

nrcventura:
I think that you'll also need to update some entries in Config.wxi, because I think that it's using some of those variables to define where things are installed.

zacharycmontoya (approved):

zacharycmontoya:
> I think that you'll also need to update some entries in Config.wxi, because I think that it's using some of those variables to define where things are installed.

If this PR is just concerned with the logs directory, then the MSI doesn't need to be changed. The MSI only affects where the automatic instrumentation files are placed on disk (Program Files) which is separate from the logs which will exist in Program Data.

pjanotti:
It still looks like we will have to revisit the wxs files but I want to merge to at least get Linux logs moved.

PR #68, opened 2021-03-03
Add development environment setup

pellared:
## Why

Make development on Linux, MacOS and Windows easier.

## What

1. See: https://github.com/pellared/opentelemetry-dotnet-instrumentation/tree/dev-env/dev
1. `build/docker/Datadog.Trace.ClrProfiler.Native.sh` should work on MacOS and probably Windows as well f someone installs cmake, clang, gcc (mingw or tdm)

I am not sure where all dlls are landing when the whole solution is built on Windows.
Someone can tell what should be the values the following env vars for Windows (I assume you do not want to play with gcc etc on Windows):

```
export CORECLR_PROFILER_PATH="${PWD}/src/Datadog.Trace.ClrProfiler.Native/bin/Debug/x64/OpenTelemetry.AutoInstrumentation.ClrProfiler.Native.${SUFIX}"
export OTEL_DOTNET_TRACER_HOME="${PWD}/src/Datadog.Trace.ClrProfiler.Native/bin/Debug/x64"
```


dev/envvars.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+export CORECLR_ENABLE_PROFILING="1"
+export CORECLR_PROFILER="{918728DD-259F-4A6A-AC2B-B85E1B658318}"
+export CORECLR_PROFILER_PATH="${PWD}/src/Datadog.Trace.ClrProfiler.Native/bin/Debug/x64/OpenTelemetry.AutoInstrumentation.ClrProfiler.Native.so"
```
dszmigielski, line 5: PWD here can be a bit misleading as you assume this will be run from parent directory

dszmigielski (commented):
It doesn't work for me on mac, but I didn't have much time to get into details. We can investigate together tomorrow :)

dev/envvars.sh
pellared: true. I can try to find some magic script to handle it that supports bash and zsh (AFAIK it is not simple) 😄

dev/instrument.sh
```diff
@@ -0,0 +1,42 @@
...
+    case "$os" in
+        windows*) echo "dll" ;;
+        linux*) echo "so" ;;
+        darwin*) echo "dylib" ;;
+        *) echo "OS: ${os} is not supported" ; exit 1 ;;
+    esac
+}
+
+ROOT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." >/dev/null && pwd )"
+echo $ROOT_DIR
```
pellared, line 25: I will remove it 😉

dszmigielski:
> It doesn't work for me on mac, but I didn't have much time to get into details. We can investigate together tomorrow :)

During our investigation on this topic we found out that the problem were not the scripts from this PR, but rather a leftover after renaming stuff to OTEL. There is separate PR for that issue: [here](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/84).

dszmigielski (approved):

nrcventura (approved):

zacharycmontoya (approved):

PR #71, opened 2021-03-04
Fix tags and zipkin tags serialization

RassK:
### **Background**
The Tags getter was only referring to one set of custom set tags. Other tags coming from class 'TagsList' implementation state were missing. GetAdditional tags has protected access - so all AdditionalTags were missing on final serialization (except DD API).  

### **Solution**
Referring to CustomTags as manually set tags, left naming AdditionalTags. Builds up a new full set of tags when calling for GetAll method.

### DIFF
--- a/src/Datadog.Trace/Tagging/TagsList.cs
+++ b/src/Datadog.Trace/Tagging/TagsList.cs
@@ -23,18 +23,24 @@ namespace Datadog.Trace.Tagging
 
             if (customTags != null)
             {
-                allTags.AddRange(customTags);
+                lock (customTags)
+                {
+                    allTags.AddRange(customTags);
+                }
             }
 
             if (additionalTags != null)
             {
-                foreach (var property in additionalTags)
+                lock (additionalTags)
                 {
-                    var value = property.Getter(this);
-
-                    if (value != null)
+                    foreach (var property in additionalTags)
                     {
-                        allTags.Add(new KeyValuePair<string, string>(property.Key, value));
+                        var value = property.Getter(this);
+
+                        if (value != null)
+                        {
+                            allTags.Add(new KeyValuePair<string, string>(property.Key, value));
+                        }
                     }
                 }
             }
--- a/src/Datadog.Trace/Tagging/TagsList.cs
+++ b/src/Datadog.Trace/Tagging/TagsList.cs
@@ -13,7 +13,7 @@ namespace Datadog.Trace.Tagging
 
         public List<KeyValuePair<string, double>> Metrics => Volatile.Read(ref _metrics);
 
-        public List<KeyValuePair<string, string>> GetAll()
+        public List<KeyValuePair<string, string>> GetAllTags()
         {
             var customTags = GetCustomTags();
             var additionalTags = GetAdditionalTags();
@@ -269,10 +269,7 @@ namespace Datadog.Trace.Tagging
 
         protected virtual IProperty<double?>[] GetAdditionalMetrics() => ArrayHelper.Empty<IProperty<double?>>();
 
-        private List<KeyValuePair<string, string>> GetCustomTags()
-        {
-            return Volatile.Read(ref _tags);
-        }
+        protected virtual IList<KeyValuePair<string, string>> GetCustomTags() => Volatile.Read(ref _tags);
 
         private int WriteTags(ref byte[] bytes, int offset)
         {
--- a/src/Datadog.Trace/Tagging/TagsList.cs
+++ b/src/Datadog.Trace/Tagging/TagsList.cs
@@ -13,7 +13,34 @@ namespace Datadog.Trace.Tagging
 
         public List<KeyValuePair<string, double>> Metrics => Volatile.Read(ref _metrics);
 
-        public List<KeyValuePair<string, string>> Tags => Volatile.Read(ref _tags);
+        public List<KeyValuePair<string, string>> GetAll()
+        {
+            var customTags = GetCustomTags();
+            var additionalTags = GetAdditionalTags();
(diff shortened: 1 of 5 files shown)


src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -13,7 +13,34 @@ internal abstract class TagsList : ITags
...
+
+            if (additionalTags != null)
+            {
+                foreach (var property in additionalTags)
+                {
+                    var value = property.Getter(this);
+
+                    if (value != null)
+                    {
+                        allTags.Add(new KeyValuePair<string, string>(property.Key, value));
```
pellared, line 37: Pottential bug:
You can get duplicates if the same tag is set via `customTag` and `additionalTag`. I see in `GetTag` that `additionalTag` has priority over `customTag`. If it is an acceptable performance trade-off I would at least leave a comment that such behaviour may occur.

src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -13,7 +13,34 @@ internal abstract class TagsList : ITags
 
         public List<KeyValuePair<string, double>> Metrics => Volatile.Read(ref _metrics);
 
-        public List<KeyValuePair<string, string>> Tags => Volatile.Read(ref _tags);
+        public List<KeyValuePair<string, string>> GetAll()
```
pellared, line 16: Not sure about the name b/c this interface contains not only tags but also metrics. However, I prefer methods in interfaces rather than properties. Maybe `GetAllTags()`?
I think it is better to wait for others' comments before doing any rename.

src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -13,7 +13,34 @@ internal abstract class TagsList : ITags
...
+        {
+            var customTags = GetCustomTags();
+            var additionalTags = GetAdditionalTags();
+            var allTags = new List<KeyValuePair<string, string>>(
+                customTags?.Count ?? 0 +
+                additionalTags?.Length ?? 0);
+
+            if (customTags != null)
+            {
+                allTags.AddRange(customTags);
```
pellared, line 26: **race condition:**
missing `lock` on `customTags`
Notice `SetTag()` and `ToString()` methods

src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -13,7 +13,34 @@ internal abstract class TagsList : ITags
...
+            }
+
+            if (additionalTags != null)
+            {
+                foreach (var property in additionalTags)
```
pellared, line 31: **race condition:**
missing `lock` on `customTags`
Notice `SetTag()` and `ToString()` methods

src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -13,7 +13,34 @@ internal abstract class TagsList : ITags
...
         public List<KeyValuePair<string, double>> Metrics => Volatile.Read(ref _metrics);
 
-        public List<KeyValuePair<string, string>> Tags => Volatile.Read(ref _tags);
+        public List<KeyValuePair<string, string>> GetAll()
+        {
+            var customTags = GetCustomTags();
+            var additionalTags = GetAdditionalTags();
+            var allTags = new List<KeyValuePair<string, string>>(
+                customTags?.Count ?? 0 +
+                additionalTags?.Length ?? 0);
```
pellared, line 22: here is also a race condition but it should not make any harm

pellared (changes requested):
I am mostly concerned about race conditions.
However, I appreciate the fix 👍

src/Datadog.Trace/Tagging/TagsList.cs
pellared, line 16: **blocker bug:**
I think this implementation breaks `ToString()` , `SerializeTo()` and maybe even other stuff.

src/Datadog.Trace/Tagging/TagsList.cs
RassK: AdditionalTags have priority on SetTag, so this already ensures there will be no duplicates. I didn't find any unit tests confirming it, so added one for SetTag.

pellared: Thx

src/Datadog.Trace/Tagging/TagsList.cs
pellared: Fixed

src/Datadog.Trace/Tagging/TagsList.cs
pellared: Fixed

src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -236,6 +269,11 @@ public override string ToString()
 
         protected virtual IProperty<double?>[] GetAdditionalMetrics() => ArrayHelper.Empty<IProperty<double?>>();
 
+        private List<KeyValuePair<string, string>> GetCustomTags()
```
pjanotti, line 272: nit: use same syntax style as the method above

src/Datadog.Trace/Tagging/TagsList.cs
pjanotti: I think I may have suggested `GetAll()` to avoid the stutter, however, what I agree with Robert's suggestion since it is symmetrical with the other `Get*Tags()` methods. Maybe later we look at renaming `ITags` to something else since it also include metrics.

pjanotti (approved):
LGTM

pellared (approved):

PR #72, opened 2021-03-04
Add support for 128bit traceid

dszmigielski:
DD implementation uses a 64-bit trace ID but OTel requires a 128-bit trace ID (see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spancontext).

I have changed previously used ulong to new struct - TraceId, that has access to lower bits in an ulong format if needed.
The type is chosen based on the current Tracer OutboundHttpConvention.


pjanotti:
@dszmigielski update the PR description to describe the changes.

src/Datadog.Trace/Util/SpanIdGenerator.cs
```diff
@@ -67,5 +69,15 @@ public ulong CreateNew()
 
             return (ulong)value & 0x7FFFFFFFFFFFFFFF;
         }
+
+        public BigInteger CreateNew128Bit()
+        {
+            var array = new byte[16];
+            using var generator = RandomNumberGenerator.Create();
```
pjanotti, line 76: Using the crypto one may be too expensive, look at https://github.com/dotnet/runtime/pull/46890 and the RandomNumberGenerator that they added there.

pjanotti (commented):
I don't think that this should rely directly on `ActivityTraceId`: it only exposes the Id as a string and although it gets cached it makes cumbersome for certain operations on critical locations. Consider a dedicated type that can keep returning `ulong` (the low part of the traceId) when appropriate.

dszmigielski:
> @dszmigielski update the PR description to describe the changes.

Yes I will, this was created only as a draft for now, so I can see how the CI works with the changes.

src/Datadog.Trace/Util/SpanIdGenerator.cs
dszmigielski: This is a leftover from previous POC, I will get rid of it in the next commit.

src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -25,12 +25,12 @@ public int Serialize(ref byte[] bytes, int offset, Span value, IFormatterResolve
 
             len += 2; // Tags and metrics
 
-            int originalOffset = offset;
+            var originalOffset = offset;
```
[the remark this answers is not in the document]
dszmigielski: Reverted

src/Datadog.Trace/Datadog.Trace.csproj
```diff
@@ -17,12 +17,11 @@
     <!-- Instrumented libraries -->
     <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.0.0" PrivateAssets="All" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Abstractions" Version="2.0.0" PrivateAssets="All" />
-
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.4.1" />
   </ItemGroup>
 
   <ItemGroup>
     <PackageReference Include="LibLog" Version="5.0.6" PrivateAssets="All" />
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.7.0" />
```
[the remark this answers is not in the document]
dszmigielski: Removed

src/Datadog.Trace/Sampling/RuleBasedSampler.cs
```diff
@@ -74,7 +75,8 @@ public void RegisterRule(ISamplingRule rule)
 
         private SamplingPriority GetSamplingPriority(Span span, float rate, bool agentSampling)
         {
-            var sample = ((span.TraceId * KnuthFactor) % TracerConstants.MaxTraceId) <= (rate * TracerConstants.MaxTraceId);
+            var traceId = (ulong)Convert.ToInt64(span.TraceId.ToString().Substring(startIndex: 0, length: 16), fromBase: 16);
```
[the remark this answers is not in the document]
dszmigielski: Fixed to use lower of the traceId

src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -25,12 +25,12 @@ public int Serialize(ref byte[] bytes, int offset, Span value, IFormatterResolve
...
 
             offset += MessagePackBinary.WriteMapHeader(ref bytes, offset, len);
 
             offset += MessagePackBinary.WriteString(ref bytes, offset, "trace_id");
-            offset += MessagePackBinary.WriteUInt64(ref bytes, offset, value.Context.TraceId);
+            offset += MessagePackBinary.WriteString(ref bytes, offset, value.Context.TraceId.ToString());
```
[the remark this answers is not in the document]
dszmigielski: I don't have any ideas how to approach it differently - this is no longer ulong, so there is no way to leave this without changes. We can't pass lower bits here, as we will be loosing information on the whole traceId.

src/Datadog.Trace/Datadog.Trace.csproj
```diff
@@ -23,6 +24,7 @@
 
   <ItemGroup>
     <PackageReference Include="LibLog" Version="5.0.6" PrivateAssets="All" />
+    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.1" />
```
nrcventura, line 27: Should this also have `PrivateAssests="All"`?

dszmigielski: Yep, fixed.

src/Datadog.Trace/CorrelationIdentifier.cs
```diff
@@ -57,11 +59,11 @@ public static string Env
         /// Gets the id of the active trace.
         /// </summary>
         /// <returns>The id of the active trace. If there is no active trace, returns zero.</returns>
-        public static ulong TraceId
+        public static TraceId? TraceId
```
pellared, line 62: @zacharycmontoya won't such change break similarly to: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/82#discussion_r594488610

src/Datadog.Trace.ServiceFabric/Remoting.cs
```diff
@@ -216,14 +216,14 @@ private static void GetMessageHeaders(EventArgs? eventArgs, out ServiceRemotingR
 
         private static void InjectContext(PropagationContext context, IServiceRemotingRequestMessageHeader messageHeaders)
         {
-            if (context.TraceId == 0 || context.ParentSpanId == 0)
+            if (context.TraceId.Equals(TraceId.Zero) || context.ParentSpanId == 0)
```
pellared, line 219: I think it would be more idiomatic if we overload the operators for `!=` or `==`:

```
context.TraceId == TraceId.Zero`
```

I was thinking of adding a property to `TraceId` named `Assigned` but I think it would just add an unnecessary layer of indirection

src/Datadog.Trace.ServiceFabric/Remoting.cs
```diff
@@ -216,14 +216,14 @@ private static void GetMessageHeaders(EventArgs? eventArgs, out ServiceRemotingR
...
             {
                 return;
             }
 
             try
             {
-                messageHeaders.TryAddHeader(HttpHeaderNames.TraceId, context, ctx => BitConverter.GetBytes(ctx.TraceId));
+                messageHeaders.TryAddHeader(HttpHeaderNames.TraceId, context, ctx => ConvertHexStringToByteArray(ctx.TraceId.ToString()));
```
pellared, line 226: I feel little worried about the performance of multiple conversions: `lower+upper ulongs` -> `hex string` -> `byte[]`

src/Datadog.Trace.ServiceFabric/Remoting.cs
```diff
@@ -243,13 +243,28 @@ private static void InjectContext(PropagationContext context, IServiceRemotingRe
...
+        {
+            byte[] data = new byte[hexString.Length / 2];
+            for (var index = 0; index < data.Length; index++)
+            {
+                string byteValue = hexString.Substring(index * 2, 2);
+                data[index] = byte.Parse(byteValue, NumberStyles.HexNumber, CultureInfo.InvariantCulture);
+            }
+
+            return data;
+        }
```
pellared, line 256: Can you try and compare the performance of https://stackoverflow.com/a/9995303/859275 ?

src/Datadog.Trace/Conventions/DatadogOutboundHttpConvention.cs
```diff
@@ -33,5 +33,10 @@ public Scope CreateScope(OutboundHttpArgs args, out HttpTags tags)
             tags.SetAnalyticsSampleRate(integrationId, _tracer.Settings, enabledWithGlobalSetting: false);
             return scope;
         }
+
+        public TraceId GenerateNewTraceId()
+        {
+            return TraceId.CreateRandom64Bit();
+        }
```
pellared, line 40: This should NOT be in `DatadogOutboundHttpConvention` class.
I was considering to make a `Convention` facade class and it can be introduced in this PR.
More: 
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63#discussion_r585289877
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63#discussion_r587331037

src/Datadog.Trace/Conventions/IOutboundHttpConvention.cs
```diff
@@ -8,5 +8,7 @@ namespace Datadog.Trace.Conventions
     internal interface IOutboundHttpConvention
     {
         Scope CreateScope(OutboundHttpArgs args, out HttpTags tags);
+
+        TraceId GenerateNewTraceId();
```
pellared, line 12: This should be removed per https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/72/files#r596037493

src/Datadog.Trace/Datadog.Trace.csproj
```diff
@@ -16,13 +17,14 @@
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'netcoreapp3.1'">
     <!-- Instrumented libraries -->
     <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.0.0" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.Abstractions" Version="2.0.0" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Abstractions" Version="2.0.0" />
```
pellared, line 20: Ouch. I remember the discussion on last SIG that we should not do such stuff. Am I correct or is this some different case?

src/Datadog.Trace/Datadog.Trace.csproj
```diff
@@ -16,13 +17,14 @@
...
 
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.4.1" />
   </ItemGroup>
 
   <ItemGroup>
     <PackageReference Include="LibLog" Version="5.0.6" PrivateAssets="All" />
+    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.1" PrivateAssets="All" />
```
pellared, line 27: Ouch. I remember the discussion on last SIG that we should not do such stuff. Am I correct or is this some different case?

src/Datadog.Trace/RandomNumberGenerator.cs
```diff
@@ -0,0 +1,68 @@
+using System;
+
+namespace Datadog.Trace
+{
+    /// <summary>
+    /// RandomNumberGenerator implementation is the 64-bit random number generator based on the Xoshiro256StarStar algorithm (known as shift-register generators).
+    /// </summary>
+    internal class RandomNumberGenerator
```
pellared, line 8: Is this class copied from some other place? If so could you add a hyperlink where is the original source? It will help to backtrack changes and issues.

src/Datadog.Trace/SpanContextPropagator.cs
```diff
@@ -166,6 +164,43 @@ public SpanContext Extract<T>(T carrier, Func<T, string, IEnumerable<string>> ge
...
+            return ParseTraceId(headerValues, headerName);
+        }
+
+        private static TraceId? ParseTraceId<T>(T carrier, Func<T, string, IEnumerable<string>> getter, string headerName)
+        {
+            var headerValues = getter(carrier, headerName).ToList();
+            return ParseTraceId(headerValues, headerName);
+        }
+
+        private static TraceId? ParseTraceId(IEnumerable<string> headerValues, string headerName)
```
pellared, line 180: It would be more orthogonal and readable to change the returning type to  `TraceId` and return `TraceId.Zero` instead of `null`. The usages would also be simplified.

src/Datadog.Trace/SpanContextPropagator.cs
```diff
@@ -166,6 +164,43 @@ public SpanContext Extract<T>(T carrier, Func<T, string, IEnumerable<string>> ge
...
+        {
+            var headerValuesList = headerValues.ToList();
+            foreach (var headerValue in headerValuesList)
+            {
+                try
+                {
+                    var traceId = TraceId.CreateFromString(headerValue);
+                    if (traceId.Equals(TraceId.Zero))
```
pellared, line 188: is there a possibility that a header would really be like that? would it be a valid Trace ID?
what do you think of getting rid of
```
catch (Exception ex) when (ex is ArgumentOutOfRangeException || ex is InvalidOperationException || ex is OverflowException || ex is FormatException)
```
and returning `TraceId.Zero` in `CreateFromString` in case of any exception?
Maybe it would be also possible to get rid of possible exceptions to improve performance?

src/Datadog.Trace/TraceId.cs
```diff
@@ -0,0 +1,155 @@
+using System;
+using System.Runtime.CompilerServices;
+
+namespace Datadog.Trace
+{
+    /// <summary>
+    /// Class representing 64 or 128 bit TraceID.
+    /// </summary>
+    public readonly struct TraceId
```
pellared, line 9: Can you please describe (e.g. in PR description) the decisions that were behind this design? I think you could even try to describe the whole design process which resulted in having this implementation of `TraceId`

I think that for backing fields we can have:
- higher + lower `ulong` (`lower == 0` means 64 bit, the collision probability is low - probably collision would not make any big harm)
- `[]byte`
- hex formatted `string`

Have you compared your implementation with https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L1747? It uses hex formatted `string` as backing field and I think it makes sense as this is the usual input and output for tracing (some headers).

src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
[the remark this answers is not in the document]
pellared: This could be `offset += MessagePackBinary.WriteString(ref bytes, offset, value.Context.TraceId.Lower` I guess

src/Datadog.Trace/SpanContextPropagator.cs
pellared, line 188: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/72/files#r596031226

test/benchmarks/Benchmarks.Trace/AgentWriterBenchmark.cs
```diff
@@ -37,8 +37,8 @@ static AgentWriterBenchmark()
 
             for (int i = 0; i < SpanCount; i++)
             {
-                Spans[i] = new Span(new SpanContext((ulong)i, (ulong)i, SamplingPriority.UserReject, "Benchmark", null), now);
-                EnrichedSpans[i] = new Span(new SpanContext((ulong)i, (ulong)i, SamplingPriority.UserReject, "Benchmark", null), now);
+                Spans[i] = new Span(new SpanContext(TraceId.CreateFromInt(i), (ulong)i, SamplingPriority.UserReject, "Benchmark", null), now);
```
pellared, line 40: are there any differences in the benchmark results?

src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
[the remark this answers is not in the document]
dszmigielski: The issue with that approach is we are losing half the data of the TraceId if we only pass the lower bits.

src/Datadog.Trace/Conventions/DatadogOutboundHttpConvention.cs
dszmigielski: Yes, that's what I thought. I will create a separate interface for it for now, and add a field to Tracer class.

src/Datadog.Trace/Datadog.Trace.csproj
dszmigielski: This line is my error, I removed that from the wrong package :|

src/Datadog.Trace/SpanContextPropagator.cs
dszmigielski: There is a unit test where 0 is passed and it should fail parsing as invalid TraceId. I could move the try/catch to TraceId class as it might be more readable, but I don't see any performance improvements as the exception will have to be thrown and caught anyway.

src/Datadog.Trace.ServiceFabric/Remoting.cs
dszmigielski: I like the == idea, but would like to hear others' opinion on that first :)

src/Datadog.Trace/TraceId.cs
pellared, line 9: I think you might have forgotten to add `: IEquatable<TraceId>`

src/Datadog.Trace/RandomNumberGenerator.cs
dszmigielski: Added.

src/Datadog.Trace/Conventions/DatadogOutboundHttpConvention.cs
dszmigielski: Done.

src/Datadog.Trace/SpanContextPropagator.cs
dszmigielski: Done.

src/Datadog.Trace/Conventions/DatadogOutboundHttpConvention.cs
pellared: Resolve

src/Datadog.Trace.ServiceFabric/Remoting.cs
dszmigielski: Changed it to one from the link, and moved it to TraceId class.

src/Datadog.Trace/TraceId.cs
dszmigielski: Added

src/Datadog.Trace.ServiceFabric/Remoting.cs
nrcventura: I like the `==` idea. I also thought about having a property or method similar to `Assigned`, maybe even `IsZero`, but in the end thought it was unnecessary too.

src/Datadog.Trace.ServiceFabric/Remoting.cs
nrcventura: That may save some time but it may also have an impact on memory if that TraceId is copied around (and then triggers creating a new string each time). At the same time, if we're only calling ToString right before a network operation happens, the difference in impact is probably small.

src/Datadog.Trace/CorrelationIdentifier.cs
zacharycmontoya: Yes, a lot of these changes modify the signatures of API's to use this new type instead of long's. I'll review the PR in its entirety so I can propose a more helpful suggestion, but we wouldn't be able to accept many of the internal API's that currently exist in the PR

src/Datadog.Trace/SpanContextPropagator.cs
```diff
@@ -166,6 +164,37 @@ public SpanContext Extract<T>(T carrier, Func<T, string, IEnumerable<string>> ge
...
+        private static TraceId ParseTraceId<T>(T headers, string headerName)
+            where T : IHeadersCollection
+        {
+            var headerValues = headers.GetValues(headerName);
+            return ParseTraceId(headerValues, headerName);
+        }
+
+        private static TraceId ParseTraceId<T>(T carrier, Func<T, string, IEnumerable<string>> getter, string headerName)
+        {
+            var headerValues = getter(carrier, headerName).ToList();
```
zacharycmontoya, line 176: I think the `ToList()` conversions are unnecessary since we're only enumerating the items

src/Datadog.Trace/SpanContextPropagator.cs
```diff
@@ -166,6 +164,37 @@ public SpanContext Extract<T>(T carrier, Func<T, string, IEnumerable<string>> ge
...
+            return ParseTraceId(headerValues, headerName);
+        }
+
+        private static TraceId ParseTraceId(IEnumerable<string> headerValues, string headerName)
+        {
+            var headerValuesList = headerValues.ToList();
```
zacharycmontoya, line 182: Same `ToList()` applies here

src/Datadog.Trace/CorrelationIdentifier.cs
zacharycmontoya: Disregarding the question of API changes, this could be a non-nullable `TraceId` where the default value is `TraceId.Zero`, since that would more closely align with the previous default value

src/Datadog.Trace/SpanContextPropagator.cs
```diff
@@ -46,7 +46,7 @@ public void Inject<T>(SpanContext context, T headers)
             // lock sampling priority when span propagates.
             context.TraceContext?.LockSamplingPriority();
```
zacharycmontoya, line 47: The SpanContextPropagator might also need to be feature-flagged based on the desired traceId implementation, because Datadog writes the traceId as a decimal number, whereas this is currently writing the traceId as a hexadecimal number

zacharycmontoya: This feature flag could also control the Extract method as well

src/Datadog.Trace/Logging/LibLogScopeEventSubscriber.cs
```diff
@@ -261,7 +261,7 @@ private void SetLogContext(ulong traceId, ulong spanId)
             }
         }
 
-        private void SetSerilogCompatibleLogContext(ulong traceId, ulong spanId)
+        private void SetSerilogCompatibleLogContext(TraceId traceId, ulong spanId)
```
zacharycmontoya, line 264: This line of code is only a DD feature for now, so we should make sure the string extracted here is in a decimal representation

zacharycmontoya: This also applies to the `SetLogContext` method

src/Datadog.Trace/Datadog.Trace.csproj
zacharycmontoya: Is this needed for the trace id generation? I'm not sure Datadog will accept this change because it then exposes an OOB dependency on .NET Framework, unless we did some sort of IL Repacking (if that would even work for this Unsafe library). Some of the functionality is built-in on .NET Core, so removing this NuGet package and doing a conditional compilation of whatever API's are built-in to .NET Core could work

src/Datadog.Trace.OpenTracing/OpenTracingSpanContext.cs
```diff
@@ -14,7 +14,7 @@ public OpenTracingSpanContext(ISpanContext context)
             Context = context;
         }
 
-        public string TraceId => Context.TraceId.ToString(CultureInfo.InvariantCulture);
+        public string TraceId => Context.TraceId.ToString();
```
zacharycmontoya, line 17: Can we emit the string as decimal representation here to maintain the same behavior?

src/Datadog.Trace.OpenTracing/HttpHeadersCodec.cs
```diff
@@ -42,7 +42,7 @@ public void Inject(global::OpenTracing.ISpanContext context, object carrier)
             else
             {
                 // any other OpenTracing.ISpanContext
-                headers.Set(HttpHeaderNames.TraceId, context.TraceId.ToString(InvariantCulture));
+                headers.Set(HttpHeaderNames.TraceId, context.TraceId);
```
zacharycmontoya, line 45: Can we emit the string as decimal representation here to maintain the same behavior?

src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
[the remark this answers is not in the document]
pellared: I quickly inspected the code and it looks like it is only used for Datadog Agent.

src/Datadog.Trace/RandomNumberGenerator.cs
pellared: Resovle

src/Datadog.Trace.OpenTracing/HttpHeadersCodec.cs
dszmigielski: Following the https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md this should be hex formatted, I can change it so that for 64bit version the decimal is returned to preserve previous datadog behaviour. Would that be ok for you?

src/Datadog.Trace.OpenTracing/OpenTracingSpanContext.cs
dszmigielski: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/72#discussion_r596684567

src/Datadog.Trace.ServiceFabric/Remoting.cs
pellared: Resolve

src/Datadog.Trace/SpanContextPropagator.cs
pellared: Resolve

src/Datadog.Trace/SpanContextPropagator.cs
pellared: 1. Where is this unit test?
2. I think it would make the code more readable and easier to use.

src/Datadog.Trace/TraceId.cs
pellared: TBH I did not think to add ALL possible backing fields 😄 
I looks like the `byte[]` is used only in context propagation so probably this one at least could be removed. And maybe changed to some extension method like `AsBytes()`.

pellared: Resolve

src/Datadog.Trace/SpanContextPropagator.cs
dszmigielski: 1. Datadog.Trace.Tests.HeadersCollectionTests.Extract_InvalidTraceId
2. I will move that.

src/Datadog.Trace/SpanContextPropagator.cs
dszmigielski: Changed that to use decimal for 64bit TraceId, with method to parse being chosen in Convention class.

src/Datadog.Trace/Logging/LibLogScopeEventSubscriber.cs
dszmigielski: 64bit version should now preserve the old behaviour (decimal format).

src/Datadog.Trace/SpanContextPropagator.cs
dszmigielski: Removed

src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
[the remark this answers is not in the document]
dszmigielski: @zacharycmontoya should I just change it back to WriteUInt64 and pass Lower bits?

src/Datadog.Trace/SpanContextPropagator.cs
dszmigielski: Yep, the IEnumerable is enumerated twice in this method, so I did that.

lucaspimentel: `headerValues` is only enumerated twice if there is a trace id header with an invalid value. The different is probably very small, but it seems like this is trading performance in the common case vs the exceptional case.

src/Datadog.Trace/TraceIdExtensions.cs
```diff
@@ -0,0 +1,40 @@
+namespace Datadog.Trace
+{
+    /// <summary>
+    /// Extensions for <see cref="TraceId"/> class.
+    /// </summary>
+    public static class TraceIdExtensions
```
lucaspimentel, line 6: Can we add tests for these methods? I couldn't find any in this PR.

dszmigielski: Good catch, thanks!

src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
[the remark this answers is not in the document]
dszmigielski: Okay, that's what I did here

src/Datadog.Trace.ServiceFabric/Remoting.cs
dszmigielski: Changed to ==

src/Datadog.Trace/Conventions/IOutboundHttpConvention.cs
dszmigielski: Done, moved to separate interface for trace id conventions.

src/Datadog.Trace/SpanContextPropagator.cs
dszmigielski: Moved

src/Datadog.Trace.ServiceFabric/Remoting.cs
```diff
@@ -243,13 +243,28 @@ private static void InjectContext(PropagationContext context, IServiceRemotingRe
...
             try
             {
-                ulong traceId = messageHeaders.TryGetHeaderValueUInt64(HttpHeaderNames.TraceId) ?? 0;
+                var traceIdAsString = messageHeaders.TryGetHeaderValueString(HttpHeaderNames.TraceId);
+                var traceId = traceIdAsString == null
+                                  ? TraceId.Zero
+                                  : TraceId.CreateFromString(traceIdAsString);
 
-                if (traceId > 0)
+                if (!traceId.Equals(TraceId.Zero))
```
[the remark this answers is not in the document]
dszmigielski: Done

src/Datadog.Trace/TraceId.cs
```diff
@@ -0,0 +1,212 @@
...
+            Lower = lower;
+            _string = $"{_higher:x16}{Lower:x16}";
+        }
+
+        private TraceId(ulong lower)
+        {
+            _is64Bit = true;
+            _higher = 0;
+            Lower = lower;
+            _string = Lower.ToString(CultureInfo.InvariantCulture);
```
zacharycmontoya, line 29: This is great, thank you for accepting and incorporating my feedback 😄

src/Datadog.Trace/TraceId.cs
```diff
@@ -0,0 +1,212 @@
...
+                return Zero;
+            }
+        }
+
+        /// <summary>
+        /// Creates traceId from given string representing 64bit traceId in decimal format.
+        /// </summary>
+        /// <param name="id">String ID to be parsed.</param>
+        /// <returns>Instance of 64bit <see cref="TraceId"/> representing the same traceId as the passed string.</returns>
+        public static TraceId CreateFromDecimalString(string id)
```
zacharycmontoya, line 144: I think it would make sense to name this something like `CreateFromStringDatadogConvention` that way it's clear when it should be used and when it shouldn't be. We can also document in the summary that this 1) Converts to 64-bit and 2) Expects a base 10 number for parsing. What do you think?

src/Datadog.Trace/TraceId.cs
```diff
@@ -0,0 +1,212 @@
...
+            var lower = (ulong)BitConverter.ToInt64(lowerBytes, startIndex: 0) & 0x7FFFFFFFFFFFFFFF;
+
+            return new TraceId(higher, lower);
+        }
+
+        /// <summary>
+        /// Creates random 64 bit traceId.
+        /// </summary>
+        /// <returns>Instance of randomly generated <see cref="TraceId"/>.</returns>
+        public static TraceId CreateRandom64Bit()
```
zacharycmontoya, line 87: Maybe this could also benefit from being named Datadog in some way. From this repo's perspective, this is only temporary and will be phased out, but it helps remove confusion because this method actually returns a trace id that only has 63-bits (the top bit is always zeroed out)

src/Datadog.Trace/TraceId.cs
zacharycmontoya: Thinking more on this logic, I think it's rather strange for me to be asking you to change the default behavior for these lower values. Perhaps we can pass in a `bool` or a `Func` to determine if the string should be represented as decimal or hex so we don't have a built-in assumption that 64-bit id == decimal. It's a strange and not intuitive assumption to have in the code base. What do you think?

zacharycmontoya (commented):
I just completed another pass through this review and it's looking pretty good! The one remaining question I have is about the string representations and setting them to good defaults/configuration

src/Datadog.Trace/TraceId.cs
dszmigielski: @zacharycmontoya I went the documentation way. I have changed summaries for the struct itself and its methods, so it is clear how it supports datadog convention. I think this will be easier to change when we want to move on to supporting only 128 bit traceid, as we will just get rid of the methods. Please take a look :)

pellared (approved):

zacharycmontoya (approved):
Looks good! I have a really minor nitpick in that all instances of "DataDog" should be "Datadog" (I know, our github repo has incorrect casing but we can't fix it 😞  ) but I won't block approval on that. This is a good set of changes @dszmigielski !

nrcventura (approved):

pjanotti (approved):

PR #74, opened 2021-03-05
Refactor after upstream catchup

pellared:
## Why

A few refactorings that I propose after reviewing https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/70/commits/2669bb542b066258865346ba31f6a045de70928f

## What

1. Rename `IAgentWriter` to `ITraceWriter`.

I think this name is better. And the fact that there is no `TraceWritter` class shows that it a real abstraction (not something artificial just to have test doubles in unit tests).

2. Introduce `IMetrics`.

Thanks to it the `ITraceWriter` implementations do not need to have `if (_statsd != null)`.
It is also a little step to decouple the code from `DogStatsd`.

### DIFF
--- a/src/Datadog.Trace/Agent/ExporterWriter.cs
+++ b/src/Datadog.Trace/Agent/ExporterWriter.cs
@@ -1,10 +1,10 @@
 using System;
 using System.Linq;
 using System.Threading.Tasks;
+using Datadog.Trace.Abstractions;
 using Datadog.Trace.DogStatsd;
 using Datadog.Trace.Logging;
 using Datadog.Trace.Util;
-using Datadog.Trace.Vendors.StatsdClient;
 
 namespace Datadog.Trace.Agent
 {
@@ -13,17 +13,17 @@ namespace Datadog.Trace.Agent
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<ExporterWriter>();
 
         private readonly ExporterWriterBuffer<Span[]> _tracesBuffer;
-        private readonly IDogStatsd _statsd;
+        private readonly IMetrics _metrics;
         private readonly Task _flushTask;
         private readonly TaskCompletionSource<bool> _processExit = new TaskCompletionSource<bool>();
 
         private readonly IExporter _exporter;
 
-        public ExporterWriter(IExporter exporter, IDogStatsd statsd, bool automaticFlush = true, int queueSize = 1000)
+        public ExporterWriter(IExporter exporter, IMetrics metrics, bool automaticFlush = true, int queueSize = 1000)
         {
             _tracesBuffer = new ExporterWriterBuffer<Span[]>(queueSize);
             _exporter = exporter;
-            _statsd = statsd;
+            _metrics = metrics;
 
             _flushTask = automaticFlush ? Task.Run(FlushTracesTaskLoopAsync) : Task.FromResult(true);
         }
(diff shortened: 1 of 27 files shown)


src/Datadog.Trace/Tracer.cs
```diff
@@ -96,18 +97,19 @@ internal Tracer(TracerSettings settings, IAgentWriter agentWriter, ISampler samp
                 Statsd = statsd ?? CreateDogStatsdClient(Settings, DefaultServiceName, Settings.DogStatsdPort);
             }
 
+            IMetrics metrics = Statsd != null ? new DogStatsdMetrics(Statsd) : new NullMetrics();
```
nrcventura, line 100: I like how this results in simplifying the other code that deals with metrics.

src/Datadog.Trace/Agent/ExporterWriterBuffer.cs
```diff
@@ -20,19 +20,14 @@ public bool Push(T item)
...
-                    _items[_count++] = item;
-                    return true;
-                }
-                else
-                {
-                    // drop a random trace.
-                    // note that Random.Next() is NOT thread-safe,
-                    // but we are running this inside a lock
-                    _items[_random.Next(_items.Length)] = item;
+                    // drop the trace as the buffer is full
```
nrcventura, line 25: This looks like it will effectively change the sampling strategy (under certain situations), and as a result we may want to consider making a change like this in a separate PR. At some point we'll also want to update the sampling strategy to match the strategies that OTel requires, so we may want to delay a change like this until then.

zacharycmontoya:
> 3. ExporterWriter does NOT replace any trace if the buffer is full.
We should remove this change from the PR since it will be obselete after porting the changes from https://github.com/DataDog/dd-trace-dotnet/pull/1151, which serializes the traces if the buffer becomes full.

src/Datadog.Trace/Agent/ExporterWriterBuffer.cs
zacharycmontoya: We can remove this change per https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/74#issuecomment-794679841

test/Datadog.Trace.Tests/ExporterWriterBufferTests.cs
```diff
@@ -35,7 +35,7 @@ public void Push_MoreThanCapacity_False()
 
             Assert.False(buffer.Push(101));
 
-            // Check that one random element of the queue was replaced
+            // Check NONE element of the buffer was replaced
```
zacharycmontoya, line 38: We should leave this file unchanged as well since it gets removed the Datadog trace buffer PR

src/Datadog.Trace/Agent/ExporterWriterBuffer.cs
pellared: Moved to https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/81 where we can have a dedicated discussion.

src/Datadog.Trace/Agent/ExporterWriter.cs
```diff
@@ -42,16 +42,13 @@ public void WriteTrace(Span[] trace)
                 Log.Warning("Trace buffer is full. Dropping a trace from the buffer.");
             }
 
-            if (_statsd != null)
-            {
-                _statsd.Increment(TracerMetricNames.Queue.EnqueuedTraces);
-                _statsd.Increment(TracerMetricNames.Queue.EnqueuedSpans, trace.Length);
+            _metrics.Increment(TracerMetricNames.Queue.EnqueuedTraces);
```
dszmigielski, line 45: This can be solved a lot easier by just adding '?' operator instead of if:
_statsd?.Increment(TracerMetricNames.Queue.EnqueuedTraces);
but I understand this is just a perk of having separate interface for metrics, that is mostly there for decoupling reasons.

RassK (approved):

src/Datadog.Trace/Agent/ExporterWriter.cs
pellared: Yeah. I do not think we want to couple our code to `DogStatsd` 😉

dszmigielski (approved):

zacharycmontoya (approved):

nrcventura (approved):

pjanotti (approved):

PR #77, opened 2021-03-08
Fix Datadog.Trace.ClrProfiler.Native.sh

pellared:
## Why

The docker-compose.yml and scripts where probably only tested in Azure Pieplines.
I think it was not tested locally.
`docker-compose run --rm build && docker-compose run --rm Profiler && docker-compose run --rm IntegrationTests` was not working properly.

## What

Use `$buildConfiguration` in `Datadog.Trace.ClrProfiler.Native.sh` and `docker-compose.yml`.

### DIFF
--- a/build/docker/Datadog.Trace.ClrProfiler.Native.sh
+++ b/build/docker/Datadog.Trace.ClrProfiler.Native.sh
@@ -10,13 +10,13 @@ PUBLISH_OUTPUT_NET31="$( pwd )/src/bin/managed-publish/netcoreapp3.1"
 
 cd src/Datadog.Trace.ClrProfiler.Native
 mkdir -p build
-(cd build && cmake ../ -DCMAKE_BUILD_TYPE=Release && make)
+(cd build && cmake ../ -DCMAKE_BUILD_TYPE=${buildConfiguration:-Debug}  && make)
 
-mkdir -p bin/Release/x64
-cp -f build/bin/Datadog.Trace.ClrProfiler.Native.so bin/Release/x64/OpenTelemetry.AutoInstrumentation.ClrProfiler.Native.so
+mkdir -p bin/${buildConfiguration:-Debug}/x64
+cp -f build/bin/Datadog.Trace.ClrProfiler.Native.so bin/${buildConfiguration:-Debug}/x64/OpenTelemetry.AutoInstrumentation.ClrProfiler.Native.so
 
-mkdir -p bin/Release/x64/netstandard2.0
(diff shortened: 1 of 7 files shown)


dszmigielski (approved):

docker-compose.yml
```diff
@@ -151,6 +151,8 @@ services:
       context: ./
       dockerfile: ./build/docker/native.dockerfile
     image: datadog-native
+    environment: 
+    - buildConfiguration=${buildConfiguration}
```
pjanotti, line 155: Should it default here too? It doesn't remove the need to default on the shell file since the script should stand on its own but gets "more" consistency.

build/docker/Datadog.Trace.ClrProfiler.Native.sh
```diff
@@ -7,16 +7,17 @@ cd "$DIR/../.."
 
 PUBLISH_OUTPUT_NET2="$( pwd )/src/bin/managed-publish/netstandard2.0"
 PUBLISH_OUTPUT_NET31="$( pwd )/src/bin/managed-publish/netcoreapp3.1"
+BUILD_TYPE=${buildConfiguration:-Debug}
```
pjanotti, line 10: Could check if other scripts are missing similar fallback?

pjanotti (commented):
Nice @pellared.

docker-compose.yml
pellared: Defaults are in `.env` file

build/docker/Datadog.Trace.ClrProfiler.Native.sh
pellared: will do

pellared: Other scripts use `publishTargetFramework` and `buildConfiguration` but they do not assign any default value.

Do you think it is worth to make something like:
```
buildConfiguration=${buildConfiguration:-Debug}
publishTargetFramework=${publishTargetFramework:-netcoreapp3.1}
```
in all these files?

It may be handy for me as I develop in a quasi-Linux environment 😄

pjanotti: Yes, go for it.

pellared: Addressed: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/77/commits/90c8350836813a27e1af6ef3c563c1baf63d5f9d

zacharycmontoya (approved):
Good find @pellared !

pjanotti (approved):

PR #81, opened 2021-03-10
Do not replace a random trace if buffer is full

pellared:
## Why

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/74#discussion_r590659538

## What

 `ExporterWriter` does NOT replace any trace if the buffer is full.

AFAIK this is how `AgentWriter` works and IMO it makes more sense. If the system cannot send traces then it may mean that something is going on strange and better not to lose the traces when something strange is happening. Traces from suspicious moments is more important than most recent stuff.


dszmigielski (approved):

pjanotti (approved):
LGTM. Always in favor of predictability :)

zacharycmontoya (approved):
LGTM

PR #82, opened 2021-03-10
Add support for OTel Status

pellared:
## Why

From [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span):
> After the Span is created, it SHOULD be possible to [...] set the Status

## What

1. Add `Status` with `StatusCode` and `Description` to `Span` per: [specs - set status](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status)
2. Refactor to use `Span.Status` instead of `Span.Error`. All existing instrumentations should work (sic!).

### Constraints

This PR does not respect:

>The status code SHOULD remain unset, except for the following circumstances:

This is tracked under https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/85

## Next PRs

1. Report status in Zipkin exporter. See: [zipkin status](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/zipkin.md#status)
1. Report status in Jaeger exporter. See: [jaeger status](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#status)
1. Handle Outboud HTTP request status as defined in [OTEL specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#status)
1. Add support for OTel exceptions (and stuff related with it like update exporters and the outbound HTTP instrumentation)
1. https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/85

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Managed/Integrations/GraphQLIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Integrations/GraphQLIntegration.cs
@@ -307,9 +307,9 @@ namespace Datadog.Trace.ClrProfiler.Integrations
 
             if (errorCount > 0)
             {
-                span.Error = true;
-
-                span.SetTag(Trace.Tags.ErrorMsg, $"{errorCount} error(s)");
+                var message = $"{errorCount} error(s)";
+                span.Status = SpanStatus.Error.WithDescription(message);
+                span.SetTag(Trace.Tags.ErrorMsg, message);
                 span.SetTag(Trace.Tags.ErrorType, errorType);
                 span.SetTag(Trace.Tags.ErrorStack, ConstructErrorMessage(executionErrors));
             }
--- a/src/Datadog.Trace/Abstractions/ISpan.cs
+++ b/src/Datadog.Trace/Abstractions/ISpan.cs
@@ -13,11 +13,6 @@ namespace Datadog.Trace.Abstractions
         /// </summary>
         SpanStatus Status { get; set; }
 
-        /// <summary>
-        /// Gets or sets a value indicating whether this span represents an error.
-        /// </summary>
-        bool Error { get; set; }
-
         ISpan SetTag(string key, string value);
 
         string GetTag(string key);
--- a/src/Datadog.Trace/Abstractions/ISpan.cs
+++ b/src/Datadog.Trace/Abstractions/ISpan.cs
@@ -9,6 +9,11 @@ namespace Datadog.Trace.Abstractions
         string Type { get; set; }
 
         /// <summary>
+        /// Gets or sets the span's execution status
+        /// </summary>
+        SpanStatus Status { get; set; }
+
+        /// <summary>
         /// Gets or sets a value indicating whether this span represents an error.
         /// </summary>
         bool Error { get; set; }
--- a/src/Datadog.Trace/Span.cs
+++ b/src/Datadog.Trace/Span.cs
@@ -64,6 +64,11 @@ namespace Datadog.Trace
         public bool Error { get; set; }
 
         /// <summary>
+        /// Gets or sets the span's execution status
+        /// </summary>
+        public SpanStatus Status { get; set; }
+
+        /// <summary>
         /// Gets or sets the service name.
         /// </summary>
         public string ServiceName
@@ -112,6 +117,7 @@ namespace Datadog.Trace
             sb.AppendLine($"Type: {Type}");
             sb.AppendLine($"Start: {StartTime}");
             sb.AppendLine($"Duration: {Duration}");
+            sb.AppendLine($"Status: {Status}");
             sb.AppendLine($"Error: {Error}");
             sb.AppendLine($"Meta: {Tags}");
 
--- a/src/Datadog.Trace/Span.cs
+++ b/src/Datadog.Trace/Span.cs
@@ -59,11 +59,6 @@ namespace Datadog.Trace
         public string Type { get; set; }
 
         /// <summary>
-        /// Gets or sets a value indicating whether this span represents an error
-        /// </summary>
-        public bool Error { get; set; }
-
-        /// <summary>
         /// Gets or sets the span's execution status
         /// </summary>
         public SpanStatus Status { get; set; }
@@ -118,7 +113,6 @@ namespace Datadog.Trace
             sb.AppendLine($"Start: {StartTime}");
             sb.AppendLine($"Duration: {Duration}");
             sb.AppendLine($"Status: {Status}");
-            sb.AppendLine($"Error: {Error}");
             sb.AppendLine($"Meta: {Tags}");
 
             return sb.ToString();
@@ -284,8 +278,6 @@ namespace Datadog.Trace
         /// <param name="exception">The exception.</param>
         public void SetException(Exception exception)
         {
-            Error = true;
-
             if (exception != null)
             {
                 // for AggregateException, use the first inner exception until we can support multiple errors.
@@ -296,9 +288,14 @@ namespace Datadog.Trace
                     exception = aggregateException.InnerExceptions[0];
                 }
 
+                Status = SpanStatus.Error.WithDescription(exception.Message);
                 SetTag(Trace.Tags.ErrorMsg, exception.Message);
                 SetTag(Trace.Tags.ErrorStack, exception.ToString());
                 SetTag(Trace.Tags.ErrorType, exception.GetType().ToString());
+            }
+            else
+            {
+                Status = SpanStatus.Error;
             }
         }
 
--- a/src/Datadog.Trace/SpanStatus.cs
+++ b/src/Datadog.Trace/SpanStatus.cs
@@ -1,0 +1,119 @@
+namespace Datadog.Trace
+{
+    /// <summary>
+    /// Span execution status.
+    /// </summary>
+    public readonly struct SpanStatus : System.IEquatable<SpanStatus>
+    {
+        /// <summary>
+        /// The operation completed successfully.
+        /// </summary>
+        public static readonly SpanStatus Ok = new SpanStatus(StatusCode.Ok);
+
+        /// <summary>
+        /// The default status.
+        /// </summary>
+        public static readonly SpanStatus Unset = new SpanStatus(StatusCode.Unset);
+
+        /// <summary>
+        /// The operation contains an error.
+        /// </summary>
+        public static readonly SpanStatus Error = new SpanStatus(StatusCode.Error);
+
+        internal SpanStatus(StatusCode statusCode, string description = null)
+        {
+            this.StatusCode = statusCode;
+            this.Description = description;
+        }
+
+        /// <summary>
+        /// Gets the canonical code from this status.
+        /// </summary>
+        public StatusCode StatusCode { get; }
+
+        /// <summary>
+        /// Gets the status description.
+        /// </summary>
+        /// <remarks>
+        /// Note: Status Description is only valid for <see
+        /// cref="StatusCode.Error"/> Status and will be ignored for all other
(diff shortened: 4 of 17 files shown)


test/Datadog.Trace.Tests/SpanStatusTests.cs
```diff
@@ -0,0 +1,104 @@
+using Xunit;
```
pellared, line 1: Taken from https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Tests/Trace/StatusTest.cs

src/Datadog.Trace/StatusCode.cs
```diff
@@ -0,0 +1,23 @@
+namespace Datadog.Trace
```
pellared, line 1: Taken from https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/Trace/StatusCode.cs

src/Datadog.Trace/SpanStatus.cs
```diff
@@ -0,0 +1,119 @@
+namespace Datadog.Trace
```
pellared, line 1: Taken from https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/Trace/Status.cs

dszmigielski (approved):
LGTM

RassK (approved):
Seems nice and clean.

pellared:
The only "failed" test is:
```
[xUnit.net 00:05:02.85]     Datadog.Trace.ClrProfiler.IntegrationTests.MultiDomainHostTests.WorksOutsideTheGAC(targetFramework: "net47") [FAIL]
  Failed Datadog.Trace.ClrProfiler.IntegrationTests.MultiDomainHostTests.WorksOutsideTheGAC(targetFramework: "net47") [42 s]
  Error Message:
   Expected at least 8 span, only received 7
Expected: True
Actual:   False
```

So the tests looks good.

src/Datadog.Trace/SpanStatus.cs
pellared: maybe I should mention it in comment?

src/Datadog.Trace/StatusCode.cs
pellared: maybe I should mention it in comment?

test/Datadog.Trace.Tests/SpanStatusTests.cs
pellared: maybe I should mention it in a comment?

pjanotti (approved):
LGTM. Just one Q.

pjanotti:
@zacharycmontoya this changes `span.Error` to a new struct but doesn't affect data serialization. PTAL.

src/Datadog.Trace/Span.cs
```diff
@@ -290,10 +288,15 @@ public void SetException(Exception exception)
...
                 }
 
+                Status = SpanStatus.Error.WithDescription(exception.Message);
                 SetTag(Trace.Tags.ErrorMsg, exception.Message);
                 SetTag(Trace.Tags.ErrorStack, exception.ToString());
                 SetTag(Trace.Tags.ErrorType, exception.GetType().ToString());
             }
+            else
+            {
+                Status = SpanStatus.Error;
```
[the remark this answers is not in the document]
pellared: I quickly browsed all 74 references. Looks like it is never null.
Even if we decide to change it, I prefer to do it in a separate PR. So that we can easily and quickly revert it if it will not work.

src/Datadog.Trace/SpanStatus.cs
```diff
@@ -0,0 +1,119 @@
...
+            var result = 1;
+            result = (31 * result) + this.StatusCode.GetHashCode();
+            result = (31 * result) + (this.Description?.GetHashCode() ?? 0);
+            return result;
+        }
+
+        /// <inheritdoc/>
+        public override string ToString()
+        {
+            return nameof(SpanStatus)
```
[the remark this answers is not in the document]
pellared: This class is copied from OTel .NET SDK. I prefer to have it without changes (I changed only the type name). Especially that `ToString()` is only used in unit tests (xUnit assertion messages) and during debugging.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/GraphQLIntegration.cs
```diff
@@ -307,9 +307,9 @@ private static void RecordExecutionErrorsIfPresent(Span span, string errorType,
...
             if (errorCount > 0)
             {
-                span.Error = true;
-
-                span.SetTag(Trace.Tags.ErrorMsg, $"{errorCount} error(s)");
+                var message = $"{errorCount} error(s)";
+                span.Status = SpanStatus.Error.WithDescription(message);
+                span.SetTag(Trace.Tags.ErrorMsg, message);
                 span.SetTag(Trace.Tags.ErrorType, errorType);
                 span.SetTag(Trace.Tags.ErrorStack, ConstructErrorMessage(executionErrors));
```
nrcventura, line 314: We'll probably need to handle these using the semantic conventions abstraction, because they are probably required for Datadog, but they are not part of the OTel semantic conventions. Either way, that work can be done separately from this PR.

src/Datadog.Trace/SpanStatus.cs
nrcventura: I like the idea of having that mentioned in a comment.

nrcventura (approved):
This looks good to me. I think that the things that were called out can be handled with separate PRs.

test/Datadog.Trace.Tests/SpanStatusTests.cs
pellared: done

src/Datadog.Trace/SpanStatus.cs
pellared: done

src/Datadog.Trace/StatusCode.cs
pellared: done

src/Datadog.Trace.ClrProfiler.Managed/Integrations/GraphQLIntegration.cs
pellared: I will mention this comment under https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/85

src/Datadog.Trace/Abstractions/ISpan.cs
```diff
@@ -9,9 +9,9 @@ internal interface ISpan
         string Type { get; set; }
 
         /// <summary>
-        /// Gets or sets a value indicating whether this span represents an error.
+        /// Gets or sets the span's execution status
         /// </summary>
-        bool Error { get; set; }
+        SpanStatus Status { get; set; }
```
zacharycmontoya, line 14: Datadog won't be able to accept this change...even though it is an internal interface, this may cause a crash in Datadog automatic instrumentation. I can speak more to it, but we can't accept this. Perhaps we can add a new interface member instead?

zacharycmontoya (changes requested):
I recommend adding the SpanStatus as a separate interface. Currently, removing a member we currently use (even though it is internal) may cause issues in Datadog and we won't be able to take the change as stands

src/Datadog.Trace/Abstractions/ISpan.cs
pellared: Can you please give repro steps and/or what is going to cause the issue? It will help me avoid similar issues in future.

zacharycmontoya: The steps for this situation are:
1. .NET Core application
2. Datadog automatic instrumentation enabled for the application, using a lower version (say 1.24.0)
3. Datadog NuGet package added to the application, using a higher version that has an internal API removed

Result: `MissingMethodException`. This occurs because the automatic instrumentation (`Datadog.Trace.ClrProfiler.Managed`) has an assembly dependency on the manual instrumentation (`Datadog.Trace`) and is allowed to make internal calls across assemblies. However, using a NuGet package allows the application developer to add a newer version than what is expected and the type loading will unify on the newer version, so issues occur when the API exists on the older `Datadog.Trace` but are removed in a later `Datadog.Trace`. We are working to resolve this issue but we have not completed that work yet.

pellared: Can you just confirm that it would be reproducible only for https://github.com/DataDog/dd-trace-go if you just merge these changes as it is? To reproduce it here I would need to create NuGet package for OTel .NET AutoInstrumentation manually.

### Option 1
Make the internal API backwards compatible.

### Option 2
https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation does not support Datadog NuGet package. It should support only the OpenTracing NuGet package at the moment. In future, it could support .NET Activity API and OpenTelemetry SDK.
Datadog can always make add the "backward compatibility" when merging this PR to the https://github.com/DataDog/dd-trace-dotnet repository.

I just wanted to propose **Option 2** as I see more and more problems connected with "API compatibility" (e.g. here https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/72). I do not want to force any solution. I just think we should think about it more and maybe even describe the options with pros and cons in some Google Docs and discuss it during the next SIG meeting(s).

zacharycmontoya: > Datadog can always make add the "backward compatibility" when merging this PR to the https://github.com/DataDog/dd-trace-dotnet repository.

Yeah I am on board with this change, we can make the API compatibility change in the dd-trace-dotnet repo. I think Paulo's comment about adding a mechanism to detect API changes would be really helpful to identify changes like this in the future.

nrcventura: Is there an Option 3, which I think is out of scope for this PR, where we start to build a new OTel specific tracer API implementation? This tracer would only need to implement an API surface area that OTel needs, and would allow the APIs to operate independently. We are starting to see more and more points of friction to maintain compatibility, and we may need to start carefully choosing where and where not to maintain compatibility to allow progress to continue.

pellared: @zacharycmontoya That is very kind. Even though I do not want to rush. I feel this is an important design decision that I that SIG would have a "quorum" agreement. Maybe there are some other ways to make life easier for all of us. Maybe we could make a brainstorming session during SIG meeting 😉 

@nrcventura I was thinking of **Option 3** even when I was analyzing the semantic conventions for attributes/tags. I felt the number of changes would be unrealistic but now I think it is worth considering it again. It is good to have as many options as possible 👍 I will edit my comment and add your option proposal.

iNikem:
@zacharycmontoya What is the status of this PR? Do I read it right, that you object to the change that is in conformance with Otel spec? Or is there something else at play that I miss?

zacharycmontoya:
> @zacharycmontoya What is the status of this PR? Do I read it right, that you object to the change that is in conformance with Otel spec? Or is there something else at play that I miss?

Hi Nikita, we had some back and forth discussions to understand how we could make this change most compatible for code flow going both ways between the originating DD repo and this repo. After some discussion in the SIG meetings, we're good to move forward with these changes and when DD ports these changes we can make them backward compatible.

This PR and some of the other PR's going on concurrently were the first significant ones to test the management of code divergence between the two repo's so there's been a lot of discussion to understand how to handle it.

pellared:
@zacharycmontoya I rebased the PR. Can you merge it, please?

zacharycmontoya (approved):

PR #84, opened 2021-03-11
Fix leftovers after renaming profiler

dszmigielski:
There were some leftovers after series of renaming PRs that we found out were not changed (some of them are new from upstream merge). So here they are.

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
+++ b/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
@@ -2653,7 +2653,7 @@ void CorProfiler::GetAssemblyAndSymbolsBytes(BYTE** pAssemblyArray, int* assembl
     for(auto i = 0; i < imgCount; i++) {
         const std::string name = std::string(_dyld_get_image_name(i));
 
-        if (name.rfind("Datadog.Trace.ClrProfiler.Native.dylib") != std::string::npos) {
+        if (name.rfind("OpenTelemetry.AutoInstrumentation.ClrProfiler.Native.dylib") != std::string::npos) {
             const mach_header_64* header = (const struct mach_header_64 *) _dyld_get_image_header(i);
 
             unsigned long dllSize;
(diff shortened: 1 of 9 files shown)


src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -2653,7 +2653,7 @@ void CorProfiler::GetAssemblyAndSymbolsBytes(BYTE** pAssemblyArray, int* assembl
     for(auto i = 0; i < imgCount; i++) {
         const std::string name = std::string(_dyld_get_image_name(i));
 
-        if (name.rfind("Datadog.Trace.ClrProfiler.Native.dylib") != std::string::npos) {
+        if (name.rfind("OpenTelemetry.AutoInstrumentation.ClrProfiler.Native.dylib") != std::string::npos) {
```
dszmigielski, line 2656: This line not being changed caused macOS profiler to fail at the application start, crashing the application with weird error if the profiler name didn't match the one in if (and it didn't since we have changed it to OTEL name):

Unhandled exception. System.OverflowException: Arithmetic operation resulted in an overflow.
   at __DDVoidMethodType__.__DDVoidMethodCall__()
   at Microsoft.DotNet.Cli.Program.Main(String[] args)

dszmigielski: Does anybody have a clue why does it fail like that, and why do we have to hardcode the name of the output file in the code?

pellared: @tonyredondo ?

pellared (approved):
Tested more or less together 😉

nrcventura (approved):

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
pellared: @dszmigielski please create an issue for it

pjanotti (approved):

PR #90, opened 2021-03-16
Make Zach a maintainer

pjanotti:
Per conversation at the last SIG meeting making Zach officially a maintainer. I will get the teams updated in parallel.

cc @zacharycmontoya @nrcventura @macrogreg

### DIFF
--- a/.azure-pipelines/integration-tests.yml
+++ b/.azure-pipelines/integration-tests.yml
@@ -6,7 +6,6 @@ trigger:
     exclude:
       - docs/*
       - .github/*
-      - README.md
 pr:
   branches:
     include:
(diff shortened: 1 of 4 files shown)


pellared (approved):

nrcventura (approved):

.azure-pipelines/integration-tests.yml
```diff
@@ -6,6 +6,7 @@ trigger:
     exclude:
       - docs/*
       - .github/*
+      - README.md
```
pjanotti, line 9: This doesn't seem to work the way I wanted :( ...

pjanotti: > Also, the path exclude parameter only works for file path instead of file

:( source: https://developercommunity.visualstudio.com/t/azure-pipelines-triggers-pathsexclude-in-pr-is-not/1046653#T-N1047494

PR #92, opened 2021-03-17
Add propagation switch and B3

RassK:
### **What's done**
- Ported B3 implementation from Splunk SignalFX. (No need to keep similarity to source in future as it's in maintenance now)
- Added propagation switch: ENV variable **OTEL_PROPAGATOR** which defaults to DD propagation.
- Changed TracingEnabled switch value to '**otel-tracing-enabled**' and moved to CommonHttpHeaderNames to remove propagator dependency
- Separated B3 and DD header names
- Added RabbitMQ support for B3 which was originally missing
- Added Cartesian Product (product of different header collections and additional params) test theories to B3

### **What's next**
- Add support to multiple propagators [(see comment)](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/92#discussion_r600330255)
- Reuse context agnostic propagator logic for ServiceFabric. (See experimental code [here](https://github.com/RassK/opentelemetry-dotnet-instrumentation/compare/context-propagation-b3...RassK:context-propagation-b3-experimental))


src/Datadog.Trace/Propagation/ContextPropagatorResolver.cs
```diff
@@ -0,0 +1,28 @@
...
+        };
+
+        public static IPropagator GetPropagator(PropagatorType propagator)
+        {
+            if (_propagatorSelector.TryGetValue(propagator, out Func<IPropagator> getter))
+            {
+                return getter();
+            }
+
+            throw new InvalidOperationException($"There is propagator registered for type '{propagator}'");
```
lucaspimentel, line 25:
```suggestion
            throw new InvalidOperationException($"There is no propagator registered for type '{propagator}'");
```

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -283,7 +284,8 @@ private void OnHostingHttpRequestInStart(object arg)
 
                 string resourceName = $"{httpMethod} {resourceUrl}";
 
-                SpanContext propagatedContext = ExtractPropagatedContext(request);
+                IPropagator propagator = ContextPropagatorResolver.GetPropagator(tracer.Settings.Propagator);
```
zacharycmontoya, line 287:
```suggestion
                IPropagator propagator = tracer.Propagator;
```

src/Datadog.Trace/Propagation/ContextPropagatorResolver.cs
zacharycmontoya, line 25:
```suggestion
            throw new InvalidOperationException($"There is no propagator registered for type '{propagator}'");
```

src/Datadog.Trace/Configuration/PropagatorType.cs
```diff
@@ -0,0 +1,23 @@
+namespace Datadog.Trace.Configuration
+{
+    /// <summary>
+    /// Enumeration for the available propagator types.
+    /// </summary>
+    public enum PropagatorType
+    {
+        /// <summary>
+        /// The default propagator.
```
pellared, line 9: Worth mentioning what is the default one.

src/Datadog.Trace.ServiceFabric/Propagators/ServiceFabricContextPropagator.cs
```diff
@@ -0,0 +1,11 @@
+using Microsoft.ServiceFabric.Services.Remoting.V2;
+
+namespace Datadog.Trace.ServiceFabric.Propagators
+{
+    internal abstract class ServiceFabricContextPropagator
```
pellared, line 5: any reason why an abstract class instead of an interface? performance?

src/Datadog.Trace/Propagation/Propagator.cs
```diff
@@ -0,0 +1,48 @@
...
+            {
+                if (!string.IsNullOrEmpty(headerValue))
+                {
+                    return headerValue;
+                }
+            }
+
+            return null;
+        }
+    }
```
pellared, line 47: Instead of doing an abstract class, `ExtractHeaderTags` could be an extension method and `ParseString` could be a regular static helper method.
I think it would be more loosely coupled and would not require making all these `public abstract` stuff just for sake of implementing the interface. Also, it makes the code a little easier to test.

test/Datadog.Trace.TestHelpers/HeadersCollectionTestExtensions.cs
```diff
@@ -0,0 +1,33 @@
+using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.Linq;
+using System.Net;
+using Datadog.Trace.ExtensionMethods;
+using Datadog.Trace.Headers;
+
+namespace Datadog.Trace.TestHelpers
+{
+    public abstract class HeadersCollectionTestExtensions
```
pellared, line 10: why not as a regular static class? e.g. `public static class HeadersCollectionTestHelper`
My rule of thumb: use inheritance if polymorphism is needed. Otherwise, using "composition" or helper functions is more maintainable.

src/Datadog.Trace/Propagation/B3/B3SpanContextPropagator.cs
```diff
@@ -0,0 +1,209 @@
...
+        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<B3SpanContextPropagator>();
+
+        private B3SpanContextPropagator()
+        {
+        }
+
+        /// <summary>
+        /// Gets the singleton instance of the propagator.
+        /// </summary>
+        public static B3SpanContextPropagator Instance { get; } = new B3SpanContextPropagator();
```
pellared, line 26: Maybe we can get rid of this singleton?

src/Datadog.Trace/Propagation/B3/B3HttpHeaderNames.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        /// <summary>
+        /// Setting used to determine whether a trace should be sampled or not.
+        /// </summary>
+        public const string B3Sampled = "x-b3-sampled";
+
+        /// <summary>
+        /// Setting used to determine whether a trace should be considered for debugging.
+        /// </summary>
+        public const string B3Flags = "x-b3-flags";
```
pellared, line 28: I think we can move these to `B3SpanContextPropagator`
Better cohesion -> no other place where these consts are used.

src/Datadog.Trace/Propagation/B3/B3HttpHeaderNames.cs
```diff
@@ -0,0 +1,30 @@
+namespace Datadog.Trace.Propagation.B3
```
pellared, line 1: Personally, I think that there is no need of `B3` folder and namespace. `B3` prefixes is names is enough

src/Datadog.Trace/Propagation/B3/B3SpanContextPropagator.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        /// <summary>
+        /// Propagates the specified context by adding new headers to a <see cref="IHeadersCollection"/>.
+        /// This locks the sampling priority for <paramref name="context"/>.
+        /// </summary>
+        /// <param name="context">A <see cref="SpanContext"/> value that will be propagated into <paramref name="headers"/>.</param>
+        /// <param name="headers">A <see cref="IHeadersCollection"/> to add new headers to.</param>
+        public override void Inject(SpanContext context, IHeadersCollection headers)
```
pellared, line 34: should this not be some extension method that reuses `public override void Inject<T>(SpanContext context, T carrier, Action<T, string, string> setter)` ?

src/Datadog.Trace/Propagation/B3/B3SpanContextPropagator.cs
```diff
@@ -0,0 +1,209 @@
...
+                setter(carrier, samplingHeader.Value.Key, samplingHeader.Value.Value);
+            }
+        }
+
+        /// <summary>
+        /// Extracts a <see cref="SpanContext"/> from the values found in the specified headers.
+        /// </summary>
+        /// <param name="headers">The headers that contain the values to be extracted.</param>
+        /// <returns>A new <see cref="SpanContext"/> that contains the values obtained from <paramref name="headers"/>.</returns>
+        public override SpanContext Extract(IHeadersCollection headers)
```
pellared, line 89: Same as https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/92/files#r600386458

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -293,6 +293,13 @@ public static class ConfigurationKeys
         /// </summary>
         public const string Convention = "OTEL_CONVENTION";
 
+        /// <summary>
+        /// Configuration key for the propagator logic to be used.
+        /// Default is <c>Datadog</c>.
+        /// <seealso cref="TracerSettings.Propagator"/>
+        /// </summary>
+        public const string Propagator = "OTEL_PROPAGATOR";
```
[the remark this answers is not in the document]
RassK: Agree, seems a spec miss. Renamed for this PR, should be addressed with new PR.

src/Datadog.Trace/Propagation/B3/B3HttpHeaderNames.cs
RassK: Agree, I first thought there will be more files there. But seems the final result was just 2, so flattened out both implementations.

test/Datadog.Trace.TestHelpers/HeadersCollectionTestExtensions.cs
RassK: It feeds XUnit MemberData. It's also possible to use ClassData, but then you need 3 classes.

src/Datadog.Trace.ServiceFabric/Propagators/ServiceFabricContextPropagator.cs
RassK: just a refactoring error.

src/Datadog.Trace/Propagation/Propagator.cs
RassK: Agree, seems much better idea. + ExtractHeaderTags doesn't have any connections to concrete implementation. Refactored.

src/Datadog.Trace/Propagation/B3/B3SpanContextPropagator.cs
RassK: Excellent idea for refactoring. I refactored propagator to be carrier agnostic - to continue with the pattern added previously, that carrier could be any kind of type that could hold metadata.

src/Datadog.Trace/Propagation/B3/B3SpanContextPropagator.cs
RassK: Yes, the propagator is now coupled with Tracer, as Tracer is coupled with settings and settings define the propagator. So for me seems logical refactor.

src/Datadog.Trace.ServiceFabric/Remoting.cs
```diff
@@ -389,5 +366,19 @@ private static void FinishSpan(EventArgs e, string spanKind)
             bool analyticsEnabled = integrationSettings.AnalyticsEnabled ?? (enabledWithGlobalSetting && tracer.Settings.AnalyticsEnabled);
             return analyticsEnabled ? integrationSettings.AnalyticsSampleRate : (double?)null;
         }
+
+        private static IServiceFabricContextPropagator GetContextPropagator(Tracer instance)
```
RassK, line 370: I think this should actually reuse existing context agnostic propagator. 
See experimental code [here](https://github.com/RassK/opentelemetry-dotnet-instrumentation/compare/context-propagation-b3...RassK:context-propagation-b3-experimental)

test/Datadog.Trace.TestHelpers/HeadersCollectionTestExtensions.cs
RassK: Agree, didn't focus on the that poor name choice. I think it wanted to be extension originally 😄 Renamed.

dszmigielski:
How should we approach testing the B3 context propagation? Should we run all our integration tests in both scenarios now (Datadog and B3)? This would make already long CI even worse, so I don't think that is a way to go for now. Any ideas guys?

RassK:
> How should we approach testing the B3 context propagation? Should we run all our integration tests in both scenarios now (Datadog and B3)? This would make already long CI even worse, so I don't think that is a way to go for now. Any ideas guys?

Already waiting on the list:
https://github.com/RassK/opentelemetry-dotnet-instrumentation/tree/context-propagation-integration-tests

Basically the solution is multi propagation >> verify both headers on one go. This adds minimal overhead.

RassK:
@zacharycmontoya  @lucaspimentel Could you review latest changes again?

src/Datadog.Trace.OpenTracing/HttpHeadersCodec.cs
```diff
@@ -34,16 +40,9 @@ public void Inject(global::OpenTracing.ISpanContext context, object carrier)
...
 
-            if (context is OpenTracingSpanContext otSpanContext && otSpanContext.Context is SpanContext ddSpanContext)
-            {
-                // this is a Datadog context
-                SpanContextPropagator.Instance.Inject(ddSpanContext, headers);
-            }
-            else
+            if (context is OpenTracingSpanContext otSpanContext && otSpanContext.Context is SpanContext spanContext)
             {
-                // any other OpenTracing.ISpanContext
```
zacharycmontoya, line 44: Any particular reason to remove this fallback path?

src/Datadog.Trace/Propagation/PropagationExtensions.cs
```diff
@@ -0,0 +1,36 @@
+using System.Collections.Generic;
+using Datadog.Trace.Headers;
+
+namespace Datadog.Trace.Propagation
+{
+    internal static class PropagationExtensions
+    {
+        public static void Inject(this IPropagator propagator, SpanContext context, IHeadersCollection headers)
+        {
+            propagator.Inject(context, headers, (carrier, header, value) => carrier.Set(header, value));
```
nrcventura, line 10: The lambdas are simple to follow and read, but there are usually fewer allocations involved when you just use a static method because these lambdas are not capturing any closures.

src/Datadog.Trace/Propagation/CommonHttpHeaderNames.cs
```diff
@@ -0,0 +1,14 @@
...
+    /// <summary>
+    /// Names of HTTP headers that are commonly or internaly used
+    /// </summary>
+    public class CommonHttpHeaderNames
+    {
+        /// <summary>
+        /// If header is set to "false", tracing is disabled for that http request.
+        /// Tracing is enabled by default.
+        /// </summary>
+        public const string TracingEnabled = "otel-tracing-enabled";
```
nrcventura, line 12: Should this be "x-otel-tracing-enabled" instead?

src/Datadog.Trace/Propagation/PropagationExtensions.cs
```diff
@@ -0,0 +1,36 @@
+using System.Collections.Generic;
+using Datadog.Trace.Headers;
+
+namespace Datadog.Trace.Propagation
+{
+    internal static class PropagationExtensions
+    {
+        public static void Inject(this IPropagator propagator, SpanContext context, IHeadersCollection headers)
```
zacharycmontoya, line 8: I love this 😍

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -293,6 +293,13 @@ public static class ConfigurationKeys
         /// </summary>
         public const string Convention = "OTEL_CONVENTION";
 
+        /// <summary>
+        /// Configuration key for the propagator logic to be used.
+        /// Default is <c>Datadog</c>.
+        /// <seealso cref="TracerSettings.Propagator"/>
+        /// </summary>
+        public const string Propagators = "OTEL_PROPAGATORS";
```
zacharycmontoya, line 301: Is there a particular reason this is the plural "Propagators" instead of the singular "Propagator"?

test/Datadog.Trace.Tests/Propagators/B3SpanContextPropagatorTests.cs
```diff
@@ -0,0 +1,201 @@
...
+
+            Assert.NotNull(resultContext);
+            Assert.Equal(context.SpanId, resultContext.SpanId);
+            Assert.Equal(context.TraceId, resultContext.TraceId);
+            Assert.Equal(samplingPriority, resultContext.SamplingPriority);
+        }
+
+        [Theory]
+        [MemberData(nameof(GetHeaderCollectionImplementations))]
+        internal void WebRequest_InjectExtract_Identity(IHeadersCollection headers)
```
zacharycmontoya, line 79: This looks like a duplicate of `HttpRequestMessage_InjectExtract_Identity` minus the changes in the traceId, spanId, and samplingPriority. Perhaps this method can be removed or just renamed so it more clearly expresses the intent of the test

test/Datadog.Trace.Tests/Propagators/B3SpanContextPropagatorTests.cs
```diff
@@ -0,0 +1,201 @@
...
+        private B3SpanContextPropagator _propagator;
+
+        public B3SpanContextPropagatorTests()
+        {
+            _propagator = new B3SpanContextPropagator();
+        }
+
+        [Theory]
+        [MemberData(nameof(GetHeaderCollectionImplementations))]
+        internal void HttpRequestMessage_InjectExtract_Identity(IHeadersCollection headers)
```
zacharycmontoya, line 21: I recommend removing the initial `HttpRequestMessage` and `WebRequest` prefixes since the test methods are testing multiple implementations

zacharycmontoya (approved):
Looks good!

src/Datadog.Trace/Propagation/CommonHttpHeaderNames.cs
pellared: 1. @zacharycmontoya: ⚠️ this should be `x-datadog-tracing-enabled` for Datadog. make sure to change if this PR would be merged as it is

2. @pjanotti @nrcventura @zacharycmontoya : is it good that the "client" can disable tracing per a single request? I am not sure if having such a feature (especially by default) is good. Consider a bad actor who intentionally is disabled tracing to reduce the chances of being caught. Moreover, I have not found anything like that in https://github.com/open-telemetry/opentelemetry-specification 

I suggest reverting this change (to have `x-datadog-tracing-enabled` header) and create a dedicated GH issue to discuss the problem and address it separately.

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
pellared: @zacharycmontoya  https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/92#discussion_r600330255

src/Datadog.Trace/Propagation/CommonHttpHeaderNames.cs
RassK: This seems to be dummy header anyway. In case if this header is in place all traces are dropped. More meant to control that exporters are not causing instrumentation loop.
I can see a problem when you are using internal / contrib integrations privately.

zacharycmontoya: This was a custom built solution by Datadog used primarily to avoid instrumenting the http payloads to our agent. I think we should follow the suggestion by @pellared and revert the value for now to `x-datadog-tracing-enabled` and we can figure out that story in a separate issue

test/Datadog.Trace.Tests/Propagators/B3SpanContextPropagatorTests.cs
RassK: Great catch!

src/Datadog.Trace/Propagation/CommonHttpHeaderNames.cs
nrcventura: I agree with the recommendation from @zacharycmontoya. There are multiple ways to solve that problem, and we can consider them separately.

src/Datadog.Trace/Propagation/PropagationHelpers.cs
```diff
@@ -0,0 +1,45 @@
...
+using Datadog.Trace.Conventions;
+using Datadog.Trace.Logging;
+
+namespace Datadog.Trace.Propagation
+{
+    internal static class PropagationHelpers
+    {
+        public static TraceId ParseTraceId<T>(T carrier, Func<T, string, IEnumerable<string>> getter, string headerName, ITraceIdConvention traceIdConvention, IDatadogLogger logger)
+        {
+            var headerValues = getter(carrier, headerName).ToList();
```
nrcventura, line 13: Do we need the call to `ToList()` here?

src/Datadog.Trace.OpenTracing/HttpHeadersCodec.cs
RassK: Seems like refactoring error. I'll add back references to DD headers but to be more correct I would add OpenTracing propagator logic.

src/Datadog.Trace/Propagation/PropagationHelpers.cs
RassK: should be good with deferred execution as well - might gain actually some performance than just an allocation win.

nrcventura (approved):

PR #98, opened 2021-03-30
Configuration source instead of settings

dszmigielski:
The idea behind this change is that we want to be able to configure exporters (and possibly other components) without adding its configuration to TracerSettings. Instead new Settings class could be created for every component that requires it, that would read configuration from IConfigurationSource. In attempt to avoid changing Tracer API I have added IConfigurationSource as a public property to the TracerSettings class, so when other components are created inside Tracer they have easy access to the configuration and can have their own Setting classes created based on the IConfigurationSource.


RassK (approved):

src/Datadog.Trace/Agent/Jaeger/JaegerOptions.cs
```diff
@@ -7,6 +9,22 @@ public class JaegerOptions
     {
         internal const int DefaultMaxPayloadSizeInBytes = 4096;
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="JaegerOptions"/> class.
+        /// </summary>
+        /// <param name="settings"><see cref="TracerSettings"/> to read configuration from</param>
+        public JaegerOptions(TracerSettings settings = null)
```
pellared, line 16: 1. This is more like "conversion" than constructing a valid object.
I suggest changing it to:
  ```
  public static  JaegerOptions FromTracerSettings(TracerSettings settings)
  ```

2. However to be honest I do not see any reason to make such stuff public. What was wrong with having `private static JaegerOptions CreateJaegerOptions` in `Tracer`? It was used only there so I do not see any reason to extract it.

src/Datadog.Trace/Tracer.cs
```diff
@@ -225,7 +223,7 @@ public static Tracer Instance
         /// <summary>
         /// Gets the default service name for traces where a service name is not specified.
         /// </summary>
-        public string DefaultServiceName { get; }
+        public string DefaultServiceName => Settings.ServiceName;
```
pellared, line 226: nice 👍

src/Datadog.Trace/Agent/Jaeger/JaegerOptions.cs
dszmigielski: You are right in this case. I did this mostly to illustrate how I see creating new setting from now on. Jaeger one uses only the settings that are already extracted from IConfigurationSource by TracerSettings, but in other cases we would probably want to add new configuration keys, then we would use the passed settings object to get to the IConfigurationSource instance nad extract what we want (`var myNewKeyValue = settings.ConfigurationSource.GetString("MY_NEW_KEY");`). I would rather not see such a code in a static Tracer method, that's why I moved it to JaegerOptions.

pellared: >  we would probably want to add new configuration keys

Maybe we should do it? See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#otlp-exporter

nrcventura: I like the `static JaegerOptions FromTracerSettings(TracerSettings settings)` suggestion. It allows further separation of the logic for mapping settings to options which can make the test code involving options easier to manage. It also gets rid of the need for having the optional settings parameter (or multiple .ctors) so that test code doesn't also need to define settings code.

src/Datadog.Trace/Agent/Jaeger/JaegerOptions.cs
```diff
@@ -31,5 +34,24 @@ public class JaegerOptions
...
+        /// <summary>
+        /// Creates an instance of <see cref="JaegerOptions"/> based on passed <see cref="FromTracerSettings"/>.
+        /// </summary>
+        /// <param name="settings"><see cref="TracerSettings"/> to read configuration from</param>
+        /// <returns>New instance of <see cref="JaegerOptions"/></returns>
+        public static JaegerOptions FromTracerSettings(TracerSettings settings)
+        {
+            var agentHost = settings.ConfigurationSource?.GetString(ConfigurationKeys.JaegerExporterAgentHost) ?? "localhost";
+            var agentPort = settings.ConfigurationSource?.GetInt32(ConfigurationKeys.AgentPort) ?? 6831;
+            var agentUri = new Uri(settings.ConfigurationSource?.GetString(ConfigurationKeys.AgentUri) ?? $"http://{agentHost}:{agentPort}");
```
pellared, line 47: this is wrong.
see:
1. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#jaeger-exporter
2. port description https://www.jaegertracing.io/docs/1.22/getting-started/

we support only UDP

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -318,6 +318,31 @@ public static class ConfigurationKeys
...
+
+        /// <summary>
+        /// Configuration key for username to be used for HTTP basic authentication.
+        /// </summary>
+        public const string JaegerExporterUser = "OTEL_EXPORTER_JAEGER_USER";
+
+        /// <summary>
+        /// Configuration key for password to be used for HTTP basic authentication.
+        /// </summary>
+        public const string JaegerExporterPassword = "OTEL_EXPORTER_JAEGER_PASSWORD";
```
pellared, line 344: I think we are not supporting these setting so I think it is better to remove them

src/Datadog.Trace/Agent/Jaeger/JaegerOptions.cs
```diff
@@ -31,5 +34,24 @@ public class JaegerOptions
...
+
+            return new JaegerOptions
+            {
+                Host = agentUri.Host,
+                Port = agentUri.Port,
```
pellared, line 52: this should be:
```
                Host = agentHost,
                Port = agentPort,
```

src/Datadog.Trace/Agent/Jaeger/JaegerOptions.cs
pellared: resolve

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
pellared: resolve

src/Datadog.Trace/Agent/Jaeger/JaegerOptions.cs
pellared: resolve

src/Datadog.Trace/Agent/Jaeger/JaegerOptions.cs
pellared: resolve

src/Datadog.Trace/Tracer.cs
pellared: resolve 😄

pellared (approved):

nrcventura (approved):

pjanotti (approved):
LGTM

PR #99, opened 2021-03-31
Upstream sync 2021-03-31

pellared:
### Enable JetBrains Tools in the Benchmarks projects ([#1318](https://github.com/DataDog/dd-trace-dotnet/pull/1318))

commit cee651f3a473d353b795779d1539fb696afeb019 (datadog/master)
Author: Tony Redondo <tony.redondo@datadoghq.com>
Date:   Tue Mar 30 23:17:57 2021 +0200

* Enable JetBrains Tools

* Change comparison.

* Updates license file.

* fix license

* Update LICENSE-3rdparty.csv

Co-authored-by: Lucas Pimentel-Ordyna <lucas.pimentel@datadoghq.com>

Co-authored-by: Lucas Pimentel-Ordyna <lucas.pimentel@datadoghq.com>

### Update ASP.NET / MVC / WebApi2 Resource Names ([#1288](https://github.com/DataDog/dd-trace-dotnet/pull/1288))

commit 8aa7722a52809e0cee8fbc4cebe868297c3a1f6a
Author: Andrew Lock <andrew.lock@datadoghq.com>
Date:   Tue Mar 30 08:49:51 2021 +0100

* Update ASP.NET MVC span resource names

Use the route template as the resource name

* Update the ASP.NET parent span resource name to be the same as the MVC span

* Add automatic trimming of optional parameters that aren't provided

Only downside of this (other than performance), is that routes where the optional parameter is provided have different resource names compared to routes where the optional parameter is ommitted. However this is current behaviour anyway, so probably not a problem

* Add feature flag for aspnet resource names

* Propagate WebApi2 resource name to root span

The static HttpContext was the only way I could find to do this, as the ControllerContext doesn't expose it

* Update WebApi2 resource names to start with a /

* Add ASP.NET integration tests for new feature flag

* Optimization: Don't set resource name of span until after the request has completed.

No point cleaning the path in general, as in most cases it will be completely replaced

* Refactor UrlHelpers

Remove unused trivial overload of GetRelativeUrl only called by other overload
Renamed remaining method to GetCleanUriPath() as that's what it really does, and was only called with tryRemoveIds set to true
Added tests for behaviour

* Fix incorrect comments

* Remove unused parameter

* Remove cases that fail on linux (as they're testing Uri implementation, not our code)

* Fix comment

Co-authored-by: Lucas Pimentel-Ordyna <lucas.pimentel@datadoghq.com>

* Rename CleanUriSegment->GetCleanUriPath

* Fix feature flag

[...]

PR #100, opened 2021-03-31
Reusing context agnostic propagator for service fabric

RassK:
### **Background**

Add propagation switch and B3 ([#92](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/92)) introduced context agnostic propagation logic. This change is rafactoring service fabric integration and bases propagation to already existing propagators.


dszmigielski (approved):
A few files of code replaced by a few lines of code? I'm in 👍

pellared:
I think this PR would conflict with #99 
Lets sync with upstream first

RassK:
> I think this PR would conflict with #99
> Lets sync with upstream first

Seems like this PR is not needed any more

PR #102, opened 2021-04-01
Add W3C propagator

dszmigielski:
This PR introduces W3C propagator. One thing to have in mind is that this feature will work only for app configured to use OpenTelemetry TraceId convention. Because of that there is a need for an iffy if in ContextPropagatorBuilder. Please have a closer look at that when reviewing, maybe someone will come up with a better idea how to handle that.


src/Datadog.Trace/Propagation/ContextPropagatorBuilder.cs
```diff
@@ -7,22 +7,31 @@ namespace Datadog.Trace.Propagation
...
             };
 
         public static IPropagator BuildPropagator(PropagatorType propagator, ITraceIdConvention traceIdConvention)
         {
-            if (_propagatorSelector.TryGetValue(propagator, out Func<ITraceIdConvention, IPropagator> getter))
+            if (PropagatorSelector.TryGetValue(propagator, out Func<ITraceIdConvention, IPropagator> getter))
             {
-                return getter(traceIdConvention);
+                // W3C propagator requires Otel TraceId convention as it's specification clearly states lengths of traceId and spanId values in the header.
+                return IsW3CButNotOtelTraceId(propagator, traceIdConvention)
```
RassK, line 24: I would throw an unsupported exception here. Cause if config is asking for W3C and gets default, it might cause more misunderstanding.

dszmigielski: Well, that actually sounds reasonable 👍

src/Datadog.Trace/Tracer.cs
```diff
@@ -119,6 +119,12 @@ internal Tracer(TracerSettings settings, ITraceWriter traceWriter, ISampler samp
             _scopeManager = scopeManager ?? new AsyncLocalScopeManager();
             Sampler = sampler ?? new RuleBasedSampler(new RateLimiter(Settings.MaxTracesSubmittedPerSecond));
 
+            // W3C propagator requires Otel TraceId convention as it's specification clearly states lengths of traceId and spanId values in the header.
+            if (Settings.Propagator == PropagatorType.W3C && Settings.Convention != ConventionType.OpenTelemetry)
```
RassK, line 123: Shouldn't it be inside ContextPropagatorBuilder?  So the builder takes care of validation and building and logic remains visible for unit tests as well.

like that >>
```
if (propagator == PropagatorType.W3C && traceIdConvention is not OtelTraceIdConvention)
...
```

nrcventura (commented):
Even though OTel doesn't have any additional data to add to the `tracestate` header, should we continue to propagate the `tracestate` header value that we initially received to provide better interoperability with systems that are not fully monitored by OTel?

src/Datadog.Trace/Propagation/W3CSpanContextPropagator.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            if (carrier == null)
+            {
+                throw new ArgumentNullException(nameof(carrier));
+            }
+
+            if (setter == null)
+            {
+                throw new ArgumentNullException(nameof(setter));
+            }
```
pellared, line 42: can you "collate" it into 3 lines, please 😄

src/Datadog.Trace/Propagation/W3CSpanContextPropagator.cs
```diff
@@ -0,0 +1,94 @@
...
+            var traceIdString = traceParentHeader.Substring(VersionPrefixIdLength, TraceIdLength);
+            var traceId = _traceIdConvention.CreateFromString(traceIdString);
+            if (traceId == TraceId.Zero)
+            {
+                Log.Warning("Could not parse {HeaderName} headers: {HeaderValues}", W3CHeaderNames.TraceParent, string.Join(",", traceParentValues));
+                return null;
+            }
+
+            var spanIdString = traceParentHeader.Substring(VersionAndTraceIdLength, SpanIdLength);
+            if (ulong.TryParse(spanIdString, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out var spanId))
```
pellared, line 85: I propose to invert the if statement so that in "main" flow we have the "positive" scenario

src/Datadog.Trace/SpanContext.cs
```diff
@@ -100,6 +116,11 @@ private SpanContext(TraceId? traceId, string serviceName)
         /// </summary>
         internal string Origin { get; }
 
+        /// <summary>
+        /// Gets the trace state for W3C propagation
+        /// </summary>
+        internal string TraceState { get; }
```
pellared, line 122: this is not following the OTEL model: https://github.com/open-telemetry/opentelemetry-specification/blob/cf59c0b868744b764ca8b9175668abc6755fecca/specification/trace/api.md#spancontext
See an example implementation here: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/Trace/SpanContext.cs

src/Datadog.Trace/SpanContext.cs
```diff
@@ -100,6 +116,11 @@ private SpanContext(TraceId? traceId, string serviceName)
         /// </summary>
         internal string Origin { get; }
 
+        /// <summary>
+        /// Gets the trace state for W3C propagation
```
pellared, line 120: Maybe something like:
```
TraceState carries vendor-specific trace identification data, represented as a list of key-value pairs. TraceState allows multiple tracing systems to participate in the same trace. It is fully described in the W3C Trace Context specification: https://www.w3.org/TR/trace-context/#tracestate-header.
```

src/Datadog.Trace/Propagation/W3CSpanContextPropagator.cs
dszmigielski: I know this is just a matter of styling, but I hate one line ifs. I would rather leave it as it is if we don't have strict rules on such things.

pellared: TBH I see no value for null checks for internals :-)

src/Datadog.Trace/Propagation/ContextPropagatorBuilder.cs
```diff
@@ -7,18 +7,25 @@ namespace Datadog.Trace.Propagation
...
 
         public static IPropagator BuildPropagator(PropagatorType propagator, ITraceIdConvention traceIdConvention)
         {
-            if (_propagatorSelector.TryGetValue(propagator, out Func<ITraceIdConvention, IPropagator> getter))
+            if (PropagatorSelector.TryGetValue(propagator, out Func<ITraceIdConvention, IPropagator> getter))
             {
+                // W3C propagator requires Otel TraceId convention as it's specification clearly states lengths of traceId and spanId values in the header.
+                if (propagator == PropagatorType.W3C && traceIdConvention is not OtelTraceIdConvention)
+                {
+                    throw new NotSupportedException($"'{PropagatorType.W3C}' propagator requires '{ConventionType.OpenTelemetry}' convention to be set");
```
pellared, line 26: @pjanotti @zacharycmontoya @nrcventura 
I am not sure if we should throw an exception and crash the instrumented application in case of improper configuration. As far as I see in all other places there is a fallback to defaults in case of improper value.

Besides, I think we should have some docs to describe the strategies for error handling e.g. when we should log vs crash etc.

src/Datadog.Trace/Propagation/W3CSpanContextPropagator.cs
dszmigielski: Yep, I did it to be consistent with other propagators. TBH if nobody opposes I would love to get rid of that :)

src/Datadog.Trace/Propagation/ContextPropagatorBuilder.cs
RassK: For me it seems very confusing if the application does not respect the values and starts behaving differently.
1. Either fail fast (+ log the reason of failure)
2. Settings should be consistent for the whole system. So instead of creating default propagator, the setting should be reverting back to the default value (+ log the reason). Otherwise other parts of the system will still see the invalid value but don't know about the actual behavior.  Which means it could lead to weird bugs.

dszmigielski: I would go with the 1. here, but after giving it a second thought that would make any .NET app crash on the machine which can be disturbing.

pellared: In general, I am also in favour of 1.
However, maybe there are reasons that we should not go that way. Especially that currently it does not work that way if I follow the code correctly.

RassK: @dszmigielski I think it's very important point - already forgot that the default behavior is to instruct **every** dotnet application. 

Wondering why the internals of tracer are not protected? If there is a fatal failure it should put the tracer to offline mode (+ log the reason) but leave the original application alive.

iNikem: First, please consult https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md

E.g. in Java most of the (auto-)configuration is done by SDK, which throws an exception and fails to start if configuration is invalid (e.g. user has requested an unknown propagator via env variable)

src/Datadog.Trace/SpanContext.cs
dszmigielski: I have redone how this is parsed.

src/Datadog.Trace/Propagation/ContextPropagatorBuilder.cs
dszmigielski: I think we fall under this:
The API or SDK may fail fast and cause the application to fail on initialization, e.g. because of a bad user config or environment, but MUST NOT cause the application to fail later at run time, e.g. due to dynamic config settings received from the Collector.

RassK: This time it seems ok, but I think using the setting value equals already "later at run time". Using the fail fast model, all of the settings (currently only TracerSettings instance) must be valid max at the end of binding. Everything else is too riski, cause if the setting value is used after initialization, its already "later at run time".

pellared: > This time it seems ok

TBH now I am not sure. We would need to check what happens if we attach the profiler e.g. to an app that only prints `Hello world` to the console.

src/Datadog.Trace/Propagation/W3CSpanContextPropagator.cs
```diff
@@ -0,0 +1,270 @@
...
+            setter(carrier, W3CHeaderNames.TraceParent, string.Format(TraceParentFormat, context.TraceId.ToString(), context.SpanId.ToString("x16")));
+            if (!string.IsNullOrEmpty(context.TraceState))
+            {
+                setter(carrier, W3CHeaderNames.TraceState, context.TraceState);
+            }
+        }
+
+        public SpanContext Extract<T>(T carrier, Func<T, string, IEnumerable<string>> getter)
+        {
+            var traceStateCollection = getter(carrier, W3CHeaderNames.TraceState);
```
zacharycmontoya, line 49: It maybe slightly more efficient to read the tracestate right before returning the final `SpanContext` so you don't spend time reading headers you may not use.

src/Datadog.Trace/Propagation/W3CSpanContextPropagator.cs
```diff
@@ -0,0 +1,270 @@
...
+            var traceState = ExtractTraceState(traceStateCollection);
+
+            var traceParentValues = getter(carrier, W3CHeaderNames.TraceParent).ToList();
```
zacharycmontoya, line 52: In the SDK, they get the `Count()` of the resulting enumerable and then access the traceparent by getting the `First()` of the resulting enumerable. That seems slightly more efficient so we don't allocate a new List when we don't really need to

src/Datadog.Trace/Propagation/ContextPropagatorBuilder.cs
zacharycmontoya: cc @lucaspimentel 

In this case, it makes sense to bubble up an exception through the Tracer constructor because some combination of configuration was invalid. However, I don't think we account for this currently. We currently use a lazy initializer and get the default `Tracer` instance: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/57919f03c92c9ccb0917b0fa0c964b85f989dca7/src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs#L63 . BTW this is called at startup when the IL rewriting aggressively finds the first JIT-compiled method and makes a call to this: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/57919f03c92c9ccb0917b0fa0c964b85f989dca7/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.cs#L42

Inside `Instrumentation.Initialize` would be a good place to
1. Catch the exception from the constructor
2. Log the exception from the constructor
3. Maybe set `Tracer.Instance = new Tracer(new TracerSettings { TraceEnabled = false } );` so it can't do anymore harm, even though the IL rewriting will still occur?

pellared: ad. 3
I think it would be better to print also the "log" to the console and close the application. Otherwise, the user may not be aware that the auto-instrumentation is not working.

src/Datadog.Trace/Propagation/W3CSpanContextPropagator.cs
dszmigielski: Hmm, this could cause multiple enumeration, so I would rather not do that.

src/Datadog.Trace/Propagation/ContextPropagatorBuilder.cs
dszmigielski: At the moment this will not fail until the first instrumentation attempt is made, so it's no near perfect. Even though I would close this PR and create an issue for fixing this behaviour separately.

pellared (approved):

pellared:
Just make GH issue for leftovers, please.

src/Datadog.Trace/Propagation/W3CSpanContextPropagator.cs
RassK: I think Zach's point is valid. `string.Join(",", traceParentValues) ` could be just `traceParentHeader`, cause `traceParentValues.Count != 1` will not let you past anyway. So you are left with `Count()` and `First()` operations - which means full list will be never needed.

RassK: ... or even better, `FirstOrDefault()` single enumerable operation.

src/Datadog.Trace/Propagation/ContextPropagatorBuilder.cs
zacharycmontoya: Yes this issue can be handled separately, and we can discuss what the side effect of bad configuration is. At Datadog, we have a strong opinion that we should absolutely minimize changes to the application's behavior when attaching the tracer, so in our tracer we would not choose to stop the application because of bad configuration.

zacharycmontoya (approved):
LGTM, some of the validation / error handling can be addressed separately

src/Datadog.Trace/Propagation/W3CSpanContextPropagator.cs
dszmigielski: The FirstOrDefault would be cool but then we have no way of checking if there was more than one value, and for now in this case we return null SpanContext.

RassK: sorry, I meant `SingleOrDefault()`, my bad.

RassK (approved):

pjanotti (approved):

nrcventura (approved):

PR #104, opened 2021-04-05
Allow multiple context propagators

RassK:
### **Background**
This PR adds support for multiple propagators that's required by the OTel standard. (following [comment](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/92#pullrequestreview-619525899) in PR #92)

### **What's next**
Configuration source needs to be validated in early startup time. After that point settings instance should be considered as guaranteed OK (no throwing later in run time).  See comments:
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/102#pullrequestreview-628734402
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/104#pullrequestreview-629247977
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/104#pullrequestreview-629248164


src/Datadog.Trace/Tracer.cs
```diff
@@ -119,7 +119,8 @@ internal Tracer(TracerSettings settings, ITraceWriter traceWriter, ISampler samp
             _scopeManager = scopeManager ?? new AsyncLocalScopeManager();
             Sampler = sampler ?? new RuleBasedSampler(new RateLimiter(Settings.MaxTracesSubmittedPerSecond));
 
-            _propagator = ContextPropagatorBuilder.BuildPropagator(Settings.Propagator, TraceIdConvention);
+            var propagators = ContextPropagatorBuilder.BuildPropagators(Settings.Propagators, TraceIdConvention);
```
RassK, line 122: Please note the upcoming configuration issue here. PropagatorType should be matched with TraceIdConvention (eg: we can go with name:value based configuration key), cause single convention type could not be accepted for all of the propagators. (See PR #102 [comment](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/102#pullrequestreview-626175666))

src/Datadog.Trace/Propagation/MultiplexSpanContextPropagator.cs
```diff
@@ -0,0 +1,40 @@
+using System;
+using System.Collections.Generic;
+
+namespace Datadog.Trace.Propagation
+{
+    internal class MultiplexSpanContextPropagator : IPropagator
```
nrcventura, line 6: We may want to use the name `CompositeTextMapPropagator` for consistency with the OTel .NET SDK.

src/Datadog.Trace/Tracer.cs
pellared: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/102#discussion_r607669452

src/Datadog.Trace/Configuration/ConfigurationSourceExtensions.cs
```diff
@@ -0,0 +1,57 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Datadog.Trace.Configuration
+{
+    internal static class ConfigurationSourceExtensions
+    {
+        private const char Splitter = ';';
```
pellared, line 9: How about `Separator` ?

src/Datadog.Trace/Configuration/ConfigurationSourceExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        public static TEnum GetTypedValue<TEnum>(this IConfigurationSource source, string key)
+            where TEnum : struct, IConvertible
+        {
+            Enum.TryParse(source?.GetString(key) ?? "default", ignoreCase: true, out TEnum typedValue);
+            return typedValue;
+        }
+
+        public static HashSet<TEnum> GetTypedValues<TEnum>(this IConfigurationSource source, string key)
+            where TEnum : struct, IConvertible
```
pellared, line 26:
```suggestion
            where TEnum : Enum
```

More: https://stackoverflow.com/a/8086788/859275

src/Datadog.Trace/Configuration/ConfigurationSourceExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        public static TEnum GetTypedValue<TEnum>(this IConfigurationSource source, string key)
+            where TEnum : struct, IConvertible
+        {
+            Enum.TryParse(source?.GetString(key) ?? "default", ignoreCase: true, out TEnum typedValue);
+            return typedValue;
+        }
+
+        public static HashSet<TEnum> GetTypedValues<TEnum>(this IConfigurationSource source, string key)
```
pellared, line 25: Why returning a `HashSet` here? It makes the code not "orthogonal" as e.g. `GetStrings` returns `IEnumerable`.
the caller can always convert to HashSet. 

Also I do not like the undocumented side-effect of setting the `default`. Especially that the default depends on the context.
See:
> Platforms such as ASP.NET may pre-configure out-of-the-box propagators. If pre-configured, Propagators SHOULD default to a composite Propagator containing the W3C Trace Context Propagator and the Baggage Propagator specified in the Baggage API. These platforms MUST also allow pre-configured propagators to be disabled or overridden.

From: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#global-propagators

src/Datadog.Trace/Configuration/ConfigurationSourceExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            return source?.GetString(key)
+                ?.Split(new[] { Splitter }, StringSplitOptions.RemoveEmptyEntries)
+                ?? Enumerable.Empty<string>();
+        }
+
+        public static TEnum GetTypedValue<TEnum>(this IConfigurationSource source, string key)
+            where TEnum : struct, IConvertible
+        {
+            Enum.TryParse(source?.GetString(key) ?? "default", ignoreCase: true, out TEnum typedValue);
```
pellared, line 21: Not sure if we want/need the `"default"` here. The zero-values of enums look fine. Is it more performant (I do not think so(?

src/Datadog.Trace/Propagation/MultiplexSpanContextPropagator.cs
pellared: Also in specs https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#composite-propagator

I think we should also start adding doc comments especially to types like this one.

src/Datadog.Trace/Propagation/ContextPropagatorBuilder.cs
```diff
@@ -15,14 +16,19 @@ internal static class ContextPropagatorBuilder
                 { PropagatorType.Default, (convention) => new DDSpanContextPropagator(convention) },
             };
 
-        public static IPropagator BuildPropagator(PropagatorType propagator, ITraceIdConvention traceIdConvention)
+        public static ICollection<IPropagator> BuildPropagators(IEnumerable<PropagatorType> propagatorTypes, ITraceIdConvention traceIdConvention)
         {
-            if (_propagatorSelector.TryGetValue(propagator, out Func<ITraceIdConvention, IPropagator> getter))
+            return propagatorTypes.Select(type => BuildPropagator(type, traceIdConvention)).ToList();
```
pellared, line 21: Any reason why `ToList()` and not `ToArray()` ?

src/Datadog.Trace/Propagation/ContextPropagatorBuilder.cs
```diff
@@ -15,14 +16,19 @@ internal static class ContextPropagatorBuilder
                 { PropagatorType.Default, (convention) => new DDSpanContextPropagator(convention) },
             };
 
-        public static IPropagator BuildPropagator(PropagatorType propagator, ITraceIdConvention traceIdConvention)
+        public static ICollection<IPropagator> BuildPropagators(IEnumerable<PropagatorType> propagatorTypes, ITraceIdConvention traceIdConvention)
```
pellared, line 19: This class is IMO badly named. It should be a `Factory` not a `Builder`.

src/Datadog.Trace/Propagation/ContextPropagatorBuilder.cs
```diff
@@ -15,14 +16,19 @@ internal static class ContextPropagatorBuilder
...
         {
-            if (_propagatorSelector.TryGetValue(propagator, out Func<ITraceIdConvention, IPropagator> getter))
+            return propagatorTypes.Select(type => BuildPropagator(type, traceIdConvention)).ToList();
+        }
+
+        public static IPropagator BuildPropagator(PropagatorType propagatorType, ITraceIdConvention traceIdConvention)
```
pellared, line 24: Should it not be changed to `private`?

src/Datadog.Trace/Propagation/MultiplexSpanContextPropagator.cs
```diff
@@ -0,0 +1,40 @@
...
+
+                // if context is missing, propagator wasn't capable to extract
+                if (context != null)
+                {
+                    return context;
+                }
+            }
+
+            // non of the propagators found specific headers
+            return null;
```
pellared, line 29: I think this is incorrect. It will return the first found context. It should aggregate them somehow:

```
            SpanContext context = null;
            foreach (IPropagator propagator in _propagators)
            {
                var extractedContext = propagator.Extract(carrier, getter);
                context = Aggregate(context, extractedContext);
            }
            return context;
```

src/Datadog.Trace/Configuration/ConfigurationSourceExtensions.cs
RassK: wow, they finally did it, thanks for the info!

src/Datadog.Trace/Configuration/ConfigurationSourceExtensions.cs
RassK: Originally just moved this part. When I'm looking into enum parsing source, the null value is more performant than invalid value. Will fix it.

src/Datadog.Trace/Propagation/ContextPropagatorBuilder.cs
RassK: Not sure there is much difference. In nano world over the time one used to be "better" than another and vice versa.
https://stackoverflow.com/questions/1105990/is-it-better-to-call-tolist-or-toarray-in-linq-queries

src/Datadog.Trace/Propagation/ContextPropagatorBuilder.cs
RassK: I would keep it as public api. In future some vendors might choose not to use composite propagation as default.

src/Datadog.Trace/Propagation/MultiplexSpanContextPropagator.cs
iNikem: You cannot aggregate SpanContext. You can just pick either the first or the last one. First seems more logical and performant.

iNikem: Also see https://github.com/open-telemetry/opentelemetry-specification/issues/496

src/Datadog.Trace/Configuration/ConfigurationSourceExtensions.cs
RassK: agree, seems like design error.

src/Datadog.Trace/Configuration/ConfigurationSourceExtensions.cs
RassK: Unfortunately TryParse doesn't use it yet. I'll keep it as is - currently the solution seems more of using KISS.

src/Datadog.Trace/Propagation/ContextPropagatorBuilder.cs
iNikem: Just a note: it is easy to change private to public at any time. It is almost impossible to change from public to private, because somebody may already use it. So YAGNI, unless actually required, make it private.

RassK: seems private it is

src/Datadog.Trace/Tracer.cs
RassK: Current result: no support for different TraceIds in same time

src/Datadog.Trace/Configuration/ConfigurationSourceExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        public static TEnum GetTypedValue<TEnum>(this IConfigurationSource source, string key)
+            where TEnum : struct, IConvertible
+        {
+            Enum.TryParse(source?.GetString(key), ignoreCase: true, out TEnum typedValue);
+            return typedValue;
+        }
+
+        public static IEnumerable<TEnum> GetTypedValues<TEnum>(this IConfigurationSource source, string key)
```
pellared, line 25: how we would know that an invalid value was provided. [more context](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/102#discussion_r607774012)

src/Datadog.Trace/Configuration/ConfigurationSourceExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+        private const char Separator = ';';
+
+        public static IEnumerable<string> GetStrings(this IConfigurationSource source, string key)
+        {
+            return source?.GetString(key)
+                ?.Split(new[] { Separator }, StringSplitOptions.RemoveEmptyEntries)
+                ?? Enumerable.Empty<string>();
+        }
+
+        public static TEnum GetTypedValue<TEnum>(this IConfigurationSource source, string key)
```
pellared, line 18: how we would know that an invalid value was provided. [more context](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/102#discussion_r607774012)

src/Datadog.Trace/Propagation/ContextPropagatorBuilder.cs
pellared: resolve

src/Datadog.Trace/Configuration/ConfigurationSourceExtensions.cs
pellared: resolve

src/Datadog.Trace/Propagation/CompositeTextMapPropagator.cs
```diff
@@ -0,0 +1,44 @@
+using System;
+using System.Collections.Generic;
+
+namespace Datadog.Trace.Propagation
+{
+    /// <summary>
+    /// Composite Propagator
+    /// See more details <see href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#composite-propagator">here</see>
```
pellared, line 8: `.` missing at the end

src/Datadog.Trace/Propagation/ContextPropagatorFactory.cs
```diff
@@ -15,14 +16,19 @@ internal static class ContextPropagatorBuilder
...
+
+        private static IPropagator BuildPropagator(PropagatorType propagatorType, ITraceIdConvention traceIdConvention)
+        {
+            if (_propagatorSelector.TryGetValue(propagatorType, out Func<ITraceIdConvention, IPropagator> getter))
             {
                 return getter(traceIdConvention);
             }
 
-            throw new InvalidOperationException($"There is no propagator registered for type '{propagator}'");
+            throw new InvalidOperationException($"There is no propagator registered for type '{propagatorType}'");
```
pellared, line 31: `.` missing at the end of the messsage

test/Datadog.Trace.Tests/Propagators/B3SpanContextPropagatorTests.cs
```diff
@@ -185,18 +184,5 @@ private static IHeadersCollection InjectContext(IHeadersCollection headers, stri
...
-            var matches = headers.GetValues(key);
-            Assert.Single(matches);
-            matches.ToList().ForEach(x => Assert.Equal(expected, x));
-        }
-
-        private static void AssertMissing(IHeadersCollection headers, string key)
-        {
-            var matches = headers.GetValues(key);
-            Assert.Empty(matches);
-        }
```
pellared, line 200: why these methods have been moved?

RassK: Reused for CompositeTextMapPropagatorTests, HeadersCollectionTestBase seems to be ideal for tests that deal with headers.

src/Datadog.Trace/Configuration/ConfigurationSourceExtensions.cs
RassK: I'm not sure this is the right place to deal with validation. As mentioned in #102 it should fail fast. I would propose configuration source validator in early startup (binding configuration to settings instance). After that point configuration should be valid for whole application.

test/Datadog.Trace.Tests/Propagators/B3SpanContextPropagatorTests.cs
pellared: I see. Only `AssertMissing` is not reused there but I think we can keep it as you refactored 👍

src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -56,16 +56,10 @@ public TracerSettings(IConfigurationSource source)
                 TraceEnabled = false;
             }
 
-            var disabledIntegrationNames = source?.GetString(ConfigurationKeys.DisabledIntegrations)
-                                                 ?.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries) ??
-                                           Enumerable.Empty<string>();
-
+            var disabledIntegrationNames = source.GetStrings(ConfigurationKeys.DisabledIntegrations);
             DisabledIntegrationNames = new HashSet<string>(disabledIntegrationNames, StringComparer.OrdinalIgnoreCase);
```
zacharycmontoya, line 60: You could make these into one-liners! 🎉

pellared (approved):
Approving. It will be good to make some follow-up on https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/104/files#r608056476

src/Datadog.Trace/Configuration/TracerSettings.cs
RassK: one-lined! 😄

test/Datadog.Trace.Tests/Propagators/B3SpanContextPropagatorTests.cs
RassK: Yep, both seemed to be generic / reusable. Seems bit weird to leave one out 😄

test/Datadog.Trace.TestHelpers/HeadersCollectionTestBase.cs
```diff
@@ -29,5 +30,18 @@ public static IEnumerable<object[]> GetHeadersInvalidSamplingPrioritiesCartesian
                    from invalidSamplingPriority in HeadersCollectionTestHelpers.GetInvalidSamplingPriorities().SelectMany(i => i)
                    select new[] { header, invalidSamplingPriority };
         }
+
+        internal static void AssertExpected(IHeadersCollection headers, string key, string expected)
```
zacharycmontoya, line 34: You could make this public so you don't have to add an `InternalsVisibleTo` attribute. This won't be consumed by developers so I don't see any reason to keep this internal

src/Datadog.Trace/Propagation/CompositeTextMapPropagator.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Datadog.Trace.Propagation
+{
+    /// <summary>
+    /// Composite Propagator.
+    /// See more details <see href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#composite-propagator">here</see>.
+    /// </summary>
+    internal class CompositeTextMapPropagator : IPropagator
+    {
+        private readonly ICollection<IPropagator> _propagators;
```
zacharycmontoya, line 12: Could this be an `IEnumerable<IPropagator>` instead? It looks like this is only used to enumerate values and doesn't do insertions/removals

src/Datadog.Trace/Configuration/ConfigurationSourceExtensions.cs
zacharycmontoya: For now, would you consider making this use `Enum.Parse` and throw an exception so we know that the result is valid? I'd prefer that this PR close with a clear indication of configuration errors, so this change might be easier than doing a configuration source validator right now

test/Datadog.Trace.TestHelpers/HeadersCollectionTestBase.cs
RassK: IHeadersCollection is also internal, that's the original reason.

src/Datadog.Trace/Propagation/CompositeTextMapPropagator.cs
RassK: I usually like to use ICollection when list is actually materialized - so it's quickly visible it's safe to use. With IEnumerable you are also introducing deferred execution, it can easily backfire with context like this.

RassK: In perfect world I would use IReadOnlyCollection, but here it doesn't seem to be a pattern.

src/Datadog.Trace/Configuration/ConfigurationSourceExtensions.cs
RassK: Originally I tried not to change parsing logic with this PR (notice the previous logic is 1:1 same as now) but I can easily change it there is interest.

pellared: I suggest doing it separately

src/Datadog.Trace/Propagation/CompositeTextMapPropagator.cs
pellared: I agree with @RassK . I used to IReadOnlyCollection for such cases

nrcventura (approved):

pjanotti (approved):

zacharycmontoya (approved):

zacharycmontoya:
Unit tests failed because of flaky test: `Datadog.Trace.Tests.SpanTests.Accurate_Duration`. Datadog just merged https://github.com/DataDog/dd-trace-dotnet/pull/1360 that hopefully resolves this permanently

zacharycmontoya:
Since the test failure is known to be flaky, I'll go ahead and merge

Issue #117, opened 2021-04-07
Extendibility via plugins


RassK:
### **Why**

Distro providers should have more simpler ways to build up their own distro. Full fork is hard to manage and keep in sync.

### **What**
Introduce plugin architecture for extendibility. Plugins could reside in `/plugins ` folder where `dll`s can be dynamically loaded. 

This feature is connected to: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/118

pellared:
We need to agree on what extension points are needed.

For Splunk, we need to be able to add additional (or modify existing) HTTP response headers for HTTP Inbound instrumentation.

Maybe we should start with this one?

pellared:
Instead of

> Plugins could reside in /plugins folder where dlls can be dynamically loaded.

I would prefer to add some env var like `OTEL_DOTNET_INSTR_PLUGIN`. Requiring to explicitly set it is more secure. It also gives the user to use different plugins for different application. It also makes it possible to harden the location where the plugin resides (e.g. aggressively restrict access to it as potential place where a malicious user could inject some sniffer).

pellared:
For reference: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/566

PR #129, opened 2021-04-15
Extensibility prototype

RassK:
### **Background**

Resolves https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/117

Plugins.json (default name) contains configuration for loadable plugin paths. Tried to use `GlobalSettings` to store plugin configuration and `PluginManager` to encapsulate initialization logic.

Requirements:
- No dependencies

Examples:
- Propagators

What's next:
- Integrations tests for IOTelPlugin
- Lookover auto configuration (https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/129#pullrequestreview-647515612)
- Security / Integrity checks for plugins (optional)


src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
```diff
@@ -59,6 +61,8 @@ public static void Initialize()
 
             try
             {
+                TryLoadVendorPlugin();
+
                 // ensure global instance is created if it's not already
```
RassK, line 66: One thing that worries me is that if in this sentence. This should be the first bootstrapping point...

src/Datadog.Trace/Configuration/GlobalSettings.cs
```diff
@@ -184,5 +223,21 @@ private static bool TryLoadHostingEnvironmentPath(out string hostingPath)
             hostingPath = default;
             return false;
         }
+
+        private static void LoadVendorFactoryConfigurator()
+        {
+            if (Source.VendorPluginLoaded)
```
RassK, line 229: Overleft `if`, will be fixed later.

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
```diff
@@ -126,5 +130,37 @@ private static void StartDiagnosticManager()
...
 #endif
+
+        private static void TryLoadVendorPlugin()
+        {
+            string pluginName = GlobalSettings.Source.VendorPluginName;
+            if (!string.IsNullOrWhiteSpace(pluginName))
+            {
+                string pluginPath = Path.Combine(
+                    Path.GetDirectoryName(typeof(Instrumentation).Assembly.Location),
+                    pluginName);
```
pellared, line 141: I feel that specifying full path more flexible and can be safer at the same time.

I think it should locate files in the same way as `OTEL_DOTNET_TRACER_HOME` and `OTEL_INTEGRATIONS`.

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -336,6 +336,11 @@ public static class ConfigurationKeys
         /// </summary>
         public const string JaegerExporterAgentPort = "OTEL_EXPORTER_JAEGER_AGENT_PORT";
 
+        /// <summary>
+        /// Configuration key for specifing vendor plugin path
+        /// </summary>
+        public const string VendorPluginName = "OTEL_VENDOR_PLUGIN";
```
pellared, line 342: Open question. Is it desired to allow multiple plugins? Especially that https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/117 tells about "plugins". I think it would be safer to think about multiple plugins even though it complicates the design. Especially taking into consideration that the plugins should EXTEND and not REPLACE the core functionality (at least in my understanding). I think that all extension points should work like enumerators/decorators/middlewares.

pellared, line 342: 1. `VENDOR` prefix does not make sense. Maybe someone would create some common, reusable plugins. 
2. We should agree on some .NET AutoInstr specific prefix. E.g. `OTEL_DOTNET_TRACER`. We do not want to use the same env vars for plugins for e.g. .NET and Java 😉 

I propose to rename this env var to: `OTEL_DOTNET_TRACER_PLUGINS`

src/Datadog.Trace/Configuration/GlobalSettings.cs
```diff
@@ -184,5 +223,21 @@ private static bool TryLoadHostingEnvironmentPath(out string hostingPath)
...
+            {
+                var factoryInterface = typeof(IFactoryConfigurator);
+                var factoryType = Source.VendorPlugin
+                    .GetTypes()
+                    .FirstOrDefault(x => factoryInterface.IsAssignableFrom(x));
+
+                if (factoryType != null)
+                {
+                    Source.FactoryConfigurator = (IFactoryConfigurator)Activator.CreateInstance(factoryType);
```
pellared, line 238: I think that, as a rule of thumb, plugins should EXTEND and not REPLACE the core functionality. Therefore, whenever possible, plugins should be registered in a form of enumerators/decorators/middlewares. Here it is for sure possible. E.g. we may want to register multiple plugins where each adds an additional propagator. E.g. for Datadog, AWS and OpenCensus. More: https://github.com/open-telemetry/opentelemetry-specification/blob/3f07b6669dbbf4625579b5a0f371768f50788cd2/specification/context/api-propagators.md#propagators-distribution

src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -356,7 +357,7 @@ public TracerSettings(IConfigurationSource source)
         /// Default is <c>Datadog</c>
         /// <seealso cref="ConfigurationKeys.Propagators"/>
         /// </summary>
-        public HashSet<PropagatorType> Propagators { get; set; }
+        public HashSet<string> Propagators { get; set; }
```
pellared, line 360: I feel that `PropagatorType` can be preserved here and plugins could have their own enums. We just need to give plugins a way to parse the tracer configuration on their own.

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
pellared: But it is not. You have to put the `TryLoadVendorPlugin();` in `Tracer` constructor AFAIK...

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
```diff
@@ -126,5 +130,37 @@ private static void StartDiagnosticManager()
...
+                {
+                    try
+                    {
+                        Assembly pluginAssembly = Assembly.LoadFrom(pluginPath);
+
+                        GlobalSettings.SetVendorPlugin(pluginAssembly);
+
+                        Log.Information("Vendor plugin assembly loaded '{0}'.", pluginAssembly.FullName);
+                    }
+                    catch (Exception ex)
```
pellared, line 153: do not catch general exceptions like that. What if we catch a runtime exception like `OutOfMemoryException`? You can check all "normal" exceptions which `Assembly.LoadFrom` can throw https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfrom?view=net-5.0 (check in all supported frameworks).

src/Datadog.Trace/Configuration/GlobalSettings.cs
pellared: `VendorPluginLoaded` is not needed as well. Correct?

src/Datadog.Trace/Configuration/GlobalSettings.cs
```diff
@@ -184,5 +223,21 @@ private static bool TryLoadHostingEnvironmentPath(out string hostingPath)
             hostingPath = default;
             return false;
         }
+
+        private static void LoadVendorFactoryConfigurator()
```
pellared, line 227: I do not like the fact that logic like this is put in `GlobalSettings` class.

src/Datadog.Trace/Configuration/GlobalSettings.cs
```diff
@@ -184,5 +223,21 @@ private static bool TryLoadHostingEnvironmentPath(out string hostingPath)
...
+        {
+            if (Source.VendorPluginLoaded)
+            {
+                var factoryInterface = typeof(IFactoryConfigurator);
+                var factoryType = Source.VendorPlugin
+                    .GetTypes()
+                    .FirstOrDefault(x => factoryInterface.IsAssignableFrom(x));
```
pellared, line 234: I do not like having logic like this in `GlobalSettings` class.

src/Datadog.Trace/Configuration/GlobalSettings.cs
```diff
@@ -55,6 +77,11 @@ internal GlobalSettings(IConfigurationSource source)
         /// </summary>
         internal bool DiagnosticSourceEnabled { get; }
 
+        /// <summary>
+        /// Gets the factory configurator.
+        /// </summary>
+        internal IFactoryConfigurator FactoryConfigurator { get; private set; }
```
pellared, line 83: I feel that this class should contain "serializable" config. What you add here is a more like "runtime" state of the object created based on settings.

pellared (commented):
Nice starting point 👍 
I tried to focus more on high-level design and general approach during the code review. However I also pointed out several implementation details which I thought is better to tell about now rather than later.

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
MikeGoldsmith: I like the `OTEL_DOTNET` prefix to help reduce the conflict with other plugin lists. It would be worth reviewing other env vars to improve consistency.

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
RassK: You can't specify full path, it should load the correct version (runtime version) of the plugin.

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
RassK: I tried to avoid exactly that, then plugins will be directly connected to `Tracer` instead of global scope.

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
RassK: Runtime folder logic is stored in Loader assembly. Not sure we want to redesign that heavily to reuse the same logic.

src/Datadog.Trace/Configuration/GlobalSettings.cs
RassK: Do you mean to scan through every plugin assembly or just forward settings and each plugin assembly will provide what it could?
Any ideas how to override existing logic? For eg some propagators have vendor specific fields not used in core functionality.

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
pellared: > Runtime folder logic is stored in Loader assembly. Not sure we want to redesign that heavily to reuse the same logic.

I think it would be good, but no worth starting with it.

So basically I think we initially support only the env var value to be equal to `OTEL_DOTNET_TRACER_HOME` value. And then we can reimplment the assembly loader so that it can be used for plugin loading. Does it make any sense?

pellared: I want to say that this should work upfront.
```
OTEL_DOTNET_TRACER_HOME=/home/me/tracer
OTEL_DOTNET_TRACER_PLUGINS=/home/me/tracer

/home/me/tracer
|
- integrations.json
- ...Native.dll
- netcoreapp3.1
|--- ...Tracer.dll
|--- ...plugin.dll 
```

And this should NOT work initially, but after could work after we reuse the assembly loader.

```
OTEL_DOTNET_TRACER_HOME=/home/me/tracer
OTEL_DOTNET_TRACER_PLUGINS=/home/plugins

/home/me/tracer
|
- integrations.json
- ...Native.dll
- netcoreapp3.1
|--- ...Tracer.dll 

/home/plugins
|
- netcoreapp3.1
|--- ...plugin.dll 
```

src/Datadog.Trace/Configuration/GlobalSettings.cs
pellared: The plugin e.g. could implement `B3ExtendedPropagator` by overriding the `B3Propagator` and the tracer would use it when `b3extended` is set in `OTEL_PROPAGATORS`.

pellared: > Do you mean to scan through every plugin assembly or just forward settings and each plugin assembly will provide what it could?

I think that we would have to scan the assemblies and we want to SOMEHOW register stuff from the plugins. Maybe we can even provide the functionality to "Add" or "Replace" via registration API. Take a look at https://blog.ploeh.dk/2014/05/19/di-friendly-framework/ for inspiration

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
nrcventura: I like the idea of have a separate plugin location. That would make it easier for people that have their own custom plugins (for whatever reason) outside of a vendor to be able to update their installation and potentially not remove their current plugins.

src/Datadog.Trace/Configuration/GlobalSettings.cs
```diff
@@ -42,6 +49,21 @@ internal GlobalSettings(IConfigurationSource source)
         /// <seealso cref="ConfigurationKeys.DebugEnabled"/>
         public bool DebugEnabled { get; private set; }
 
+        /// <summary>
+        /// Gets the vendor plugin path.
+        /// </summary>
+        public string VendorPluginName { get; private set; }
```
nrcventura, line 55: We may want to keep the PluginName more generic and remove the `Vendor` prefix, because there might be plugin use cases that are not vendor specific.

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
RassK: Should we divert back to the original idea? Instead of specifying plugin files, we'll just specify plugin dir. We can use the same logic to determine the runtime folder without bigger changes. 
Any concerns about security? - it will autoload all dlls in plugin dir.

pellared: > Any concerns about security? - it will autoload all dlls in plugin dir.

At a minimum, the mechanism has to be documented. Because it can always be used as an attack vector. No matter how it can be secured, it always allows ACE (Arbitraty Code Execution).

If we are not happy with autoloading all dlls in a plugin directory, we can introduce more sophisticated `plugins.json` (like `integrations.json`) where we could allow specifying plugins in a more strict manner. We could even introduce stuff like plugin checksums to make sure that nothing has modified the plugins and that they are exactly what we expect. I cannot imagine a better (yet still being quite simple) defensive mechanism.

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
zacharycmontoya: This is on the loading path, so it will be called on startup before anything else, but theoretically you could always call this later to ensure initialization

src/Datadog.Trace/Configuration/GlobalSettings.cs
zacharycmontoya: Which part of the logic do you not like in this class?

Unrelated, do we only expect one factory per plugin? I could see us getting all types that implement `IFactoryConfigurator` so the factories could configure settings independently.

src/Datadog.Trace/Configuration/TracerSettings.cs
zacharycmontoya: Hmm can you elaborate on this? I'm not sure I understand how we could mix and match enums

src/Datadog.Trace/Configuration/GlobalSettings.cs
zacharycmontoya: +1 on this

zacharycmontoya:
I took an initial look and the direction looks good 👍🏼

src/Datadog.Trace/Configuration/TracerSettings.cs
pellared: I think that `TracerSettings` should only contain "core" settings.
The "core" should not know anything about the plugins apart from the contract (extension point interface).
The plugins can parse the env vars to create their own settings.
For example, if the user would set `OTEL_PROPAGATORS=tracecontext,baggage,splunk` then the "core" should return two enums for both w3c propagators `tracecontext,baggage` and do NOP for the third one `splunk`. The third `splunk` one can be then processed by the plugin. The plugin should simply return an additional propagator when this value is set.
MOre: https://github.com/open-telemetry/opentelemetry-specification/blob/b46bcab5fb709381f1fd52096a19541370c7d1b3/specification/sdk-environment-variables.md#general-sdk-configuration

src/Datadog.Trace/Configuration/GlobalSettings.cs
pellared: `GlobalSettings` should be responsible for parsing the configuration. It should not invoke logic like loading the plugins. It can load "core" configuration for plugins like where are they stored.

pellared: > Unrelated, do we only expect one factory per plugin? I could see us getting all types that implement IFactoryConfigurator so the factories could configure settings independently

Good question. 

We can have one "God" interface with all possible extension points. We can always provide an abstract class for ease of use. It would be a "non-SOLID" design however it could be slightly more performance and most importantly - **easier to find out potential extension points for the plugin devs**.

Make an interface per extension point like advised [here](https://blog.ploeh.dk/2014/05/19/di-friendly-framework/). It is a more loose-coupled design.

I do not have any strong opinion here. Internally we would probably still want to have these kind of "little" interfaces. Right now we have one extension point so we can always postpone this discussion and decision until we have more experience. Especially that I believe that as a rule of thumb, we will advise rebuilding the plugin for each version of do OTel Tracer. AFAIK this is what Java AutoInstr plugins do as well. E.g. https://github.com/signalfx/splunk-otel-java/releases/tag/v0.10.0

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
pellared: @zacharycmontoya So this comment is misleading? Because how I understand your statement is that this is the first place where the `Tracer` instance would be created.

RassK: @pellared well the trick is that, it's created on static call. I tracked this place as a first static call. However reading the comment, I can't be 100% sure that this is the same for every platform. I was just running Asp.Net Core on console. Another point is if it's protecting a rare race condition somewhere etc.

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
RassK: Json config sounds better, regarding that we can have different security options there. 
Then there is an option to use also full paths to plugins - there could be a use case where plugin is missing for certain framework versions. 

```
{
   ...
   "net45": [
      path/to/my/plugin1.dll,
   ],
   ...
   "netcoreapp3.1": [
      path/to/my/plugin1.dll,
      { path: path/to/my/plugin2.dll, checksum: <check_sum> }
   ]
}
```

src/Datadog.Trace/Configuration/GlobalSettings.cs
RassK: I would prefer "God" interface cause of the performance. There is only single assembly scan and `Activator.CreateInstance` call per plugin dll. Probably should rename that interface also, reflecting more "plugin" in it's name.

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
zacharycmontoya: Yes I suppose the comment is misleading. The first time the `Tracer` instance is created will be here

src/Datadog.Trace/Configuration/GlobalSettings.cs
zacharycmontoya: @pellared makes sense, thanks! We can start with one interface and revisit the decision later if needed

src/Datadog.Trace/Configuration/GlobalSettings.cs
```diff
@@ -121,15 +131,28 @@ internal static CompositeConfigurationSource CreateDefaultConfigurationSource()
             return configurationSource;
         }
 
-        private static bool TryLoadJsonConfigurationFile(IConfigurationSource configurationSource, out IConfigurationSource jsonConfigurationSource)
+        internal static bool TryLoadPluginJsonConfigurationFile(IConfigurationSource configurationSource, out JsonConfigurationSource jsonConfigurationSource)
```
RassK, line 134: accessor error, should be private

src/Datadog.Trace/Plugins/IOTelPlugin.cs
```diff
@@ -0,0 +1,16 @@
+using Datadog.Trace.Propagation;
+
+namespace Datadog.Trace.Plugins
+{
+    /// <summary>
+    /// Provides extendibility points for tracer
```
RassK, line 6: Provides extendibility points for tracer**.**

src/Datadog.Trace/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,108 @@
...
+                return;
+            }
+
+            var runtimeName = GetPluginsRuntimeName();
+
+            Log.Debug("Executing plugins configuration: {0}", pluginsConfig);
+            Log.Information("Trying to load plugins for '{0}' runtime.", runtimeName);
+
+            // TODO: Detect if objects instead of path
+            var pluginFiles = pluginsConfig?.GetValue<JToken>($"['{runtimeName}']").ToObject<string[]>();
```
RassK, line 32: Optional security options could be added. Currently parses only path, in future could be object with path and checksum etc.

src/Datadog.Trace/Propagation/B3SpanContextPropagator.cs
```diff
@@ -10,7 +10,7 @@ namespace Datadog.Trace.Propagation
     /// <summary>
     /// Class that handles B3 style context propagation.
     /// </summary>
-    internal class B3SpanContextPropagator : IPropagator
+    public class B3SpanContextPropagator : IPropagator
```
RassK, line 13: Standard propagator made extendable.

src/Datadog.Trace/Propagation/OTelPropagatorsProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+            return PropagatorSelector.ContainsKey(propagatorId);
+        }
+
+        /// <summary>
+        /// Builds the propagator with given spec.
+        /// </summary>
+        /// <param name="propagatorId">Propagator id.</param>
+        /// <param name="traceIdConvention">Trace id convention.</param>
+        /// <returns>Context propagator.</returns>
+        public virtual IPropagator GetPropagator(string propagatorId, ITraceIdConvention traceIdConvention)
```
RassK, line 35: should not be virtual

src/Datadog.Trace/Propagation/W3CSpanContextPropagator.cs
```diff
@@ -9,7 +9,10 @@
 
 namespace Datadog.Trace.Propagation
 {
-    internal class W3CSpanContextPropagator : IPropagator
+    /// <summary>
+    /// Class that handles W3C style context propagation.
+    /// </summary>
+    public class W3CSpanContextPropagator : IPropagator
```
RassK, line 15: Standard propagator made extendable.

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
RassK: Added prototype to handle different exceptions. Currently can't find any existing pattern, lot of places are catching general exception.

src/Datadog.Trace/Configuration/TracerSettings.cs
RassK: I could not find any good way to preserve enum type. I think the same logic should use the same design line, so some of the base logic could be reused.

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
pellared: @RassK Can you fix the comment please?

src/Datadog.Trace/Plugins/PluginManager.cs
pellared: As far as I see, currently, if user would provide improper JSON then it will silently fail, because of:

```
            try
            {
                // Creates GlobalSettings instance and loads plugins
                PluginManager.TryLoadPlugins(GlobalSettings.Source.PluginsConfiguration);

                // First call to create Tracer instace
                _ = Tracer.Instance;
            }
            catch
            {
                // ignore
            }
```

Also, the tracer may even not be initialized or initialize at some strange time?

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
```diff
@@ -59,7 +60,10 @@ public static void Initialize()
 
             try
             {
-                // ensure global instance is created if it's not already
+                // Creates GlobalSettings instance and loads plugins
+                PluginManager.TryLoadPlugins(GlobalSettings.Source.PluginsConfiguration);
```
pellared, line 64: **OPTION 1**
Can we move this plugin loading inside the tracer constructor?
It looks like the constructor of the Tracer is currently the only reasonable [composition root](https://blog.ploeh.dk/2011/07/28/CompositionRoot/) for the managed code as we cannot inject anything inside Tracer.

I have also other reasons why I would advise you to do so.
Access to tracer + this Instrumentation class has synchronization mechanisms, while the `PluginManager` does not. It is very hard to reason if accessing `PluginManager.Loaded` is threat safe. I think it is but I am not sure and trying to figure it out took me quite a lot of time. Also, we may easily break the thread saftiness during refactoring (assuming everything is OK currently). The other thing I do not like is that the `PluginManager` becomes another singleton.

**OPTION 2**
We could also try changing the `Tracer.Instance` design to not use the default constructor for initialization. And pass the loaded plugins to the constructor. 

I fear that this change might be a lot harder and it may be a minefield.

src/Datadog.Trace/Configuration/GlobalSettings.cs
```diff
@@ -121,15 +131,28 @@ internal static CompositeConfigurationSource CreateDefaultConfigurationSource()
             return configurationSource;
         }
 
-        private static bool TryLoadJsonConfigurationFile(IConfigurationSource configurationSource, out IConfigurationSource jsonConfigurationSource)
+        private static bool TryLoadPluginJsonConfigurationFile(IConfigurationSource configurationSource, out JsonConfigurationSource jsonConfigurationSource)
+        {
+            var configurationFileName = configurationSource?.GetString(ConfigurationKeys.PluginConfigurationFileName) ??
+                                        Path.Combine(GetCurrentDirectory(), "plugins.json");
```
pellared, line 137: Not sure if we want to use `Path.Combine(GetCurrentDirectory(), "plugins.json")` if nothing is provided.
I always fear such stuff because it is a great place that someone may use for ACE (Arbitrary Code Execution). The bad-guy must only have a possibility to create some file in the same directory (it can be created by a different user!). The user may be not aware of the risk. I would prefer to require explicitly enabling via ENV VAR. Non-admin user is unable to set env var for different users.

src/Datadog.Trace/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        private static IReadOnlyCollection<IOTelPlugin> TryLoadPlugins(string[] pluginFiles)
+        {
+            var loaded = new List<IOTelPlugin>();
+
+            foreach (string file in pluginFiles)
+            {
+                string fullPath = Path.GetFullPath(file);
+
+                if (File.Exists(fullPath))
```
pellared, line 56: I really think it is not needed.
`Assembly.LoadFrom()` should throw `FileNotFoundException` if the file is missing.
There is always a non-zero possibility for a "race" when the file exists here and does not exist when you load the assembly. So I think it is enough to handle this case only during assembly loading

src/Datadog.Trace/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        private static IOTelPlugin ConvertToPlugin(Assembly assembly)
+        {
+            var pluginType = typeof(IOTelPlugin);
+
+            var pluginInstance = assembly
+                .GetTypes()
+                .Where(p => pluginType.IsAssignableFrom(p))
+                .Select(p => (IOTelPlugin)Activator.CreateInstance(p))
+                .FirstOrDefault();
```
pellared, line 99: any reason why not keeping it as enumerable so someone can implement multiple `IOTelPlugin`s? the problem I see is that currently if someone implements the interface multiple times then we do not know which one would be loaded. it may be hard when troubleshooting as we even have nothing in logs

src/Datadog.Trace/Configuration/TracerSettings.cs
pellared: I think it makes more sense to create an issue for it and tackle it later (but not too late) if nobody finds an easy way how to address the issue now.

src/Datadog.Trace/Propagation/CompositePropagatorsProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+                .Select(x => x.GetPropagator(propagatorId, traceIdConvention))
+                .FirstOrDefault();
+
+            if (propagator == null)
+            {
+                throw new InvalidOperationException($"There is no propagator registered for type '{propagatorId}'.");
+            }
+
+            return propagator;
+        }
```
pellared, line 58: I believe that this class does not have to implement `IPropagatorsProvider`

src/Datadog.Trace/Propagation/CompositePropagatorsProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+        public IPropagator GetPropagator(string propagatorId, ITraceIdConvention traceIdConvention)
+        {
+            var propagator = _providers
+                .Where(x => x.CanProvide(propagatorId, traceIdConvention))
+                .Select(x => x.GetPropagator(propagatorId, traceIdConvention))
+                .FirstOrDefault();
+
+            if (propagator == null)
+            {
+                throw new InvalidOperationException($"There is no propagator registered for type '{propagatorId}'.");
```
pellared, line 54: have you tested what will happen E2E in such case? I have a feeling that it will silently crash the tracer without any logging.

src/Datadog.Trace/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Linq;
+using System.Reflection;
+using Datadog.Trace.Configuration;
+using Datadog.Trace.Logging;
+using Datadog.Trace.Util;
+using Datadog.Trace.Vendors.Newtonsoft.Json.Linq;
+
+namespace Datadog.Trace.Plugins
+{
+    internal static class PluginManager
```
pellared, line 13: I would love to have this code tested. I think it is so complex and critical at the same time.

pellared (commented):
This PR is a LOT better after your changes 👍 
Left some comments and resolved comments that I found addressed.

src/Datadog.Trace/Plugins/PluginManager.cs
RassK: Currently it is a hard requirement / limitation 1 plugin per assembly (see requirements). I don't know any good use case for multiple plugins in the same assembly. I think allowing less initially is better than allowing more that is not used mostly.

src/Datadog.Trace/Configuration/GlobalSettings.cs
RassK: Initially almost the same logic as for json configuration file, just tries to load different file (see `TryLoadJsonConfigurationFile`). Just removing `plugins.json` defaults wouldn't give much benefits, as the ACE could be executed through autoloading `datadog.json` which in turn can direct to plugin by overloading unspecified plugin path.

```
var configurationFileName = configurationSource.GetString(ConfigurationKeys.ConfigurationFileName) ??
      configurationSource.GetString("OTEL_DOTNET_TRACER_CONFIG_FILE") ??
      Path.Combine(GetCurrentDirectory(), "datadog.json");
```

src/Datadog.Trace/Plugins/PluginManager.cs
RassK: Currently it's a performance optimization (if vs exception). Caused of defaults in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/129#pullrequestreview-647515612

src/Datadog.Trace/Configuration/GlobalSettings.cs
pellared: 👍 Then I would say the same mechanism be removed for `datadog.json` 😉

src/Datadog.Trace/Plugins/PluginManager.cs
pellared: I am OK with limiting to 1 plugin per assembly, but we should at least log if there were more (warning) and which one was really loaded.

src/Datadog.Trace/Plugins/PluginManager.cs
RassK: Actually not because of the try catch there but because of the existing logic moved to `TryLoadJsonConfigurationFile`.

src/Datadog.Trace/Propagation/CompositePropagatorsProvider.cs
RassK: Refactored.

src/Datadog.Trace/Propagation/CompositePropagatorsProvider.cs
RassK: Silent in case of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/129#discussion_r622494798 and unknown if lazy initialized later. 

added log msg.

src/Datadog.Trace/Plugins/PluginManager.cs
RassK: Added warning

src/Datadog.Trace/Configuration/GlobalSettings.cs
RassK: Sounds reasonable. Should we refactor json configuration loading in different PR? - as this addresses completely different logic than current one. + the parsing issue https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/129#discussion_r622942562.

src/Datadog.Trace/Plugins/PluginManager.cs
pellared: (for me this can be done in different PR)

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
RassK: Trying out the option 2 with help of `ManualResetEventSlim` class.

src/Datadog.Trace/Tracer.cs
```diff
@@ -212,15 +212,20 @@ public static Tracer Instance
         {
             get
             {
-                return LazyInitializer.EnsureInitialized(ref _instance, ref _globalInstanceInitialized, ref _globalInstanceLock);
+                _globalManualReset.Wait();
+
+                lock (_globalInstanceLock)
```
pellared, line 217: locks should not be needed if you we ResetEvents for synchronization

src/Datadog.Trace/Tracer.cs
```diff
@@ -212,15 +212,20 @@ public static Tracer Instance
...
+                {
+                    return _instance;
+                }
             }
 
             set
```
pellared, line 223: can we get rid of the setter?

RassK: We have `AspNetCoreDiagnosticObserverTests` and `CorrelationIdentifierTests` that are using it. Maybe even more, but these are the first ones that pop up.

src/Datadog.Trace/Tracer.cs
RassK: These were there for the 2x set synchronization (`Initialize `and property `set`). ResetEvent should be for get & set synchronization. Theoretically only tests would be affected.

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
RassK: Unfortunately had to revert it. There are loads of tests and classes that hide `Tracer.Instance` calls that also expect it to create the tracer (only in case of tests). Originally the idea seem to work (running in lab).

pellared: My only idea of how to refactor it to Option 2 without changing loads of tests would be to do something like:

```
        private static Tracer _instance = new Tracer();
        private static ManualResetEventSlim _globalInstanceResetEvent = new ManualResetEventSlim(true);
        private static object _globalInstanceLock = new object();

        /// <summary>
        /// Gets or sets the global <see cref="Tracer"/> instance.
        /// Used by all automatic instrumentation.
        /// </summary>
        public static Tracer Instance
        {
            get
            {
                _globalInstanceResetEvent.Wait();
                return _instance;
            }

            set
            {
                
                lock (_globalInstanceLock)
                {
                    _globalInstanceResetEvent.Reset();
                    _instance = value;
                    _globalInstanceResetEvent.Set();
                }
            }
        }
```

then in `Instrumentation.Initilize()` could set the `Tracer.Instance = new Tracer(plugins)` instead of hacky `_ = Tracer.Instance`

However `private static Tracer _instance = new Tracer();` would make the initialization time of the tracer about 2x longer. Also, I imagine there could be other issues as the AutoInstrumentation may start using the initial version of the tracer before it is swapped in the `Instrumentation.Initilize` method. I would personally be afraid of merging something like that 😄 

I propose to proceed with **OPTION 1** and create an issue to try to refactor to **OPTION 2** as it would require changes in a lot of tests that you mentioned.

I still think it is worth refactoring to **OPTION 1**, because:
1. We would get rid of a static class, properties, fields in `PluginManager`
2. For sure it would be thread-safe
3. The code would be covered by tests. Even if it would be a NOP or load an example plugin - at least one scenario could be "covered".

Therefore I propose to proceed with **OPTION 1** and we can create

src/Datadog.Trace/Configuration/GlobalSettings.cs
pellared: Did you not mean https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/129#discussion_r622507858 ? 

For me, it can be a separate PR + GH issue (just to not forget about it)

src/Datadog.Trace/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,118 @@
...
+                        {
+                            Log.Warning("Could not load {0} from '{1}'.", nameof(IOTelPlugin), pluginAssembly.FullName);
+                        }
+                    }
+                    catch (Exception ex) when (
+                        ExceptionUtil.IsAssemblyLoadException(ex) ||
+                        ExceptionUtil.IsDynamicInvocationException(ex))
+                    {
+                        Log.Warning(ex, "Plugin assembly could not be loaded.");
+                        Log.Information("Skipping vendor plugin load.");
```
pellared, line 79: Any reason why it is not a single `Warning` log line?

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
RassK: It seems to work more close to the original way, but the original was more performant. Plus in case of race condition you could have 2 different outcomes - it either loads Tracer with plugins or fallback to default constructor. 

We need to separate initialization and getter. The initialization needs to happen in controlled way and all other places (in case of race condition) need to wait for initialization. When its 2 in 1, it's impossible to make wrong places to wait for initialization. 

@zacharycmontoya any docs about high level execution tree from native to managed level would be highly useful.

pellared: > Plus in case of race condition you could have 2 different outcomes - it either loads Tracer with plugins or fallback to default constructor.

If the race conditions is possible then even the current codebase may not work properly.

I would say that the approach I proposed is even safer as if the race condition would happen the plugins would eventually loaded. In case of your PR they would never be loaded + possible race condition when accessing `PluginManager.Loaded`.

RassK: I mean the race between `Tracer.Instance` and `Tracer.Initialize()`, Otherwise I do not see any point to change LazyInitializer, as the result won't change?

I might have to try again, otherwise things won't get better. I might have to add `Tracer.Initialize()` to a few test classes constructor and this should do the trick.

src/Datadog.Trace/Configuration/GlobalSettings.cs
RassK: I'm currently documenting some configuration setup after a client issue. This reminds me that we actually cannot remove `datadog.json` autoload (although it should be renamed for OTel context). `OTEL_TRACE_CONFIG_FILE` default can be overloaded only via environment variable which transforms it to the global scope. Clients with multiple applications in the same environment do not have a way to configure app specific values after we remove autoload path. Autoload will ensure that `{App1Path}/datadog.json` and `{App2Path}/datadog.json` can be loaded.

src/Datadog.Trace.ClrProfiler.Managed/Instrumentation.cs
RassK: This resolves also possible `PluginManager.Loaded` issue cause plugins will be loaded before controlled initialization `Tracer.Initialize()`.  `Tracer.Instance` can't ensure controlled initialization.

pellared: I do NOT suggest to add Tracer.Initialize() but  assign via Tracer.Instance in Instrumentation.Initialize. According to @zacharycmontoya there should be no race in case of production usage

RassK: Yep, that should work also, so less synchronization is needed. Good point 👍

src/Datadog.Trace/Configuration/GlobalSettings.cs
pellared: > can be overloaded only via environment variable which transforms it to the global scope

The env var setting could be scooped to a terminal scope or even single .NET process execution. Or am I missing something?

RassK: We had the issue with IIS hosting multiple .NET Framework web apps, that could not be env scoped - although this issue was solved with web.config / app.config.
I'm not 100% sure, that everything could be solved with env scoping or app.config / web.config, there is loads of tech there that are not directly hostable. 

Maybe just add another var to enable config load? - so by default it will be off.

pellared: > I'm not 100% sure, that everything could be solved with env scoping or app.config / web.config.
> [...] Maybe just add another var to enable config load? - so by default it will be off.

I would rather add a var to enable config load if we are certain that we cannot solve it any other way. _Less is more_ 😉

RassK: I would agree if it's about adding functionality. But more concerned when we are reducing publicly available functionality. Anyway we can discuss it under a new issue when it's open and more detailed.

pellared: I am ok with moving it to an issue and discussing it during SIG.

src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -640,18 +641,18 @@ internal string GetServiceName(Tracer tracer, string serviceName)
             return ServiceNameMappings.GetServiceName(tracer.DefaultServiceName, serviceName);
         }
 
-        private static HashSet<PropagatorType> GetPropagators(IConfigurationSource source)
+        private static HashSet<string> GetPropagators(IConfigurationSource source)
         {
-            var propagators = source.GetTypedValues<PropagatorType>(ConfigurationKeys.Propagators);
+            var propagators = source.GetStrings(ConfigurationKeys.Propagators);
 
             if (!propagators.Any())
```
dszmigielski, line 648: This will cause multiple enumeration (at `.Any()` and then at `HashSet` creation). I know there will always be an array underneath but I would add `.ToArray()` in that case just for the sake of it. We won't risk it breaking if the implementation changes.

src/Datadog.Trace/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,117 @@
...
+                Log.Warning("Detected {0} plugins in the assembly '{1}'. Loading only the first type '{2}'.", pluginTypes.Count, assembly.FullName, pluginInstance.GetType().FullName);
+            }
+
+            return pluginInstance;
+        }
+
+        private static string GetPluginsRuntimeName()
+        {
+            // returns the runtime directory of the current running tracer
+            return Directory.GetParent(typeof(PluginManager).Assembly.Location).Name;
```
dszmigielski, line 114: This can cause NullReferenceException in theory, I would change it to:
```csharp
return Directory.GetParent(typeof(PluginManager).Assembly.Location)?.Name;
```
so it's all safe

src/Datadog.Trace/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,117 @@
...
+                return Loaded;
+            }
+
+            var runtimeName = GetPluginsRuntimeName();
+
+            Log.Debug("Executing plugins configuration: {0}", pluginsConfig);
+            Log.Information("Trying to load plugins for '{0}' runtime.", runtimeName);
+
+            // TODO: Detect if objects instead of path
+            var pluginFiles = pluginsConfig?.GetValue<JToken>($"['{runtimeName}']").ToObject<string[]>();
```
dszmigielski, line 32:
```csharp
var pluginFiles = pluginsConfig.GetValue<JToken>($"['{runtimeName}']").ToObject<string[]>();
```
The `?` is not needed as you have already checked against null a few lines above.

src/Datadog.Trace/Util/ExceptionUtil.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Runtime.InteropServices;
+using System.Security;
+
+namespace Datadog.Trace.Util
+{
+    internal static class ExceptionUtil
+    {
+        public static bool IsAssemblyLoadException(Exception ex)
+        {
+            return
```
dszmigielski, line 13: I would go with
```csharp
return
    ex is ArgumentNullException
    or FileNotFoundException
    or FileLoadException
    or BadImageFormatException
    or SecurityException
    or ArgumentException
    or PathTooLongException;
```
seems more readable to me

src/Datadog.Trace/Propagation/CompositePropagatorsProvider.cs
```diff
@@ -0,0 +1,62 @@
+using System;
```
dszmigielski, line 1: I don't see any unit test for that, should we add some here, or will it be separate PR?

dszmigielski (approved):
LGTM - I really like the overall design, I have added some comments, but none of them is blocking for me.

src/Datadog.Trace/Propagation/CompositePropagatorsProvider.cs
RassK: added tests

src/Datadog.Trace/Plugins/PluginManager.cs
RassK: Well in theory if `?.Name` returns `null` then it's a way bigger problem 😄 One option is to convert `NullReferenceException` to `InvalidOperationException` etc, but for me it doesn't give much.

src/Datadog.Trace/Configuration/TracerSettings.cs
RassK: Originally it was intentional due perf and allocation benefits (Enumerate first + enumerate all vs allocate new list / array). @pellared any ideas which way is more preferable?

pellared:
@RassK Could you please create **What's next** section? And probably **Concerns**  with some stuff that we were discussing during today's meeting.

src/Datadog.Trace/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,115 @@
...
+        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(PluginManager));
+
+        internal static IReadOnlyCollection<IOTelPlugin> TryLoadPlugins(JsonConfigurationSource pluginsConfig)
+        {
+            if (pluginsConfig == null)
+            {
+                return ArrayHelper.Empty<IOTelPlugin>();
+            }
+
+            var runtimeName = GetPluginsRuntimeName();
```
zacharycmontoya, line 24: This doesn't seem like the best way to get the name of the runtime, especially if running in the GAC will be a scenario. What about using the `FrameworkDescription.Instance` to get the name?

src/Datadog.Trace/Plugins/PluginManager.cs
zacharycmontoya: This doesn't seem like the best way to get the name of the runtime, especially if running in the GAC will be a scenario. What about using the `FrameworkDescription.Instance` to get the name?

src/Datadog.Trace/Propagation/CompositePropagatorsProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+            _providers.Add(provider);
+
+            return this;
+        }
+
+        public CompositePropagatorsProvider RegisterProviderFromPlugins(IEnumerable<IOTelPlugin> plugins)
+        {
+            foreach (var plugin in plugins)
+            {
+                _providers.Add(plugin.GetPropagatorsProvider());
```
zacharycmontoya, line 32: I'm not sure if I'm overengineering here, but if every plugin needs to have a  `GetPropagatorsProvider()` and a plugin may not be affecting propagators, this could return `null`. Would it make sense to add a `HasPropagatorsProvider` to the `IOTelPlugin` interface right now?

pellared: @zacharycmontoya We need to have a single small interface per extension point. Why? To ensure future compatibility. We do not want to require rebuilding and shipping a new plugin if OTel .NET AutoInstr provides a new extension point in the same interface.

@RassK In general it is good to describe to write some docs as part of PR. It not only helps to maintain the docs up to date but also helps a lot during the review process. Especially that the PR's description does not tell much 😉

src/Datadog.Trace/Plugins/PluginManager.cs
pellared: Any reason why not getting the fully qualified assembly name:

```csharp
return typeof(PluginManager).Assembly.FullName.ToString()
```
?

pellared: @zacharycmontoya  Is it possible to run the tracer from the GAC? Any docs?

src/Datadog.Trace/Propagation/CompositePropagatorsProvider.cs
RassK: Had to completely refactor this. At first it was quite hard to imagine exact use case of multiple extension points, but I think we are getting closer.

src/Datadog.Trace/Plugins/PluginManager.cs
RassK: Had to rework it completely. Should be more proper solution for finding out the target framework. Thanks @zacharycmontoya for the direction!

src/Datadog.Trace/Plugins/IExtendPropagators.cs
```diff
@@ -0,0 +1,16 @@
+using Datadog.Trace.Propagation;
+
+namespace Datadog.Trace.Plugins
+{
+    /// <summary>
+    /// Provides propagators extendability point.
+    /// </summary>
+    public interface IExtendPropagators : IOTelExtension
```
pellared, line 8: This interface looks very strange. It is like a provider for a provider. Can't we simply `IPropagatorsProvider` as the extendability point?

pellared: The good thing is that if you use `IPropagatorsProvider` as the extensibility point then you can even dogfood it for the Datadog.Trace assembly and then we have the code already tested via integration test 😉

RassK: removed that middle layer.

src/Datadog.Trace/Configuration/TracerSettings.cs
pellared: I think it is good as it is 👍  Resolving

src/Datadog.Trace/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                return ArrayHelper.Empty<IOTelExtension>();
+            }
+
+            var targetFramework = FrameworkDescription.Instance.TargetFramework;
+
+            Log.Debug("Executing plugins configuration: {0}", pluginsConfig);
+            Log.Information("Trying to load plugins with target framework '{0}'.", targetFramework);
+
+            // TODO: Detect if objects instead of path
```
pellared, line 29: I do not understand this comment 😄

src/Datadog.Trace/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,99 @@
...
+            var pluginFiles = pluginsConfig.GetValue<JToken>($"['{targetFramework}']").ToObject<string[]>();
```
pellared, line 30: I think it will throw an exception in case of case of bad JSON structure.
https://github.com/JamesNK/Newtonsoft.Json/blob/f7e7bd05d9280f17993500085202ff4ea150564a/Src/Newtonsoft.Json/Linq/JToken.cs#L1961
There would be lo log message in such a case.

test/Datadog.Trace.Tests/Propagators/CompositePropagatorsProviderTests.cs
```diff
@@ -0,0 +1,101 @@
...
+            public SpanContext Extract<T>(T carrier, Func<T, string, IEnumerable<string>> getter)
+            {
+                throw new NotImplementedException();
+            }
+
+            public void Inject<T>(SpanContext context, T carrier, Action<T, string, string> setter)
+            {
+                throw new NotImplementedException();
+            }
+        }
```
pellared, line 99: Why not using Moq?

src/Datadog.Trace/Plugins/PluginManager.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/129#discussion_r616554758
That's for the future, where we discussed about security / integrity of the loaded plugin.

src/Datadog.Trace/Plugins/PluginManager.cs
RassK: Seems right, fixed.

src/Datadog.Trace/Plugins/PluginManager.cs
RassK: Changed a bit, should be less cryptic 😄

pellared (approved):
👍 🥇 

Can you add some screenshots that it is working with `VendorSpanContextPropagator` to give other reviewers more confidence? 😉

RassK:
> 👍 🥇
> 
> Can you add some screenshots that it is working with `VendorSpanContextPropagator` to give the maintainers more confidence? I want it merged 😉

I have a simple lab setup: 
1. Main application
2. API application

Main application makes request to API endpoint. That specific endpoint stores headers from last request which can be accessed with another API endpoint. Here is the capture of last requests with `b3` and `vendor` propagator results.

propagators_capture

src/Datadog.Trace/Configuration/TracerSettings.cs
pjanotti: Yeah, didn't like it either but let's look at it separately.

src/Datadog.Trace/Configuration/TracerSettings.cs
pjanotti: I would not worry much about the perf for such cases since it is executed once per application lifetime. Of course, I'm not saying to ignore performance and to not do easy/obvious savings but the impact here will be very minimal anyway.

src/Datadog.Trace/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,111 @@
...
+            {
+                // TODO: Here additional metadata could be loaded (eg: for security and integrity)
+                // instead of just string path
+                pluginFiles = pluginsConfig.GetValue<JToken>($"['{targetFramework}']").ToObject<string[]>();
+            }
+            catch (ArgumentException ex)
+            {
+                Log.Warning(ex, "Could not parse list of plugin paths. Invalid plugin configuration provided.");
+
+                return ArrayHelper.Empty<IOTelExtension>();
```
pjanotti, line 41: Since this is a configuration file from a "distribution" my preference will be to fail to start as a whole.

src/Datadog.Trace/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,111 @@
...
+            }
+
+            if (pluginFiles == null || !pluginFiles.Any())
```
pjanotti, line 44: nit: since it is an array why use `Any()`? Just check the length.

dszmigielski (approved):

src/Datadog.Trace/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,111 @@
...
+
+        private static IReadOnlyCollection<IOTelExtension> TryLoadPlugins(string[] pluginFiles)
+        {
+            var loaded = new List<IOTelExtension>();
+
+            foreach (string file in pluginFiles)
+            {
+                string fullPath = Path.GetFullPath(file);
+
+                if (File.Exists(fullPath))
```
pjanotti, line 66: nit: the if is pretty big wrapping a try-catch a better if !file.Exists { log; continue; } makes the code easier to read.

src/Datadog.Trace/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,111 @@
...
+                    catch (Exception ex) when (
+                        ExceptionUtil.IsAssemblyLoadException(ex) ||
+                        ExceptionUtil.IsDynamicInvocationException(ex))
+                    {
+                        Log.Warning(ex, "Plugin assembly could not be loaded. Skipping vendor plugin load.");
+                    }
+                }
+                else
+                {
+                    Log.Warning("Plugin path is defined but could not find the path '{0}'.", fullPath);
```
pjanotti, line 93: As mentioned somewhere else for plugins I think the right call is to fail the start.

pjanotti (approved):
LGTM. A few suggestions and follow-ups but let's merge as it is if it passes the tests.

pjanotti:
I'm seeing a bunch of `Failed to fetch https://packages.microsoft.com/debian/10/prod/dists/buster/InRelease  Could not connect to packages.microsoft.com:443 (13.82.67.141), connection timed out` on Az Pipelines. Re-kicking.

Issue #127, opened 2021-04-14
Unsafe reference: System.Runtime.CompilerServices.Unsafe


pellared:
See: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/68d5b3775ca55eeff28e2fe5da04ae9bf3fafe36/src/Datadog.Trace/Datadog.Trace.csproj#L28

What will happen if the instrumented code would reference e.g. 5.0.0 or some version that will be incompatible?

I guess it can lead to a runtime exception.

Can anyone confirm my assumptions?

zacharycmontoya:
This is one specific issue we faced: https://github.com/DataDog/dd-trace-dotnet/issues/771

Since this class of issue has not happened in some time, I don't have a great explanation for you at the top of my head on the exact scenario and the runtime decisions that cause the issue.

lucaspimentel:
Version 4.5.1 of this package includes binaries for `netcoreapp2.0`, `netstandard1.0`, and `netstandard2.0`, but this project targets `net45` and `net461` which are not compatible.

If `Unsafe` is only used in .NET Core, you should probably add a condition to this `PackageReference` or move it into a different `ItemGroup`, like this one: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/68d5b3775ca55eeff28e2fe5da04ae9bf3fafe36/src/Datadog.Trace/Datadog.Trace.csproj#L17

You probably also want to deploy `System.Runtime.CompilerServices.Unsafe.dll` with the installer packages.

There are other issues in .NET Framework (like strict assembly binding or referencing `netstandard2.0` from projects that target .NET Framework 4.7.1 or below), but I don't think they apply here.

pellared:
> This is one specific issue we faced: [DataDog/dd-trace-dotnet#771](https://github.com/DataDog/dd-trace-dotnet/issues/771)
> 
> Since this class of issue has not happened in some time, I don't have a great explanation for you at the top of my head on the exact scenario and the runtime decisions that cause the issue.

@zacharycmontoya 
Maybe you have removed this reference? 😉 
Compare: https://github.com/DataDog/dd-trace-dotnet/search?l=XML&q=Unsafe&type=code to https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/search?l=XML&q=Unsafe

zacharycmontoya:
[...]

PR #130, opened 2021-04-16
Remove Unsafe package

dszmigielski:
Changing code a bit to get rid of the System.Runtime.CompilerServices.Unsafe package. This should resolve [Issue#127](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/127).

### DIFF
--- a/src/Datadog.Trace/TraceId.cs
+++ b/src/Datadog.Trace/TraceId.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Globalization;
-using System.Runtime.CompilerServices;
 
 namespace Datadog.Trace
 {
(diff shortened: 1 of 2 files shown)


src/Datadog.Trace/TraceId.cs
```diff
@@ -91,12 +84,8 @@ public static TraceId CreateRandom()
         /// <returns>Instance of randomly generated <see cref="TraceId"/>.</returns>
         public static TraceId CreateRandomDataDogCompatible()
         {
-            var lowerBytes = new byte[8];
-
             var randomNumberGenerator = RandomNumberGenerator.Current;
-            Unsafe.WriteUnaligned(ref lowerBytes[0], randomNumberGenerator.Next());
-
-            var lower = (ulong)BitConverter.ToInt64(lowerBytes, startIndex: 0) & 0x7FFFFFFFFFFFFFFF;
+            var lower = (ulong)randomNumberGenerator.Next() & 0x7FFFFFFFFFFFFFFF;
```
pellared, line 88: Side comment:
AFAIK `ulong` is not [CLS Compliant](https://docs.microsoft.com/en-us/dotnet/api/system.clscompliantattribute?view=net-5.0). Is it not an issue? Shouldn't we add `[assembly: CLSCompliant(true)]` to the Tracer?

pellared (approved):

src/Datadog.Trace/TraceId.cs
dszmigielski: Should we create issue for that? We would have to change spanId to be long as well, as for now we are using ulong for that as well.

pellared: Done https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/131

lucaspimentel (approved):

pjanotti (approved):

Issue #111, opened 2021-04-07
Document installation and usage


pellared:
## Why

The repository lacks information regarding the installation, configuration and usage.

## What 

Create `docs/USAGE.md` which will describe how to install, configure and use the auto-instrumentation.
This may help:
- https://docs.datadoghq.com/tracing/setup_overview/setup/dotnet-framework/?tab=environmentvariables
- https://github.com/signalfx/signalfx-dotnet-tracing

PR #132, opened 2021-04-20
Add documentation on installation and usage

dszmigielski:
**What**
The USAGE.md was added to the docs directory. It includes installation steps for autoinstrumentation as well as env variables for configuration.

**Why**
For now this repository has no information about the installation, configuration or usage. This PR should resolve [Issue#111](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/111).


docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+
+Use these environment variables to configure the tracing library:
+
+| Environment variable | Default value | Description |
+|-|-|-|
+| `OTEL_SERVICE_NAME` |  | The name of the service. |
+| `OTEL_ENV` |  | The value for the `environment` tag added to every span. This determines the environment in which the service is available in OTEL µAPM.  |
+| `OTEL_TRACE_ENABLED` | `true` | Enable to activate the tracer. |
+| `OTEL_TRACE_DEBUG` | `false` | Enable to activate debugging mode for the tracer. |
+| `OTEL_TRACE_AGENT_URL` | `http://localhost:9080/v1/trace` | The hostname and port for a OTEL Smart Agent or OpenTelemetry Collector. |
```
nrcventura, line 17: The example value and description do not match. One has a full URL and the other says that just a hostname and port should be provided. Which one is correct?

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
+# Usage
+
+**WARNING**: Please notice that no official release has been created for this repo yet, so installation instructions below would require you to build the code manually beforehand.
+
+## Configure the OpenTelemetry Tracing Library for .NET
+
+### Configuration values
+
+Use these environment variables to configure the tracing library:
```
pellared, line 9: Missing documentation for:
- OTEL_TRACE_CONFIG_FILE
- OTEL_VERSION
- ...

Please double-check everything with: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/src/Datadog.Trace/Configuration/ConfigurationKeys.cs 😉

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+
+| Environment variable | Default value | Description |
+|-|-|-|
+| `OTEL_SERVICE_NAME` |  | The name of the service. |
```
pellared, line 13: Should it not be `OTEL_SERVICE` ?

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+
+**WARNING**: Please notice that no official release has been created for this repo yet, so installation instructions below would require you to build the code manually beforehand.
+
+## Configure the OpenTelemetry Tracing Library for .NET
+
+### Configuration values
+
+Use these environment variables to configure the tracing library:
+
+| Environment variable | Default value | Description |
```
pellared, line 11: Can we change the table formatting to the one used here: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md to be consistent?

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+    ```batch
+    setx COR_PROFILER "{918728DD-259F-4A6A-AC2B-B85E1B658318}" /m
+    ```
+   - For .NET Core applications:
+   ```batch
+   setx CORECLR_PROFILER "{918728DD-259F-4A6A-AC2B-B85E1B658318}" /m
+   ```
+3. Set the "service name" that better describes your application:
+   ```batch
+   setx OTEL_SERVICE_NAME MyServiceName /m
```
pellared, line 106: I guess it should be `OTEL_SERVICE`

docs/USAGE.md
pjanotti: I tend to prefer `OTEL_SERVICE_NAME` - eventually this should come from resources, anyway, I think that the suffix `_NAME` makes it a bit more intuitive. Happy to be convinced otherwise.

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+## Configure the OpenTelemetry Tracing Library for .NET
+
+### Configuration values
+
+Use these environment variables to configure the tracing library:
+
+| Environment variable | Default value | Description |
+|-|-|-|
+| `OTEL_SERVICE_NAME` |  | The name of the service. |
+| `OTEL_ENV` |  | The value for the `environment` tag added to every span. This determines the environment in which the service is available in OTEL µAPM.  |
```
pjanotti, line 14: IIRC `environment` is an OpenTracing convention, we could actually start to use this one https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/deployment_environment.md

```suggestion
| `OTEL_ENV` |  | The value for the `environment` tag added to every span. |
```

docs/USAGE.md
pjanotti, line 17: For the time being:
```suggestion
| `OTEL_TRACE_AGENT_URL` | `http://localhost:9080/v1/trace` | The URL to where send the traces. |
```
Could you please open an issue to use env vars from the exporters instead?

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+| `OTEL_TRACE_GLOBAL_TAGS` |  | Comma-separated list of key-value pairs to specify global span tags. For example: `"key1:val1,key2:val2"` |
+| `OTEL_LOGS_INJECTION` | `false` | Enable to inject trace IDs, span IDs, service name and environment into logs. This requires a compatible logger or manual configuration. |
+| `OTEL_MAX_LOGFILE_SIZE` | `10 MB` | The maximum size for tracer log files, in bytes. |
+| `OTEL_TRACE_LOG_PATH` | Linux: `/var/log/OTEL/dotnet/dotnet-profiler.log`<br>Windows: `%ProgramData%"\OTEL .NET Tracing\logs\dotnet-profiler.log` | The path of the profiler log file. |
+| `OTEL_DIAGNOSTIC_SOURCE_ENABLED` | `true` | Enable to generate troubleshooting logs with the `System.Diagnostics.DiagnosticSource` class. |
+| `OTEL_DISABLED_INTEGRATIONS` |  | The integrations you want to disable, if any, separated by a semi-colon. These are the supported integrations: AspNetMvc, AspNetWebApi2, DbCommand, ElasticsearchNet5, ElasticsearchNet6, GraphQL, HttpMessageHandler, IDbCommand, MongoDb, NpgsqlCommand, OpenTracing, ServiceStackRedis, SqlCommand, StackExchangeRedis, Wcf, WebRequest |
```
pjanotti, line 23: Could you please double-check if it is really `;` ? I remember being mentioned in some other PR that we should standardized to `,`. If currently is semi-colon please open an issue to fix it.

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+| `OTEL_TRACE_DOMAIN_NEUTRAL_INSTRUMENTATION` | `false` |  Sets whether to intercept method calls when the caller method is inside a domain-neutral assembly. This is recommended when instrumenting IIS applications. |
```
pjanotti, line 24: I don't think this one exists anymore.

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+| `OTEL_PROFILER_PROCESSES` |  | Sets the filename of executables the profiler can attach to. If not defined (default), the profiler will attach to any process. Supports multiple values separated with semi-colons, for example: `MyApp.exe;dotnet.exe` |
```
pjanotti, line 25: Same regarding previous env var using `;` as a separator - not to be fixed on this PR but please take the chance to access the current state of such env vars. Same for next env var.

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+| `OTEL_PROFILER_EXCLUDE_PROCESSES` |  | Sets the filename of executables the profiler cannot attach to. If not defined (default), the profiler will attach to any process. Supports multiple values separated with semi-colons, for example: `MyApp.exe;dotnet.exe` |
+| `OTEL_CONVENTION` | `Datadog` | Sets the outbound http and trace id convention for tracer. Available values are: `Datadog` (64bit trace id), `OpenTelemetry` (128 bit trace id). |
```
pjanotti, line 27: also the semantic convention for tracing

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+| `OTEL_PROPAGATOR` | `Datadog` | Sets the propagator for tracer. Available values are: `Datadog`, `B3`, `W3C`. |
```
pjanotti, line 28: We can use separators here, right? The order is also relevant, correct?

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+### Linux
+
+After downloading the library, install the CLR Profiler and its components
+via your system's package manager.
+
+1. Download the latest release of the library.
+2. Install the CLR Profiler and its components with your system's package
+manager:
+    ```bash
+    # Use dpkg:
```
pjanotti, line 39: Sometimes people are using docker containers and they are not sure about which distro their Linux is based on. Perhaps add some not along the lines of https://github.com/signalfx/signalfx-dotnet-tracing#linux-instrumentation-not-working

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+    ```
+3. Set Zipkin exporter:
+    ```bash
+    $ export OTEL_EXPORTER='Zipkin'
+    ```
+4. Set OpenTelemetry conventions:
+    ```bash
+    $ export OTEL_CONVENTION='OpenTelemetry'
+    ```
+5. Set the endpoint of a Smart Agent or OpenTelemetry Collector:
```
pjanotti, line 67: Do not use Smart Agent. In this case something like "Set the endpoint, e.g. OpenTelemetry Collector:"

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+    ```bash
+    $ export OTEL_TRACE_AGENT_URL='http://<YourSmartAgentOrCollector>:9080/v1/trace'
```
pjanotti, line 69:
```suggestion
    $ export OTEL_TRACE_AGENT_URL='http://localhost:9080/v1/trace'
```

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+    ```
+   - For .NET Core applications:
+   ```batch
+   setx CORECLR_PROFILER "{918728DD-259F-4A6A-AC2B-B85E1B658318}" /m
+   ```
+3. Set the "service name" that better describes your application:
+   ```batch
+   setx OTEL_SERVICE_NAME MyServiceName /m
+   ```
+4. Set the endpoint of a Smart Agent or OpenTelemetry Collector that will forward
```
pjanotti, line 108: See previous comment

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+the trace data:
+   ```batch
+   setx OTEL_TRACE_AGENT_URL http://localhost:9080/v1/trace /m
+   ```
+5. Optionally, enable trace injection in logs:
+   ```batch
+   setx OTEL_LOGS_INJECTION true /m
+   ```
+6. Optionally, if instrumenting IIS applications add the following environmet variable set to `true`:
```
pjanotti, line 117: IIRC does not apply anymore.

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+    }
+    ```
+
+## Troubleshooting
+
+Check if you are not hitting one of the issues listed below.
+
+### IIS applications not instrumenting expected services
+
+Set the environment variable `OTEL_TRACE_DOMAIN_NEUTRAL_INSTRUMENTATION` to `true` - without it
```
pjanotti, line 221: As mentioned above I don;t think this applies anymore.

docs/USAGE.md
pjanotti: Ah I see it below, perhaps add a link on item # 2

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+$ cat /proc/version
+```
+
+### High CPU usage
+
+The default installation of auto-instrumentation enables tracing all .NET processes on the box.
+In the typical scenarios (dedicated VMs or containers), this is not a problem.
+Use the environment variables `OTEL_PROFILER_EXCLUDE_PROCESSES` and `OTEL_PROFILER_PROCESSES`
+to include/exclude applications from the tracing auto-instrumentation.
+These are ";" delimited lists that control the inclusion/exclusion of processes.
```
pjanotti, line 243: Update here if needed regarding the actual separator.

docs/USAGE.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Custom instrumentation not being captured
+
+If the code accessing `GlobalTracer.Instance` executes before any auto-instrumentation is injected
```
pjanotti, line 247: We need to double-check if this applies here too.

docs/USAGE.md
dszmigielski: I don't what to change that in this PR as this is purely documentation issue. Please create GH issue if you feel the other name would be more fitting.

docs/USAGE.md
dszmigielski: Yes, this uses `;`:
`internal static class ConfigurationSourceExtensions
    {
        private const char Separator = ';';`

docs/USAGE.md
dszmigielski: This is accurate, they still use `;`

docs/USAGE.md
pellared: @RassK FYI

docs/USAGE.md
```diff
@@ -0,0 +1,311 @@
...
+complete implementation of the OpenTracing API.
+
+The auto-instrumentation provides a base you can build on by adding your own
+custom instrumentation. By using both instrumentation approaches, you'll be
+able to present a more detailed representation of the logic and functionality
+of your application, clients, and framework.
+
+1. Add the OpenTracing dependency to your project:
+    ```xml
+    <PackageReference Include="OpenTracing" Version="0.12.0" />
```
pellared, line 216:
```suggestion
    <PackageReference Include="OpenTracing" Version="0.12.0" />
```

there is a newer version

docs/USAGE.md
pellared: It "applies" in the sense that we still give the possibility to the instrument using OpenTracing.

However, based on https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/129#discussion_r616875205 I think this section is not needed

@zacharycmontoya could you confirm, please?

docs/USAGE.md
```diff
@@ -0,0 +1,311 @@
...
+| `OTEL_HTTP_CLIENT_ERROR_STATUSES` | The application's client http statuses to set spans as errors by. | `400-499` |
+| `OTEL_EXPORTER_JAEGER_AGENT_HOST` | Hostname for the Jaeger agent. | `localhost` |
+| `OTEL_EXPORTER_JAEGER_AGENT_PORT` | Port for the Jaeger agent. | `6831` |
+| `OTEL_TRACE_PARTIAL_FLUSH_ENABLED` | Enable to activate sending partial traces to the agent. | `false` |
+| `OTEL_TRACE_PARTIAL_FLUSH_MIN_SPANS` | The minimum number of closed spans in a trace before it's partially flushed. `OTEL_TRACE_PARTIAL_FLUSH_ENABLED` has to be enabled for this to take effect. | `500` |
+| `OTEL_SERVICE` | Application's default service name. |  |
+| `OTEL_ENV` | The value for the `environment` tag added to every span. |  |
+| `OTEL_TRACE_ENABLED` | Enable to activate the tracer. | `true` | 
+| `OTEL_TRACE_DEBUG` | Enable to activate debugging mode for the tracer. | `false` | 
+| `OTEL_TRACE_AGENT_URL` | The URL to where send the traces. | `http://localhost:9080/v1/trace` |
```
pjanotti, line 52: Isn't the default dependent on the OTel exporter?

docs/USAGE.md
pjanotti: @dszmigielski resolving the review, please do not forget to create the issue.

docs/USAGE.md
```diff
@@ -0,0 +1,311 @@
...
+| ` 
+| `OTEL_MAX_LOGFILE_SIZE` | The maximum size for tracer log files, in bytes. | `10 MB` |
+| `OTEL_TRACE_LOG_PATH` | The path of the profiler log file. | Linux: `/var/log/OTEL/dotnet/dotnet-profiler.log`<br>Windows: `%ProgramData%"\OTEL .NET Tracing\logs\dotnet-profiler.log` |
+| `OTEL_DIAGNOSTIC_SOURCE_ENABLED` | Enable to generate troubleshooting logs with the `System.Diagnostics.DiagnosticSource` class. | `true` |
+| `OTEL_DISABLED_INTEGRATIONS` | The integrations you want to disable, if any, separated by a semi-colon. These are the supported integrations: AspNetMvc, AspNetWebApi2, DbCommand, ElasticsearchNet5, ElasticsearchNet6, GraphQL, HttpMessageHandler, IDbCommand, MongoDb, NpgsqlCommand, OpenTracing, ServiceStackRedis, SqlCommand, StackExchangeRedis, Wcf, WebRequest |  |
+| `OTEL_CONVENTION` | Sets the outbound http and trace id convention for tracer. Available values are: `Datadog` (64bit trace id), `OpenTelemetry` (128 bit trace id). |  `Datadog` |
+| `OTEL_PROPAGATORS` | Semicollon-separated list of the propagators for tracer. Available propagators are: `Datadog`, `B3`, `W3C`. The Tracer will try to execute extraction in the passed order. | `Datadog` |
```
pjanotti, line 61: Comma, no? If not:
```suggestion
| `OTEL_PROPAGATORS` | Semicolon separated list of the propagators for the tracer. Available propagators are: `Datadog`, `B3`, `W3C`. The Tracer will try to execute extraction in the given order. | `Datadog` |
```

docs/USAGE.md
```diff
@@ -0,0 +1,311 @@
...
+| `OTEL_TRACE_DOMAIN_NEUTRAL_INSTRUMENTATION` |  Sets whether to intercept method calls when the caller method is inside a domain-neutral assembly. This is recommended when instrumenting IIS applications. | `false` |
```
pjanotti, line 62: I don't think this env var exists anymore.

docs/USAGE.md
pjanotti: > It "applies" in the sense that we still give the possibility to the instrument using OpenTracing.

yeah, I meant in the sense that I'm not sure if this is still an issue or not.

docs/USAGE.md
pjanotti: @dszmigielski I see version 0.12.0 being used by Datadog.Trace.OpenTracing.csproj, it shouldn't be a problem if we use 0.12.1 as suggested by @pellared but it is good to confirm. To keep the PR scope: create an issue to validate the version issue.

docs/USAGE.md
```diff
@@ -0,0 +1,311 @@
...
+| ` 
+| `OTEL_MAX_LOGFILE_SIZE` | The maximum size for tracer log files, in bytes. | `10 MB` |
+| `OTEL_TRACE_LOG_PATH` | The path of the profiler log file. | Linux: `/var/log/OTEL/dotnet/dotnet-profiler.log`<br>Windows: `%ProgramData%"\OTEL .NET Tracing\logs\dotnet-profiler.log` |
+| `OTEL_DIAGNOSTIC_SOURCE_ENABLED` | Enable to generate troubleshooting logs with the `System.Diagnostics.DiagnosticSource` class. | `true` |
+| `OTEL_DISABLED_INTEGRATIONS` | The integrations you want to disable, if any, separated by a semi-colon. These are the supported integrations: AspNetMvc, AspNetWebApi2, DbCommand, ElasticsearchNet5, ElasticsearchNet6, GraphQL, HttpMessageHandler, IDbCommand, MongoDb, NpgsqlCommand, OpenTracing, ServiceStackRedis, SqlCommand, StackExchangeRedis, Wcf, WebRequest |  |
+| `OTEL_CONVENTION` | Sets the outbound http and trace id convention for tracer. Available values are: `Datadog` (64bit trace id), `OpenTelemetry` (128 bit trace id). |  `Datadog` |
```
pjanotti, line 60: I know that the OTEL conventions are not fully implemented but then we don't need to review the doc as we add each OTEL convention.
```suggestion
| `OTEL_CONVENTION` | Sets the semantic and trace id conventions for the tracer. Available values are: `Datadog` (64bit trace id), `OpenTelemetry` (128 bit trace id). |  `Datadog` |
```

pjanotti (commented):
A few minor things to clear

docs/USAGE.md
dszmigielski: It's present in environment_variables.h and is used further in some cpp code.

docs/USAGE.md
dszmigielski: I'll remove that

docs/USAGE.md
dszmigielski: Well it seems it's not, at least not in this env variable, but this is still incorrect. It should be `http://localhost:8126` following this:

`var agentUri = source?.GetString(ConfigurationKeys.AgentUri) ??
                           // default value
                           $"http://{agentHost}:{agentPort}";`

if the agnetHost and agentPort are not set in env variables then it takes constant defaults: `localhost` and `8126` for port.

Jaeger uses its own configuration key (`OTEL_EXPORTER_JAEGER_AGENT_HOST`). Zipkin and default one use `OTEL_TRACE_AGENT_URL`. I don't know if it should work that way.

docs/USAGE.md
dszmigielski: This is same as [here](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/132#discussion_r620029086). It is there in cpp code.

pellared (approved):
Can you also add something like:

```
## Usage

See [USAGE.md](USAGE.md) for installation, usage and configuration instructions.
```

to the mail `README.md` ?

docs/USAGE.md
```diff
@@ -0,0 +1,285 @@
...
+| `OTEL_TRACE_PARTIAL_FLUSH_MIN_SPANS` | The minimum number of closed spans in a trace before it's partially flushed. `OTEL_TRACE_PARTIAL_FLUSH_ENABLED` has to be enabled for this to take effect. | `500` |
+| `OTEL_SERVICE` | Application's default service name. |  |
+| `OTEL_ENV` | The value for the `environment` tag added to every span. |  |
+| `OTEL_TRACE_ENABLED` | Enable to activate the tracer. | `true` | 
+| `OTEL_TRACE_DEBUG` | Enable to activate debugging mode for the tracer. | `false` | 
+| `OTEL_TRACE_AGENT_URL` | The URL to where send the traces for `Zipkin` and `DatadogAgent` exporters (see: `OTEL_EXPORTER`). | `http://localhost:8126` | 
+| `OTEL_TAGS` | Comma-separated list of key-value pairs to specify global span tags. For example: `"key1:val1,key2:val2"` |  |
+| `OTEL_LOGS_INJECTION` | Enable to inject trace IDs, span IDs, service name and environment into logs. This requires a compatible logger or manual configuration. | `false` | `OTEL_EXPORTER` | The exporter to be used. The Tracer uses it to encode and dispatch traces. Available values are: `DatadogAgent`, `Zipkin`, `Jeager`. | `DatadogAgent` |
+| `
```
nrcventura, line 55: Looks like we're adding an extra blank row in the table.

docs/USAGE.md
pjanotti: Thanks for pointing that out.

docs/USAGE.md
pjanotti: We need to fix Zipkin to also use its own env var. Now that config sources are available that shouldn't be a problem. Could you please open an issue for that?

docs/USAGE.md
pjanotti: The comment on the cpp code it is pretty good: "environment variable is not needed when running on .NET Framework 4.5.2 or higher, and will be ignored.". Adding something like: "only applies to .NET Framework prior to version 4.5.2" seems a good idea.

pjanotti (approved):

docs/USAGE.md
dszmigielski: Thanks for the catch! Fixed :)

nrcventura (approved):

docs/USAGE.md
```diff
@@ -0,0 +1,284 @@
...
+
+**Warning**: Pay close attention to the scope of environment variables. Ensure they
+are properly set prior to launching the targeted process. The following steps set the
+environment variables at the machine level, with the exception of the variables used
+for finer control of which processes will be instrumented, which are set in the current
+command session.
+
+1. Install the CLR Profiler using an installer file (`.msi` file) from the latest release.
+Choose the installer (x64 or x86) according to the architecture of the application
+you're instrumenting.
```
zacharycmontoya, line 130:
```suggestion
Choose the installer (x64 or x86) according to the architecture of the operating
system where it will be running.
```

zacharycmontoya (approved):

docs/USAGE.md
zacharycmontoya: This is because the x64 MSI will install both the 32-bit and 64-bit profiler onto the machine. The x86 MSI will only install the 32-bit profiler onto the machine, since it can only run 32-bit processes.

pjanotti (approved):

PR #172, opened 2021-05-21
PoC: minimizing the code: only calling the loader

pjanotti:
Keep only the managed loader and native profiler: remove other managed components and keep only the tests used in the PoC script. The main solution was cleaned up but minimal sln and proj still are referring the removed projects. 

Keeping a minimal GH workflow to ensure that the branch is building correctly.


pellared (approved):
Quick review summary:

1. I was able to build and run the `poc.sh` on Windows and WSL2
1. `Datadog.Trace.sln` works fine in Visual Studio in Windows
1. `Datadog.Trace.sln` is usable in Visual Studio Code in Windows and WSL2

Proposed next cleanup steps (low priority) that can be done next PR:
1. Remove other `.sln` files
2. Move `test-applications` folder one level higher as the https://github.com/pjanotti/opentelemetry-dotnet-instrumentation/blob/3e80df906e5cfe9675d48e531537f41af896a26b/test/Directory.Build.props#L13-L19 does not make sense for these apps

PR #174, opened 2021-05-21
Add GH job for MacOS

pellared:
## Why

A proper CI pipeline is always handy.

## What

- Add build for MacOS
- Unify the build for all OSes


pjanotti (approved):
Nice!

PR #177, opened 2021-05-24
PoC: Add env var for tracer init

dszmigielski:
Adding trace initialisation based on the env variable. Most of the initialisation code is based on @RassK work 🥇


src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/TraceConfigurationHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+    internal static class TraceConfigurationHelper
+    {
+        public static TracerProviderBuilder UseEnvironmentVariables(this TracerProviderBuilder builder, Settings settings)
+        {
+            return builder.SetResourceBuilder(ResourceBuilder.CreateDefault()
+                                                             .AddService(settings.ServiceName, serviceVersion: settings.ServiceVersion))
+                          .SetExporter(settings);
+        }
+
+        public static TracerProviderBuilder AddAspNetInstrumentation(this TracerProviderBuilder builder)
```
RassK, line 15: `AddAspNetInstrumentation` will cause circular reference with this extension and OTel.SDK ASP.NET extension in line 18.

dszmigielski: How you found a solution for that?

dszmigielski: I am not able to test for framework at the moment :(

build.sh
```diff
@@ -49,3 +48,13 @@ case "$os" in
...
+
+# build Managed
+cd $DIR
+OUTDIR="$( pwd )/src/Datadog.Trace.ClrProfiler.Native/bin/${BUILD_TYPE}/x64"
+
+mkdir -p "$OUTDIR/netstandard2.0"
+mkdir -p "$OUTDIR/netcoreapp3.1"
+
+dotnet publish -f netstandard2.0 -c ${BUILD_TYPE} src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed.csproj -o "$OUTDIR/netstandard2.0"
+dotnet publish -f netcoreapp3.1 -c ${BUILD_TYPE} src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed.csproj -o "$OUTDIR/netcoreapp3.1"
```
pjanotti, line 60: Instead of doing the steps for each target framework, I suggest defining one variable on the script with the set of target frameworks according to the OS of the build machine:

- On Windows: `net452 net46 net461 netstandard2.0 netcoreapp3.1`
- Others: `netstandard2.0 netcoreapp3.1`

and then use the variable on a loop:
```suggestion
for framework in $sdkTargetFrameworks ; do
    dotnet publish -f $framework -c ${BUILD_TYPE} src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed.csproj -o "$OUTDIR/$framework"
done
```

If we eventually adopt the SDK we should invest in extracting the target from the SDK NuGet versions and adjust build and code accordingly. However, for the PoC manually changing according to the SDK package is good enough.

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/TraceConfigurationHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+                    var agentPort = settings.JaegerExporterAgentPort;
+
+                    builder.AddJaegerExporter(options =>
+                    {
+                        options.AgentHost = agentHost;
+                        options.AgentPort = agentPort;
+                    });
+
+                    break;
+#endif
```
pjanotti, line 49: Add a default case and throw an exception if it is reached.

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/ConfigurationKeys.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// dispatch traces.
+        /// Default is <c>"Zipkin"</c>.
+        /// </summary>
+        public const string Exporter = "OTEL_EXPORTER";
+
+        /// <summary>
+        /// Configuration key for the Agent URL where the Tracer can send traces.
+        /// Default value is "http://localhost:8126".
+        /// </summary>
+        public const string AgentUri = "OTEL_TRACE_AGENT_URL";
```
pjanotti, line 36: nit: this doesn't matter much for the PoC but I would rather use `OTEL_EXPORTER_ZIPKIN_ENDPOINT` as defined at https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#zipkin-exporter

pjanotti (commented):
Some suggestions below. I will see if I can debug it on Windows.

Related to the targets that I'm suggesting for Windows: you should change the code at https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/83eca20d9e0435ee7ed7cc5a93cf41cf049255e6/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetFramework.cs#L25 to:

```csharp
            var tracerFrameworkDirectory = corlibVersion < corlib461Version ? "net452" : "net461";
```

this way it matches the SDK target frameworks on Windows we can simply use the same names. One idea will be to do some kind of "best" match here to avoid having to change in both places but this is good enough for the PoC.

pjanotti:
BTW, this wasn't build against the latest `poc-otel-sdk` branch.

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/TraceConfigurationHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+using OpenTelemetry.Trace;
+
+namespace OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed.Configuration
+{
+    internal static class TraceConfigurationHelper
+    {
+        public static TracerProviderBuilder UseEnvironmentVariables(this TracerProviderBuilder builder, Settings settings)
+        {
+            return builder.SetResourceBuilder(ResourceBuilder.CreateDefault()
+                                                             .AddService(settings.ServiceName, serviceVersion: settings.ServiceVersion))
```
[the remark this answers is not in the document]
pjanotti: Here you also need a fallback:

```suggestion
            return builder.SetResourceBuilder(ResourceBuilder.CreateDefault()
                                                             .AddService(settings.ServiceName ?? "UNKNOWN_SERVICE_NAME", serviceVersion: settings.ServiceVersion))
```

Later we can add the code to get the service name from the process (like it is done today).

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Instrumentation.cs
```diff
@@ -0,0 +1,53 @@
...
+            try
+            {
+                var settings = Settings.Instance;
+
+                if (settings.LoadTracerAtStartup)
+                {
+                    var builder = Sdk.CreateTracerProviderBuilder()
+                                     .UseEnvironmentVariables(settings)
+                                     .AddAspNetInstrumentation()
+                                     .AddHttpClientInstrumentation()
```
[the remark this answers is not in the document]
pjanotti: In my case, it was crashing due to missing the service name. Fixing that and doing the proper rename of the ASP.NET instrumentation extension method I got the console exporter working.

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/TraceConfigurationHelper.cs
pjanotti: Here the implementation that I got to work on Windows:
```csharp
        {
#if NET452 || NET461
            return builder.AddAspNetInstrumentation();
#elif NETCOREAPP3_1_OR_GREATER
            return builder.AddAspNetCoreInstrumentation();
#else
            return builder;
#endif
        }
```

build.sh
pjanotti: The package used at https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/177#discussion_r638450690 doesn't have a version for `net46` so this is something that will need to be handled. For the moment I am tempted to remove `net46` from the SDK targets and dig deeper at this later.

build.sh
```diff
@@ -49,3 +48,13 @@ case "$os" in
     mkdir -p ${OUTDIR}
     cp -f build/bin/Datadog.Trace.ClrProfiler.Native.${SUFIX} ${OUTDIR}/OpenTelemetry.AutoInstrumentation.ClrProfiler.Native.${SUFIX}
 esac
+
+# build Managed
+cd $DIR
+OUTDIR="$( pwd )/src/Datadog.Trace.ClrProfiler.Native/bin/${BUILD_TYPE}/x64"
```
pellared, line 54: maybe move the 
```
OUTDIR="$( pwd )/src/Datadog.Trace.ClrProfiler.Native/bin/${BUILD_TYPE}/x64"
```
declaration before building native dll's

and remove `OUTDIR="bin/${BUILD_TYPE}/x64"` as it may lead to tedious bugs if someone changes only one delcation by mistake

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/TraceConfigurationHelper.cs
dszmigielski: Took your code @pjanotti :)

build.sh
dszmigielski: I have removed `net46` from the list

src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.cs
```diff
@@ -33,6 +33,23 @@ static Startup()
...
+
+            try
+            {
+                var assembly = Assembly.Load("OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed, Version=0.0.1.0, Culture=neutral, PublicKeyToken=34b8972644a12429");
+
+                if (assembly != null)
+                {
+                    // call method Datadog.Trace.ClrProfiler.Instrumentation.Initialize()
+                    var type = assembly.GetType("OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed.Instrumentation", throwOnError: false);
+                    var method = type?.GetRuntimeMethod("Initialize", new Type[0]);
```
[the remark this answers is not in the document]
dszmigielski: What logger should we be using? Previously we had Datadog one in Tracer solution, should we have same now?

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Instrumentation.cs
```diff
@@ -0,0 +1,53 @@
...
+                                     .AddSqlClientInstrumentation()
+                                     .SetSampler(new AlwaysOnSampler())
+                                     .AddSource("OpenTelemetry.AutoInstrumentation.*")
+                                     .AddConsoleExporter();
+
+                    _tracerProvider = builder.Build();
+                }
+            }
+            catch (Exception)
```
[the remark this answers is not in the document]
dszmigielski: I had the same problem as above - what logger should we be using?

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed.csproj
```diff
@@ -0,0 +1,19 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFrameworks>netcoreapp3.1;netstandard2.0</TargetFrameworks>
```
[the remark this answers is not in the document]
dszmigielski: I have added that, excluded `net46` as above

pjanotti (approved):
There are still a few things to change/improve but I think we should merge this one and follow up on the suggestions.

build.sh
```diff
@@ -49,3 +50,12 @@ case "$os" in
...
+
+for framework in ${SDK_TARGET_FRAMEWORKS} ; do
+    mkdir -p "$OUTDIR/$framework"
+    dotnet publish -f $framework -c ${BUILD_TYPE} src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed.csproj -o "$OUTDIR/$framework"
+done
```
pellared, line 61: it is always good to end with a new line when dealing with *nix stuff 😉 
Reference: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

pellared (approved):

pellared:
Tested on Windows and WSL

pellared:
Does not work on MacOS
- the `Initialize()` method is not called
- there are not logs for any processes other than `dotnet`

RassK (approved):
Tested with custom sample and using builds 4.6.1 and netcoreapp3.1.
Requires DS 5.0.1 references from sample (expected).

PR #180, opened 2021-05-27
Add `OTEL_DOTNET_TRACER_FORCE` env var

pellared:
Adds `OTEL_DOTNET_TRACER_FORCE` (enabled by default) so that the .NET Profiler invokes `RunILStartupHook` during the first "acceptable" JIT compilation.

It is a hacky/dirty implementation. Just to make it work. I am open to suggestions on how to improve it.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -48,7 +48,7 @@ Use these environment variables to configure the tracing library:
 | `OTEL_SERVICE` | Application's default service name. |  |
 | `OTEL_ENV` | The value for the `environment` tag added to every span. |  |
 | `OTEL_TRACE_ENABLED` | Enable to activate the tracer. | `true` |
-| `OTEL_DOTNET_TRACER_FORCE` | Enable the tracer even of no integrations is set using `OTEL_INTEGRATIONS`. | `true` |
+| `OTEL_DOTNET_TRACER_FORCE` | Enable the tracer even if no integrations is set using `OTEL_INTEGRATIONS`. | `true` |
 | `OTEL_TRACE_DEBUG` | Enable to activate debugging mode for the tracer. | `false` |
 | `OTEL_TRACE_AGENT_URL` | The URL to where send the traces for `Zipkin` and `DatadogAgent` exporters (see: `OTEL_EXPORTER`). | `http://localhost:8126` |
 | `OTEL_TAGS` | Comma-separated list of key-value pairs to specify global span tags. For example: `"key1:val1,key2:val2"` |  |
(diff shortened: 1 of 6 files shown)


zacharycmontoya (approved):
LGTM 👍🏼 If the overhead of scanning all JIT compilations becomes an issue, that can be optimized later

docs/USAGE.md
```diff
@@ -47,9 +47,10 @@ Use these environment variables to configure the tracing library:
 | `OTEL_TRACE_PARTIAL_FLUSH_MIN_SPANS` | The minimum number of closed spans in a trace before it's partially flushed. `OTEL_TRACE_PARTIAL_FLUSH_ENABLED` has to be enabled for this to take effect. | `500` |
 | `OTEL_SERVICE` | Application's default service name. |  |
 | `OTEL_ENV` | The value for the `environment` tag added to every span. |  |
-| `OTEL_TRACE_ENABLED` | Enable to activate the tracer. | `true` | 
-| `OTEL_TRACE_DEBUG` | Enable to activate debugging mode for the tracer. | `false` | 
-| `OTEL_TRACE_AGENT_URL` | The URL to where send the traces for `Zipkin` and `DatadogAgent` exporters (see: `OTEL_EXPORTER`). | `http://localhost:8126` | 
+| `OTEL_TRACE_ENABLED` | Enable to activate the tracer. | `true` |
+| `OTEL_DOTNET_TRACER_FORCE` | Enable the tracer even of no integrations is set using `OTEL_INTEGRATIONS`. | `true` |
```
pjanotti, line 51:
```suggestion
| `OTEL_DOTNET_TRACER_FORCE` | Enable the tracer even if no integrations is set using `OTEL_INTEGRATIONS`. | `true` |
```

pjanotti (approved):
LGTM

pjanotti:
@pellared something that could be improved later is that the code now collects information for all modules and later it just decides to not instrument a method because it doesn't find the method in the set to be instrumented. This could be improved by reverting to the old code path after the tracer/SDK is injected.

pellared:
Also, the env var `OTEL_DOTNET_TRACER_FORCE` value could be loaded only once and stored in a field 😉

pjanotti:
> Also, the env var OTEL_DOTNET_TRACER_FORCE value could be loaded only once and stored in a field 😉

The macro `ToBooleanWithDefault` takes care of that, no?

pellared:
> > Also, the env var OTEL_DOTNET_TRACER_FORCE value could be loaded only once and stored in a field 😉
> 
> The macro `ToBooleanWithDefault` takes care of that, no?

Yes. You are correct 👍

PR #186, opened 2021-05-31
PoC: Add support to testing OpenTracing integration

pjanotti:
Added another env var to the PoC script, `openTracingWrapper` to test the integration with `OpenTelemetry.Shims.OpenTracing`. This adds an extra span to have the extra layer in between but I think give better visibility and covers the cases of the OT spans as parent and child.

### DIFF
--- a/samples/ConsoleApp/Program.cs
+++ b/samples/ConsoleApp/Program.cs
@@ -13,7 +13,7 @@ namespace ConsoleApp
         {
             using (var activity = MyActivitySource.StartActivity("Main"))
             {
-#if OpenTracingWrapper
+#if openTracingWrapper
                 await OpenTracingLibrary.Wrapper.WithOpenTracingSpanAsync("client", RunAsync);
 #else
                 await RunAsync();
--- a/samples/ConsoleApp/Program.cs
+++ b/samples/ConsoleApp/Program.cs
@@ -13,11 +13,7 @@ namespace ConsoleApp
         {
             using (var activity = MyActivitySource.StartActivity("Main"))
             {
-#if openTracingWrapper
                 await OpenTracingLibrary.Wrapper.WithOpenTracingSpanAsync("client", RunAsync);
-#else
-                await RunAsync();
-#endif
             }
 
             return 0;
--- a/samples/ConsoleApp/Program.cs
+++ b/samples/ConsoleApp/Program.cs
@@ -14,18 +14,18 @@ namespace ConsoleApp
             using (var activity = MyActivitySource.StartActivity("Main"))
             {
 #if OpenTracingWrapper
-                OpenTracingLibrary.Wrapper.WithOpenTracingSpan("client", async () => await Run());
+                await OpenTracingLibrary.Wrapper.WithOpenTracingSpanAsync("client", RunAsync);
 #else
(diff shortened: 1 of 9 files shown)


samples/ConsoleApp/Program.cs
```diff
@@ -13,6 +13,20 @@ private static async Task<int> Main()
         {
             using (var activity = MyActivitySource.StartActivity("Main"))
             {
+#if openTracingWrapper
```
pellared, line 16: is there any reason why we do not want to have it always enabled?
the `openTracingWrapper` adds some additional complexity in scripts and project file as well

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Instrumentation.cs
```diff
@@ -43,14 +45,38 @@ public static void Initialize()
                         .AddSource("OpenTelemetry.AutoInstrumentation.*");
 
                     _tracerProvider = builder.Build();
-                    Console.WriteLine($">>>>>>>>>>>>>>>>>>>>>>> Process: {p.ProcessName}({p.Id}), initialized");
+                    Log("OpenTelemetry SDK loaded.");
```
pellared, line 48: nit
```suggestion
                    Log("OpenTelemetry tracer initialized.");
```

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Instrumentation.cs
```diff
@@ -43,14 +45,38 @@ public static void Initialize()
...
                 }
             }
-            catch (Exception)
+            catch (Exception ex)
             {
-                // TODO: Should we have our own logger like datadog has?
+                Log($"OpenTelemetry SDK load exception: {ex}");
```
pellared, line 53: do we need these `try/catches` at all?
the expcetion is thrown and should crash the application, right?

samples/ConsoleApp/Program.cs
```diff
@@ -24,10 +38,8 @@ private static async Task<int> Main()
 
                 var client = new HttpClient();
                 Console.WriteLine("Calling client.GetAsync");
-                await regularHttpClient.GetAsync("http://127.0.0.1:8080");
+                await client.GetAsync("http://127.0.0.1:8080");
```
pellared, line 41: 😄

pellared (approved):
nice 👍 
just left a few minor comments

samples/ConsoleApp/Program.cs
pjanotti: yeah, we can have this on by default - I started defensively anticipating some problem since we don't have any there is no reason for the env var.

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Instrumentation.cs
pjanotti: Right, but I think it makes it easier to pinpoint any issue: the code is small but a few minutes here and there accumulate...

pellared (approved):

PR #189, opened 2021-06-02
Configuration to control tracer settings

dszmigielski:
*What*
Adding some configuration to control Tracer settings. For now this contains:
* enable/disable specific instrumentations (HttpClient, AspNet, SqlClient);
* enable/disable console exporter;
* add additional activitysource names.

*Why*
To facilitate further experimentation it will be convenient to have control of the SDK bootstrap/configuration via environment variables.


src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs
```diff
@@ -1,20 +1,38 @@
...
             var resourceBuilder = ResourceBuilder
                 .CreateDefault()
                 .AddService(settings.ServiceName ?? "UNKNOWN_SERVICE_NAME", serviceVersion: settings.ServiceVersion);
 
-            return builder
+            builder
                 .SetResourceBuilder(resourceBuilder)
                 .SetExporter(settings);
+
+            foreach (var enabledInstrumentation in settings.EnabledInstrumentations)
```
RassK, line 28: I think blacklisting will be more appropriate for this scenario?

dszmigielski: Well, I was considering both, and came to a conclusion that it doesn't matter that much. Both have pros and cons, for now I think this is good enough.

RassK: I mean you mostly would include all integrations and disable the ones that don't work for you (prob some rare case). I think that's why the original solution uses blacklisting. 
Otherwise you have to list all integrations to disable a single one.

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/ConfigurationKeys.cs
```diff
@@ -60,6 +60,21 @@ public class ConfigurationKeys
...
 
+        /// <summary>
+        /// Configuration key for whether the console exporter is enabled.
+        /// </summary>
+        public const string ConsoleExporterEnabled = "OTEL_CONSOLE_EXPORTER_ENABLED";
+
+        /// <summary>
+        /// Configuration key for comma separated list of disabled instrumentations.
+        /// </summary>
+        public const string EnabledInstrumentations = "OTEL_DISABLED_INSTRUMENTATIONS";
```
pellared, line 71: rename the property 😉

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs
```diff
@@ -1,20 +1,38 @@
...
+            {
+                AddInstrumentation[enabledInstrumentation](builder);
+            }
+
+            builder.AddSource(settings.ActivitySources.ToArray());
```
pellared, line 33: I think we should add the same for `AddLegacySource`
can be in a separate PR

docs/USAGE.md
```diff
@@ -63,6 +65,8 @@ Use these environment variables to configure the tracing library:
...
 | `OTEL_PROFILER_EXCLUDE_PROCESSES` | Sets the filename of executables the profiler cannot attach to. If not defined (default), the profiler will attach to any process. Supports multiple values separated with comma, for example: `MyApp.exe,dotnet.exe` |  |
 | `OTEL_DOTNET_TRACER_LOAD_AT_STARTUP` | Defines whether the tracer is created by the auto instrumentation library or not. The default value is `true`. For applications with manual instrumentation in place and tracer being initialized in the app this should be changed to `false`. Note that if this is disabled, the application code will be in charge of setting up the exporters and any other desired instrumentation, e.g. AddAspNetInstrumentation (see [custom instrumentation section](#configure-custom-instrumentation)). | `true` | 
+| `OTEL_DISABLED_INSTRUMENTATIONS` | Comma separated list of disabled instrumentation. The available values are: `HttpClient`, `AspNet`, `SqlClient`. |  |
+| `OTEL_ADDITIONAL_SOURCES` | Comma separated list of additional activitysource names to be added to the tracer at the startup. |  |
```
pellared, line 69: I think we should use `OTEL_DOTNET_TRACER` as a prefix.
Here is why: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#language-specific-environment-variables

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs
dszmigielski: I will add it to this one

docs/USAGE.md
dszmigielski: Do we want that on every single env variable that has OTEL prefix right now?

pellared: Not ALL. Some env var names are defined in the specs: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md

We can do it in a seperate PR

dszmigielski: I have changed the ones that were added in this PR for now.

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/InstrumentationType.cs
```diff
@@ -0,0 +1,23 @@
+namespace OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed.Configuration
+{
+    /// <summary>
+    /// Enum representing all instrumentations.
+    /// </summary>
+    public enum InstrumentationType
```
RassK, line 6: There is already `IntegrationIds` enum. We should unite these.

docs/USAGE.md
```diff
@@ -52,17 +52,23 @@ Use these environment variables to configure the tracing library:
...
 | `OTEL_MAX_LOGFILE_SIZE` | The maximum size for tracer log files, in bytes. | `10 MB` |
 | `OTEL_TRACE_LOG_PATH` | The path of the profiler log file. | Linux: `/var/log/OTEL/dotnet/dotnet-profiler.log`<br>Windows: `%ProgramData%"\OTEL .NET Tracing\logs\dotnet-profiler.log` |
 | `OTEL_DIAGNOSTIC_SOURCE_ENABLED` | Enable to generate troubleshooting logs with the `System.Diagnostics.DiagnosticSource` class. | `true` |
-| `OTEL_DISABLED_INTEGRATIONS` | The integrations you want to disable, if any, separated by a comma. These are the supported integrations: AspNetMvc, AspNetWebApi2, DbCommand, ElasticsearchNet5, ElasticsearchNet6, GraphQL, HttpMessageHandler, IDbCommand, MongoDb, NpgsqlCommand, OpenTracing, ServiceStackRedis, SqlCommand, StackExchangeRedis, Wcf, WebRequest |  |
+| `OTEL_DOTNET_TRACER_DISABLED_INTEGRATIONS` | The integrations you want to disable, if any, separated by a comma. These are the supported integrations: AspNetMvc, AspNetWebApi2, DbCommand, ElasticsearchNet5, ElasticsearchNet6, GraphQL, HttpMessageHandler, IDbCommand, MongoDb, NpgsqlCommand, OpenTracing, ServiceStackRedis, SqlCommand, StackExchangeRedis, Wcf, WebRequest |  |
```
RassK, line 61: Currently the list is only: AspNet, MSSql, HttpClient, MongoDb

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/Settings.cs
```diff
@@ -35,12 +36,52 @@ public Settings(IConfigurationSource source)
...
+                foreach (var instrumentation in disabledConfigurations.Split(separator: ','))
+                {
+                    if (Enum.TryParse(instrumentation, out InstrumentationType parsedType))
+                    {
+                        EnabledInstrumentations.Remove(parsedType);
+                    }
+                    else
+                    {
+                        // TODO replace with proper logging
+                        Console.WriteLine($"Could not parse instrumentation \"{instrumentation}\". Skipping...");
```
RassK, line 56: You can use `ConsoleLogger` with `ILogger` interface. Later it's easier to replace implementing type.

docs/USAGE.md
dszmigielski: Wait a second, I mixed that up. We have two separate env variables at the moment, one in cpp, on in csharp for disabled integrations. I renamed the wrong one in the documentation. How should we go about that? I don't think we want that in 2 places.

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/InstrumentationType.cs
dszmigielski: I have added fields from InstrumentationType to IntegrationIds

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/Settings.cs
dszmigielski: Done

docs/USAGE.md
RassK: I think both layers can use the same var like previously? There is anyway "soft" string matching, so overleft values should not cause problems for any layer?

dszmigielski: 🤦

dszmigielski: Done

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs
```diff
@@ -1,20 +1,45 @@
...
                 .AddService(settings.ServiceName ?? "UNKNOWN_SERVICE_NAME", serviceVersion: settings.ServiceVersion);
 
-            return builder
+            builder
                 .SetResourceBuilder(resourceBuilder)
                 .SetExporter(settings);
+
+            foreach (var enabledInstrumentation in settings.EnabledIntegrations)
+            {
+                if (AddInstrumentation.ContainsKey(enabledInstrumentation))
```
pjanotti, line 30: nitpicking since in this context this happens only at initialization but the code ends up with two lookups on the dictionary a `TryGetValue` will be better.

Also in the spirit of breaking in case of wrong configuration if `TryGetValue` returns false it should crash the app.

docs/USAGE.md
```diff
@@ -52,17 +52,22 @@ Use these environment variables to configure the tracing library:
...
+| `OTEL_DOTNET_TRACER_CONSOLE_EXPORTER_ENABLED` | Defines whether the console exporter is enabled or not. | `true` |
 | `OTEL_MAX_LOGFILE_SIZE` | The maximum size for tracer log files, in bytes. | `10 MB` |
 | `OTEL_TRACE_LOG_PATH` | The path of the profiler log file. | Linux: `/var/log/OTEL/dotnet/dotnet-profiler.log`<br>Windows: `%ProgramData%"\OTEL .NET Tracing\logs\dotnet-profiler.log` |
 | `OTEL_DIAGNOSTIC_SOURCE_ENABLED` | Enable to generate troubleshooting logs with the `System.Diagnostics.DiagnosticSource` class. | `true` |
-| `OTEL_DISABLED_INTEGRATIONS` | The integrations you want to disable, if any, separated by a comma. These are the supported integrations: AspNetMvc, AspNetWebApi2, DbCommand, ElasticsearchNet5, ElasticsearchNet6, GraphQL, HttpMessageHandler, IDbCommand, MongoDb, NpgsqlCommand, OpenTracing, ServiceStackRedis, SqlCommand, StackExchangeRedis, Wcf, WebRequest |  |
+| `OTEL_DOTNET_TRACER_DISABLED_INTEGRATIONS` | The integrations you want to disable, if any, separated by a comma. These are the supported integrations: `AspNet`, `HttpClient`, `SqlClient`, `MongoDb` |  |
```
pjanotti, line 61: This is coming from the current terminology but here on the PoC it will be better to call these "instrumentations"
```suggestion
| `OTEL_DOTNET_TRACER_DISABLED_INSTRUMENTATIONS` | The instrumentations you want to disable, if any, separated by a comma. These are the supported integrations: `AspNet`, `HttpClient`, `SqlClient`, `MongoDb` |  |
```

docs/USAGE.md
```diff
@@ -52,17 +52,22 @@ Use these environment variables to configure the tracing library:
 | `OTEL_DOTNET_TRACER_FORCE` | Enable the tracer even if no integrations is set using `OTEL_INTEGRATIONS`. | `true` |
 | `OTEL_EXPORTER_ZIPKIN_ENDPOINT` | The URL to where send the traces for `Zipkin` and `DatadogAgent` exporters (see: `OTEL_EXPORTER`). | `http://localhost:8126` | 
 | `OTEL_TAGS` | Comma-separated list of key-value pairs to specify global span tags. For example: `"key1:val1,key2:val2"` |  |
-| `OTEL_LOGS_INJECTION` | Enable to inject trace IDs, span IDs, service name and environment into logs. This requires a compatible logger or manual configuration. | `false` | `OTEL_EXPORTER` | The exporter to be used. The Tracer uses it to encode and dispatch traces. Available values are: `DatadogAgent`, `Zipkin`, `Jeager`. | `DatadogAgent` |
+| `OTEL_LOGS_INJECTION` | Enable to inject trace IDs, span IDs, service name and environment into logs. This requires a compatible logger or manual configuration. | `false` |
```
pjanotti, line 55: Side note: this is a feature on SDK that we should validate, not related to this PR though.

docs/USAGE.md
```diff
@@ -52,17 +52,22 @@ Use these environment variables to configure the tracing library:
...
 | `OTEL_PROFILER_PROCESSES` | Sets the filename of executables the profiler can attach to. If not defined (default), the profiler will attach to any process. Supports multiple values separated with comma, for example: `MyApp.exe,dotnet.exe` |  |
 | `OTEL_PROFILER_EXCLUDE_PROCESSES` | Sets the filename of executables the profiler cannot attach to. If not defined (default), the profiler will attach to any process. Supports multiple values separated with comma, for example: `MyApp.exe,dotnet.exe` |  |
 | `OTEL_DOTNET_TRACER_LOAD_AT_STARTUP` | Defines whether the tracer is created by the auto instrumentation library or not. The default value is `true`. For applications with manual instrumentation in place and tracer being initialized in the app this should be changed to `false`. Note that if this is disabled, the application code will be in charge of setting up the exporters and any other desired instrumentation, e.g. AddAspNetInstrumentation (see [custom instrumentation section](#configure-custom-instrumentation)). | `true` | 
+| `OTEL_DOTNET_TRACER_ADDITIONAL_SOURCES` | Comma separated list of additional activitysource names to be added to the tracer at the startup. |  |
```
pjanotti, line 68:
```suggestion
| `OTEL_DOTNET_TRACER_ADDITIONAL_SOURCES` | Comma separated list of additional `ActivitySource` names to be added to the tracer at the startup. |  |
```

src/Datadog.Trace.ClrProfiler.Native/environment_variables.h
```diff
@@ -60,7 +60,7 @@ const WSTRING service_version = WStr("OTEL_VERSION");
 // enabled. If not set (default), all integrations are enabled. Supports
 // multiple values separated with comma, for example:
 // "ElasticsearchNet,AspNetWebApi2"
-const WSTRING disabled_integrations = WStr("OTEL_DISABLED_INTEGRATIONS");
+const WSTRING disabled_integrations = WStr("OTEL_DOTNET_TRACER_DISABLED_INTEGRATIONS");
```
pjanotti, line 63:
```suggestion
const WSTRING disabled_integrations = WStr("OTEL_DOTNET_TRACER_DISABLED_INSTRUMENTATIONS");
```

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/ConfigurationKeys.cs
```diff
@@ -60,6 +60,26 @@ public class ConfigurationKeys
...
 
+        /// <summary>
+        /// Configuration key for whether the console exporter is enabled.
+        /// </summary>
+        public const string ConsoleExporterEnabled = "OTEL_DOTNET_TRACER_CONSOLE_EXPORTER_ENABLED";
+
+        /// <summary>
+        /// Configuration key for comma separated list of disabled integrations.
+        /// </summary>
+        public const string DisabledIntegrations = "OTEL_DOTNET_TRACER_DISABLED_INTEGRATIONS";
```
pjanotti, line 71:
```suggestion
        public const string DisabledIntegrations = "OTEL_DOTNET_TRACER_DISABLED_INSTRUMENTATIONS";
```

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/ConfigurationKeys.cs
```diff
@@ -60,6 +60,26 @@ public class ConfigurationKeys
...
+
+        /// <summary>
+        /// Configuration key for additional activitysource names to be added to the tracer at the startup.
```
pjanotti, line 74:
```suggestion
        /// Configuration key for additional <see cref="ActivitySource"/> names to be added to the tracer at the startup.
```

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/Settings.cs
```diff
@@ -35,12 +39,52 @@ public Settings(IConfigurationSource source)
...
+                foreach (var instrumentation in disabledIntegrations.Split(separator: ','))
+                {
+                    if (Enum.TryParse(instrumentation, out IntegrationIds parsedType))
+                    {
+                        EnabledIntegrations.Remove(parsedType);
+                    }
+                    else
+                    {
+                        // TODO replace with proper logging
+                        _logger.Warning($"Could not parse instrumentation \"{instrumentation}\". Skipping...");
```
pjanotti, line 59: Let's crash with a proper exception.

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/Settings.cs
```diff
@@ -35,12 +39,52 @@ public Settings(IConfigurationSource source)
...
+                    }
+                }
+            }
+
+            var additionalSources = source.GetString(ConfigurationKeys.AdditionalSources);
+            if (additionalSources != null)
```
pjanotti, line 65: Don't change this for this PR but this code is repeated for these "enumerative" configurations, it seems to deserve common code to handle it. Anyway, not on this PR.

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs
```diff
@@ -1,20 +1,45 @@
...
         public static TracerProviderBuilder UseEnvironmentVariables(this TracerProviderBuilder builder, Settings settings)
         {
             var resourceBuilder = ResourceBuilder
                 .CreateDefault()
                 .AddService(settings.ServiceName ?? "UNKNOWN_SERVICE_NAME", serviceVersion: settings.ServiceVersion);
 
-            return builder
+            builder
                 .SetResourceBuilder(resourceBuilder)
                 .SetExporter(settings);
```
pellared, line 26: minor issue

```suggestion
            builder = builder
                .SetResourceBuilder(resourceBuilder)
                .SetExporter(settings);
```

this would be safer if for any reason in the future the builder would be immutable.

Probably it would never be the case. However, at least as a reader, I do not have to think if not assigning the value is a bug or not.

If you want to address that then you would need to also change `builder = addInstrumentation(builder);` and other places as well

pellared (approved):

RassK (approved):
LGTM

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/IntegrationIds.cs
```diff
@@ -1,8 +1,29 @@
 // ReSharper disable InconsistentNaming - Name is used for integration names
 namespace OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed.Configuration
 {
-    internal enum IntegrationIds
+    /// <summary>
+    /// Enum representing supported integrations.
+    /// </summary>
+    public enum IntegrationIds
```
pjanotti, line 7: nit: rename to instrumentations.

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/Settings.cs
```diff
@@ -35,12 +37,51 @@ public Settings(IConfigurationSource source)
             JaegerExporterAgentHost = source.GetString(ConfigurationKeys.JaegerExporterAgentHost) ?? "localhost";
             JaegerExporterAgentPort = source.GetInt32(ConfigurationKeys.JaegerExporterAgentPort) ?? 6831;
 
+            LoadTracerAtStartup = source.GetBool(ConfigurationKeys.LoadTracerAtStartup) ?? true;
+
+            ConsoleExporterEnabled = source.GetBool(ConfigurationKeys.ConsoleExporterEnabled) ?? true;
+
+            EnabledInstrumentations = Enum.GetValues(typeof(IntegrationIds)).Cast<IntegrationIds>().ToList();
```
pjanotti, line 44: Is the Enum.GetValues exception readable? I mean does it points exactly what is wrong in this case?

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/Settings.cs
```diff
@@ -35,12 +37,51 @@ public Settings(IConfigurationSource source)
...
+            {
+                foreach (var instrumentation in disabledInstrumentations.Split(separator: ','))
+                {
+                    if (Enum.TryParse(instrumentation, out IntegrationIds parsedType))
+                    {
+                        EnabledInstrumentations.Remove(parsedType);
+                    }
+                    else
+                    {
+                        throw new ArgumentOutOfRangeException($"The \"{instrumentation}\" is not recognized as supported instrumentation and cannot be disabled");
```
pjanotti, line 56: nit: ArgumentException, no?

pjanotti (approved):

PR #194, opened 2021-06-09
POC: Document DS version requirements

RassK:
- Document DS version requirements. 
- Add debug log to output DS version in sample instrumented application

### DIFF
--- a/samples/ConsoleApp/Program.cs
+++ b/samples/ConsoleApp/Program.cs
@@ -11,6 +11,9 @@ namespace ConsoleApp
 
         private static async Task<int> Main()
         {
+            Console.WriteLine("Ensure right version of DiagnosticSource is installed");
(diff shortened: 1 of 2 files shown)


pellared (approved):

samples/ConsoleApp/Program.cs
```diff
@@ -11,6 +11,9 @@ internal class Program
 
         private static async Task<int> Main()
         {
+            Console.WriteLine("Ensure right version of DiagnosticSource is installed");
+            Console.WriteLine(typeof(ActivitySource).Assembly.FullName);
```
pjanotti, line 15: I will test this myself later but I'm assuming that with our always inject option the SDK was already injected by this point. Anyway I must confirm that.

pjanotti (approved):

PR #195, opened 2021-06-11
Add OpenTracing call to sample app for shim verification

dszmigielski:
*What*
Adding manual OpenTracing instrumentation code to sample app.

*Why*
We should still support OpenTracing manual instrumentation.

*Testing*
Testes on Linux poc-otel-sdk branch. The span is there in the correct trace.
Screenshot 2021-06-11 at 15 32 10
Screenshot 2021-06-11 at 15 33 38

### DIFF
--- a/samples/ConsoleApp/Program.cs
+++ b/samples/ConsoleApp/Program.cs
@@ -45,7 +45,12 @@ namespace ConsoleApp
                 Console.WriteLine("Called client.GetAsync");
             }
             
-            var request = new HttpRequestMessage { RequestUri = new Uri("https://www.example.com/"), Method = HttpMethod.Post, Content = new StringContent(string.Empty, Encoding.UTF8), };
+            var request = new HttpRequestMessage
+            {
+                RequestUri = new Uri("https://www.example.com/"),
+                Method = HttpMethod.Post,
+                Content = new StringContent(string.Empty, Encoding.UTF8),
+            };
 
             var tracer = GlobalTracer.Instance;
             using (var scope = tracer.BuildSpan("Client POST " + request.RequestUri)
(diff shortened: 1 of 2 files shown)


samples/ConsoleApp/Program.cs
```diff
@@ -34,12 +39,47 @@ private static async Task RunAsync()
...
                 Console.WriteLine("Called regularHttpClient.GetAsync");
-
-                var client = new HttpClient();
+                
                 Console.WriteLine("Calling client.GetAsync");
                 await client.GetAsync("http://127.0.0.1:8080/api/mongo");
                 Console.WriteLine("Called client.GetAsync");
             }
+            
+            var request = new HttpRequestMessage { RequestUri = new Uri("https://www.example.com/"), Method = HttpMethod.Post, Content = new StringContent(string.Empty, Encoding.UTF8), };
```
RassK, line 48: seems this part needs more separation (to functions or even classes).

pjanotti, line 48: nit: break this in multiple lines

samples/ConsoleApp/Program.cs
```diff
@@ -34,12 +39,47 @@ private static async Task RunAsync()
...
+
+                using var response = await client.SendAsync(request);
+
+                scope.Span.SetTag("http.status_code", (int)response.StatusCode);
+
+                var responseContent = await response.Content.ReadAsStringAsync();
+                scope.Span.SetTag("response.content", responseContent);
+                scope.Span.SetTag("response.length", responseContent.Length);
+
+                foreach (var header in response.Headers)
```
pjanotti, line 71: For our purposes here the two tags above are enough: we don't need the other tags added in this foreach, they make nice for a sample but not really adding value here.

pjanotti (approved):
@dszmigielski 2 minor things but LGTM

pjanotti (approved):

PR #196, opened 2021-06-18
Add SDK tracer extensibility

pellared:
## Why

One should have the possibility to add additional configuration to the OTel SDK Tracer. It could be done per application.

## What

Add support for `OTEL_DOTNET_TRACER_INSTRUMENTATION_PLUGINS` env var which can be used to select additional plugins used to extend the SDK tracer configuration. More info in the PR docs.

Add two examples of its usage:
1. plugin embedded in an app - it adds [Redis instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.StackExchangeRedis/README.md)!
2. plugin distributed in a separate assembly

## Testing

From `./poc.sh` output:

```sh
++ waitForServer
Hello from Samples.AspNetCoreMvc.OtelSdkPlugin
Hello from Vendor.Distro.Plugin
>>>>>>>>>>>>>>>>>>>>>>> Process: Samples.AspNetCoreMvc31(13412): OpenTelemetry tracer initialized.
>>>>>>>>>>>>>>>>>>>>>>> Process: Samples.AspNetCoreMvc31(13412): OpenTracingShim loaded.

[...]

Hello from Vendor.Distro.Plugin
>>>>>>>>>>>>>>>>>>>>>>> Process: ConsoleApp(31088): OpenTelemetry tracer initialized.
>>>>>>>>>>>>>>>>>>>>>>> Process: ConsoleApp(31088): OpenTracingShim loaded
```

The `Hello from ...` messages are coming from plugins.

Also, take notice the span from Redis instrumentation

image


src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/Settings.cs
```diff
@@ -143,6 +152,11 @@ private Settings(IConfigurationSource source)
         /// </summary>
         public IList<Instrumentation> EnabledInstrumentations { get; }
 
+        /// <summary>
+        /// Gets the list of plugins' repesented by <see cref="Type.AssemblyQualifiedName"/>.
```
dszmigielski, line 156: nit: apostrophe is not needed here

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/PluginsConfigurationHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+                builder = builder.InvokePlugin(assemblyQualifiedName);
+            }
+
+            return builder;
+        }
+
+        private static TracerProviderBuilder InvokePlugin(this TracerProviderBuilder builder, string pluginAssemblyQualifiedName)
+        {
+            // get the type and method
+            var t = Type.GetType(pluginAssemblyQualifiedName);
```
dszmigielski, line 22: I assume if we get exception here, crashing app is exactly what we want to happen (fail fast)?

dszmigielski (approved):
LGTM

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/PluginsConfigurationHelper.cs
pellared: I think so. Maybe it would be good to catch it and make some more readable error, but I did not want to waste time on it as this is a PoC.

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/Settings.cs
pellared: fixed. thx

samples/Samples.AspNetCoreMvc21/Controllers/RedisController.cs
```diff
@@ -0,0 +1,20 @@
+using Microsoft.AspNetCore.Mvc;
+using StackExchange.Redis;
+
+namespace Samples.AspNetCoreMvc.Controllers
+{
+    [Route("api/redis")]
+    public class RedisController : ControllerBase
+    {
+        public static ConnectionMultiplexer Connection = ConnectionMultiplexer.Connect("localhost");
```
pellared, line 9: I think using `static` here makes sense, becasue:

> The central object in StackExchange.Redis is the ConnectionMultiplexer class in the StackExchange.Redis namespace; this is the object that hides away the details of multiple servers. Because the ConnectionMultiplexer does a lot, it is designed to be shared and reused between callers. You should not create a ConnectionMultiplexer per operation. It is fully thread-safe and ready for this usage. In all the subsequent examples it will be assumed that you have a ConnectionMultiplexer instance stored away for re-use. But for now, let's create one. This is done using ConnectionMultiplexer.Connect or ConnectionMultiplexer.ConnectAsync, passing in either a configuration string or a ConfigurationOptions object. The configuration string can take the form of a comma-delimited series of nodes, so let's just connect to an instance on the local machine on the default port (6379):

From: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.StackExchangeRedis/README.md

RassK (approved):
LGTM

dszmigielski (approved):

docs/USAGE.md
```diff
@@ -64,10 +64,10 @@ Use these environment variables to configure the tracing library:
...
+| `OTEL_DOTNET_TRACER_LOAD_AT_STARTUP` | Defines whether the tracer is created by the auto instrumentation library or not. The default value is `true`. For applications with manual instrumentation in place and tracer being initialized in the app this should be changed to `false`. Note that if this is disabled, the application code will be in charge of setting up the exporters and any other desired instrumentation, e.g. AddAspNetInstrumentation (see [custom instrumentation section](#configure-custom-instrumentation)). | `true` |
 | `OTEL_DOTNET_TRACER_ADDITIONAL_SOURCES` | Comma separated list of additional `ActivitySource` names to be added to the tracer at the startup. |  |
 | `OTEL_DOTNET_TRACER_LEGACY_SOURCES` | Comma separated list of additional legacy source names to be added to the tracer at the startup. |  |
-
+| `OTEL_DOTNET_TRACER_PROVIDER_PLUGINS` | Colon (:) separated list of OTel SDK tracer plugins repesented by `System.Type.AssemblyQualifiedName`. See more in dedicated section. |
```
pjanotti, line 70: nit: it is indeed a "tracer provider builder" plugin but I think it will be clear if we say that it is an instrumentation (also there is a typo).
```suggestion
| `OTEL_DOTNET_INSTRUMENTATION_PACKAGES` | Colon (:) separated list of OTel SDK instrumentation packages represented by `System.Type.AssemblyQualifiedName`. See more in dedicated section. |
```

samples/Samples.AspNetCoreMvc21/OtelSdkPlugin.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+using OpenTelemetry.Trace;
+using Samples.AspNetCoreMvc.Controllers;
+
+namespace Samples.AspNetCoreMvc
+{
+    public class OtelSdkPlugin
+    {
+        public TracerProviderBuilder ConfigureTracerProvider(TracerProviderBuilder builder)
```
pjanotti, line 9: When we are more confident in this interface we should go back to SDK and add it to the instrumentations and ask people implementing instrumentations to add it.

pjanotti (approved):
Nice!

This PR the additional instrumentations are done via the app, which is very nice. I assume that to ship instrumentations with the profiler itself it is just a matter of packing them with the distribution (details TBD).

samples/Samples.AspNetCoreMvc21/OtelSdkPlugin.cs
pellared: I will add it to the design docs

samples/Vendor.Distro/Vendor.Distro.csproj
```diff
@@ -0,0 +1,11 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net452;net461;netstandard2.0;netcoreapp3.1</TargetFrameworks>
```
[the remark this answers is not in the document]
pellared: We have this as well: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/e774b4d94c6e59d7a4f7eb64d8acb7874bfe4ef4/src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed.csproj#L4

I will test it tomorrow if I will have time and make a PR if it is not needed

docs/USAGE.md
pellared: How about `OTEL_DOTNET_TRACER_INSTRUMENTATION_PLUGINS`? I am not sure about the `PACKAGES` suffix as for me a package is the same as the assembly.

pellared: Renamed here https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/196/commits/0417fffe2fe1a72d6f4e71f9ac222a4db340326e

samples/Vendor.Distro/Vendor.Distro.csproj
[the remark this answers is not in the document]
pellared:
```
Process: Samples.AspNetCoreMvc31(12336): OpenTelemetry SDK load exception: System.NullReferenceException: Object reference not set to an instance of an object.
```

I think the tracer.loader would need the assembly resolve fallback mechanism mentioned during last SIG meeting.

PR #198, opened 2021-06-23
Add binding redirection sample app

dszmigielski:
## What
Adding documentation and sample application for binding redirection of `Diagnostic Source` package.

## Why
On .NET Framework strong name signing can force multiple versions of the same assembly being loaded on the same process. This causes a separated hierarchy of Activity objects (i.e. spans) not seeing the proper relationship between. Proper documentation will help customers to avoid these kind of problems.

My tests have shown that all that has to be done is adding reference to the version of `Diagnostic Source` that the SDK is using (`5.0.0.1` for now) to the application. This will cause runtime to generate binding redirection automatically. This can also be added manually to `app.config` but can cause warning at compilation.

### Referencing lower version:
log_without_redirection
Without redirect

### With redirect:
log_redirect
with_redirection


pellared (approved):

samples/OldReference/InstrumentedHttpCall.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace OldReference
+{
+    public static class InstrumentedHttpCall
+    {
+        private static readonly ActivitySource MyActivitySource = new("OpenTelemetry.AutoInstrumentation.BindingRedirect");
+
+        public static async Task Get(string url)
```
lucaspimentel, line 13: Should be called `GetAsync` according to C# convention?

docs/USAGE.md
```diff
@@ -300,6 +314,23 @@ Use the environment variables `OTEL_PROFILER_EXCLUDE_PROCESSES` and `OTEL_PROFIL
 to include/exclude applications from the tracing auto-instrumentation.
 These are ";" delimited lists that control the inclusion/exclusion of processes.
 
+### No proper relatioship between spans
+
+On .NET Framework strong name signing can force multiple versions of the same assembly being loaded on the same process. This causes a separate hierarchy of Activity objects. If you are referencing packages in your application that use different version of the `System.Diagnostics.DiagnosticSource` than the `OpenTelemetry.Api` used by autoinstrumentation (`5.0.1`) you have to explicitly reference the `System.Diagnostics.DiagnosticSource` package in the correct version in your application (see [custom instrumentation section](#configure-custom-instrumentation)). This will cause automatic binding redirection to occur resolving the issue. You can also manually add binding redirection to the `App.config` file:
```
lucaspimentel, line 319: Automatic binding redirection can be disabled. You may want to mention this and direct users to the docs on how to enable it.
https://docs.microsoft.com/en-us/dotnet/framework/configure-apps/how-to-enable-and-disable-automatic-binding-redirection

lucaspimentel (approved):
LGTM aside from some non-blocking comments.

PR #205, opened 2021-07-27
[PoC] Remove strong name 

pjanotti:
This more to open a conversation to see if we really want to keep strong name signing on the assemblies that we build.

Auto-instrumentation doesn't publish packages so it doesn't need to have a strong name signature. Removing it can simplify (just a little) build and tests. That said is possible that I'm missing some scenario in which we need the assemblies to be signed. If nothing comes to mind at this point I suggest that we remove the strong name and add it back only when we have a concrete reason.


pellared:
Have you tested if instrumenting an app that is signed with snk works?

pjanotti:
> Have you tested if instrumenting an app that is signed with snk works?

@pellared good question. I did, but separately from here but thinking about it I'm making the `ConsoleApp` strong named so we can validate this in more complex scenarios.

pellared (approved):
I tested it as well. Works GREAT 👍

PR #206, opened 2021-07-28
Improve BindingRedirect sample app

pjanotti:
The BindingRedirect sample app was relying "more" on the updated package reference then on the binding redirect. Changed the BindingRedirect app to use an older version of S.D.DS and to have it more close to the scenario that the app was already build and we can just change the app.config file.

The BindingRedirect sample will fail if running without bytecode instrumentation and the shipped app.config.


samples/OldReference/InstrumentedHttpCall.cs
```diff
@@ -20,15 +19,21 @@ public static async Task GetAsync(string url)
             
             Console.WriteLine(string.Join("\n", loaded));
 
-            using (var activity = MyActivitySource.StartActivity("RunAsync"))
+            var activity = new Activity("RunAsync");
```
pellared, line 22: why these changes are done?
the sample was build based on the official sample: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/docs/trace/getting-started/Program.cs

pellared (approved):

PR #208, opened 2021-07-29
Add OTLP exporter support

dszmigielski:
*What*
Add support for OTLP


src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed.csproj
```diff
@@ -15,6 +15,7 @@
     <PackageReference Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.0.0-rc7" Condition="'$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'netcoreapp3.1'" />
     <PackageReference Include="OpenTelemetry.Exporter.Jaeger" Version="1.2.0-alpha1" />
     <PackageReference Include="OpenTelemetry.Exporter.Zipkin" Version="1.1.0" />
+    <PackageReference Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.1.0" />
```
pellared, line 18: I think it should be

```suggestion
    <PackageReference Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.2.0-alpha1" />
```

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs
```diff
@@ -75,6 +75,18 @@ private static TracerProviderBuilder SetExporter(this TracerProviderBuilder buil
                         options.ExportProcessorType = ExportProcessorType.Simple; // for PoC
                     });
                     break;
+                case "OTLP":
```
pellared, line 78:
```suggestion
                case "otlp":
```

from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#exporter-selection

pellared (approved):

pellared:
It is not working. We are not sure exactly where is the problem but looks that we are not alone: https://github.com/open-telemetry/opentelemetry-collector/issues/3713

We failed to send traces to OTel Collector via `jaeger` (and when sending directly to jaeger everything is working fine) and `otlp` receivers. For `zipkin` it was working - strange...

Here is the OTel Collector config:

```yaml
extensions:
  health_check:
  pprof:
    endpoint: :1777
  zpages:
    endpoint: :55679

receivers:
  otlp:
    protocols:
      grpc:
      http:
  opencensus:
  jaeger:
    protocols:
      grpc:
      thrift_binary:
      thrift_compact:
      thrift_http:
  zipkin:

exporters:
  logging:
    logLevel: debug
  jaeger:
    endpoint: jaeger:14250
    insecure: true
  prometheus:
    endpoint: "0.0.0.0:8889"
    namespace: promexample
    const_labels:
      label1: value1

service:
  pipelines:
    traces:
      receivers: [otlp, opencensus, jaeger, zipkin]
      exporters: [logging, jaeger]
    metrics:
      receivers: [otlp, opencensus]
      exporters: [prometheus]
      #exporters: [logging]

  extensions: [health_check, pprof, zpages]
```

@dszmigielski will take a deeper dive on Monday and I will assist him if needed.

src/OpenTelemetry.AutoInstrumentation.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs
pjanotti: @dszmigielski remove the case for "OTLP"

nrcventura:
> We failed to send traces to OTel Collector via `jaeger` (and when sending directly to jaeger everything is working fine) and `otlp` receivers. For `zipkin` it was working - strange...

How are you attempting to run this test? Both the collector and test app in docker containers via docker compose? Just the collector in a docker container? Just the test app in a docker container? 

I suspect the problem in the linked collector issue is with the test app and collector both running in docker containers so they should not use localhost for the test app to talk to the collector container (because localhost just points back to the test app container).

dszmigielski:
> > We failed to send traces to OTel Collector via `jaeger` (and when sending directly to jaeger everything is working fine) and `otlp` receivers. For `zipkin` it was working - strange...
> 
> How are you attempting to run this test? Both the collector and test app in docker containers via docker compose? Just the collector in a docker container? Just the test app in a docker container?
> 
> I suspect the problem in the linked collector issue is with the test app and collector both running in docker containers so they should not use localhost for the test app to talk to the collector container (because localhost just points back to the test app container).

We were running the collector in the docker container and the instrumented app locally. Same configuration for all tests was used - including zipkin that worked. I have also tested golang examples with otlp configured and it worked (collector received the spans correctly), so I don't think that is a setup issue. Seems something is wrong with the SDK OTLP implementation.

dickiebowuk:
>>I suspect the problem in the linked collector issue is with the test app and collector both running in docker containers so they should not use localhost for the test app to talk to the collector container (because localhost just points back to the test app container).

I was running the collector in a docker container via docker compose with the instrumented app running locally hence using localhost. I pulled the latest 'main' branch of the collector yesterday and tried again. I have successfully exported to the collector from my app using otlp collector and received data in Zipkin, Jaeger and Prometheus. I did a further test to push data into New Relic from the collector and that worked as well.

pellared:
@dickiebowuk It is already fixed here: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/211 . Anyway thanks for your input 👍

PR #209, opened 2021-08-05
[PoC] Fix poc script on Linux

pjanotti:
There was a copy command on `poc.sh` that was Windows only and the rename of the CLR profiler binary was incomplete (still generating upstream name on non-Windows).

### DIFF
--- a/build/nuke/ArtifactNames.cs
+++ b/build/nuke/ArtifactNames.cs
@@ -1,0 +1,4 @@
+public static class ArtifactNames
+{
+    public const string NativeProfiler = "OpenTelemetry.AutoInstrumentation.ClrProfiler.Native";
+}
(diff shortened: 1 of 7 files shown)


pellared (approved):

build/nuke/ArtifactNames.cs
```diff
@@ -0,0 +1,4 @@
+public static class ArtifactNames
```
pellared, line 1: What about renaming to `Artifacts` to have symmetry to the `Projects` class name?

pjanotti: There is a `Artifacts` member under the `Build` class, I didn't look much but initially I when I created it with the name `Artifacts` I had to add the qualifier `global::` to avoid conflict and I wanted to avoid that. I will give it a second look.

pellared:
:shipit:

PR #214, opened 2021-08-25
Add loader tests

dszmigielski:
Adding unit test for Loader project


zacharycmontoya:
These binaries were added to git. I think this was unintentional:
- test/OpenTelemetry.ClrProfiler.Managed.Loader.Tests/Profiler/net461/OpenTelemetry.ClrProfiler.Managed.dll
- test/OpenTelemetry.ClrProfiler.Managed.Loader.Tests/Profiler/netcoreapp3.1/OpenTelemetry.ClrProfiler.Managed.dll
- test/OpenTelemetry.ClrProfiler.Managed.Loader.Tests/Profiler/netstandard2.0/OpenTelemetry.ClrProfiler.Managed.dll

dszmigielski:
> These binaries were added to git. I think this was unintentional:
> 
> * test/OpenTelemetry.ClrProfiler.Managed.Loader.Tests/Profiler/net461/OpenTelemetry.ClrProfiler.Managed.dll
> * test/OpenTelemetry.ClrProfiler.Managed.Loader.Tests/Profiler/netcoreapp3.1/OpenTelemetry.ClrProfiler.Managed.dll
> * test/OpenTelemetry.ClrProfiler.Managed.Loader.Tests/Profiler/netstandard2.0/OpenTelemetry.ClrProfiler.Managed.dll

This is actually intentional - these are "mock" dlls for the tests so we don't load the real otel dll and call its method.

test/OpenTelemetry.ClrProfiler.Managed.Loader.Tests/StartupTests.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        [Fact]
+        public void Ctor_LoadsManagedAssembly()
+        {
+            var director = Directory.GetCurrentDirectory();
+            Environment.SetEnvironmentVariable("OTEL_DOTNET_TRACER_HOME", @$"{director}\Profiler");
+
+            var exception = Record.Exception(() => Startup.ManagedProfilerDirectory);
+            // That means the assembly was loaded successfully and Initialize method was called.
+            Assert.Null(exception);
```
zacharycmontoya, line 17: > // That means the assembly was loaded successfully and Initialize method was called.

Maybe you could also iterate over the list of loaded assemblies to also assert that `OpenTelemetry.ClrProfiler.Managed.dll` has been loaded?

zacharycmontoya:
> This is actually intentional - these are "mock" dlls for the tests so we don't load the real otel dll and call its method.

Oh I see. So this is built from some other source that has the same Initialize method, but essentially no other implementation, is that correct?

dszmigielski:
> > This is actually intentional - these are "mock" dlls for the tests so we don't load the real otel dll and call its method.
> 
> Oh I see. So this is built from some other source that has the same Initialize method, but essentially no other implementation, is that correct?

Exactly that ⬆️

test/OpenTelemetry.ClrProfiler.Managed.Loader.Tests/StartupTests.cs
zacharycmontoya: I'm trying to think of an additional signal to assert that we've loaded everything correctly, in addition to no exceptions happening

dszmigielski: I actually couldn't come up with anything, if there is a problem with loading the assembly or calling the method the exception will be thrown, so I assume that is a valid thing to check against.

dszmigielski:
> > > > This is actually intentional - these are "mock" dlls for the tests so we don't load the real otel dll and call its method.
> > > 
> > > 
> > > Oh I see. So this is built from some other source that has the same Initialize method, but essentially no other implementation, is that correct?
> > 
> > 
> > Exactly that ⬆️
> 
> Got it, thanks for clarifying! This works just fine, but if we wanted to make this more self-explanatory in the repo, I think you could do the following:
> 
> 1. Checked in the new project with source
> 2. Add the project as a project reference for the test project
> 3. Do the necessary file copying as an MSBuild Target after the Build target

That seemed like a nice idea, so I did that - added mocks directory with the project. It is built and published in the correct directory by the nuke now.

test/OpenTelemetry.ClrProfiler.Managed.Loader.Tests/StartupTests.cs
dszmigielski: I have added checking if the assembly is present in the list of the loaded assemblies.

nrcventura:
It looks like there are still a few build errors for these tests. Here's the linux build error.
```
CMake Error at CMakeLists.txt:76 (message):
  OpenTelemetry.ClrProfiler.Managed.Loader must be build first.
```

RassK:
> > > > > This is actually intentional - these are "mock" dlls for the tests so we don't load the real otel dll and call its method.
> > > > 
> > > > 
> > > > Oh I see. So this is built from some other source that has the same Initialize method, but essentially no other implementation, is that correct?
> > > 
> > > 
> > > Exactly that ⬆️
> > 
> > 
> > Got it, thanks for clarifying! This works just fine, but if we wanted to make this more self-explanatory in the repo, I think you could do the following:
> > 
> > 1. Checked in the new project with source
> > 2. Add the project as a project reference for the test project
> > 3. Do the necessary file copying as an MSBuild Target after the Build target
> 
> That seemed like a nice idea, so I did that - added mocks directory with the project. It is built and published in the correct directory by the nuke now.

I guess the binary form should be removed from git then? Not good to include binaries to git.

dszmigielski:
@RassK yes, I did not push yesterday changes for some reason. The dlls are gone by now.

dszmigielski:
> It looks like there are still a few build errors for these tests. Here's the linux build error.
> 
> ```
> CMake Error at CMakeLists.txt:76 (message):
>   OpenTelemetry.ClrProfiler.Managed.Loader must be build first.
> ```

Fixed :)

nrcventura (approved):

PR #221, opened 2021-09-16
[POC] Mongodb integration tests

RassK:
* Adds MongoDb integration tests are are runnable locally and in Ci

Notice: 
Windows can have docker in 2 modes (Windows containers and Linux containers). Windows must be in Windows containers mode to run tests.


test/integration-tests/IntegrationTests.MongoDB/MongoDbCollection.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                ShutDownMongoContainer(_container);
+            }
+        }
+
+        private bool ShouldLaunchContainer()
+        {
+            var environment = EnvironmentHelper.GetIntegrationsEnvironment();
+
+            if (environment == IntegrationsEnvironment.CI)
```
pellared, line 51: do you suspect any other environments?

if not then we could simply check the `CI` env var. so just change the logic if it is running on CI to something like:

```csharp
private bool IsRunningOnCI() {
  string env = Environment.GetEnvironmentVariable("INTEGRATIONS_ENVIRONMENT");
  return !string.IsNullOrEmpty(env);
}
```

github actions sets `CI=true`. See: https://docs.github.com/en/actions/learn-github-actions/environment-variables

I am only proposing b/c I think it will simplify the code.

If you disagree just resolve this comment

test/integration-tests/IntegrationTests.MongoDB/MongoDbCollection.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            var container = mongoContainersBuilder.Build();
+            container.StartAsync().Wait();
+
+            return container;
+        }
+
+        private void ShutDownMongoContainer(TestcontainersContainer container)
+        {
+            container.CleanUpAsync().Wait();
```
pellared, line 87: is there no method like `container.CleanUp()`? There is no benefit of using `async` API here. I remember I was fixing deadlocks for code like this.

https://www.markheath.net/post/async-antipatterns

if there is no other way of doing that then resolve this comment

take notice that this comment also applies to other places:

```csharp
container.StartAsync().Wait();
// ...
container.DisposeAsync().AsTask().Wait();
```

src/OpenTelemetry.ClrProfiler.Managed/AssemblyInfo.cs
```diff
@@ -0,0 +1,4 @@
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("IntegrationTests.Helpers")]
+[assembly: InternalsVisibleTo("IntegrationTests.MongoDB")]
```
pellared, line 4: why is it needed? maybe it would be better to hardcode something or make it public?

src/OpenTelemetry.ClrProfiler.Managed/Instrumentations/MongoDb/MongoDbIntegration.cs
```diff
@@ -15,9 +16,11 @@ namespace OpenTelemetry.ClrProfiler.Managed.Instrumentations.MongoDb
     /// <summary>
     /// Tracing integration for MongoDB.Driver.Core.
     /// </summary>
+    [Browsable(false)]
+    [EditorBrowsable(EditorBrowsableState.Never)]
```
pellared, line 20: I am curious why is it needed

test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
```diff
@@ -56,6 +57,36 @@ public Process StartSample(int traceAgentPort, string arguments, string packageV
                 processToProfile: executable);
         }
 
+        public ProcessResult RunSampleAndWaitForExit(int traceAgentPort, int? statsdPort = null, string arguments = null, string packageVersion = "", string framework = "", int aspNetCorePort = 5000)
+        {
+            var process = StartSample(traceAgentPort, arguments, packageVersion, aspNetCorePort: aspNetCorePort, statsdPort: statsdPort, framework: framework);
+
+            using var helper = new ProcessHelper(process);
+
+            process.WaitForExit();
```
pellared, line 66: what happens if the process hangs? how about having some default timeout?

test/test-applications/integrations/Samples.MongoDB/Samples.MongoDB.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <ApiVersion Condition="'$(ApiVersion)' == ''">2.13.1</ApiVersion>
```
pellared, line 8: Right now the pipeline always uses `2.13.1`. Correct?

test/integration-tests/IntegrationTests.MongoDB/MongoDbTests.cs
```diff
@@ -0,0 +1,88 @@
...
+
+            int agentPort = TcpPortProvider.GetOpenPort();
+
+            using (var agent = new MockZipkinCollector(Output, agentPort))
+            using (var processResult = RunSampleAndWaitForExit(agent.Port, arguments: $"--mongo-db {_mongoDb.Port}"))
+            {
+                Assert.True(processResult.ExitCode >= 0, $"Process exited with code {processResult.ExitCode} and exception: {processResult.StandardError}");
+
+                var spans = agent.WaitForSpans(3, 500);
+                Assert.True(spans.Count >= 3, $"Expecting at least 3 spans, only received {spans.Count}");
```
pellared, line 46: Maybe you want to use [FluentAssertions](https://fluentassertions.com/introduction)?

pellared (approved):
Great work 👍 I had no possibility to run it on my own machine as I was testing other PR in parallel.

I left some comments which may be resolved if you disagree.
You may also create issues out of some or address them in the next PRs - whatever you prefer.

test/integration-tests/IntegrationTests.Helpers/ProcessHelper.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            process.BeginOutputReadLine();
+            process.BeginErrorReadLine();
+        }
+
+        public string StandardOutput => _outputBuffer.ToString();
+
+        public string ErrorOutput => _errorBuffer.ToString();
+
+        public bool Drain(int timeout = Timeout.Infinite)
```
pellared, line 31: How about using `TimeSpan` instead of `int` for `timeout`? Then the caller does not need to think what is the unit.

test/integration-tests/IntegrationTests.MongoDB/MongoDbCollection.cs
RassK: Great find, I suspected that there is something already, but could not find it again.

src/OpenTelemetry.ClrProfiler.Managed/AssemblyInfo.cs
RassK: Seems leftover. I guess I was reusing some utils from that dll, and did not want to create separate utils dll just yet.

src/OpenTelemetry.ClrProfiler.Managed/Instrumentations/MongoDb/MongoDbIntegration.cs
pjanotti: They hide it from the editor/IDE so it doesn't offer this on the auto-complete and related IDE stuff - likely to avoid someone directly referencing the instrumentation code by accident, one needs to be intentional about it.

src/OpenTelemetry.ClrProfiler.Native/logger.h
```diff
@@ -16,7 +16,7 @@ struct TracerLoggerPolicy
 #ifdef _WIN32
     // this field will be removed once merged with the profiler in order to have
     // the same product folder name
-    inline static const WSTRING folder_path = WStr(R"(Datadog .NET Tracer\logs)");
+    inline static const WSTRING folder_path = WStr(R"(OpenTelemetry .NET AutoInstrumentation\logs)");
```
pjanotti, line 19: Not for now but we may want to review this folder name is really long ... I guess I suggested it in the past but I changed my mind 😬

test/integration-tests/IntegrationTests.Helpers/ProcessHelper.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            _errorMutex.Dispose();
+            _outputMutex.Dispose();
+        }
+
+        private static void DrainOutput(DataReceivedEventArgs e, StringBuilder buffer, ManualResetEventSlim mutex)
+        {
+            if (e.Data == null)
+            {
+                mutex.Set();
```
pjanotti, line 51: I like golang style instead of else branch. It can help with size and branch prediction. Not that important on test code but gets us in the good habits. 
```c#
            var data = e.Data;
            if (data != null)
            {
                buffer.AppendLine(data);
                return;
            }
            
            mutex.Set();
```
See [IL/JITAsm](https://sharplab.io/#v2:EYLgtghglgdgNAFxBAzmOATEBqAPgAQCYBGAWACh8AGAAn2IDoARKCAcxgHsUEoBjFAG4K1OowAqAUwAeCYZVr0G4gBYAnSRAyw28ijAhhJKAA4Q+kmgEkYCSWzURenGFJ4oGACUkAbE5LUUCgBvChpwmgB6aJoAHhQAVzBINQBPAD4wiOjImiZHWBQaBBVLHggYDAg1DBoK2oC1TjUaTgSEE3bWgDM6mhMmixQg8giomNjIxOTqjKzw/ABmOkIaAAVB4xRvPwCaEGsWU24IYB9JeZpQ0bGIgagANydLDS0XH1SaAFkKhIgfABKxkkCAAog9JLYAMo+KBgGgAfUazS+7RkNAAvDQYJIAO4ACgAlPJbnc1I9njRXhh3p8fjA/oDgWCIdDYfCEW0Ou1UXZpJjsXiiSTSf1yU87FTNDSYB8aFCEOSYGwAEIJKA+DB7TntToINXdbp7LE4gnEy5je4Sl7S2nyxU6NUarUtJFqJpqA1GlomoXm8gWiJLdabYY7fxqfEbThDIoDGNbQmB8LXUW3eOxhgAeV17SYTggQIsj0ktWwWPxCLgNEkhMx6TyBRgOe5CHxkmrOtbXoCna5et5Mn9actoY8oPdzXzCELkmLELLFarNbrGIb+WgzdzbY7iORnoSht7e8nakH0n9yfTY4YKvssBbeqBWgAMrBJMKr6OE8Nb/eYBOHrPhgb44p+NykgAvhQV7BvQtAKvU1QYI+XRroi/btD2ajKJwCpKmw4Gwcs8E0IBzSoQg9Ynh62G4fhOhERBtzBsAnCcD4jabvisBUbwRhcgK4hwpIXIMDY3SwFAdhJsxYypiO4RQL0+L8aJXQAIRYsJAntOJMCSTA0m1l+EQKYpYxqYJkRYoQIqKdBAZyaK+AAOwYdu54MAA6tAbZWe0dYAGRBTRKJotIPl+apIlcsOoqOcRdAACx5FAxwoB+slpuZI5uh6XlHCY3BZfZeWYQghXpcVmXgWmiXOeEVqUvQABsKVcQ+274tOs7zqW4KQggACCahsEUu4McqTqanswCHt61b0oyQKZSyQ0wnCNBgBF2WirlabKTQ7bMAWmImgkPg+HtikHRZO18gwUIgnVFmORZNY+Jlpkpj9tzzUeOHDSY/iVKBH6SKdM7xQ5V4NWmzWSm1HUAOIcRUbAKqk5wbl1rY9QWRaSCWGCDbYo3jTW1ZTaq6qzS0AOLd8vz/KtIJkwgm3wg9Q5XndpJPC0VQzgKkO9WVh0qcLEA0Fp2KXddf38yOjMBAwwOgyB774tLMMfXQrkSwlf1/TzkXPW2etjO9ESOZBQA===)

pjanotti: In the case above the IL is a tiny bit smaller but the ASM is the same size :) bad example.

dszmigielski (approved):

test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
RassK: I set process response timeout to 30 minutes (same for standard and error output). Whole process default timeout is now 60 minutes. 

Although I find the response checking timeout bit weird... Shouldn't it actually set both mutexes? In case of successful standard output, error mutex timeout is still set and could fire false positive.

test/integration-tests/IntegrationTests.MongoDB/MongoDbTests.cs
RassK: I'll pass this at the moment. Don't have good structure for FluentAssertions in my mind at the moment. I think it should be different, cause `Assert.*` is more eye-catching than `Should()`. So FluentAssertions should be more grouped together to separate normal code and assertions with naked eye.

test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
pellared: > error mutex timeout is still set and could fire false positive.

that would be bad 😋

pellared: > I set process response timeout to 30 minutes (same for standard and error output). Whole process default timeout is now 60 minutes.

Then you should do: `timeout /= 2` otherwise it is very confusing

test/integration-tests/IntegrationTests.MongoDB/MongoDbTests.cs
pellared: OK

test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
RassK: refactored, to make sure that wait timeout is equally distributed.

test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
```diff
@@ -56,6 +59,36 @@ public Process StartSample(int traceAgentPort, string arguments, string packageV
...
 
+        public ProcessResult RunSampleAndWaitForExit(int traceAgentPort, int? statsdPort = null, string arguments = null, string packageVersion = "", string framework = "", int aspNetCorePort = 5000)
+        {
+            var process = StartSample(traceAgentPort, arguments, packageVersion, aspNetCorePort: aspNetCorePort, statsdPort: statsdPort, framework: framework);
+
+            using var helper = new ProcessHelper(process);
+
+            process.WaitForExit(DefaultProcessTimeoutMs);
+            helper.Drain();
+            var exitCode = process.ExitCode;
```
pellared, line 70: See: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.waitforexit?view=net-5.0#System_Diagnostics_Process_WaitForExit_System_Int32_

> Returns true if the associated process has exited; otherwise, false.

I think that `process.WaitForExit(int)` does not terminate the process. I guess that we should kill it. Otherwise the process may leak and probably `process.ExitCode` will throw an exception.

test/integration-tests/IntegrationTests.Helpers/ProcessHelper.cs
```diff
@@ -0,0 +1,61 @@
...
+        private static void DrainOutput(DataReceivedEventArgs e, StringBuilder buffer, ManualResetEventSlim mutex)
+        {
+            string data = e.Data;
+            if (data == null)
+            {
+                mutex.Set();
+                return;
+            }
+
+            buffer.AppendLine(data);
```
pellared: It would be safer to add `lock (buffer)` around this line and add locks when on `StandardError` and `StandardOutput` fields.

test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
RassK: true, fixed.

test/integration-tests/IntegrationTests.Helpers/ProcessHelper.cs
RassK: So the drainage is not meant to read before process exits. Added exception if somebody tries to read before, this saves us locking time.

test/integration-tests/IntegrationTests.Helpers/ProcessHelper.cs
RassK: Also more readable.... but there were bigger changes anyway.

test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
```diff
@@ -3,12 +3,16 @@
...
 using System.Linq;
+using IntegrationTests.Helpers.Models;
 using Xunit.Abstractions;
 
 namespace IntegrationTests.Helpers
 {
     public abstract class TestHelper
     {
+        // Warning: Long timeouts can cause integer overflow!
+        private static readonly TimeSpan DefaultProcessTimeout = TimeSpan.FromMinutes(60);
```
pellared, line 14: Do we really want to have that long timeout for the sample application?
Maybe 5 minutes?

test/integration-tests/IntegrationTests.Helpers/ProcessHelper.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Threading;
+
+namespace IntegrationTests.Helpers
+{
+    /// <summary>
+    /// Drains the standard and error output of a process
+    /// </summary>
+    public class ProcessHelper : IDisposable
+    {
+        private const int DefaultTimeoutMinutes = 30;
```
pellared, line 13: Do we really want to have that long timeout for draining the output?
Maybe 1 minute?

pellared (approved):

test/integration-tests/IntegrationTests.Helpers/ProcessHelper.cs
RassK: Output is generated through the lifetime of the sample. I guess this and process timeout is pretty much connected.

pellared: Gotcha then maybe 5 min as well?

RassK: ok, I changed both to 5 mins, let's try, we can adjust always later if there are issues.

PR #222, opened 2021-09-20
Upstream pull on main

pjanotti:
Enabled minimal AZ pipelines, see https://dev.azure.com/opentelemetry/pipelines/_build/results?buildId=3827&view=results

Picks up the changes from upstream before we switch to the SDK branch.

Remove a few benchmarks to make the pipeline faster (DataDog/dd-trace-dotnet#1497)

Increase the crank scenario duration to 4 minutes. (DataDog/dd-trace-dotnet#1505)

Cleanup project files (redux) (DataDog/dd-trace-dotnet#1499)

* delete redundant props file

* remove files from solution file

* don't build test apps that won't be used

* remove conditions for net5.0

* remove checks for empty values

* remove <RuntimeIdentifier> and <RuntimeIdentifiers>

left over from the 2.x era

* move import to the top so we can override default values

* delete redundant import

* use the recommended import declaration

AgentWriter FlushTracesAsync changes (DataDog/dd-trace-dotnet#1501)

* Change behavior of the FlushTracesAsync method to enqueue all flush continuations per buffer

* AgentWriterEnqueueFlushTasks unit test.

CIApp: Update CircleCI spec (DataDog/dd-trace-dotnet#1503)

* Update CircleCI spec

* fix typo

Add Nuke build project and update consolidated pipeline (DataDog/dd-trace-dotnet#1476)

* Stop git reporting errors in MakeFile when there aren't any

There's still one annoying one I can't get rid of:
ar: creating archive obj/libre2.a

* Add Samples.AspNet472.LoaderOptimizationRegKey to sln so that it's restored with everything else

Use "root" packages folder in root instead of own packages subfolder in LoaderOptimization project

* Add additional target framework to RazorPages project

Causes some issues when skipping dependencies later unfortunately, this was the easiest fix for now

* Add Nuke build project

Includes build.cmd etc for easily running
Includes docker.sh for running inside docker using the same images as we do in CI
I haven't added the build project to the solution yet, mostly to avoid rebasing woes, but will add it during cleanup
To avoid breaking other pipelines, I added new targets to the proj file. These can be trimmed in the future.

* Update the ultimate-pipeline to use Nuke

Splits things into extra stages to increase throughput when we're waiting on a particular machine type linux/windows/mac
Many more optimizations possible, but may not be worth the effort yet
Added new docker-compose targets for now. Can clean these up later
There are some steps which haven't been fully converted yet, but this is a good start IMO

* git update-index --chmod=+x docker.sh build.sh

* Update version number in Nuke script

* Skip building Kafka multiple times

* Try fix CI test apps - msbuild faffing

* Add wait-for-it for dependencies in docker-compose

Should add back some resiliency (kafka can sometimes be a bit flakey to start up)

* Apply suggestions from code review

Co-authored-by: Lucas Pimentel-Ordyna <lucas.pimentel@datadoghq.com>

* Ensure we clean the "deps" directory too

* Fix macos publish directory

* Rename docker.sh -> build_in_docker.sh

* Port Arm64 throughput tests

* Rename IntegrationTests2 -> NukeIntegrationTests

* Restore ReferenceAssemblies reference for dependency libs

The .NET Framework 2.0 apps need them, as the SDK doesn't seem to include them  by default

* Tweak IIS sample runner utility

* Skip building Oracle projects in CompileSamples for now (no drive space)

* Print drive info to try and see where it's all going

* Don't copy CI data around so much

Saves 0.5GB of drive space

* Add option to print drive space for every target

* Revert "Don't copy CI data around so much"

This reverts commit fd10d3a528f44234fccaebe0aeeecc86a7083831.

* Allow overriding the test filter in integration tests

* Revert "Revert "Don't copy CI data around so much""

This reverts commit 82a8afed1ced4ea65d2be5f296679733b5273ef7.

* Fix publish logs when multiple job attempts

Publish logs for other test steps

* Update Version

Co-authored-by: Lucas Pimentel-Ordyna <lucas.pimentel@datadoghq.com>

Include Nuke build project in the sln  (DataDog/dd-trace-dotnet#1508)

* Fix invalid project solution configurations

* Add Nuke project to Sln file

* Fix typo in SetAllVersions

CIApp: Correctness and stability fixes. (DataDog/dd-trace-dotnet#1504)

* Fixes an scenario when a NUnit test suite fails on the OneTimeSetUp method gets reported different than the console output.

* Fixes NUnit integrations.

* Improve render of parameters.

* Improvements.

* Maps InconclusiveException to a SkipStatus.

* Remove unused property.

* Increase NUnit test cases with Assert.Pass and Assert.Inconclusive tests.

Fix race condition in runner pipeline (DataDog/dd-trace-dotnet#1510)

Ensure both span buffer are flushed in the FlushTracesAsync() call. (DataDog/dd-trace-dotnet#1514)

Avoid the trace copy when flushing spans (DataDog/dd-trace-dotnet#1502)

Modify GetReJITParameters callback method to never return E_FAIL HRESULT (DataDog/dd-trace-dotnet#1517)

Add net5.0 support to dd-trace (DataDog/dd-trace-dotnet#1516)

* Add net5.0 target framework.

* Adds a temporal basic nuget description.

Crank sample app profiler detector. (DataDog/dd-trace-dotnet#1518)

* Adds a profiler detector to the crank sample app.

* Exit with an error if the profiler is not correctly attached.

Enable calltarget instrumentation by default on all architectures (DataDog/dd-trace-dotnet#1509)

Delete project NugetProfilerVersionMismatch (DataDog/dd-trace-dotnet#1520)

* remove unused project NugetProfilerVersionMismatch

* update nuke pipeline

Enable ultimate pipeline for PRs (DataDog/dd-trace-dotnet#1515)

* Run ultimate pipeline on all PRs and tags

* Schedule ultimate-pipeline to run every night

Skip stages that aren't required for schedule

* Disable old pipeline triggers

* Create a more obvious `benchmarksOnly` variable, and allow "manually" triggering a benchmarking run

* Allow skipping benchmarks using skip_benchmarks

* Fix environment variable evaluation

Convert GitLab package signing build to use Nuke targets (DataDog/dd-trace-dotnet#1512)

* Add a package signing target, for use in GitLab

* Amend gitlab build script to call Nuke targets

* Tweak info task to run first

* Remove old gitlab build files

* Fix build in Gitlab

Use explicit AnyCPU platform, as otherwise dotnet publish looks in the wrong place
Gitlab doesn't install MSBuild in a well-known location, so set the path manually

* Update GitLab Docker image

Incorporates changes from DataDog/datadog-agent-buildimages#135
Adds web build tools and makes a couple of optimisations

* Try setting VSToolsPath explicitly

* Ensure we sign the dlls that go into the msi and nupkgs

* Sign the NuGet bin folder dlls as they get packed

Previously we were only signing the "published" dlls, but they aren't used during dotnet pack

* Fix default naming of windows-tracer-home.zip

Cosmos Db support (DataDog/dd-trace-dotnet#1473)

* Cosmos Db support

- Call target instrumentation only
- Test only run on windows, since the emulator is not supported on other
  platforms

Upgrade spdlog to 1.4 and move to Static lib version (DataDog/dd-trace-dotnet#1507)

Fix Nuke pipeline and remove some unused sample projects (DataDog/dd-trace-dotnet#1521)

* Fix windows-tracer-home artefact naming

* Replace usage of samples-iis in Nuke pipeline

Replace with direct publish of aspnet projects instead

* Delete unused projects from main sln file

* Remove references to non-existant folders

* Delete files from unused projects on disk

* Split CompileManagedUnitTests in two

Should slightly reduce the amount of space required in integration builds, as we only need to build the test helper projects

* Switch to creating symlinks between build output

Instead of explicit copying to save space

Fix CosomsDb tests and disable PR triggers for old pipelines (DataDog/dd-trace-dotnet#1523)

* Disable PR triggers for old pipelines

Disable scheduled triggers for benchmark/throughput (run by ultimate pipeline)

* Cosmosdb buids package versions

* Start cosmosdb emulator in test pipeline

* Add LinuxUnsupported filter to linux integration tests

* Fix Clean target to handle reparse/junction points

Add automatic instrumentation for AWS SQS (DataDog/dd-trace-dotnet#1471)

- Instrumented Operations
  - CreateQueue
  - DeleteQueue
  - PurgeQueue
  - ReceiveMessage
  - SendMessage
  - SendMessageBatch
- Samples.AWS.SQS added to automated integration tests
- Verified that the instrumentation adds `aws.region` and `aws.requestId` tags when sending SQS requests to an AWS account

Small build improvements (DataDog/dd-trace-dotnet#1524)

* xunit/nunit/mstest samples are built using multi api

So no point building them standalone

* Fix occasional flaky build for xunit sample on arm64

For some (unknown) reason, the arm64 integration test build was occasionally failing with "Type 'AutoGeneratedProgram' already defines a member called 'Main' with the same parameter types"

I can reproduce this locally by running

```
./build_in_docker.sh Clean BuildTracerHome BuildLinuxIntegrationTests
./build_in_docker.sh CompileMultiApiPackageVersionSamples
```

Which makes sense, because the generated .cs file in _other_ obj folders are included when building the samples

What doesn't really make sense is why those obj folders sometimes exist when we're doing a clean checkout

The solution in this commit resolves the problem anyway

* Ignore NuGet package related warnings in CompileMultiApiPackageVersionSamples

These are "expected", in that we're using net461 packages on netcoreapp deliberately, and are using "out of bounds" versions of Newtonsoft.Json because we're deliberately using downgrade versions of dependencies.

The warnings are all surfaced in the AzDo UI, so this recuces the clutter

* Simplify Nuke project yml includes

* Set ARFLAGS so that we don't get stderr messages for things that are fine

I just repeated the same flags they were trying to use here https://github.com/google/re2/blob/master/Makefile#L23 but aren't using them properly (I think)

* Fix AWS SQS tests (not starting SQS container)

* Ignore some known warnings when building samples

* Don't use obsolete interface

* Remove unnecessary global /nowarn:NU1608

* Fix ValueTuple reference

* Ignore use of obsolete Mongo API which we have to use

* Fix version conflict instead of ignoring it

* enable -> restore

Fix xUnit 2.2.0 test flakyness (DataDog/dd-trace-dotnet#1526)

Custom native IsAlreadyLoaded() implementation for .NET 4.5.x (DataDog/dd-trace-dotnet#1528)

* Split IsAlreadyLoaded() implementation depending of the .NET Framework version (<4.6)

* change comment.

Adds CorLib detection for assembly ref (DataDog/dd-trace-dotnet#1522)

* Adds CorLib detection and use that on assembly ref instead of forcing mscorlib reference.

* Change AppDomain.Load to Assembly.Load

* Changes from review.

* fix comment.

Downgrade the version of cmake used for MacOs builds (DataDog/dd-trace-dotnet#1529)

Version 3.20.x causes the loader dll and pdb to not be embedded correctly.
You can confirm this behaviour by running `xcrun size -x -l -m Datadog.Trace.ClrProfiler.Native.dylib` and checking for the `Segment binary` section. In a bad build (with 3.20.x) it will be missing

Revert "Enable calltarget instrumentation by default on all architectures (#1509)" (DataDog/dd-trace-dotnet#1532)

This reverts commit 1ec618a.

Small build improvements (DataDog/dd-trace-dotnet#1531)

* Increase dependency warmup timeout to reduce flakiness

* Ignore CMake temporary build files (macos)

* Don't need the static.a file in the tracer home

* Simplify/Unify the build tracer home step for linux + mac

Previously we were copying files _into_ the native project's bin folder, and then publishing from there. This change makes things a bit simpler in the pipeline as well as being more consistent wih the output in general.

- Only publish the net4 targets in the managed build step on Windows (not required for linux)
- Copy the files _to_ the tracer home directly,  _from_ the bin output, instead of the other way around

* Add build step for generating the tracer home directory in the dd-trace format

dd-trace uses a different folder layout for dd-trace.
This adds an extra Nuke target for creating a directory with the required structure (required for Linux and Mac)

* Add pipeline variable for ddTracerHome

* Fix comment

* Try fix flakiness in Arm64 Linux Integration builds

* Suppress build warnings on macos

NUnit integration improvements (DataDog/dd-trace-dotnet#1533)

* - Adds support for parameterized fixtures.
- Adds support for test declared in a base class.

* - Adds support for test.message tag.

* fix comment.

* Add `test.message`  test

* - Sets the service name to the repo name in case no value is set.

* remove comment.

* - Adds the case when the repo doesn't end with .git
- Adds the repository name for callsite.

* null check.

Add an OWIN test application to the integration-tests pipeline (DataDog/dd-trace-dotnet#1525)

* Treat 501-503 as error codes

* Add test application Samples.Owin.WebApi2

 Uses same endpoints as SamplesAspNetMvc5 WebApi2 endpoints, for easier testing

* Add snapshot tests for Owin app using Verify

* Add verify snapshots for owin test

* Don't build Samples.Owin.WebApi2 on Linux

* Delete duplicate test declaration

* Fix link in build project

* Switch to using WaitHandle instead of Console.Read, as that was failing in AzDo (as didn't actually wait)

Add some extra logging

* Remove the invalid HTTP status codes in aspnetcore tests

Move the tests cases to TracerSettingsTests instead

Fix Enable32Bit flag for IIS container (DataDog/dd-trace-dotnet#1536)

For some reason, this _has_ to be uppercase to be picked up.
Unlike the other arg. Which must _not_ be uppercase...

Fix spans flushing on Testing framework instrumentations (DataDog/dd-trace-dotnet#1535)

* Change instrumentation for flushing spans to the Xunit.Sdk.TestAssemblyFinished ctor

* Fix MsTest RunCleanup integration.


zacharycmontoya:
@pjanotti to fix a lot of the linux integration test failures, you'll need to either set `TestAllPackageVersions=true` or take some of the `Build.Steps.cs` changes from this PR: https://github.com/DataDog/dd-trace-dotnet/pull/1818

The change makes sure that when `TestAllPackageVersions=false`, that all test applications will still be built.

zacharycmontoya:
You may also consider removing the `Add $(dotnetToolTag) build tag` step in `minimal-pipeline.yml` to make the `dotnet_tool` stage pass. The task is not important for your purposes

pjanotti:
Ok, except for CodeQL all green.

pellared (approved):

rajkumar-rangaraj (approved):

PR #223, opened 2021-09-23
Bootstrapping tests

dszmigielski:
Added tests for bootstrapping the tracer via our own code or in the client application. Because of the way the tracer works, every single test has to be run in separate process - this is done in the nuke build.


build/nuke/Build.Steps.cs
```diff
@@ -220,26 +225,57 @@ partial class Build
             );
         });
 
-    private AbsolutePath GetResultsDirectory(Project proj) => BuildDataDirectory / "results" / proj.Name;
```
RassK, line 223: I guess that and following accessors should not be removed

test/OpenTelemetry.ClrProfiler.Managed.Bootstrapping.Tests/InstrumentationTests.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    public class InstrumentationTests
+    {
+        private readonly ActivitySource _otelActivitySource = new("OpenTelemetry.ClrProfiler.*");
+        private readonly ActivitySource _customActivitySource = new("Custom");
+
+        public InstrumentationTests()
+        {
+            // Exporter should always be set to a valid value
+            Environment.SetEnvironmentVariable("OTEL_EXPORTER", "zipkin");
```
pellared, line 16: We should rather support a scenario where the exporter is not set via env var. One can always use a plugin to load it. Can you create an issue for it and reference it in this comment?

dszmigielski: This is a very simple fix, I will just include it here

pellared:
Why `OpenTelemetry.ClrProfiler.Managed.Bootstrapping.Tests.csproj` and not  `OpenTelemetry.ClrProfiler.Managed.Tests.csproj`? Is it to reduce the possible impact of global variables? If there is a reason then I suggest adding a `README.md` explaining it. Otherwise, I think we can name it simply  `OpenTelemetry.ClrProfiler.Managed.Tests.csproj`.

PS. Update the branch 😉

dszmigielski:
> Why `OpenTelemetry.ClrProfiler.Managed.Bootstrapping.Tests.csproj` and not `OpenTelemetry.ClrProfiler.Managed.Tests.csproj`? Is it to reduce the possible impact of global variables? If there is a reason then I suggest adding a `README.md` explaining it. Otherwise, I think we can name it simply `OpenTelemetry.ClrProfiler.Managed.Tests.csproj`.
> 
> PS. Update the branch 😉

I wanted this to be separate project because these tests are quite unique in how every single test has to run in a separate process. If we add more tests to ClrProfilerManaged it probably won't have to do that, causing a lot of mess in how we run the tests. So I just wanted it to be its own thing.

test/OpenTelemetry.ClrProfiler.Managed.Bootstrapping.Tests/InstrumentationTests.cs
```diff
@@ -0,0 +1,69 @@
+using System;
+using System.Diagnostics;
+using OpenTelemetry.Trace;
+using Xunit;
+
+namespace OpenTelemetry.ClrProfiler.Managed.Bootstrapping.Tests
+{
+    public class InstrumentationTests
```
pellared, line 8: wouldn't these tests fail if all are being run using `dotnet test`? 

if so would it not better to e.g. require some env var to run a test like `BOOSTRAPPING_TESTS` and use https://josephwoodward.co.uk/2019/01/skipping-xunit-tests-based-on-runtime-conditions to check if is set?

build/nuke/Build.Steps.cs
```diff
@@ -224,21 +229,52 @@ partial class Build
...
+    {
+        var project = Solution.GetProject(Projects.Tests.ClrProfilerManagedBootstrappingTests);
+
+        const string testPrefix = "OpenTelemetry.ClrProfiler.Managed.Bootstrapping.Tests.InstrumentationTests";
+        var testNames = new[] {
+            "Initialize_WithDisabledFlag_DoesNotCreateTracerProvider",
+            "Initialize_WithDefaultFlag_CreatesTracerProvider",
+            "Initialize_WithEnabledFlag_CreatesTracerProvider",
+            "Initialize_WithPreviouslyCreatedTracerProvider_WorksCorrectly"
+        }.Select(name => $"{testPrefix}.{name}");
```
pellared, line 263: how about using reflection to get all the test names? otherwise, we can easily stop running some tests if e.g. we add more of them

pellared: or at least give a comment in `test/OpenTelemetry.ClrProfiler.Managed.Bootstrapping.Tests/InstrumentationTests.cs` to make sure that the test names are synced

test/OpenTelemetry.ClrProfiler.Managed.Bootstrapping.Tests/InstrumentationTests.cs
pellared: I see that this pattern is used in other places `[Trait("Something", "True")]`

dszmigielski: I'm not really sure how that would work. I mean, yes if you just run dotnet test on the whole projects 3 out of 4 tests will fail, but I don't know how adding the env variable can resolve that.

pellared: I can try making a PR tomorrow morning.

pellared (approved):

pellared:
@pjanotti PTAL

test/OpenTelemetry.ClrProfiler.Managed.Bootstrapping.Tests/InstrumentationTests.cs
nrcventura: Another alternative is to consider this as more of an integration test and in the test you can programmatically start a new process with the appropriate environment variables. It will likely be more complicated if you need more "bootstrapping" code to get a test like that working.

nrcventura: If you are able to add a layer of indirection to the call to `OpenTracing.Util.GlobalTracer.RegisterIfAbsent` in the `Initialize` method you'd be able to swap out that call with a mocked call here and not have to worry about managing a different process or using a different test project. You just need to ensure that the original call is replaced after the test completes.

pellared: I pushed my proposal directly to your branch.
Let me know what yoy think: 9eaac502f

pellared: > Another alternative is to consider this as more of an integration test and in the test you can programmatically start a new process with the appropriate environment variables

I prefer the current approach as running as a separate process would also test the Profiler+Loader code. The current approach is more granural.

PR #224, opened 2021-09-27
ASP.NET (.NET Framework) integration tests

RassK:
### Key features

* Hosts ASP.NET test application in Windows Docker container (with IIS)
* `MockZipkinCollector` lives in the host, so could be quickly spinned up and reuse built image for a new container.


test/test-applications/integrations/aspnet/Samples.AspNet/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM mcr.microsoft.com/dotnet/framework/aspnet:4.8-windowsservercore-ltsc2019
```
zacharycmontoya, line 1: @RassK I think you may be running into the issue presented here, and here's a link to a workaround script you can run at startup: https://github.com/docker/for-win/issues/1976#issuecomment-585423014

The issue is present only on windowsservercore docker images. If it was possible to switch to a nanoserver image you could switch to that, but I don't think there are aspnet images based on nanoserver...

zacharycmontoya: I am surprised that you are still running into issues after setting up a docker network like you did because the linked issue pertains to the internal DNS resolving `gateway.docker.internal` and `host.docker.internal`, but maybe it is worth trying the workaround

RassK: In a mean time I simplified it a bit. Creating it's own nat network with fixed gateway address, but even a fix IP could not get connection up. I have created 2 labs W10 based and WS2019 based, both work locally, but in GHA it just don't connect.

RassK: Update: I can see that host VM definitely attaches the gateway IP and health check request from host to host gateway endpoint works. So the questions remains what blocks the request from container to gateway IP.

RassK: Update: Seems not a network issue too. Thanks to the Github support verified that nanoserver image can run curl against healt-check endpoint. Still not sure why request gets blocked in Github actions with windows server core image.

RassK: @pellared @zacharycmontoya from the latest updates seems it's a classic issue: Profiler is not attaching. Somehow in local environments it's fine, but specifically in Github Actions Windows VM + profiled application in docker (aspnet container) causing a problem. Could you have a look?

build/nuke/Extensions/SolutionExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+    public static class SolutionExtensions
+    {
+        public static Project[] GetIntegrationTests(this Solution solution)
+        {
+            return solution
+                .GetProjects("IntegrationTests.*")
+                // Skip helpers project
+                .Where(x => !x.Name.Equals(Projects.Tests.IntegrationTestsHelpers, StringComparison.OrdinalIgnoreCase))
+                // Skip .NET Fx integrations
+                .Where(x => !x.Path.ToString().Contains("aspnet", StringComparison.OrdinalIgnoreCase))
```
nrcventura, line 16: Should the name we're checking here be more specific? It seems like this could accidentally exclude aspnet core tests.

RassK: @nrcventura true, what do you think of checking directory path instead of project path? 
Changing `x => x.Path` to `x => x.Directory`. It's not going to eliminate the problem, but should reduce naming conflicts greatly - as only folders match.

nrcventura: Would it be just as easy to use the TargetFrameworks [Project extension method](https://nuke.build/api/Nuke.Common/Nuke.Common.ProjectModel.ProjectExtensions.html#Nuke_Common_ProjectModel_ProjectExtensions_GetTargetFrameworks_Nuke_Common_ProjectModel_Project_) to perform the check? I think that would more closely match your intention.

RassK: @nrcventura thanks! I have been looking for this functionality for a while. That's of curse much better.

RassK: UPD: @nrcventura unfortunately TargetFramework doesn't tell us if it's ASP.NET FX or .NET FX Console. Currently all except StartupHook is including .net461. If you have any better ideas, we can try these but otherwise I would go with `x.Directory` option and come back later to figure out how to group asp.net fx, .net fx console and others

nrcventura: I was thinking that we might be able to do something like (in pseudocode) `.Where(x => !x.TargetFrameworks.All(f => f.ContainsSubstring("net4"))`, so that we are excluding projects where all of the target frameworks are .net framework (as opposed to .net core, .net standard, or .net 5+).

build/nuke/Build.Steps.cs
```diff
@@ -304,15 +306,16 @@ private void RunBootstrappingTests()
 
     private void RunIntegrationTests()
     {
-        Project[] integrationTests = Solution
-            .GetProjects("IntegrationTests.*")
-            .ToArray();
+        Project[] integrationTests = Solution.GetIntegrationTests();
+
+        string filter = IsWin ? null : "WindowsOnly=true";
```
zacharycmontoya, line 311: I don't really understand this condition. Is this supposed to exclude IIS or Windows-only tests when being run on other OSes?

build/nuke/Extensions/SolutionExtensions.cs
zacharycmontoya: Another solution could be to rename the folder to `aspnetfx`, so we can distinguish between that and `aspnetcore`

zacharycmontoya (approved):
LGTM, nice work!

build/nuke/Extensions/SolutionExtensions.cs
RassK: `project.GetTargetFrameworks()` actually gets me idea, that it's based on `project.GetProperty()` probably. which means I might be able to do something like `.Where(x => !x.GetProperty<bool>("IsAspNetIntegration"))`

build/nuke/Build.Steps.cs
RassK: Yep, this always happens with filter naming, depends if that filter is inclusive or exclusive. This filter here is exclusive, so it filters out tests with metadata `"WindowsOnly=true"`. Any ideas to improve readability are welcome.

build/nuke/Extensions/SolutionExtensions.cs
zacharycmontoya: If you have some way to define the `IsAspNetIntegration` property without needing to remember to update each project file, that could work. I'd recommend checking for a property that is built-in or relying on our folder/naming structure so it will be hard to mess up

build/nuke/Build.Steps.cs
zacharycmontoya: I thought these XUnit filters were inclusive. It looks like in the GitHub Actions runs for this PR, in the Ubuntu run for example, a lot of tests are not being run because it's only selecting tests with the "WindowsOnly=true" attribute. Maybe you could do the following

```suggestion
        string filter = IsWin ? null : "WindowsOnly!=true";
```

build/nuke/Extensions/SolutionExtensions.cs
RassK: The plan is to add props file to aspnet folder, so all projects auto gain that prop. Have to test it first, in theory it works, but if in practice `project.GetProperty()` is pure csproj xml parsing then I guess it wont work. The same thing is actually with TargetFrameworks as these are inherited also from some lower level .props and not explicitly defined in csproj.

build/nuke/Build.Steps.cs
pjanotti: My understand is if you add a trait it will only run the test if you explicitly set the trait. If the test doesn't have any trait it runs unless the test is filtered out. So it seems to me: 

1. Add the `[Trait("WindowsOnly", "True")]`
2. Set the filter to `WindowsOnly=True`

Is this incorrect?

test/integration-tests/IntegrationTests.Helpers/DockerNetworkHelper.cs
```diff
@@ -0,0 +1,57 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Docker.DotNet;
+using Docker.DotNet.Models;
+
+namespace IntegrationTests.Helpers
+{
+    internal static class DockerNetworkHelper
```
pjanotti, line 9: It will be good to have some comments describing what this class is doing.

pjanotti (approved):
Very good @RassK - some comments for the helper types will be good.

build/nuke/Extensions/SolutionExtensions.cs
nrcventura: If we are able to use the prop approach, should the prop be more generic than `IsAspNetIntegration` and follow the `WindowsOnly` approach used elsewhere?

build/nuke/Build.Steps.cs
RassK: Fixed.

build/nuke/Extensions/SolutionExtensions.cs
RassK: Added `WindowsOnly` prop to `Directory.Build.props`, seems working fine.

nrcventura (approved):

zacharycmontoya:
Merging now that the Mongo fix has been merged

PR #227, opened 2021-09-30
Bootstrap OpenTelemetry SDK Using Dotnet Startup Hook

### LINKED ISSUE #225: Bootstrap OpenTelemetry SDK Using Dotnet Startup Hook
<H3>.NET Core Startup Hook Proposal</H3>

.NET Core 3+ and .NET 5+ provides a low-level hook that allows injecting managed code to run before the main application's entry point. Take advantage of dotnet startup hook to bootstrap OpenTelemetry SDK to an application.

Currently, OpenTelemetry Dotnet Instrumentation has a requirement to include System.Diagnostics.DiagnosticSource, or other instrumentation package reference at build time to instrument an application with OpenTelemetry SDK. Proposal here is to use dotnet startup hook to load OpenTelemetry SDK dynamically into an application without taking package references. 

```text
+-------------------+      DOTNET STARTUP HOOK     +-------------------------+
|    Application    |   <-----------------------   |    OpenTelemetry SDK    |
+-------------------+                              +-------------------------+
```
<H3>Advantages</H3>

- **Zero-Touch Enablement**: Set `DOTNET_STARTUP_HOOKS` environment variable value to the location of startup hook assembly to inject OpenTelemetry SDK to an application
- **Better performance**: Low performance cost for an application that does not require byte-code instrumentation
- **Low maintenance**: This helps always point to the latest version of instrumentation library. Customer don't need to worry about updating Nuget package reference in their app

<H3>Limitation</H3>

- If the application has a reference to lower version of `System.Diagnostics.DiagnosticSource` than the one referenced in OpenTelemetry SDK. This approach may not work. Workaround is to refer to the same `System.Diagnostics.DiagnosticSource` in package reference as of OpenTelemetry. This limitation could be overcome by creating a proxy or adaptor to convert the `Activity` created in lower version to OpenTelemetry version. Will propose this in a new issue.  

<H3>Reference</H3>

[Dotnet Host Startup Hook](https://github.com/dotnet/runtime/blob/main/docs/design/features/host-startup-hook.md)

rajkumar-rangaraj:
Fixes #225 

* Added dotnet startup hook project. This project loads `OpenTelemetry.ClrProfiler.Managed` assembly through reflection and calls `Instrumentation.Initialize()` to load the OpenTelemetrySDK
* Added sample app `Samples.StartupHook.AspNetCoreMvc31` which has ability to send data to both console/zipkin exporters with ASP.NET and HTTP Client Instrumentation.
* Uses DOTNET_STARTUP_HOOKS environment variable to take the location of startup hook assembly.

Below is the snapshots for data logged in zipkin

image


samples/Samples.StartupHook.AspNetCoreMvc31/Properties/launchSettings.json
```diff
@@ -0,0 +1,49 @@
...
+  },
+  "profiles": {
+    "IIS Express": {
+      "commandName": "IISExpress",
+      "launchBrowser": true,
+      "environmentVariables": {
+        "ASPNETCORE_ENVIRONMENT": "Development",
+
+        "DOTNET_STARTUP_HOOKS": "$(StartupHookOutputDirectory)\\netcoreapp3.1\\OpenTelemetry.Instrumentation.StartupHook.dll",
+        "OTEL_MANAGED_PROFILER_PATH": "$(ManagedProfilerOutputDirectory)\\netcoreapp3.1",
```
RassK, line 18: This part is a bit conflicting with the nature of auto instrumentation.  Currently we do not specify managed framework, it's picked up automatically by the loader. Requiring this would need app local settings always to be present, instead of using global one.

RassK (commented):
I think the loader included in ClrProfiler.Native must be disabled as well in case of netcoreapp

RassK (commented):
"OTEL_MANAGED_PROFILER_PATH" could be replaced with existing "OTEL_DOTNET_TRACER_HOME" if netframework folder could be auto chosen.

samples/Samples.StartupHook.AspNetCoreMvc31/Program.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Samples.StartupHook.AspNetCoreMvc31
+{
+    public class Program
```
pellared, line 12: How about adding an integration test that would use this sample to check if `OpenTelemetry.Instrumentation.StartupHook` is working?

pellared (commented):
I just made a very quick look. I will review it more carefully later, but looks very nice 👍

pellared:
Have you tested what happens if both StartupHook and .NET Profiler are configured (at the same time)?

RassK:
> Have you tested what happens if both StartupHook and .NET Profiler are configured (at the same time)?

Both of the solutions are based on `Assembly.Load`, I guess the default behaviour was to ignore already loaded assembly. 
I think the better question is what happens with native layer, if it's not properly configured.

samples/Samples.StartupHook.AspNetCoreMvc31/Properties/launchSettings.json
rajkumar-rangaraj: I agree. Called this out in the PR description. Will use `OTEL_DOTNET_TRACER_HOME`.

samples/Samples.StartupHook.AspNetCoreMvc31/Program.cs
rajkumar-rangaraj: Great idea! Will get this covered in the next PR.

rajkumar-rangaraj:
> > Have you tested what happens if both StartupHook and .NET Profiler are configured (at the same time)?
> 
> Both of the solutions are based on `Assembly.Load`, I guess the default behaviour was to ignore already loaded assembly. I think the better question is what happens with native layer, if it's not properly configured.

I might need your help @pellared @RassK to check this out. Will reach out to you on slack.

src/OpenTelemetry.Instrumentation.StartupHook/AssemblyResolver.cs
```diff
@@ -0,0 +1,34 @@
...
+        internal static string OtelManagedProfilerPath => Environment.GetEnvironmentVariable("OTEL_DOTNET_TRACER_HOME");
+
+        public static Assembly LoadAssemblyFromSharedLocation(AssemblyLoadContext context, AssemblyName assemblyName)
+        {
+            string sharedAssemblyPath = null;
+            try
+            {
+                // TODO: If dotnet does not like the provided assembly, it will retry this event in a loop.
+                // Add a logic to avoid cyclic loop.
+                sharedAssemblyPath = Path.Combine(OtelManagedProfilerPath, $"{assemblyName.Name}.dll");
```
pjanotti, line 19: This code is very similar to the `Managed.Loader` likely it can be combined. That said I think it is better to do it in a separate PR. Perhaps after adding the tests suggested by Robert.

pjanotti (commented):
@rajkumar-rangaraj Could you, please, add a md file to the folder of he app describing how to use the sample and the scenarios that it tries to address? (eg. not using CLR Profiler, etc)

src/OpenTelemetry.Instrumentation.StartupHook/StartupHook.cs
```diff
@@ -0,0 +1,50 @@
...
+/// </summary>
+public class StartupHook
+{
+    /// <summary>
+    /// Load and initialize OpenTelemetry.ClrProfiler.Managed assembly to bring OpenTelemetry SDK
+    /// with a pre-defined set of exporters, shims, and instrumentations.
+    /// </summary>
+    public static void Initialize()
+    {
+        AssemblyLoadContext.Default.Resolving += AssemblyResolver.LoadAssemblyFromSharedLocation;
```
zacharycmontoya, line 18: FYI this is a behavior change from the existing loader that adds an event handler to `AppDomain.CurrentDomain.AssemblyResolve`. Should this change be adopted now in this PR or added separately?

src/OpenTelemetry.ClrProfiler.Managed/OpenTelemetry.ClrProfiler.Managed.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFrameworks>net461;netstandard2.0;netcoreapp3.1</TargetFrameworks>
     <Version>0.0.1</Version>
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
```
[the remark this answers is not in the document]
rajkumar-rangaraj: This is not needed, removed it. Using `PublishManagedProfiler` to copy the assembly to tracer home folder.

src/OpenTelemetry.Instrumentation.StartupHook/StartupHook.cs
rajkumar-rangaraj: Will get this covered in different PR, Paulo has also recommended the same.

src/OpenTelemetry.Instrumentation.StartupHook/AssemblyResolver.cs
```diff
@@ -0,0 +1,34 @@
+using System;
+using System.IO;
+using System.Reflection;
+using System.Runtime.Loader;
+
+namespace OpenTelemetry.Instrumentation.StartupHook
+{
+    internal class AssemblyResolver
+    {
+        internal static string OtelManagedProfilerPath => Environment.GetEnvironmentVariable("OTEL_MANAGED_PROFILER_PATH");
```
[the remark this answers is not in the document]
rajkumar-rangaraj: @pellared @RassK Removed reference to `OTEL_DOTNET_TRACER_HOME`. Startup hook will live in the same location as of managed profiler. Changed `PublishManagedProfiler` scripts to include StartupHook library.

src/OpenTelemetry.Instrumentation.StartupHook/AssemblyResolver.cs
rajkumar-rangaraj: When we move on to DiagnosticSource Activity proxy, this part will go through a big change to address the context. We will use LoadFile instead of LoadFrom. It's better to combine this as a part of DiagnosticSource proxy work.

rajkumar-rangaraj:
> > Have you tested what happens if both StartupHook and .NET Profiler are configured (at the same time)?
> 
> Both of the solutions are based on `Assembly.Load`, I guess the default behaviour was to ignore already loaded assembly. I think the better question is what happens with native layer, if it's not properly configured.

Here are my test results

Startup hook and .NET profiler - .NET profiler kicks in first and loads all the modules. StartupHook still calls Initialize, but it will return immediately as the tracer is already configured. It is as good as startuphook is ignored.

src/OpenTelemetry.Instrumentation.StartupHook/AssemblyResolver.cs
pjanotti: @rajkumar-rangaraj this next step is interesting: how to use the OTel SDK when using the Activity proxy? I guess that it comes in the next PR :)

pjanotti (approved):
LGTM - with a minor change to the README.md

src/OpenTelemetry.Instrumentation.StartupHook/StartupHook.cs
```diff
@@ -0,0 +1,74 @@
...
+            StartupHookEventSource.Log.Error($"Error in StartupHook initialization: ProfilerFolderLocation: {StartupAssemblyLocation}, Error: {ex}");
+        }
+    }
+
+    private static string GetStartupAssemblyLocation()
+    {
+        try
+        {
+            var startupAssemblyFilePath = Assembly.GetExecutingAssembly().Location;
+            if (startupAssemblyFilePath.StartsWith(@"\\?\"))
```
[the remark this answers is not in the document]
rajkumar-rangaraj: FullPath doesn't simplify path for relative UNC paths, thus removing the prefix. Issue happens where MAX_PATH gets with long path and relative path combined

Example: 
Console.WriteLine(Path.GetFullPath(Path.Combine(@"c:\a\b\c", @"..\")))
// c:\a\b\
Console.WriteLine(Path.GetFullPath(Path.Combine(@"\\?\c:\a\b\c", @"..\")))
// \\?\c:\a\b\c\..\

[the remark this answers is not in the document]
pjanotti: ah, makes sense. Thanks.

PR #229, opened 2021-10-06
Test against .NET 5.0

RassK:
Finalizes half way supported net5.0


nrcventura:
One thing that isn't clear to me is how this improves the net5.0 support. The previous code would load the netcoreapp3.1 code which should be compatible. If this does improve the support, how should we handle newer versions like net6.0?

rajkumar-rangaraj:
> One thing that isn't clear to me is how this improves the net5.0 support. The previous code would load the netcoreapp3.1 code which should be compatible. If this does improve the support, how should we handle newer versions like net6.0?

I got the same question. We should build a library against the lower version of supported .NET Core. This can be loaded by application that are running on .NET5.0 or 6.0. 
We might need to have our tests app validating all the available framework.

RassK:
I just created this PR to start discussions around target frameworks. It's a quick PR and can be thrown away if not needed.

Currently the biggest benefit is just running unit / integrations tests against net 5.0.
Not sure what are the performance gains if an end user actually references 5.0 libs instead of fallback libs. Usually there are benefits using latest framework, but I'm not sure if they apply in this case.

RassK:
> +1 to adding tests against .NET 5.0
> 
> Let's stick with just building for netcoreapp3.1 unless we want to adopt newer libraries for optimizations, etc.

As @pjanotti pointed out earlier to me, NET 5.0 is not LTS version, so we might think again when .NET 6.0 is released.

build/nuke/Extensions/TargetFrameworksExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+        public static IEnumerable<TargetFramework> FilterNetStandard(this IEnumerable<TargetFramework> frameworks)
+        {
+            return frameworks.Except(TargetFramework.NetStandard);
+        }
+
+        public static IEnumerable<TargetFramework> CrossPlatformTestable(this IEnumerable<TargetFramework> frameworks)
+        {
+            return frameworks
+                .FilterWindowsOnly()
+                .FilterNetStandard();
```
nrcventura, line 22: Why are we excluding netstandard 2.0 from the corss platform tests? Netstandard 2.0 libraries should work across multiple platforms.

build/nuke/Extensions/TargetFrameworksExtensions.cs
```diff
@@ -0,0 +1,25 @@
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Extensions
+{
+    public static class TargetFrameworksExtensions
+    {
+        public static IEnumerable<TargetFramework> FilterWindowsOnly(this IEnumerable<TargetFramework> frameworks)
```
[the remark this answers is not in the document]
pellared: How about `ExceptNetFramework`?

build/nuke/Extensions/TargetFrameworksExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            return frameworks.Except(TargetFramework.WindowsOnly);
+        }
+
+        public static IEnumerable<TargetFramework> FilterNetStandard(this IEnumerable<TargetFramework> frameworks)
```
pellared, line 13: How about renaming to `ExceptNetStandard`?

build/nuke/TargetFramework.cs
```diff
@@ -9,11 +9,21 @@
...
+    public static readonly TargetFramework NETCOREAPP2_1 = new TargetFramework { Value = "netcoreapp2.1" };
+    public static readonly TargetFramework NETCOREAPP3_1 = new TargetFramework { Value = "netcoreapp3.1" };
+    public static readonly TargetFramework NET5_0 = new TargetFramework { Value = "net5.0" };
+
+    public static readonly TargetFramework[] NetStandard = new[]
+    {
+        NETSTANDARD2_0
+    };
+
+    public static readonly TargetFramework[] WindowsOnly = new[]
```
pellared, line 23: How about renaming to `NetFramework` ?

samples/ConsoleApp.SelfBootstrap/ConsoleApp.SelfBootstrap.csproj
```diff
@@ -15,7 +15,7 @@
     <PackageReference Include="OpenTelemetry.Instrumentation.Http" Version="1.0.0-rc7" />
     <PackageReference Include="OpenTelemetry.Instrumentation.SqlClient" Version="1.0.0-rc7" />
     <PackageReference Include="OpenTelemetry.Instrumentation.AspNet" Version="1.0.0-rc7" Condition="'$(TargetFramework)' == 'net461'" />
-    <PackageReference Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.0.0-rc7" Condition="'$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'netcoreapp3.1'" />
+    <PackageReference Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.0.0-rc7" Condition="'$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'netcoreapp3.1' OR '$(TargetFramework)' == 'net5.0'" />
```
pellared, line 18: maybe this will be simpler?

```suggestion
<PackageReference Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.0.0-rc7" Condition="'$(TargetFramework)' != 'net461'" />
```

test/OpenTelemetry.ClrProfiler.Managed.Loader.Tests/StartupTests.cs
```diff
@@ -11,17 +11,17 @@ public class StartupTests
         public void Ctor_LoadsManagedAssembly()
         {
             var directory = Directory.GetCurrentDirectory();
-            Environment.SetEnvironmentVariable("OTEL_DOTNET_TRACER_HOME", $"{directory}/Profiler");
-
-            Console.WriteLine("Setting directory >> " + $"{directory}/Profiler");
+            Environment.SetEnvironmentVariable("OTEL_DOTNET_TRACER_HOME", Path.Combine(directory, "..", "Profiler"));
```
pellared, line 14: is this changed caused by this one in `build/nuke/Build.Steps.cs`:
```
.SetOutput(TestsDirectory / Projects.Tests.ClrProfilerManagedLoaderTests / "bin" / BuildConfiguration / "Profiler" / framework)));
```

?

build/nuke/Extensions/TargetFrameworksExtensions.cs
pellared: I think that `netstandard2.0` is not needed at all. See: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/232

rajkumar-rangaraj (approved):
LGTM

PR #230, opened 2021-10-07
StartupHook - Replace instrumentation with loader.

rajkumar-rangaraj:
Follow up PR #227 

CLR native profiler loads the managed loader which initializes the Instrumentation. Proposal in this PR is to follow the same guidelines for StartupHook. Removed the instrumentation initialization and adding it through loader. 

* Renamed `OpenTelemetry.Instrumentation.DotnetStartupHook` to `OpenTelemetry.Instrumentation.StartupHook`
* Add loader library to tracer home for startuphook to consume
* Removed StartupHook assembly resolver

### DIFF
--- a/build/nuke/Build.Steps.cs
+++ b/build/nuke/Build.Steps.cs
@@ -155,8 +155,10 @@ partial class Build
                     .SetFramework(framework)
                     .SetOutput(TracerHomeDirectory / framework)));
 
+            // StartupHook is supported starting .Net Core 3.1.
+            // We need to emit StartupHook and ClrProfilerManagedLoader assemblies only for .NET Core 3.1 target framework.
             DotNetPublish(s => s
-                .SetProject(Solution.GetProject(Projects.DotnetStartupHook))
+                .SetProject(Solution.GetProject(Projects.StartupHook))
                 .SetConfiguration(BuildConfiguration)
                 .SetTargetPlatformAnyCPU()
                 .EnableNoBuild()
@@ -164,6 +166,7 @@ partial class Build
                 .SetFramework(TargetFramework.NETCOREAPP3_1)
                 .SetOutput(TracerHomeDirectory / TargetFramework.NETCOREAPP3_1));
 
+            // ClrProfilerManagedLoader publish is needed only for .Net Core 3.1 to support load from StartupHook.
             DotNetPublish(s => s
                 .SetProject(Solution.GetProject(Projects.ClrProfilerManagedLoader))
                 .SetConfiguration(BuildConfiguration)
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry.Instrumentation.StartupHook/StartupHook.cs
```diff
@@ -1,58 +1,46 @@
...
             else
             {
-                StartupHookEventSource.Log.Error("StartupHook initialization has failed!");
+                StartupHookEventSource.Log.Trace("OpenTelemetry.ClrProfiler.Managed.Instrumentation initialized before startup hook");
             }
         }
         catch (Exception ex)
         {
-            StartupHookEventSource.Log.Error($"Error in StartupHook initialization: ProfilerFolderLocation: {StartupAssemblyLocation}, Error: {ex}");
+            StartupHookEventSource.Log.Error($"Error in StartupHook initialization: LoaderFolderLocation: {loaderAssemblyLocation}, Error: {ex}");
```
pellared, line 39: TL;DR; I just suggest adding a `throw` below and also do the same in other places.

I think we have agreed that if the initialization goes wrong then we throw an exception instead of trying to continue. Example here: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/poc-otel-sdk/src/OpenTelemetry.ClrProfiler.Managed/Instrumentation.cs

For sure we should document this behavior. If we want to change the behavior we should have a discussion around it, change it, and document it again.

PS. There is an ongoing discussion about what OTel SDK should do if an improper env var is set. See: https://github.com/open-telemetry/opentelemetry-dotnet/pull/2395

pellared (commented):
nice 👍 just one aspect (the error-handling) needs to be addressed

build/nuke/Build.Steps.cs
```diff
@@ -163,6 +163,15 @@ partial class Build
                 .EnableNoRestore()
                 .SetFramework(TargetFramework.NETCOREAPP3_1)
                 .SetOutput(TracerHomeDirectory / TargetFramework.NETCOREAPP3_1));
+
+            DotNetPublish(s => s
```
RassK, line 167: note: one version of the same assembly is already embedded to native.dll (not sure if worth separating?)

src/OpenTelemetry.Instrumentation.StartupHook/StartupHook.cs
rajkumar-rangaraj: No need to throw from `GetLoaderAssemblyLocation()` we return null if directory path is not present and it continues it execution.

rajkumar-rangaraj: Changed a code, little to throw from `GetLoaderAssemblyLocation()`.

build/nuke/Build.Steps.cs
RassK: I'm not saying it's directly issue with this PR, just pointing out on the way some comments. There is no better way than direct marks in the code. 

All I'm saying is that thanks to StartupHook, the loader is only necessary for Windows net461. Currently it's binary is shipped embedded to the dll + separately in netcoreapp3.1 deploy (so it's just a question of optimized shipping). Linux and Mac deploy could actually remove it completely simplifying native build, but then StartupHook would need some refactoring. 

At first I thought StartupHook is going to be the main way loading, because net461 is just the only exceptional case.

PR #247, opened 2021-10-15
Test if CMake downgrade is still needed

pellared:
We may run this from time to time to monitor when CMake downgrade will not be needed.


RassK:
> We may run this from time to time to monitor when CMake downgrade will not be needed.

There is a step for MacOS also. I see currently Linux step is removed only.

PR #257, opened 2021-10-19
Fix CMake issue with embedding loader

zacharycmontoya:
The existing issue is that upgrading CMake to 3.20.x causes the managed loader to not be embedded properly into the profiler dll. This fixes the root issue in the CMake file to correctly declare the object to embed.

The fix is ported from https://github.com/DataDog/dd-trace-dotnet/pull/1899
This should enable us to remove the CMake downgrade as proposed in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/247


pellared:
Can you remove the downgrade steps and scripts as well?

rajkumar-rangaraj (approved):

zacharycmontoya:
Closing in favor of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/259, which can be up-to-date with main per the branch protection rules

PR #259, opened 2021-10-20
Fix CMake issue with embedding loader (updated PR)

zacharycmontoya:
Supercedes #257 because the branch needed to be updated to main, and the branch policies disallowed branch updates
Supercedes #247 by implementing the removal of CMake downgrade scripts

The existing issue is that upgrading CMake to 3.20.x causes the managed loader to not be embedded properly into the profiler dll. This fixes the root issue in the CMake file to correctly declare the object to embed.

The fix is ported from DataDog/dd-trace-dotnet#1899

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,9 +71,6 @@ jobs:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v2.3.5
-      - name: Update GitHub PATH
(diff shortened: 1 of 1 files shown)


pjanotti (approved):

pjanotti (approved):
Re-approving since now it is even better and already removes the CMake downgrade :)

PR #267, opened 2021-11-09
[WIP] Diagnostic Source Proxy Skeleton

### LINKED ISSUE #260: Diagnostic Source assembly versioning Issue
<H3>Dotnet Assembly Load Explained</H3>

Dotnet core runtime automatically load assemblies with a higher version at runtime. Below is an example of loading  `System.Diagnostics.DiagnosticSource` assembly. 

* Framework - `System.Diagnostics.DiagnosticSource.dll` is automatically referenced from framework.
* PackageReference  -  `System.Diagnostics.DiagnosticSource.dll` is referenced from NuGet package.

| Scenario                                  | Framework | PackageReference |
|-------------------------------------------|:---------:|:-------:|
| No PackageReference in project           | ✔️        |         |
| PackageReference <= Framework | ✔️        |         |
| PackageReference > Framework |           | ✔️      |

Dynamically loading `System.Diagnostics.DiagnosticSource.dll` with a higher version than the one from framework/package will crash the process. The  table below covers the scenario.

| Dynamic Load Scenario                               | Status |
|---------------------------------------------|--------|
| Dynamic Load <= Framework/PackageReference  | ✔️     |
| Dynamic Load > Framework/PackageReference    | ❌      |

✔️ 🠮 Success | ❌ 🠮 Crashes an application

<H3>Demo</H3>

* This console app is built using .NET Core 3.1 and does **not** have any package reference. 
* At build time, .NET creates a Trusted Platform Assembly (TPA) list. This list has an entry to load DiagnosticSource from the framework (`C:\Program Files\dotnet\shared\Microsoft.NETCore.App\3.1.20\System.Diagnostics.DiagnosticSource.dll, Version: System.Diagnostics.DiagnosticSource, Version=4.0.5.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51`). 
* Dynamically loading higher version of `System.Diagnostics.DiagnosticSource` will crash an app. Executing this app will exhibit the behavior.
* The behavior is same if we try to load with Startup Hook or CLR Profiler.

```xml

<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp3.1</TargetFramework>
  </PropertyGroup>

</Project>

```

```csharp
static void Main(string[] args)
{
    try
    {
        string path = @"C:\Users\userprofile\.nuget\packages\system.diagnostics.diagnosticsource\5.0.1\lib\netstandard1.3\System.Diagnostics.DiagnosticSource.dll";
        Debug.Assert(File.Exists(path));
        Assembly.LoadFrom(path);
    }
    catch (Exception e)
    {
        Console.WriteLine($"Exception: {e}");
    }
}

```

```
Exception: System.IO.FileLoadException: Could not load file or assembly 'System.Diagnostics.DiagnosticSource, Version=5.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'.
   at System.Runtime.Loader.AssemblyLoadContext.LoadFromPath(IntPtr ptrNativeAssemblyLoadContext, String ilPath, String niPath, ObjectHandleOnStack retAssembly)
   at System.Runtime.Loader.AssemblyLoadContext.LoadFromAssemblyPath(String assemblyPath)
   at System.Reflection.Assembly.LoadFrom(String assemblyFile)
   at consoleappdiag.Program.Main(String[] args) in D:\temp\consoleappdiag\Program.cs:line 16
```

<H3>Workaround</H3>

* Use dotnet [Additional-deps](https://github.com/dotnet/runtime/blob/main/docs/design/features/additional-deps.md) to provide the reference to `System.Diagnostics.DiagnosticSource` used by the OpenTelemetry SDK. 
* This approach will work if the app does not have a direct or indirect package reference to `System.Diagnostics.DiagnosticSource`. 
* Dotnet supports Framework Dependent apps and Self-Contained apps. Additional-deps will work
only for Framework Dependent apps.

<H3>Proposal</H3>

Load OpenTelemetry referenced `System.Diagnostics.DiagnosticSource.dll` in a different assembly context using `Assembly.LoadFile`. Create a proxy to listen to `Activity`s emitted by a lower version of `DiagnosticSource` and create a clone of the same `Activity` in the OpenTelemetry supported `DiagnosticSource`.  

Enable the proxy feature when an application uses a lower version of `DiagnosticSource` than the OpenTelemetry referenced DiagnosticSource.

Use `Additional-deps` in supported environments.

```text
+--------------------+             +-------------------+    
| DS (Lower Version) |             | DS (OTel Version) |---------
+--------------------+             +-------------------+         |
        |                                 ^                      | OTel supported Activity
        | Activity                        |                      v
        |                +-------+        |             +----------------------+
         --------------> | Proxy | -------              | Otel Instrumentation |
                         +-------+                      +----------------------+
				 
```

rajkumar-rangaraj:
Fixes #260 

Added structure for DiagnosticSource Subscriber/Listener with HttpClient listener. Want to take initial feedback on the project name and the structure of the project.

**Usage**

If the version `DiagnosticSource` referenced by an app is less than OpenTelemetry referenced `DiagnosticSource` then we will perform the following initialization.
```C#
// TODO: Change it to builder pattern to initialize only the listeners that are needed.
var httpClientSubscriber = new DiagnosticSourceSubscriber(new HttpClientDiagnosticSourceListener(), null);
httpClientSubscriber.Subscribe();
```
Every Instrumentation in OpenTelemetry SDK will have its own DiagnosticSourceListener. 
For example, `AddHttpClientInstrumentation` has `HttpClientDiagnosticSourceListener.` `AddAspNetCoreInstrumentation` will have a mapping to `AspnetCoreDiagnosticSourceListener` and so on.


pellared:
Do I understand correctly that this approach would work only with the `DiagnosticSourceListener` that are implemented in this repository? Which means that each instrumentation would need such a listener?

rajkumar-rangaraj:
> Do I understand correctly that this approach would work only with the `DiagnosticSourceListener` that are implemented in this repository? Which means that each instrumentation would need such a listener?

This is correct. We will have listener for each instrumentation. 
Currently, Instrumentation library is getting loaded in app context. We will change it to load it in a different context and create a proxy to cover this scenario.

nrcventura:
The name and overall structure looks good to me.

nrcventura:
One thing to consider, and it might prevent the need for creating/defining listeners for each instrumentation, is following the approach previously described by @macrogreg. An implementation of the approach can be seen in https://github.com/DataDog/dd-trace-dotnet/tree/master/shared/src/managed-lib/DynamicDiagnosticSourceBindings.

rajkumar-rangaraj:
Earlier, I reviewed this code - https://github.com/DataDog/dd-trace-dotnet/tree/master/shared/src/managed-lib/DynamicDiagnosticSourceBindings, it was part of PR #18 . 

To use this approach, we need to fork OpenTelemetry SDK and change it diagnostic source reference to vendored DiagnosticSource library. @macrogreg Please correct me if I'm wrong.

Also, it did not have the solution to remove the DiagnosticSource reference completely from user app. This needs to be handled separately.

pellared:
Per the last SIG meeting, I have a feeling that this PR can be converted to a draft PR. Also, it might be good to add a `[WIP]` of `[Do Not Merge]` prefix. @rajkumar-rangaraj does it make sense?

rajkumar-rangaraj:
Will reopen with actual implementation later.

PR #268, opened 2021-11-10
Cleanup GitHub Actions workflow

pellared:
## What

Refactor the GH Actions workflow to use a "matrix".

See the results: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/1443822406

## Why

1. Less likely to make an error when "copy-pasting" steps between the jobs
1. Easier to ensure consistency
1. Less noise
1. Even currently we are not using the GH workflow generated by Nuke


.github/workflows/ci.yml
```diff
@@ -1,90 +1,26 @@
...
-
 name: ci
 
 on:
   push:
-    branches:
-      - main
-      - 'refs/tags/*'
-    paths:
-      - '!docs/*'
```
zacharycmontoya, line 25: Did you mean to remove the docs exclusion? It seems useful to exclude docs-only changes (unless some checks are being enforced and therefore they would still have to run)

zacharycmontoya (approved):
LGTM

.github/workflows/ci.yml
pellared: > unless some checks are being enforced 

In "legacy main" the builds were required. Now there are not, but I am not sure if this is as expected.

If we do not want to require any builds then I can bring it back.

zacharycmontoya: There's no harm in leaving this change. If we feel like it slows down development, we can always add the exclusion back in later

rajkumar-rangaraj (approved):

nrcventura (approved):

Issue #261, opened 2021-10-25
Unable to auto-instrument aspnetcore 3.1 application


nrcventura:
**Describe the bug**
When attempting to auto-instrument an aspnetcore 3.1+ application you can get an error about a missing dependency.
```
>>>>>>>>>>>>>>>>>>>>>>> Process: dotnet(7582): OpenTelemetry SDK load exception: System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.AspNetCore.Http.Abstractions, Version=3.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.

File name: 'Microsoft.AspNetCore.Http.Abstractions, Version=3.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'
   at OpenTelemetry.Instrumentation.AspNetCore.Implementation.HttpInListener..ctor(AspNetCoreInstrumentationOptions options)
   at OpenTelemetry.Trace.TracerProviderBuilderExtensions.AddAspNetCoreInstrumentation(TracerProviderBuilder builder, AspNetCoreInstrumentationOptions options, Action`1 configure)
   at OpenTelemetry.Trace.TracerProviderBuilderExtensions.AddAspNetCoreInstrumentation(TracerProviderBuilder builder, Action`1 configureAspNetCoreInstrumentationOptions)
   at OpenTelemetry.ClrProfiler.Managed.Configuration.EnvironmentConfigurationHelper.AddSdkAspNetInstrumentation(TracerProviderBuilder builder) in /Users/cventura/github/nrcventura/opentelemetry-dotnet-instrumentation/src/OpenTelemetry.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs:line 56
   at OpenTelemetry.ClrProfiler.Managed.Configuration.EnvironmentConfigurationHelper.<>c.<.cctor>b__5_1(TracerProviderBuilder builder) in /Users/cventura/github/nrcventura/opentelemetry-dotnet-instrumentation/src/OpenTelemetry.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs:line 16
[...]

PR #272, opened 2021-11-12
Additional-deps and runtime package store

rajkumar-rangaraj:
Fixes proposed workaround in #260 

* Added a new project to create [Additional-deps](https://github.com/dotnet/runtime/blob/main/docs/design/features/additional-deps.md) and [runtime package store](https://docs.microsoft.com/en-us/dotnet/core/deploying/runtime-store) in trace-home.
* Updated the sample `samples/Samples.StartupHook.AspNetCoreMvc31/Properties/launchSettings.json` with new environment variable for demonstration. 

```
        "DOTNET_ADDITIONAL_DEPS": "$(SolutionDir)bin\\tracer-home\\additionalDeps",
        "DOTNET_SHARED_STORE": "$(SolutionDir)bin\\tracer-home\\store",
```
* Removed reference to System.Diagnostics.DiagnosticSource from sample app `Samples.StartupHook.AspNetCoreMvc31`, this uses additional deps and loads the library from dotnet shared store.

**TODO**: AdditionalDeps could be easily extended to solve the issue #261

**Tracer-Home directory structure**

image


src/OpenTelemetry.Instrumentation.AdditionalDeps/OpenTelemetry.Instrumentation.AdditionalDeps.csproj
```diff
@@ -0,0 +1,40 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
RassK, line 1: Just out of curiosity, is this `OpenTelemetry.Instrumentation.AdditionalDeps.dll` also important, or it's just matter of cleanup?

RassK:
Are the net5.0 references under `netcoreapp3.1/store/[Platform]/net5.0` important? ... to me it seems like it's not connected only to netcoreapp3.1, maybe worth moving to root dir?

build/nuke/Build.Steps.cs
```diff
@@ -251,6 +256,36 @@ partial class Build
...
+            DotNetPublish(s => s
+                .SetProject(Solution.GetProject(Projects.AdditionalDeps))
+                .SetConfiguration(BuildConfiguration)
+                .SetTargetPlatformAnyCPU()
+                .SetProperty("TracerHomePath", (TracerHomeDirectory / TargetFramework.NETCOREAPP3_1).ToString())
+                .EnableNoBuild()
+                .EnableNoRestore()
+                .CombineWith(TestFrameworks.ExceptNetFramework(), (p, framework) => p
+                    .SetFramework(framework)
+                    .SetOutput(AdditionalDepsDirectory / "shared" / "Microsoft.NETCore.App" / framework.ToString()[^3..] + ".0")));
```
RassK, line 281: Maybe just me but `framework.ToString()[^3..] + ".0"` seems a bit cryptic. Also the additional .0 that seems unusual in first sight. A comment might help there to expalin.

rajkumar-rangaraj:
> Are the net5.0 references under `netcoreapp3.1/store/[Platform]/net5.0` important? ... to me it seems like it's not connected only to netcoreapp3.1, maybe worth moving to root dir?

We need the folder for all support runtimes. Currently we have netcoreapp3.1 and net5.0. When we add support for .net6.0 there will be an additional folder that gets created here. 
Dotnet Runtime Store functionality forces us to have folder structure for all supported runtime to load correctly.

src/OpenTelemetry.Instrumentation.AdditionalDeps/OpenTelemetry.Instrumentation.AdditionalDeps.csproj
rajkumar-rangaraj: `OpenTelemetry.Instrumentation.AdditionalDeps.dll` is a part of deps.json, hence we need to add this to the store to load. There are no functionality in this library that gets invoked at runtime. This library has `DiagnosticSource` dependency and entire probing path is build based on it.

We could remove the need for `OpenTelemetry.Instrumentation.AdditionalDeps.dll`, but the deps.json need to go through a manual edit to remove the entry for `OpenTelemetry.Instrumentation.AdditionalDeps.dll`. This could be achieved by tweaking build steps. If we are inclined towards this approach, want to get this done as a part of follow up PR.

samples/Samples.StartupHook.AspNetCoreMvc31/Properties/launchSettings.json
```diff
@@ -16,6 +16,9 @@
 
         "DOTNET_STARTUP_HOOKS": "$(ProjectDir)$(OutputPath)profiler-lib\\OpenTelemetry.Instrumentation.StartupHook.dll",
 
+        "DOTNET_ADDITIONAL_DEPS": "$(ProjectDir)$(OutputPath)profiler-lib\\additionalDeps",
```
RassK, line 19: These references seem to not work any more as 'profiler-lib' is removed. Seems this project was left out of refactoring. 
Base can be referenced now as `$(SolutionDir)bin\\tracer-home\\`

src/OpenTelemetry.Instrumentation.AdditionalDeps/OpenTelemetry.Instrumentation.AdditionalDeps.csproj
RassK: > > Are the net5.0 references under `netcoreapp3.1/store/[Platform]/net5.0` important? ... to me it seems like it's not connected only to netcoreapp3.1, maybe worth moving to root dir?
> 
> We need the folder for all support runtimes. Currently we have netcoreapp3.1 and net5.0. When we add support for .net6.0 there will be an additional folder that gets created here. Dotnet Runtime Store functionality forces us to have folder structure for all supported runtime to load correctly.

The framework folders sounds reasonable, just if it should be inside netcoreapp3.1 dir? I was thinking more like `[root]/store/[platform]/[framework]/...`

RassK:
> > Are the net5.0 references under `netcoreapp3.1/store/[Platform]/net5.0` important? ... to me it seems like it's not connected only to netcoreapp3.1, maybe worth moving to root dir?
> 
> We need the folder for all support runtimes. Currently we have netcoreapp3.1 and net5.0. When we add support for .net6.0 there will be an additional folder that gets created here. Dotnet Runtime Store functionality forces us to have folder structure for all supported runtime to load correctly.

Thanks for explaining the framework part. If it's necessary for every framework, could we move the store dir into root then? I'm just thinking if there is ideologically connection to netcoreapp3.1 dir? so simply `[root]/store/[platform]/[framework]`

rajkumar-rangaraj:
> > > Are the net5.0 references under `netcoreapp3.1/store/[Platform]/net5.0` important? ... to me it seems like it's not connected only to netcoreapp3.1, maybe worth moving to root dir?
> > 
> > 
> > We need the folder for all support runtimes. Currently we have netcoreapp3.1 and net5.0. When we add support for .net6.0 there will be an additional folder that gets created here. Dotnet Runtime Store functionality forces us to have folder structure for all supported runtime to load correctly.
> 
> Thanks for explaining the framework part. If it's necessary for every framework, could we move the store dir into root then? I'm just thinking if there is ideologically connection to netcoreapp3.1 dir? so simply `[root]/store/[platform]/[framework]`

I'm happy to move this to root directory. This reduces the length of the path too. Not sure, if we will have any impact when we package it.

src/OpenTelemetry.Instrumentation.AdditionalDeps/OpenTelemetry.Instrumentation.AdditionalDeps.csproj
RassK: Just an additional idea for later. If we lift `OpenTelemetry.Instrumentation.AdditionalDeps`'s logic to `OpenTelemetry.ClrProfiler.Managed`, does this mean we can fix also all OTel.SDK packages and their dependencies automatically? ... also we don't double load different versions, in case user code uses the same packages?

nrcventura:
> > > > Are the net5.0 references under `netcoreapp3.1/store/[Platform]/net5.0` important? ... to me it seems like it's not connected only to netcoreapp3.1, maybe worth moving to root dir?
> > > 
> > > 
> > > We need the folder for all support runtimes. Currently we have netcoreapp3.1 and net5.0. When we add support for .net6.0 there will be an additional folder that gets created here. Dotnet Runtime Store functionality forces us to have folder structure for all supported runtime to load correctly.
> > 
> > 
> > Thanks for explaining the framework part. If it's necessary for every framework, could we move the store dir into root then? I'm just thinking if there is ideologically connection to netcoreapp3.1 dir? so simply `[root]/store/[platform]/[framework]`
> 
> I'm happy to move this to root directory. This reduces the length of the path too. Not sure, if we will have any impact when we package it.

If we move this dir structure to the root dir, it might affect the way we look up the managed tracer code. I don't know if that makes the loading logic more complicated than it's worth. For now, I think it's easier to reason about the separate directories for the different purposes. I'm not opposed to experimenting with other folder structures though.

RassK:
> > > > > Are the net5.0 references under `netcoreapp3.1/store/[Platform]/net5.0` important? ... to me it seems like it's not connected only to netcoreapp3.1, maybe worth moving to root dir?
> > > > 
> > > > 
> > > > We need the folder for all support runtimes. Currently we have netcoreapp3.1 and net5.0. When we add support for .net6.0 there will be an additional folder that gets created here. Dotnet Runtime Store functionality forces us to have folder structure for all supported runtime to load correctly.
> > > 
> > > 
> > > Thanks for explaining the framework part. If it's necessary for every framework, could we move the store dir into root then? I'm just thinking if there is ideologically connection to netcoreapp3.1 dir? so simply `[root]/store/[platform]/[framework]`
> > 
> > 
> > I'm happy to move this to root directory. This reduces the length of the path too. Not sure, if we will have any impact when we package it.
> 
> If we move this dir structure to the root dir, it might affect the way we look up the managed tracer code. I don't know if that makes the loading logic more complicated than it's worth. For now, I think it's easier to reason about the separate directories for the different purposes. I'm not opposed to experimenting with other folder structures though.

To me it seems these 2 folder structures are unrelated. Should not affect loading at all, these 2 new env's are taking care of the path finding and clr takes care from which folder to load the dependency. 
I'm thinking it like that, you don't need a dependency store for every shipped version, but you need a single dependency store with different .net versions providing the specific package.

pellared:
I feel that I still need some time to get more familiar with `DOTNET_ADDITIONAL_DEPS` and `DOTNET_SHARED_STORE` mechanics. However, in my opinion, this approach is totally acceptable.

Could you please:
- add some information to the `docs/USAGE.md` for the end-users
- Add some more details  in `docs/DESIGN.md`. There is an "Assembly Conflict Resolution" section which could be a good place to describe it.

Sorry for waiting so long for the review.

zacharycmontoya:
> > > > > > Are the net5.0 references under `netcoreapp3.1/store/[Platform]/net5.0` important? ... to me it seems like it's not connected only to netcoreapp3.1, maybe worth moving to root dir?
> > > > > 
> > > > > 
> > > > > We need the folder for all support runtimes. Currently we have netcoreapp3.1 and net5.0. When we add support for .net6.0 there will be an additional folder that gets created here. Dotnet Runtime Store functionality forces us to have folder structure for all supported runtime to load correctly.
> > > > 
> > > > 
> > > > Thanks for explaining the framework part. If it's necessary for every framework, could we move the store dir into root then? I'm just thinking if there is ideologically connection to netcoreapp3.1 dir? so simply `[root]/store/[platform]/[framework]`
> > > 
> > > 
> > > I'm happy to move this to root directory. This reduces the length of the path too. Not sure, if we will have any impact when we package it.
> > 
> > 
> > If we move this dir structure to the root dir, it might affect the way we look up the managed tracer code. I don't know if that makes the loading logic more complicated than it's worth. For now, I think it's easier to reason about the separate directories for the different purposes. I'm not opposed to experimenting with other folder structures though.
> 
> To me it seems these 2 folder structures are unrelated. Should not affect loading at all, these 2 new env's are taking care of the path finding and clr takes care from which folder to load the dependency. I'm thinking it like that, you don't need a dependency store for every shipped version, but you need a single dependency store with different .net versions providing the specific package.
> 
> (it just currently happens that we ship netcoreapp3.1 only, when you add net6.0 then it might seem clearer - just an example, there might be no reasons yet to ship additional fx specific package. @zacharycmontoya might help here, I see they have experience shipping multiple versions)

I think Rasmus is correct that the placement of the store is independent of the managed loader logic. For that reason, it would likely be easier to keep the store/deps at the root level so the path doesn't need to change if new frameworks are added by us.

src/OpenTelemetry.Instrumentation.AdditionalDeps/OpenTelemetry.Instrumentation.AdditionalDeps.csproj
```diff
@@ -0,0 +1,40 @@
...
+    <PropertyGroup>
+      <AdditionalDepsPath>$(OutputDir)\AdditionalDeps</AdditionalDepsPath>
+      <StoreOutputBasePath>$(OutputDir)\store\</StoreOutputBasePath>
+      <AdditionalDepsPath Condition="'$(TracerHomePath)' != ''">$(TracerHomePath)\AdditionalDeps</AdditionalDepsPath>
+      <StoreOutputBasePath Condition="'$(TracerHomePath)' != ''">$(TracerHomePath)\store</StoreOutputBasePath>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <StoreRuntime Include="x86">
+        <PlatformTargetForStore>x86</PlatformTargetForStore>
```
pjanotti, line 21: nit: since it is later used with the prefix `StoreRuntime` shorter names could be used, e.g. `Platform` and `Path`

src/OpenTelemetry.Instrumentation.AdditionalDeps/OpenTelemetry.Instrumentation.AdditionalDeps.csproj
```diff
@@ -0,0 +1,40 @@
...
+        <PlatformTargetForStore>x64</PlatformTargetForStore>
+        <StorePath>$(StoreOutputBasePath)</StorePath>
+      </StoreRuntime>
+    </ItemGroup>
+    
+    <Message Text="Calling ComposeStore for TargetFramework='$(TargetFramework)'" Importance="high" />
+    
+    <ItemGroup>
+      <ProjectsToBuild Include="Package.xml">
+        <Properties>TargetFramework=$(TargetFramework);PlatformTarget=%(StoreRuntime.PlatformTargetForStore);RuntimeIdentifier=%(StoreRuntime.Identity);SkipOptimization=true;ComposeDir=%(StoreRuntime.StorePath);CreateProfilingSymbols=false;Version=$(SemanticVersion4parts);NuGetPackageRootOriginal=$(NuGetPackageRoot);Environment=$(Environment)</Properties>
```
pjanotti, line 34: These are understandable by their names but is there any doc with the full set options? It looks like it is a call to `dotnet store` but some of the names don't match exactly with those. If there is a doc with more info it seems worth a link here.

src/OpenTelemetry.Instrumentation.AdditionalDeps/Package.xml
```diff
@@ -0,0 +1,19 @@
...
+    <RestoreSources>https://api.nuget.org/v3/index.json;$(TracerHomePath)\AdditionalDeps;</RestoreSources>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Condition="'$(TracerHomePath)' != ''" Include="OpenTelemetry.Instrumentation.AdditionalDeps" Version="*" />
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="6.0.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
pjanotti, line 12: I *think* I understand the effect of this when building a NuGet package but I'm not sure of what it does when building a store. Could you clarify that?

pjanotti (approved):
LGTM

@pellared suggestions at https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/272#issuecomment-977876618 are very good. That said they can be done separately (especially item 2 that may be a bit larger).

src/OpenTelemetry.Instrumentation.AdditionalDeps/OpenTelemetry.Instrumentation.AdditionalDeps.csproj
pjanotti: Related Q.: `RuntimeIdentifier` feels like a `RID` but here is just the platform target? Will this work if there is a dependency to a specific RID?

src/OpenTelemetry.Instrumentation.AdditionalDeps/OpenTelemetry.Instrumentation.AdditionalDeps.csproj
rajkumar-rangaraj: This will make both AdditionalDeps and RuntimeStore very complex. It's better to keep these separated and on a  need basis update the package that we are interested in `OpenTelemetry.Instrumentation.AdditionalDeps`.

src/OpenTelemetry.Instrumentation.AdditionalDeps/Package.xml
rajkumar-rangaraj: Good point, I added it for test and missed removing it. This is not needed.

src/OpenTelemetry.Instrumentation.AdditionalDeps/OpenTelemetry.Instrumentation.AdditionalDeps.csproj
```diff
@@ -0,0 +1,40 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp3.1;net5.0</TargetFrameworks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="6.0.0" />
```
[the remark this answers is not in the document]
rajkumar-rangaraj: Fixed

src/OpenTelemetry.Instrumentation.AdditionalDeps/OpenTelemetry.Instrumentation.AdditionalDeps.csproj
rajkumar-rangaraj: You're correct. It is a call to dotnet store, did not find a doc on how to call from the project file. I will leave a comment stating it is a call to dotnet store.

Yes, RID here is a platform target. It won't work if there is a dependency to specific RID. If we want to target a specific target then the runtime store folders needs to be created for those many RIDs that are referenced.

samples/Samples.StartupHook.AspNetCoreMvc31/Properties/launchSettings.json
```diff
@@ -14,7 +14,10 @@
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development",
 
-        "DOTNET_STARTUP_HOOKS": "$(ProjectDir)$(OutputPath)profiler-lib\\OpenTelemetry.Instrumentation.StartupHook.dll",
+        "DOTNET_STARTUP_HOOKS": "$(SolutionDir)bin\\tracer-home\\netcoreapp3.1\\OpenTelemetry.Instrumentation.StartupHook.dll",
+
+        "DOTNET_ADDITIONAL_DEPS": "$(SolutionDir)bin\\tracer-home\\additionalDeps",
```
RassK, line 19: Shouldn't `additionalDeps` be `AdditionalDeps` in case for Linux tests?

```suggestion
        "DOTNET_ADDITIONAL_DEPS": "$(SolutionDir)bin\\tracer-home\\AdditionalDeps",
```

RassK (approved):
Looks great!

nrcventura (approved):

PR #275, opened 2021-11-15
Add startup hook exclusions

### LINKED ISSUE #270: Support excluding certain processes when using the startup hook
When enabling the startup hook, depending on how you define the corresponding the environment variable, you can run into errors when trying to build or run your application. This is a result of .net attempting to load the startup hook, but not being able to also load the dependencies that are expected. 

To make the user experience smoother we should provide an environment variable, like we do for the profiler, to prevent the startup hook from running (and needing other dependencies) when the current process should be excluded.

nrcventura:
Resolves #270 

This change uses the existing `OTEL_PROFILER_EXCLUDE_PROCESSES` environment variable (to maintain consistency with the profiler exclusion logic) so that the Startup Hook will prevent itself from loading the managed tracer and OTel SDK.

Questions that likely require further thought/input:

1. Should this logic belong in the startup hook, or in the managed loader? It is currently in the startup hook, so that we do not attempt to load any additional dependencies (including the managed loader) for applications that should not be monitored (like dotnet.exe).
2. Should we use a separate environment variable for this exclusion logic, or reuse the environment variable that is also used by the profiler?
3. If we decide to reuse the environment variable, should we also reuse `OTEL_PROFILER_PROCESSES`? If so I think we should handle that use case in a separate PR.


src/OpenTelemetry.Instrumentation.StartupHook/StartupHook.cs
```diff
@@ -14,6 +16,17 @@ internal class StartupHook
     /// </summary>
     public static void Initialize()
     {
+        var applicationName = GetApplicationName();
+        StartupHookEventSource.Log.Trace($"StartupHook loaded for application with name {applicationName}.");
+
+        if (IsApplicationInExcludeList(applicationName))
+        {
+            StartupHookEventSource.Log.Trace("Application is in the exclusion list. Skipping initialization.");
```
rajkumar-rangaraj, line 24: Instead of failing silently should we consider throwing an exception?

nrcventura: I don't necessarily see this path as a failure case. Depending on how the application is started, I expect this path to get triggered and purposely not monitor that "application". For example, if the application is started with the command `dotnet run` instead of `dotnet myapp.dll`, two processes are ultimately started: 1. The `dotnet` process with an `AppDomain` of `dotnet`, and 2. the `myapp` process with an `AppDomain` of `myapp`. When `dotnet myapp.dll` is used, one `dotnet` process is started, but the `AppDomain` is `myapp`. When `dotnet run` is used to start the application we don't want to instrument the `dotnet` process/application.

One of the problems that we were seeing was that if we tried to enable the `AspNet` instrumentation library, and started the AspNet Core application using `dotnet run`, the startup hook ultimately attempted to trigger the loading of the `AspNet` instrumentation within the `dotnet` `AppDomain`. The AspNet Core dependencies were not available for that process because it is not an AspNet Core app, so an exception was thrown when we attempted to load the `AspNet` instrumentation. That exception also prevents the starting of the `myapp` process, which would have also attempted to load the startup hook.

I do think that we should provide some documentation on how to see the StartupHook log messages so that these types of startup issues can be easier to diagnose in the future.

rajkumar-rangaraj (approved):

pellared:
> Should this logic belong in the startup hook, or in the managed loader? It is currently in the startup hook, so that we do not attempt to load any additional dependencies (including the managed loader) for applications that should not be monitored (like dotnet.exe).

Having it "high" in startup hook makes sense to me. We could give a comment in code why it is there to avoid loading additional dependencies in runtime.

> Should we use a separate environment variable for this exclusion logic, or reuse the environment variable that is also used by the profiler?

I think it is OK to reuse the same env var. We could also rename it so that it has the `OTEL_DOTNET_TRACER_` prefix but it can be a subject for a different PR.

> If we decide to reuse the environment variable, should we also reuse OTEL_PROFILER_PROCESSES? If so I think we should handle that use case in a separate PR.

Yes and yes 😄 

I have not seen the code in the PR. It is on my TODO list...

src/OpenTelemetry.Instrumentation.StartupHook/StartupHook.cs
```diff
@@ -65,4 +78,46 @@ private static string GetLoaderAssemblyLocation()
...
+    private static bool IsApplicationInExcludeList(string applicationName)
+    {
+        return GetExcludedApplicationNames().Contains(applicationName);
+    }
+
+    private static IList<string> GetExcludedApplicationNames()
+    {
+        var excludedProcesses = new List<string>();
+
+        var environmentValue = Environment.GetEnvironmentVariable("OTEL_PROFILER_EXCLUDE_PROCESSES");
```
pellared, line 106: how about creating a static class with constants for all env vars? it can act as a backup for documentation. I like having "public contract" separated somewhere. We could also reuse this class e.g. in `ClearProfilerEnvironmentVariables`.

src/OpenTelemetry.Instrumentation.StartupHook/StartupHook.cs
```diff
@@ -65,4 +78,46 @@ private static string GetLoaderAssemblyLocation()
...
+        {
+            return excludedProcesses;
+        }
+
+        foreach (var processName in environmentValue.Split(','))
+        {
+            if (!string.IsNullOrWhiteSpace(processName))
+            {
+                excludedProcesses.Add(processName.Trim());
+            }
```
pellared, line 118: We can deal with whitespaces only once

```suggestion
            processName = processName.Trim();
            if (!string.IsNullOrEmpty(processName))
            {  
                excludedProcesses.Add(processName);
            }
```

src/OpenTelemetry.Instrumentation.StartupHook/StartupHook.cs
```diff
@@ -65,4 +78,46 @@ private static string GetLoaderAssemblyLocation()
...
+
+        return applicationName;
+    }
+
+    private static bool IsApplicationInExcludeList(string applicationName)
+    {
+        return GetExcludedApplicationNames().Contains(applicationName);
+    }
+
+    private static IList<string> GetExcludedApplicationNames()
```
pellared, line 102: I see no reason to return an interface. It would also improve performance a little 😄 

```suggestion
    private static List<string> GetExcludedApplicationNames()
```

test/integration-tests/IntegrationTests.StartupHook/StartupHookTests.cs
```diff
@@ -24,6 +24,30 @@ public StartupHookTests(ITestOutputHelper output)
...
+                applicationNameToExclude = EnvironmentHelper.FullSampleName;
+                expectedSpanCount = 0;
+            }
+
+            SetEnvironmentVariable("OTEL_PROFILER_EXCLUDE_PROCESSES", $"dotnet,dotnet.exe,{applicationNameToExclude}");
+
+            RunTestAndValidateResults(expectedSpanCount);
+        }
+
+        private void RunTestAndValidateResults(int expectedSpanCount)
```
pellared, line 50: Proposal no 1: rename to `RunSampleAndAssertSpanCount`

Proposal no 2: refactor the tests so we have more readable tests like:

```csharp
using (var collector = RunSample(startupHook: true))
{
    AssertSpanCount(expectedSpanCount, collector);
}
```

then I think having a helper to extract two lines would be useless. Also, I find a method having `And` as a code smell.

pellared (approved):
Just some refactoring proposals. Could be addressed in separate PR or ignored if you disagree. Nice work 👍

src/OpenTelemetry.Instrumentation.StartupHook/StartupHook.cs
nrcventura: I think that it might get tricky to use that approach for all environment variables, because we need to be careful about which dependencies the StartupHook takes on. So I don't think that we could share that code with the StartHook and the Tracer, but we could definitely share some code between the StartupHook and the tests. This would leave us with 3 locations where environment variables are defined: 1. the native profiler, 2. the managed tracer, 3. the startup hook. And some of the environment variables will be shared by more than one of those locations, such as the process inclusion/exclusion environment variables used by the native profiler and the startup hook.

src/OpenTelemetry.Instrumentation.StartupHook/StartupHook.cs
nrcventura: In this case, if you perform the `.Trim()` call before the null check, you can trigger an exception. The code in this method has the same logic as its native counterpart where the profiler checks the same environment variable, so that the code handles extra whitespace more consistently.

test/integration-tests/IntegrationTests.StartupHook/StartupHookTests.cs
nrcventura: Yes, that part of the code definitely needs some cleaning up.

src/OpenTelemetry.Instrumentation.StartupHook/StartupHook.cs
zacharycmontoya: `Environment.GetEnvironmentVariable` can throw a `SecurityException` (probably only .NET Framework if permissions re locked down?) so I recommend putting the method call inside a try/catch block

zacharycmontoya (approved):
LGTM 👍🏼

pellared:
I think this can be merged 😉

test/integration-tests/IntegrationTests.StartupHook/StartupHookTests.cs
nrcventura: I'll get this cleaned up in a follow up PR.

src/OpenTelemetry.Instrumentation.StartupHook/StartupHook.cs
```diff
@@ -65,4 +78,44 @@ private static string GetLoaderAssemblyLocation()
...
+
+    private static string GetApplicationName()
+    {
+        try
+        {
+            return AppDomain.CurrentDomain.FriendlyName;
+        }
+        catch (Exception ex)
+        {
+            StartupHookEventSource.Log.Error($"Error getting AppDomain.CurrentDomain.FriendlyName: {ex}");
```
pjanotti, line 90: Under normal usage of a start up hook I don't think we can hit this exception although we can hit it in code that are calling the startup hook directly. Are we not risking hiding improver usage with this catch here?

pjanotti (approved):
LGTM, just Qs about the catch for the `FriendlyName`.

src/OpenTelemetry.Instrumentation.StartupHook/StartupHook.cs
pjanotti: Related: what do we get for an app started with `dotnet run` that still has to build the app? Any difference between the friendly names in Windows and Linux? Perhaps, it is worth a comment on the code or link to the docs.

PR #277, opened 2021-11-17
[WIP] Update MongoDB driver

RassK:
Updates MongoDB driver to resolve deprecated OP codes issue.

### DIFF
--- a/test/test-applications/integrations/Samples.MongoDB/Samples.MongoDB.csproj
+++ b/test/test-applications/integrations/Samples.MongoDB/Samples.MongoDB.csproj
@@ -5,7 +5,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


pellared:
How about pinning the MongoDB docker container version instead of using `latest`?

RassK:
> How about pinning the MongoDB docker container version instead of using `latest`?

Unfortunately in Windows it's baked into the image instead. But there should be no issues moving to container based server.

nrcventura:
What I'm not understanding is why I don't see the same issues locally. The CI pipeline and our local machines should be pulling the same container image and using the same driver version.

RassK:
> What I'm not understanding is why I don't see the same issues locally. The CI pipeline and our local machines should be pulling the same container image and using the same driver version.

Except they didn't release 5.1 in Docker yet, but GH team upgraded the integrated version. Maybe there is even product difference Enterprise vs Community.

RassK:
I'm going to close this as now important right now.

PR #279, opened 2021-11-18
Fix MongoDB Server image version

RassK:
Sets a fix version to MongoDB Docker container image.
(unfortunately not for Mac, as Docker is not supported on GHA Mac image)

### DIFF
--- a/test/integration-tests/IntegrationTests.MongoDB/MongoDbCollection.cs
+++ b/test/integration-tests/IntegrationTests.MongoDB/MongoDbCollection.cs
@@ -16,6 +16,7 @@ namespace IntegrationTests.MongoDB
     public class MongoDbFixture : IDisposable
     {
         private const int MongoDbPort = 27017;
+        private const string MongoDbImage = "mongo:5.0.4";
 
         private TestcontainersContainer _container;
 
@@ -47,12 +48,12 @@ namespace IntegrationTests.MongoDB
         {
             if (EnvironmentHelper.IsRunningOnCI())
             {
-                if (EnvironmentTools.IsWindows() ||
-                    EnvironmentTools.IsMacOS())
+                if (EnvironmentTools.IsMacOS())
                 {
                     return false;
                 }
-                else if (EnvironmentTools.IsLinux())
+                else if (EnvironmentTools.IsWindows() ||
+                         EnvironmentTools.IsLinux())
                 {
                     return true;
                 }
@@ -68,7 +69,7 @@ namespace IntegrationTests.MongoDB
                 : Wait.ForUnixContainer();
 
             var mongoContainersBuilder = new TestcontainersBuilder<TestcontainersContainer>()
-              .WithImage("mongo")
+              .WithImage(MongoDbImage)
               .WithName($"mongo-db-{port}")
               .WithPortBinding(port, MongoDbPort)
               .WithWaitStrategy(waitOS.UntilPortIsAvailable(MongoDbPort));


nrcventura (approved):
🥇  LGTM

test/integration-tests/IntegrationTests.MongoDB/MongoDbCollection.cs
```diff
@@ -68,7 +69,7 @@ private TestcontainersContainer LaunchMongoContainer(int port)
                 : Wait.ForUnixContainer();
 
             var mongoContainersBuilder = new TestcontainersBuilder<TestcontainersContainer>()
-              .WithImage("mongo")
+              .WithImage(MongoDbImage)
```
rajkumar-rangaraj, line 72: nit: leaving a comment here will help understand why we use the hard-coded version.

rajkumar-rangaraj (approved):

nrcventura:
> (unfortunately not for Mac, as Docker is not supported on GHA Mac image)

It looks like we could still leverage a specific mongodb version in the Mac pipeline if we follow the instructions in https://docs.mongodb.com/manual/tutorial/install-mongodb-on-os-x/ , because homebrew and mongodb allow side by side installs. But that can be done separately because this gets our checks green again.

test/integration-tests/IntegrationTests.MongoDB/MongoDbCollection.cs
RassK: I think that actually leaving it without a fix version was a bug, that caused uncontrollable environment.
Better way should be integrating it with dependabot somehow, so when PR is opened with version update, we can actually see what it causes.

RassK:
> > (unfortunately not for Mac, as Docker is not supported on GHA Mac image)
> 
> It looks like we could still leverage a specific mongodb version in the Mac pipeline if we follow the instructions in https://docs.mongodb.com/manual/tutorial/install-mongodb-on-os-x/ , because homebrew and mongodb allow side by side installs. But that can be done separately because this gets our checks green again.

This is also until they release a new brew package. I 

> > (unfortunately not for Mac, as Docker is not supported on GHA Mac image)
> 
> It looks like we could still leverage a specific mongodb version in the Mac pipeline if we follow the instructions in https://docs.mongodb.com/manual/tutorial/install-mongodb-on-os-x/ , because homebrew and mongodb allow side by side installs. But that can be done separately because this gets our checks green again.

Seems also just a lucky shot that it didn't break yet. Imo Homebrew is another half baked package manager, where you can just install latest versions. So once they release new brew package version, it's again game over. (remember the issue with Make, impossible to downgrade via Homebrew)

zacharycmontoya (approved):

nrcventura:
> > > (unfortunately not for Mac, as Docker is not supported on GHA Mac image)
> > 
> > 
> > It looks like we could still leverage a specific mongodb version in the Mac pipeline if we follow the instructions in https://docs.mongodb.com/manual/tutorial/install-mongodb-on-os-x/ , because homebrew and mongodb allow side by side installs. But that can be done separately because this gets our checks green again.
> 
> Seems also just a lucky shot that it didn't break yet. Imo Homebrew is another half baked package manager, where you can just install latest versions. So once they release new brew package version, it's again game over. (remember the issue with Make, impossible to downgrade via Homebrew)

Luckily, it looks like MongoDb provides versioned taps for homebrew so that you don't have to always use "latest". There's a `@5.0` tap that can be used to prevent `5.1` from being installed and run.

PR #281, opened 2021-11-24
Add compatibility section in README.md

pellared:
Structure and content based on the following:
- https://github.com/open-telemetry/opentelemetry-dotnet
- https://github.com/open-telemetry/opentelemetry-go

Initially, I have written that we support all Linux OSes that .NET supports. But together with @RassK we decided to be more hesitant and created https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/282

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -21,8 +21,11 @@ except for versions lower than `.NET Framework 4.6.1`.
 
 ### Supported OSes
 
-OpenTelemetry .NET Auto-Instrumentation should work on all operating systems
-that given .NET version supports.
+The following operating systems are currently supported:
+
+- Windows,
+- MacOS,
+- Ubuntu.
 
 ## Usage
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -21,8 +21,8 @@ except for versions lower than `.NET Framework 4.6.1`.
 
 ### Supported OSes
 
-OpenTelemetry .NET Auto-Instrumentation is tested against Windows and Ubuntu.
-MacOS is not supported.
+OpenTelemetry .NET Auto-Instrumentation should work on all operating systems
+that given .NET version support.
 
 ## Usage
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -27,6 +27,8 @@ The following operating systems are currently supported:
 - MacOS,
 - Ubuntu.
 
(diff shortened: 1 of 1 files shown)


dszmigielski (approved):

Kielek (approved):

docs/README.md
```diff
@@ -21,8 +21,11 @@ except for versions lower than `.NET Framework 4.6.1`.
 
 ### Supported OSes
 
-OpenTelemetry .NET Auto-Instrumentation should work on all operating systems
-that given .NET version supports.
+The following operating systems are currently supported:
+
+- Windows,
+- MacOS,
+- Ubuntu.
```
pellared, line 29:
```suggestion

However, it should also work on other Linux OSes that given .NET version supports.

```

@RassK Checked that it also works e.g. on Alpine. Do you think that adding ☝️  makes sense?

rajkumar-rangaraj (approved):

docs/README.md
nrcventura: I think that a statement like that is useful.

docs/README.md
```diff
@@ -4,18 +4,35 @@
...
+### Supported .NET versions
+
+OpenTelemetry .NET Auto-Instrumentation attempts to work with all officially
+supported versions of [.NET Core](https://dotnet.microsoft.com/download/dotnet-core),
+and [.NET Framework](https://dotnet.microsoft.com/download/dotnet-framework)
+except for versions lower than `.NET Framework 4.6.1`.
+
+### Supported OSes
+
+The following operating systems are currently supported:
```
nrcventura, line 24: When we use the word `supported` here, what are we implying? Are we implying that we have automated tests against these platforms, provide usage instructions and documentation, installation methods, or other things? Should we be more transparent about why we are calling out these operating systems?

pellared: That is what I had in mind. Not sure if it is worth writing it down.

pjanotti: @pellared it seems worth to spell it out. My suggestion would be to change the title to "Compatible OSes" and state that the project is automatically tested against such and such OSes and keep the line about that it should work with other Linux flavors.

pellared: I tried to make it shorter and more concise

pjanotti (approved):

nrcventura (approved):

Issue #290, opened 2021-12-06
`TraceProvider` is not disposed


pellared:
This following field is not disposed at all:

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/09ea838561d4e8d466c0ee53e975fd3cd0c89848/src/OpenTelemetry.ClrProfiler.Managed/Instrumentation.cs#L19

This causes loss of exported spans when the batch export processor is used.

Probably the easiest thing to fix it would be creating some object with a finalizer that would dispose the `TraceProvider`.

nrcventura:
Another option is to potentially subscribe to the shutdown/terminate event and then trigger the dispose. If we subscribe to the correct event(s), I think that we'll have about 30 seconds (or something similar) to export as much data as possible. New Relic uses the [`ProcessExit` and `DomainUnload` events to trigger similar things](https://github.com/newrelic/newrelic-dotnet-agent/blob/2ba99ea91ac20f79af40259e485949c6b3b1aac3/src/Agent/NewRelic/Agent/Core/AgentInitializer.cs#L29-L31) (The `OnExit` property in that code is later updated with the real shutdown logic that is desired).

PR #294, opened 2021-12-08
TraceProvider Dispose

RassK:
### What
Disposes the SDK tracer on AppDomain events.
Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/290

### DIFF
--- a/src/OpenTelemetry.ClrProfiler.Managed/Instrumentation.cs
+++ b/src/OpenTelemetry.ClrProfiler.Managed/Instrumentation.cs
@@ -113,13 +113,9 @@ namespace OpenTelemetry.ClrProfiler.Managed
                 return;
             }
 
-            if (_tracerProvider is not null)
-            {
-                _tracerProvider.Dispose();
-                _tracerProvider = null;
+            _tracerProvider.Dispose();
 
-                Log("OpenTelemetry tracer exit.");
-            }
+            Log("OpenTelemetry tracer exit.");
         }
 
         private static void Log(string message)
--- a/src/OpenTelemetry.ClrProfiler.Managed/Instrumentation.cs
+++ b/src/OpenTelemetry.ClrProfiler.Managed/Instrumentation.cs
@@ -15,6 +15,7 @@ namespace OpenTelemetry.ClrProfiler.Managed
     {
         private static readonly Process _process = Process.GetCurrentProcess();
         private static int _firstInitialization = 1;
+        private static int _isExiting = 0;
 
         private static TracerProvider _tracerProvider;
 
@@ -65,6 +66,10 @@ namespace OpenTelemetry.ClrProfiler.Managed
 
                     _tracerProvider = builder.Build();
                     Log("OpenTelemetry tracer initialized.");
+
+                    // Register to shutdown events
+                    AppDomain.CurrentDomain.ProcessExit += OnExit;
+                    AppDomain.CurrentDomain.DomainUnload += OnExit;
                 }
             }
             catch (Exception ex)
@@ -97,6 +102,23 @@ namespace OpenTelemetry.ClrProfiler.Managed
             {
                 Log($"OpenTracingShim exception: {ex}");
                 throw;
+            }
+        }
+
+        private static void OnExit(object sender, EventArgs e)
+        {
+            if (Interlocked.Exchange(ref _isExiting, value: 1) != 0)
+            {
+                // OnExit() was already called before
+                return;
+            }
+
+            if (_tracerProvider is not null)
+            {
+                _tracerProvider.Dispose();
+                _tracerProvider = null;
+
+                Log("OpenTelemetry tracer exit.");
             }
         }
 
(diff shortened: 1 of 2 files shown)


pellared:
I committed https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/294/commits/c54a8db21fb5e1aa61e67730f4e98e703818da44 which basically was a workaround for the missing disposal.

src/OpenTelemetry.ClrProfiler.Managed/Instrumentation.cs
```diff
@@ -100,6 +105,23 @@ public static void Initialize()
...
+                return;
+            }
+
+            if (_tracerProvider is not null)
+            {
+                _tracerProvider.Dispose();
+                _tracerProvider = null;
+
+                Log("OpenTelemetry tracer exit.");
+            }
```
pellared, line 122: the code looks as if it could be accessed from multiple threads. if this is so then access to `_tracerProvider` is not synchronized at all

proposals:
1. check if it is not the main thread that is always calling `Initialize` and `OnExit` - if so then maybe we can remove the `Interlocked.Exchange()` and note in comments that the class is accessed always from the same thread, therefore, it does not need to be thread-safe
2. add simple `lock` around `_tracerProvider` usage. it should also help in a situation when `OnExit()` is called before `Initialize()` has set the `_tracerProvider` value. In practice, it should not be a problem but I think it is always better to have thread-safe code. I dislike when the code is "partially thread-safe".

pellared: maybe it will be easier to go for "proposal 2" and make an issue to check if "proposal 1" is valid

RassK: Currently OnExit is not subscribed before _traceProvider is initialized.

pellared: you are right. then `if (_tracerProvider is not null)` can be removed

pellared (approved):
Only one minor comment probably worth addressing:

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/294#discussion_r764799319

src/OpenTelemetry.ClrProfiler.Managed/Instrumentation.cs
RassK: true, gives wrong impression.

nrcventura (approved):

src/OpenTelemetry.ClrProfiler.Managed/Instrumentation.cs
```diff
@@ -65,6 +66,10 @@ public static void Initialize()
 
                     _tracerProvider = builder.Build();
                     Log("OpenTelemetry tracer initialized.");
+
+                    // Register to shutdown events
+                    AppDomain.CurrentDomain.ProcessExit += OnExit;
+                    AppDomain.CurrentDomain.DomainUnload += OnExit;
```
rajkumar-rangaraj, line 72: .NET calls these events on a best effort basis. We might reduce the impact of the loss of exported spans, but it does not guarantee the that dispose will complete.

Do we know what is the recommendation from OpenTelemetry .NET SDK to handle such scenarios?

RassK: Just more info about the issue >
https://stackoverflow.com/questions/20675024/when-will-appdomain-processexit-not-get-called

Not sure if something similar applies to IIS also, it seems additional layer of magic.

rajkumar-rangaraj: We might not have another event which could help us dispose an object during shutdown. My earlier comment was to say there is no guarantee these events are triggered in all cases. But this approach should work with most of the applications.

Yes. This is applicable to IIS and have seen similar issue in the past.

nrcventura: I believe that the situations that would terminate the app but not trigger these events would also likely cause not all data to be exported when the Simple processor is used instead of the batch processor. I think it's just a question of how likely is this problem to occur, and how much data would not get exported if that happens? In practice, I've also seen the time window that we have to send the data when the event is triggered, generally being long enough to send all of the data.

RassK: Should we add "Data loss" section to the docs? 
Currently I understand these situations will cause a data loss:
- abnormal termination of the process
- unhandled exception terminates the process

Could reduce the impact (but not avoiding):
- use Simple processor
- use local OTel Collector for faster data handover

pellared: I think is a good idea. Just instead of using simple processor I would suggest setting some BSP env var values

Kielek (approved):

lucaspimentel (approved):

src/OpenTelemetry.ClrProfiler.Managed/Instrumentation.cs
nrcventura: > Do we know what is the recommendation from OpenTelemetry .NET SDK to handle such scenarios?

I'm not aware of any official guidance there. Most of the situations that the SDK is involved in, you are able to leverage a using block or statement to get the desired behavior, by initializing the SDK in an appropriate location in your code. This question is probably worth asking some of the SDK maintainers about.

PR #311, opened 2022-01-10
Remove Callsite instrumentation

### LINKED ISSUE #306: Remove CallSite auto-instrumentation
Automatic instrumentation was changed to use CallTarget by default (instead of CallSite) in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/181. At this point, CallTarget instrumentation is stable enough that we can keep CallTarget and completely remove CallSite instrumentation from the native code. This will reduce the maintenance burden of the native code.

EDIT: Updated by Zach Montoya

zacharycmontoya:
Fixes #306 by removing all managed and native code related to CallSite instrumentation and also removing code paths related to versions lower than net46. This PR is based on the following PR from DataDog/dd-trace-dotnet: https://github.com/DataDog/dd-trace-dotnet/pull/1958.

In summary, the following code paths were pruned from the instrumentation
- Branches where `is_calltarget_enabled == false`
- Branches where `is_net46_or_greater == false`

Follow-up items:
- [ ] Force the automatic instrumentation to use ICorProfilerInfo7 (net461) instead of ICorProfilerInfo6 (net46)
- [ ] Enable NGEN by default


src/OpenTelemetry.ClrProfiler.Native/integration_loader.h
```diff
@@ -17,30 +17,21 @@ using json = nlohmann::json;
...
-                             const bool isCallTargetEnabled, const bool isNetstandardEnabled,
-                             const std::vector<WSTRING>& disabledIntegrationNames);
+    void IntegrationFromJson(const json::value_type& src, std::vector<IntegrationMethod>& integrationMethods, const std::vector<WSTRING>& disabledIntegrationNames);
 
     void MethodReplacementFromJson(const json::value_type& src, const WSTRING& integrationName,
-                                   std::vector<IntegrationMethod>& integrationMethods,
-                                   const bool isCallTargetEnabled, const bool isNetstandardEnabled);
+                                   std::vector<IntegrationMethod>& integrationMethods);
 
     MethodReference MethodReferenceFromJson(const json::value_type& src, const bool is_target_method,
```
pjanotti, line 36: It seems that the usage of this method is now always with the books for `is_target_method` and `is_wrapper_method` either as `true, false` or `false, true` which makes me think that we could live with a single boolean. Anyway, better consider it separate, just a thought that crossed my mind.

pjanotti (approved):
LGTM. Thanks @zacharycmontoya

src/OpenTelemetry.ClrProfiler.Native/integration_loader.h
zacharycmontoya: Good insight Paulo. Yes let's consider this for a later PR

src/OpenTelemetry.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -109,9 +105,11 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
     ICorProfilerInfo6* info6;
     hr = cor_profiler_info_unknown->QueryInterface(__uuidof(ICorProfilerInfo6), (void**)&info6);
 
-    if (SUCCEEDED(hr)) {
-        Logger::Debug("Interface ICorProfilerInfo6 found.");
-        is_net46_or_greater = true;
+    // Fail if net46+ not detected
+    // TODO: Update this to ICorProfilerInfo7 which corresponds to net461 and make CorProfilerBase.info_ have type ICorProfilerInfo7*
+    if (FAILED(hr)) {
+        Logger::Warn("DATADOG TRACER DIAGNOSTICS - Failed to attach profiler: interface ICorProfilerInfo6 not found.");
```
nrcventura, line 111: We should probably change this to something like the following to be consistent with the other logging messages.
```suggestion
        Logger::Warn("OpenTelemetry TRACER DIAGNOSTICS - Failed to attach profiler: interface ICorProfilerInfo6 not found.");
```

nrcventura (commented):
The changes look good to me, but the native tests might need more updates (for the failing tests).

src/OpenTelemetry.ClrProfiler.Native/cor_profiler.cpp
zacharycmontoya: Fixed in 6862af3feb8e502d2247047e6db8a0ef54d4a6a3 . This should be the only log message that accidentally got introduced with Datadog in the log

pjanotti (approved):

Kielek (approved):

pellared (approved):
Can you also update https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/DESIGN.md#bytecode-instrumentations as a follow-up? We can get rid of the "call site" description.

rajkumar-rangaraj (approved):

nrcventura (approved):

PR #338, opened 2022-01-24
Add target framework for managed library test project reference

nrcventura:
Fixes issue with inconsistent test behavior between the CI and local test runners for the native tests. It also allows us to run the native tests from within Visual Studio

The problem was caused by the native tests containing project references to multi-targeted managed projects. The different test runners handle the selection of the managed target framework differently. This PR forces the consistent usage of the net461 target framework, because the native tests are already assuming that the managed libraries are targeting .net framework. If we want to run these native tests against multiple library target frameworks we need to also update the tests so that they are not assuming the presence of mscorlib or other .net framework specifics.


zacharycmontoya (approved):
Nice work! I think net461 only is fine for now because I expect it to take quite a bit more work to start running native tests on other OSes

Kielek (approved):

pjanotti (approved):
LGTM

PR #340, opened 2022-01-26
Fix starting container for aspnet tests

### LINKED ISSUE #295: Flaky test `IntegrationTests.AspNet.AspNetTests.SubmitsTraces`
https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/runs/4461872914?check_suite_focus=true

```
  Error: [xUnit.net 00:07:35.37]     IntegrationTests.AspNet.AspNetTests.SubmitsTraces [FAIL]
    Failed IntegrationTests.AspNet.AspNetTests.SubmitsTraces [7 m 27 s]
    Error Message:
     System.AggregateException : One or more errors occurred.
  ---- System.Threading.Tasks.TaskCanceledException : A task was canceled.
    Stack Trace:
       at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
     at IntegrationTests.AspNet.AspNetTests.SubmitsTraces() in D:\a\opentelemetry-dotnet-instrumentation\opentelemetry-dotnet-instrumentation\test\integration-tests\aspnet\IntegrationTests.AspNet\AspNetTests.cs:line 32
  ----- Inner Stack Trace -----
  
    Standard Output Messages:
   Platform: X64
   Configuration: Release
   TargetFramework: net461
   .NET Core: False
   Found profiler at D:\a\opentelemetry-dotnet-instrumentation\opentelemetry-dotnet-instrumentation\bin\tracer-home\win-x64\OpenTelemetry.ClrProfiler.Native.dll.
   Profiler DLL: D:\a\opentelemetry-dotnet-instrumentation\opentelemetry-dotnet-instrumentation\bin\tracer-home\win-x64\OpenTelemetry.ClrProfiler.Native.dll
   PS> New-NetFirewallRule -DisplayName 'TraceAgent-50329' -Direction Inbound -LocalPort 50329 -Protocol TCP -Action Allow -Profile Any
   StandardOutput:
   
   
   Name                  : {7fb397b0-b8b2-42c6-8378-ff75acd3d750}
   DisplayName           : TraceAgent-50329
   Description           : 
   DisplayGroup          : 
   Group                 : 
   Enabled               : True
   Profile               : Any
   Platform              : {}
   Direction             : Inbound
   Action                : Allow
   EdgeTraversalPolicy   : Block
   LooseSourceMapping    : False
   LocalOnlyMapping      : False
   Owner                 : 
   PrimaryStatus         : OK
   Status                : The rule was parsed successfully from the store. (65536)
   EnforcementStatus     : NotApplicable
   PolicyStoreSource     : PersistentStore
   PolicyStoreSourceType : Local
   
   
   
   
   [MockZipkinCollector]: Running on port '50329'
   Zipkin Endpoint: http://10.1.1.1:50329/api/v2/spans
   Collecting docker logs to: D:\a\opentelemetry-dotnet-instrumentation\opentelemetry-dotnet-instrumentation\build_data\profiler-logs
   Container was started successfully: False
   [MockZipkinCollector]: Shutting down. Total spans received: '0'
   PS> Remove-NetFirewallRule -DisplayName TraceAgent-50329
```

I saw it a few times. I think here is the problem: `Container was started successfully: False`

@RassK any ideas? Should we do some retry?

Kielek:
Should fix #295

## Why
Try to stabilize flaky test.

## What
Change the way how we testing readiness of container.
It is hard to proof, but I think that the strategy `.WithWaitStrategy(waitOS.UntilPortIsAvailable(80)` is error prone. Internally it is trying to connect to the container and it is waiting for the given port.

It should be enough to verify it from the external process (powershell).

## Tests
CI.
When I removed the strategy the container started under 1m.
Probably it is not enough proof, on my fork it usually passed within 5 minutes, even with the strategy.

## Notes

I would like to give a chance for these changes. If the test still will be flaky we can revisit.

### DIFF
--- a/test/integration-tests/IntegrationTests.Helpers/PowershellHelper.cs
+++ b/test/integration-tests/IntegrationTests.Helpers/PowershellHelper.cs
@@ -5,7 +5,7 @@ namespace IntegrationTests.Helpers
 {
     public static class PowershellHelper
     {
-        public static void RunCommand(string psCommand, ITestOutputHelper outputHelper)
+        public static (string StandardOutput, string ErrorOutput) RunCommand(string psCommand, ITestOutputHelper outputHelper)
         {
             ProcessStartInfo startInfo = new ProcessStartInfo();
             startInfo.FileName = @"powershell.exe";
@@ -22,6 +22,8 @@ namespace IntegrationTests.Helpers
 
             outputHelper.WriteLine($"PS> {psCommand}");
             outputHelper.WriteResult(helper);
+
+            return (helper.StandardOutput, helper.ErrorOutput);
         }
     }
 }
--- a/test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
+++ b/test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Diagnostics;
 using System.IO;
-using System.Linq;
 using DotNet.Testcontainers.Containers.Builders;
 using DotNet.Testcontainers.Containers.Modules;
 using DotNet.Testcontainers.Containers.OutputConsumers;
--- a/test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
+++ b/test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
@@ -5,6 +5,7 @@ using DotNet.Testcontainers.Containers.Builders;
 using DotNet.Testcontainers.Containers.Modules;
 using DotNet.Testcontainers.Containers.OutputConsumers;
 using DotNet.Testcontainers.Containers.WaitStrategies;
+using FluentAssertions;
 using IntegrationTests.Helpers.Models;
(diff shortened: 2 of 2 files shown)


nrcventura (approved):
This seems like it is definitely worth trying out. If this consistently speeds up the setup steps of this test, I think that we might still run into some flakiness issues. There's an additional error that appears less frequently where the docker network setup fails because it says that the network rule is already in place. If that error becomes more frequent, we can address it separately.

zacharycmontoya (approved):
Worth a shot 👍🏼

RassK (approved):

RassK:
> This seems like it is definitely worth trying out. If this consistently speeds up the setup steps of this test, I think that we might still run into some flakiness issues. There's an additional error that appears less frequently where the docker network setup fails because it says that the network rule is already in place. If that error becomes more frequent, we can address it separately.

Seems there are issues when something fails in the middle, then it's not fully cleanup 🤔

PR #324, opened 2022-01-19
[WIP] Use OTLP exporter by default

pellared:
Fixes #325

Changes proposed in this pull request:

- Use OTLP exporter by default.

### Issues

1. On .NET Core for some reason the spans from the OTLP exporter do not reach the Collector. Possibly caused by bad configuration.

2. OTLP Exporter vs .NET Framework:

```
Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.TypeInitializationException: The type initializer for 'OpenTelemetry.ClrProfiler.Managed.Loader.Startup' threw an exception. ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.IO.FileNotFoundException: Error loading native library. Not found in any of the possible locations: C:\Users\rpajak\repos\opentelemetry-dotnet-instrumentation\bin\tracer-home\net461\grpc_csharp_ext.x86.dll,C:\Users\rpajak\repos\opentelemetry-dotnet-instrumentation\bin\tracer-home\net461\runtimes/win/native\grpc_csharp_ext.x86.dll,C:\Users\rpajak\repos\opentelemetry-dotnet-instrumentation\bin\tracer-home\net461\../..\runtimes/win/native\grpc_csharp_ext.x86.dll
   at Grpc.Core.Internal.UnmanagedLibrary.FirstValidLibraryPath(String[] libraryPathAlternatives)
   at Grpc.Core.Internal.UnmanagedLibrary..ctor(String[] libraryPathAlternatives)
   at Grpc.Core.Internal.NativeExtension.LoadUnmanagedLibrary()
   at Grpc.Core.Internal.NativeExtension.LoadNativeMethods()
   at Grpc.Core.Internal.NativeExtension..ctor()
   at Grpc.Core.Internal.NativeExtension.Get()
   at Grpc.Core.GrpcEnvironment.GrpcNativeInit()
   at Grpc.Core.GrpcEnvironment..ctor()
   at Grpc.Core.GrpcEnvironment.AddRef()
   at Grpc.Core.Channel..ctor(String target, ChannelCredentials credentials, IEnumerable`1 options)
   at OpenTelemetry.Exporter.OtlpExporterOptionsExtensions.CreateChannel(OtlpExporterOptions options)
[...]

PR #343, opened 2022-02-01
OTLP Exported as default for .NET (Core)

### LINKED ISSUE #325: Use OTLP gRPC Exporter by default
Per https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/255#issuecomment-970706772

Kielek:
Partially Fixes #325 
Relates to #324

Use the OTLP exporter as default for the .NET (Core) and Zipkin for .NET Framework

## Tests

CI
Local tests for ConsoleApp, works fine for
* .NET 4.6.1 (Zipkin exporter)
* .NET Core App 4.1 (otlp)

Works with workaround for
* .NET 5.0 (otlp) 

## Issue to be discussed 

https://www.nuget.org/packages/Grpc.Net.Client/2.32.0 used by https://www.nuget.org/packages/OpenTelemetry.Exporter.OpenTelemetryProtocol/1.2.0-rc1 is released for
* .NET Standard 2.1
* .NET 5.0

Otel-dotnet-inturmenation, for now, is released as .NET 4.6.1 and .NET Core App 3.1.

.NET Core App 3.1 is using .NET Standard 2.1 (deploys Grpc.Net.Client to the tracer-home).

.NET Standard 2.1 of Grpc.Net.Client is not working correctly with .NET 5.0. It leads to following error

```logs
info    zapgrpc/zapgrpc.go:174  [transport] transport: loopyWriter.run returning. connection error: desc = "transport is closing"       {"grpc_log": true}
```

As a workaround, there is a possibility to force load Grpc.Net.Client for .NET 5.0 from the instrumented application. I think that it is not the long-term sollution.

I think that we should consider separate version of otel-intrumentation for .NET 5.0+.

## Alternative solution - defaulting to Otel over http/protobuf

Is it possible, and it is working correctly both for .NET Framework 4.6.1, NET Core App 3.1 and .NET 5.0 without any hacks for the end user, but we should consider following things:
* defaults are different than prepared by the OTEL .NET SDK,
* defaults protocol for otlp exporter should be http/protobuf. The grpc is allowed only when SDK was already released https://github.com/open-telemetry/opentelemetry-specification/blob/ce50e4634efcba8da445cc23523243cb893905cb/specification/protocol/exporter.md?plain=1#L111-L114
* most of the code to manage default values is internal API, we should copy (and maintatin!) it to our code call it by reflections https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpTraceExporterHelperExtensions.cs#L61
* in the future we should expect more env-variables to manage on our side, see OTEL_EXPORTER_OTLP_TRACES_ENDPOINT, OTEL_EXPORTER_OTLP_METRICS_ENDPOINT, OTEL_EXPORTER_OTLP_TRACES_ENDPOINT, OTEL_EXPORTER_OTLP_METRICS_ENDPOINT, etc.

To reduce the complication on our side we can consider fixing https://github.com/open-telemetry/opentelemetry-dotnet/issues/2857 and waiting for the next RC candidate. If it will cover updates by the Protocol setter it will reduce complication of the code on our side.

## Discusion about changing default exporter to HTTP
https://github.com/open-telemetry/opentelemetry-dotnet/discussions/2855

### DIFF
--- a/samples/ConsoleApp/ConsoleApp.csproj
+++ b/samples/ConsoleApp/ConsoleApp.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
@@ -15,7 +15,8 @@
 
     <!-- Hack to fix SDK dependencies -->
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="6.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.1.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.0.3" />
+
   </ItemGroup>
 
   <ItemGroup>
--- a/samples/ConsoleApp/ConsoleApp.csproj
+++ b/samples/ConsoleApp/ConsoleApp.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
@@ -17,6 +17,8 @@
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="6.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.0.3" />
 
+    <!-- Hack to make the otel exportert working for .NET 5.0 -->
+    <PackageReference Include="Grpc.Net.Client" Version="2.32.0" Condition="'$(TargetFramework)' == 'net5.0'"/>
   </ItemGroup>
 
   <ItemGroup>
--- a/src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
+++ b/src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
@@ -23,7 +23,13 @@ namespace OpenTelemetry.ClrProfiler.Managed.Configuration
                 throw new ArgumentNullException(nameof(source));
             }
 
-            Exporter = source.GetString(ConfigurationKeys.Exporter);
+            Exporter = source.GetString(ConfigurationKeys.Exporter) ??
+#if NETFRAMEWORK
+                       "zipkin";
+#else
+                       "otlp";
+#endif
+
             LoadTracerAtStartup = source.GetBool(ConfigurationKeys.LoadTracerAtStartup) ?? true;
             ConsoleExporterEnabled = source.GetBool(ConfigurationKeys.ConsoleExporterEnabled) ?? true;
 
--- a/dev/envvars.sh
+++ b/dev/envvars.sh
@@ -58,7 +58,6 @@ fi
 export OTEL_DOTNET_TRACER_HOME="${CURDIR}/bin/tracer-home"
 export OTEL_INTEGRATIONS="${CURDIR}/bin/tracer-home/integrations.json"
 export OTEL_TRACE_DEBUG="1"
-export OTEL_EXPORTER="zipkin"
 export OTEL_DUMP_ILREWRITE_ENABLED="0"
 export OTEL_CLR_ENABLE_INLINING="1"
 export OTEL_PROFILER_EXCLUDE_PROCESSES="dotnet.exe,dotnet"
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -65,7 +65,7 @@ The exporter is used to output the telemetry.
 
 | Environment variable | Description | Default |
 |-|-|-|
-| `OTEL_EXPORTER` | The exporter to be used. The Tracer uses it to encode and dispatch traces. Available values are: `zipkin`, `jeager`, `otlp`. | |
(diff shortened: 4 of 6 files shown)


Kielek:
SIG notes: Try to use OTLP over HTTP exporter - check specification and possibility to execute it for all .NET (Core/Framework) versions

Kielek:
@nrcventura, @pjanotti,
I have checked the possibility to use `otlp` exporter over `http/protobuf`.
Is it possible, and it is working correctly both for .NET Framework 4.6.1, NET Core App 3.1 and .NET 5.0 without any hacks for the end user, but we should consider following things
 * defaults are different than prepared by the OTEL .NET SDK,
 * defaults protocol for `otlp` exporter should be `http/protobuf`. The `grpc` is allowed only when SDK was already released https://github.com/open-telemetry/opentelemetry-specification/blob/ce50e4634efcba8da445cc23523243cb893905cb/specification/protocol/exporter.md?plain=1#L111-L114
 * most of the code to manage default values is internal API, we should copy (and maintatin!) it to our code call it by reflections https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpTraceExporterHelperExtensions.cs#L61
 * in the future we should expect more env-variables to manage on our side, see `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`, `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`, `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`, `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`, etc.
 
 

I do not think that we will have possibility to change the defaults after first official release, so it looks like the long term decision.

Personally, I think that we should follow strategies directly from the .NET OTEL SDK.
The cosnsistency will be better for our endusers what is more, we will have less work to maintain duplicated code.

WIP changes,
https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/compare/main...Kielek:otlp-ver-http-as-default-exporter-net-core?expand=1

When the decision will be made, I will be happy to finish this one or create the new one PR with http/protobud as default.

Kielek:
I will update the first comment to update curent  state. It will be esier to follow our options.

pjanotti:
I opened a discussion with SDK for considering a change for the OTLP default from gRPC to HTTP, see https://github.com/open-telemetry/opentelemetry-dotnet/discussions/2855

Kielek:
> I will update the first comment to update curent state. It will be esier to follow our options.

First comment updated.

nrcventura:
> * most of the code to manage default values is internal API, we should copy (and maintatin!) it to our code call it by reflections

Can you explain this a little more? If we only support http/protobuf can we just get away with "hardcoding" some of the public configuration options like users of the SDK might do? I just want to get a better understanding of some of the problems that you are trying to solve.

samples/ConsoleApp/ConsoleApp.csproj
```diff
@@ -15,7 +15,10 @@
 
     <!-- Hack to fix SDK dependencies -->
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="6.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.1.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.0.3" />
+
+    <!-- Hack to make the otel exportert working for .NET 5.0 -->
+    <PackageReference Include="Grpc.Net.Client" Version="2.32.0" Condition="'$(TargetFramework)' == 'net5.0'"/>
```
pellared, line 21: If we want this merged then it should be documented in `USAGE.md`

pellared:
I am good with merging this PR if we describe how it can be used (and gotchas) in `USAGE.md`. 

As far as I understand this would require describing that:
- gRPC OTLP protocol is not working for .NET Framework. but one can always set the HTTP OTLP protocol
- gRPC OTLP protocol requires additional package references on .NET 5 and .NET 6
- OTLP endpoint must be manually set if HTTP OTLP protocol is set


We can have a separate PR to switch to HTTP OTLP protocol by default (also for .NET Framework). It would be easier to do it after https://github.com/open-telemetry/opentelemetry-dotnet/issues/2857 is fixed a new version is released. Also, I think that we can keep the support for the gRPC OTLP protocol (with workarounds and constraints described above).

src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
```diff
@@ -23,7 +23,13 @@ private Settings(IConfigurationSource source)
                 throw new ArgumentNullException(nameof(source));
             }
 
-            Exporter = source.GetString(ConfigurationKeys.Exporter);
+            Exporter = source.GetString(ConfigurationKeys.Exporter) ??
+#if NETFRAMEWORK
+                       "zipkin";
```
pellared, line 28: alternatively, for.NET Framework, we could still use the OTLP exporter, but set HTTP protocol

samples/ConsoleApp/ConsoleApp.csproj
pellared: FYI https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/350

Kielek:
> Can you explain this a little more? If we only support http/protobuf can we just get away with "hardcoding" some of the public configuration options like users of the SDK might do? I just want to get a better understanding of some of the problems that you are trying to solve.

I think that, I have found pretty cheap alternative to avoid copying whole code related to managing url (lack od auto-extending endpoints with suffix such as /v1/traces or /v1/metrics). I consider is a good tradeoff, especially if https://github.com/open-telemetry/opentelemetry-dotnet/pull/2868 will be merged shortly. 

Closing this one, #351 constrains expected changes.

PR #350, opened 2022-02-07
Describe OTLP exporter usage

pellared:
## Why

Currently, trying to use the OTLP exporter usually leads to problems.

Related issue https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/325

Related PR https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/343

## What

Describe the current state of the OTLP exporter support. Proper support of even deperecation of the gRPC protocol can be done later.

## Remarks

For .NET Framework, it is probably more complicated. When I added 

```xml
<PackageReference Include="Grpc.Net.Client" Version="2.32.0" />
```

to the sample app I got:

```csharp
Unhandled Exception: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory 
is corrupt.
   at Grpc.Core.Internal.CompletionQueueSafeHandle.ReleaseHandle()
   at System.Runtime.InteropServices.SafeHandle.InternalDispose()
   at System.Runtime.InteropServices.SafeHandle.Dispose(Boolean disposing)
   at System.Runtime.InteropServices.SafeHandle.Dispose()
   at Grpc.Core.Internal.AsyncCall`2.UnaryCall(TRequest msg)
   at Grpc.Core.DefaultCallInvoker.BlockingUnaryCall[TRequest,TResponse](Method`2 method, String host, CallOptions options, TRequest request)    
   at Grpc.Core.Interceptors.InterceptingCallInvoker.<BlockingUnaryCall>b__3_0[TRequest,TResponse](TRequest req, ClientInterceptorContext`2 ctx) 
   at Grpc.Core.ClientBase.ClientBaseConfiguration.ClientBaseConfigurationInterceptor.BlockingUnaryCall[TRequest,TResponse](TRequest request, ClientInterceptorContext`2 context, BlockingUnaryCallContinuation`2 continuation)
   at Grpc.Core.Interceptors.InterceptingCallInvoker.BlockingUnaryCall[TRequest,TResponse](Method`2 method, String host, CallOptions options, TRequest request)
   at Opentelemetry.Proto.Collector.Trace.V1.TraceService.TraceServiceClient.Export(ExportTraceServiceRequest request, CallOptions options)      
   at Opentelemetry.Proto.Collector.Trace.V1.TraceService.TraceServiceClient.Export(ExportTraceServiceRequest request, Metadata headers, Nullable`1 deadline, CancellationToken cancellationToken)
[...]

Issue #217, opened 2021-09-02
Option to disable the System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport


pellared:
Per: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/211#discussion_r689874809

The user should be able to set some configuration that will make sure that the following code will not be executed:

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/3876971f102c9150a996aed7b6909a1ce493e7e3/src/OpenTelemetry.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs#L89

pellared:
@nrcventura What do you think of moving it post-beta? I think it is important for RC/GA, but I do not find it critical for the beta. Especially that .NET Core 3.1 is planned to be supported only until the end of 2022: https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core.

nrcventura:
> @nrcventura What do you think of moving it post-beta? I think it is important for RC/GA, but I do not find it critical for the beta. Especially that .NET Core 3.1 is planned to be supported only until the end of 2022: https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core.

I'm still nervous about us making a "security" related setting change on behalf of someone else without their consent. Even if this setting only applies to .NET Core 3.1, I think that we should have it. I believe that the bulk of .NET Core usage is still .NET Core 3.1, with people starting to migrate to .NET 6 now.

As an alternative, I'd rather drop "unencrypted" support for .NET Core 3.1 by removing that switch. In theory, someone could create an additional startup hook that only sets that setting, and have that startup hook be the first one in the startup hooks environment variable. Although, it would probably be less code to have an environment variable to control that setting.

pellared:
Let's keep it in the scope then. It is so easy to implement it that I find it is not worth any further discussion 😉

Kielek:
Please check if the second part of the condition is still needed here https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/0fb8f2851b3bbd013bd4a5a9c8128fc1d1f86ad1/src/OpenTelemetry.AutoInstrumentation/Configuration/Settings.cs#L165-L168

PR #351, opened 2022-02-08
OTLP over http/protobuf as default exporter

### LINKED ISSUE #325: Use OTLP gRPC Exporter by default

Kielek:
Fixes #325
Replaces #343 

Sets the OTLP over http/protobuf exporter as default for all .NET versions.

There is a possibility to set up otlp over grpc as a exporter. Manual steps are needed, see #350.

Code related to manage endpoint should be removed when release containing https://github.com/open-telemetry/opentelemetry-dotnet/pull/2868 is available.
Code related to the managing default values for the protocol exporter can be removed if https://github.com/open-telemetry/opentelemetry-dotnet/discussions/2855 ends with changes in OTEL .NET SDK.


src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
```diff
@@ -164,5 +170,32 @@ internal static Settings FromDefaultSources()
...
+        {
+            // the defualt in SDK is grpc. It is oposite that here.
+            var exporterOtlpProtocol = source.GetString(ConfigurationKeys.ExporterOtlpProtocol);
+
+            if (exporterOtlpProtocol == "grpc")
+            {
+                return OtlpExportProtocol.Grpc;
+            }
+
+            return OtlpExportProtocol.HttpProtobuf;
```
pellared, line 184: what do you think of changing to something more or less like:

```csharp
var exporterOtlpProtocol = source.GetString(ConfigurationKeys.ExporterOtlpProtocol);
if string.IsNullOrEmpty(exporterOtlpProtocol)
{
    exporterOtlpProtocol  = "http/protobuf";
}
switch (exporterOtlpProtocol) {
case "http/protobuf":
   return OtlpExportProtocol.HttpProtobuf;
case "grpc":
   return tlpExportProtocol.Grpc;
default:
   throw new Exception(blalbla);
```

the problem is that the current code does not handle invalid values -> e.g. when one sets `blabla` as the env var

the other possibility could be doing something like

```csharp
var exporterOtlpProtocol = source.GetString(ConfigurationKeys.ExporterOtlpProtocol);
if string.IsNullOrEmpty(exporterOtlpProtocol)
{
    return OtlpExportProtocol.HttpProtobuf; // override the default
}
return null; // keep the existing value, do not override anything
```

then the exporter setup code would have to be changed to:

```csharp
builder.AddOtlpExporter(options =>
{
    if (settings.OtlpExportEndpoint != null) options.Protocol = settings.OtlpExportProtocol;
    // ...
});
```

I like the second approach more as it does not require parsing the values and it would not require any changes if support for `http/json` protocol will be added in the SDK

src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
```diff
@@ -164,5 +170,32 @@ internal static Settings FromDefaultSources()
...
+        }
+
+        private static Uri GetExporterOtlpEndpoint(IConfigurationSource source, OtlpExportProtocol otlpExportProtocol)
```
pellared, line 187: do not we need to change the default only if `exporterOtlpEndpoint` is null or empty and `otlpExportProtocol == OtlpExportProtocol.HttpProtobuf` ? otherwise the SDK should provide good value

src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
```diff
@@ -164,5 +170,32 @@ internal static Settings FromDefaultSources()
...
+        {
+            var exporterOtlpProtocol = source.GetString(ConfigurationKeys.ExporterOtlpEndpoint);
```
pellared, line 189: rename `exporterOtlpProtocol` to `exporterOtlpEndpoint` 😉

Kielek: fixed

src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
Kielek: Technically, you are right, but I expect to remove the code just after next release of OTEL .NET SDK.
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2868

I would like to keep it as it is.

src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
```diff
@@ -164,5 +170,34 @@ internal static Settings FromDefaultSources()
...
+            {
+                // override settings only for http/protobuf
+                // the second part of the condition is needed to override default endpoint as long as https://github.com/open-telemetry/opentelemetry-dotnet/pull/2868 is not fixed
+                return OpenTelemetry.Exporter.OtlpExportProtocol.HttpProtobuf;
+            }
+
+            return null;
+        }
+
+        private static Uri GetExporterOtlpEndpoint(IConfigurationSource source, OtlpExportProtocol? otlpExportProtocol)
```
pellared, line 189: consider adding a comment that this would not be needed when https://github.com/open-telemetry/opentelemetry-dotnet/pull/2868 is fixed

pellared (approved):

src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
Kielek: There is a comment near the `OtlpExportEndpoint` property. I think that it is enough.

Issue created #352

src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
```diff
@@ -23,7 +24,10 @@ private Settings(IConfigurationSource source)
                 throw new ArgumentNullException(nameof(source));
             }
 
-            Exporter = source.GetString(ConfigurationKeys.Exporter);
+            Exporter = source.GetString(ConfigurationKeys.Exporter) ?? "otlp";
+            OtlpExportProtocol = GetExporterOtlpProtocol(source);
+            OtlpExportEndpoint = GetExporterOtlpEndpoint(source, OtlpExportProtocol);
```
nrcventura, line 29: There's some non-trivial logic in these 2 new methods. Do we think that it's worth introducing some unit tests, even though some of that code will go away when the SDK is using the correct port by default?

rajkumar-rangaraj (approved):

src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
```diff
@@ -164,5 +170,34 @@ internal static Settings FromDefaultSources()
...
+            var exporterOtlpProtocol = source.GetString(ConfigurationKeys.ExporterOtlpProtocol);
+
+            if (string.IsNullOrEmpty(exporterOtlpProtocol) || exporterOtlpProtocol == "http/protobuf")
+            {
+                // override settings only for http/protobuf
+                // the second part of the condition is needed to override default endpoint as long as https://github.com/open-telemetry/opentelemetry-dotnet/pull/2868 is not fixed
+                return OpenTelemetry.Exporter.OtlpExportProtocol.HttpProtobuf;
+            }
+
+            return null;
```
pjanotti, line 186: Returning `null` here is surprising without the full context of the PR and interaction with the SDK. I think it deserves a comment.

src/OpenTelemetry.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs
```diff
@@ -70,21 +71,32 @@ private static TracerProviderBuilder SetExporter(this TracerProviderBuilder buil
...
+                            options.Endpoint = settings.OtlpExportEndpoint;
+                        }
                     });
                     break;
                 case "":
                 case null:
                     break;
                 default:
-                    throw new ArgumentOutOfRangeException("The exporter name is not recognised");
+                    throw new ArgumentOutOfRangeException("The exporter name is not recognized");
```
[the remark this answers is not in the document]
Kielek: Sure, I will change it also to `FormatException`. It is type of exception raised by the OTEL .NET SDK when the env. vars cannot be parsed/accepted.

src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
Kielek: Done,

src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
```diff
@@ -164,5 +170,35 @@ internal static Settings FromDefaultSources()
...
+            var exporterOtlpEndpoint = source.GetString(ConfigurationKeys.ExporterOtlpEndpoint);
+
+            if (!otlpExportProtocol.HasValue || otlpExportProtocol.Value != OpenTelemetry.Exporter.OtlpExportProtocol.HttpProtobuf || !string.IsNullOrWhiteSpace(exporterOtlpEndpoint))
+            {
+                // null value here means that it will be handled by OTEL .NET SDK
+                return null;
+            }
+
+            // override endpoint only for otlp over http/protobuf
+            return new Uri("http://localhost:4318/v1/traces");
```
pellared, line 201: how about the following in order to have the flow similar to the one in `GetExporterOtlpProtocol`?

```suggestion
            if (otlpExportProtocol == OpenTelemetry.Exporter.OtlpExportProtocol.HttpProtobuf && string.IsNullOrWhiteSpace(exporterOtlpEndpoint))
            {
                // override endpoint only for otlp over http/protobuf
                return new Uri("http://localhost:4318/v1/traces");
            }

            // null value here means that it will be handled by OTEL .NET SDK
            return null;
```

also I think the condition is simpler a little bit

src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
Kielek: I hope that most of the changes will be removed shortly. It was the only reason to not include the UnitTests.

Added, both scaffolding for the Managed.Tests and tests.

test/OpenTelemetry.ClrProfiler.Managed.Tests/Configuration/SettingsTests.cs
```diff
@@ -0,0 +1,86 @@
...
+        [InlineData("nonExistingProtocol")]
+        public void FromDefaultSources_OtlpExportEndpointIsNullWhenCorrespondingEnvVarIsSet(string otlpProtocol)
+        {
+            Environment.SetEnvironmentVariable(ConfigurationKeys.ExporterOtlpProtocol, otlpProtocol);
+            Environment.SetEnvironmentVariable(ConfigurationKeys.ExporterOtlpEndpoint, "http://customurl:1234");
+
+            var settings = Settings.FromDefaultSources();
+
+            settings.OtlpExportEndpoint.Should().BeNull();
+        }
```
pellared, line 78: As a reader, I would not really understand what is the purpose of those tests (apart from code coverage). They do not say why some of the values are so strange.

I want to have some information like "overriding as we want to have a different default value than SDK" or "changing the value to mitigate a bug in SDK"  or "leaving the values as null as SDK will handle it"

WDYT? Do you want a code suggestion?

Kielek: Discussed offline. Added some comments.

pellared (approved):

src/OpenTelemetry.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs
```diff
@@ -70,21 +71,32 @@ private static TracerProviderBuilder SetExporter(this TracerProviderBuilder buil
...
+                            options.Endpoint = settings.OtlpExportEndpoint;
+                        }
                     });
                     break;
                 case "":
                 case null:
                     break;
                 default:
-                    throw new ArgumentOutOfRangeException("The exporter name is not recognised");
+                    throw new FormatException($"The exporter name '{settings}' is not recognized");
```
zacharycmontoya, line 99:
```suggestion
                    throw new FormatException($"The exporter name '{settings.Exporter}' is not recognized");
```
```suggestion
                    throw new FormatException($"The exporter name '{settings}' is not recognized");
```

nrcventura (approved):

src/OpenTelemetry.ClrProfiler.Managed/Configuration/ConfigurationKeys.cs
```diff
@@ -22,10 +22,22 @@ public class ConfigurationKeys
...
         public const string Exporter = "OTEL_EXPORTER";
 
+        /// <summary>
+        /// Configuration key for the otlp protocol to be used.
+        /// Default is <c>"http/protobuf"</c>.
+        /// </summary>
+        public const string ExporterOtlpProtocol = "OTEL_EXPORTER_OTLP_PROTOCOL";
+
+        /// <summary>
+        /// Configuration key for the otlp protocol to be used.
```
zacharycmontoya, line 36: I think this doc comment is incorrect, the endpoint information should have information about the URL's needed to reach the collector

src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
```diff
@@ -164,5 +165,35 @@ internal static Settings FromDefaultSources()
 
             return new Settings(configurationSource);
         }
+
+        private static OtlpExportProtocol? GetExporterOtlpProtocol(IConfigurationSource source)
+        {
+            // the defualt in SDK is grpc. http/protobuf should be default for our purposes
+            var exporterOtlpProtocol = source.GetString(ConfigurationKeys.ExporterOtlpProtocol);
+
+            if (string.IsNullOrEmpty(exporterOtlpProtocol) || exporterOtlpProtocol == "http/protobuf")
```
zacharycmontoya, line 174: It seems a little strange to me to accept a configuration key or environment variable with a `/` in it. Are there any concerns about this?

zacharycmontoya (approved):
LGTM, just left some small questions

src/OpenTelemetry.ClrProfiler.Managed/Configuration/Settings.cs
nrcventura: It's [standardized in the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#specify-protocol).

src/OpenTelemetry.ClrProfiler.Managed/Configuration/ConfigurationKeys.cs
Kielek: Fixed, thanks!

pjanotti (commented):
@Kielek sorry for being late, could you please check the Q about unencrypted gRPC?

src/OpenTelemetry.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs
```diff
@@ -70,21 +71,32 @@ private static TracerProviderBuilder SetExporter(this TracerProviderBuilder buil
...
-                    // Adding the OtlpExporter creates a GrpcChannel.
-                    // This switch must be set before creating a GrpcChannel/HttpClient when calling an insecure gRPC service.
-                    // See: https://docs.microsoft.com/aspnet/core/grpc/troubleshoot#call-insecure-grpc-services-with-net-core-client
-                    AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", true);
+                    if (!settings.OtlpExportProtocol.HasValue || settings.OtlpExportProtocol != OtlpExportProtocol.HttpProtobuf)
+                    {
+                        // Adding the OtlpExporter creates a GrpcChannel.
+                        // This switch must be set before creating a GrpcChannel/HttpClient when calling an insecure gRPC service.
+                        // See: https://docs.microsoft.com/aspnet/core/grpc/troubleshoot#call-insecure-grpc-services-with-net-core-client
+                        AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", true);
```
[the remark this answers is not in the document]
nrcventura: That's an interesting idea. I suppose that you are more interested in detecting the usage of `http` instead of `https`, as opposed to some other scheme like `grpc` or a named pipe? If that's the case, perhaps we should add that as a note to #217? 

When taken within the context of #217, do we even need all of those other checks in the `if` statement if we are requiring someone to set an environment variable to opt-in to enabling that global switch?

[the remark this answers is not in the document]
pellared: > When taken within the context of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/217, do we even need all of those other checks in the if statement if we are requiring someone to set an environment variable to opt-in to enabling that global switch?

I agree here with @nrcventura

src/OpenTelemetry.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs
```diff
@@ -70,21 +71,32 @@ private static TracerProviderBuilder SetExporter(this TracerProviderBuilder buil
...
+                            options.Endpoint = settings.OtlpExportEndpoint;
+                        }
                     });
                     break;
                 case "":
                 case null:
                     break;
                 default:
-                    throw new ArgumentOutOfRangeException("The exporter name is not recognised");
+                    throw new FormatException($"The exporter name '{settings.Exporter}' is not recognized");
```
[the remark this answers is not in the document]
pellared: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/351#discussion_r802933611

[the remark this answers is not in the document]
pellared: Related comment https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/351#discussion_r802287375

src/OpenTelemetry.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs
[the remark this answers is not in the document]
pjanotti: Yeah, I forgot about #217. Having the setting as an opt-in is a better option.

src/OpenTelemetry.ClrProfiler.Managed/Configuration/EnvironmentConfigurationHelper.cs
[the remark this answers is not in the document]
pjanotti: Ok, as I said there is opinion weakly held :)

PR #354, opened 2022-02-10
bump unit tests dependencies

Kielek:
Leftover from #351 

bump unit tests dependencies.

I do not want to mix upgrade changes together with regular development in one PR.


pellared (approved):
It would be better to configure dependabot to bump the dependencies within the .props files.

pellared (approved):

PR #362, opened 2022-02-15
Support .NET 6.0

### LINKED ISSUE #278: Support .NET 6
Add support for .NET 6.

It will require bumping the OTel .NET SDK packages to 1.2.x. (not yet released).

We need to run integration tests against .NET 6 instead of .NET 5.

pellared:
Fixes #278

- Run tests additionally against .NET 6.0
- AdditionalDeps and shared store for .NET 6.0


.github/workflows/ci.yml
```diff
@@ -15,6 +15,12 @@ jobs:
     runs-on: ${{ matrix.machine }}
     steps:
       - uses: actions/checkout@v2.4.0
+      - uses: actions/setup-dotnet@v1.9.0
+        with:
+          dotnet-version: | 
+            3.1.x
+            5.0.x
+            6.0.x
```
pellared, line 23: In future, we may consider parallelize the GitHub workflow so that each job runs on a concrete .NET version

RassK (approved):
LGTM

rajkumar-rangaraj (approved):
LGTM

Earlier plan was to change the .NET 5.0 to .NET 6.0 as .NET 5.0 is reaching end of support sooner. Do we plan to keep .NET 5.0 and 6.0 for now and clean up 5.0 later?

nrcventura:
> > Earlier plan was to change the .NET 5.0 to .NET 6.0 as .NET 5.0 is reaching end of support sooner. Do we plan to keep .NET 5.0 and 6.0 for now and clean up 5.0 later?
> 
> In principle, I prefer those as separate PRs. However, unless someone has some reason to keep supporting 5.0, I think we should do it rather sooner than later due to the resources that it takes on CI to run the 5.0 pass.

I think that the .NET 5.0 end of life is coming up soon enough that I don't have any problems with targeting .NET 6.0 instead of 5.0. I usually prefer to just target LTS versions with my CI to minimize the number of versions always getting tested. I usually prefer separate PRs (for adding a new version and subsequently removing a version) to make it a little easier to track down problems that might occur when versions are being changed. But in this case, I don't think that will be a problem.

Kielek:
@pjanotti, @rajkumar-rangaraj, as we based on OTEL .NET SDK we should follow this project. As long as SDK is supporting .NET 5.0 we should do this. I do not see huge benefits for removing support. Especially if supporting means only executing tests in additional configuration (we still support .NET Core 3.1).

pjanotti:
Create https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/366 to track the issue about removing .NET 5.0 support.

pellared:
Let's not merge this PR before it gets reviewed by both @nrcventura and @zacharycmontoya

RassK:
Seems good to have tests at least for all active frameworks. When frameworks goes out of support and wasn't LTS, I think it could be easily dropped.

Issue #332, opened 2022-01-21
Remove PoC leftovers


pellared:
We should get rid of `POC_README.md` and `poc.sh`. However, we need to make sure that we do not miss anything important.

PR #363, opened 2022-02-15
Rename assemblies to OpenTelemetry.AutoInstrumentation

### LINKED ISSUE #349: Proposal: Renaming assemblies and related packages
This issue is in response to the questions in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/314#discussion_r797531853.

We need to come up with names for our assemblies (and ultimately the corresponding the packages) that are consistent with both the OTel .NET SDK and the .NET community, and are reasonably short so as to play nicer with path size limits in Windows. As part of that consistency we should not prefix our current assemblies with `OpenTelemetry.Instrumentation`.

The current set of assemblies to consider when thinking of this naming strategy include:
- OpenTelemetry.ClrProfiler.Managed.Core
- OpenTelemetry.ClrProfiler.Managed.Loader
- OpenTelemetry.ClrProfiler.Managed
- OpenTelemetry.ClrProfiler.Native
- OpenTelemetry.Instrumentation.StartupHook

Here are some ideas to get the discussion started:

1. Use the `OpenTelemetry.AutoInstrumentation` prefix for assemblies. This would give us names that look like:
   - OpenTelemetry.AutoInstrumentation.Core
   - OpenTelemetry.AutoInstrumentation.Loader
   - OpenTelemetry.AutoInstrumentation
   - OpenTelemetry.AutoInstrumentation.Native
   - OpenTelemetry.AutoInstrumentation.StartupHook

2. Use the `OpenTelemetry.NoCode` prefix for assemblies. This would give us names that look like:
   - OpenTelemetry.NoCode.Core
   - OpenTelemetry.NoCode.Loader
   - OpenTelemetry.NoCode
   - OpenTelemetry.NoCode.Native
   - OpenTelemetry.NoCode.StartupHook
 
3. Use the `OpenTelemetry.Auto` prefix.
   - OpenTelemetry.Auto.Core
   - OpenTelemetry.Auto.Loader
   - OpenTelemetry.Auto
   - OpenTelemetry.Auto.Native
   - OpenTelemetry.Auto.StartupHook

Other things to consider when naming the Startup Hook assembly. Should it be considered as an OpenTelemetry extension and use the `OpenTelemetry.Extensions` prefix? Should it be simply `OpenTelemetry.StartupHook`?

I also think that the `Managed` distinction in the original assembly names are not adding extra value to the project, so the example names I listed above dropped that part of the assembly name.

***My current recommendation is to go with approach 1.*** I also think that ***the Startup Hook should use the same naming convention as the other assemblies***, because that assembly is coupled to the loader assembly in this project. If we break that coupling, and the startup hook can be used outside of this project, then we should follow one of the other suggested naming conventions. I do not see any value in providing a NuGet package for any of these assemblies in isolation. As a result we could have a single NuGet package with the name `OpenTelemetry.AutoInstrumentation`.

Approach 2 is even shorter, but I don't think it accurately represents all use cases of this project (there will still be cases where code or build changes need to be made).

Kielek:
Fixes #349

Use the OpenTelemetry.AutoInstrumentation prefix for assemblies. 

- OpenTelemetry.AutoInstrumentation.Core
- OpenTelemetry.AutoInstrumentation.Loader
- OpenTelemetry.AutoInstrumentation
- OpenTelemetry.AutoInstrumentation.Native
- OpenTelemetry.AutoInstrumentation.StartupHook

### Notes
`ArrayHelper.Empty<T>()` is replaced byt Array.Empty<T>.  Side effect of resolving issues after namespace rename.
`POC_REAMDE` file has references to old names/is only partially updates. We have separate issue to clean up it #332


RassK (approved):
LGTM

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
```diff
@@ -32,21 +31,21 @@ public static void Initialize()
...
+                string loaderFilePath = Path.Combine(loaderAssemblyLocation, "OpenTelemetry.AutoInstrumentation.Loader.dll");
                 Assembly loaderAssembly = Assembly.LoadFrom(loaderFilePath);
-                loaderAssembly.CreateInstance("OpenTelemetry.ClrProfiler.Managed.Loader.Startup");
+                loaderAssembly.CreateInstance("OpenTelemetry.AutoInstrumentation.Loader.Startup");
                 StartupHookEventSource.Log.Trace("StartupHook initialized successfully!");
             }
             else
             {
-                StartupHookEventSource.Log.Trace("OpenTelemetry.ClrProfiler.Managed.Instrumentation initialized before startup hook");
+                StartupHookEventSource.Log.Trace("OOpenTelemetry.AutoInstrumentation.Instrumentation initialized before startup hook");
```
nrcventura:
```suggestion
                StartupHookEventSource.Log.Trace("OpenTelemetry.AutoInstrumentation.Instrumentation initialized before startup hook");
```
An extra `O` got introduced with the rename.

pjanotti: There are at least a few of those.

build/nuke/Build.Steps.cs
```diff
@@ -172,20 +172,20 @@ partial class Build
                     .SetFramework(framework)
                     .SetOutput(TracerHomeDirectory / framework)));
 
-            // StartupHook is supported starting .Net Core 3.1.
-            // We need to emit StartupHook and ClrProfilerManagedLoader assemblies only for .NET Core 3.1 target framework.
+            // AutoInstrumentationStartupHook is supported starting .Net Core 3.1.
```
pjanotti, line 175: This refers to the StartupHook as a .NET concept and shouldn't be changed.

build/nuke/Build.Steps.cs
```diff
@@ -172,20 +172,20 @@ partial class Build
...
+            // We need to emit AutoInstrumentationStartupHook and AutoInstrumentationLoader assemblies only for .NET Core 3.1 target framework.
             DotNetPublish(s => s
-                .SetProject(Solution.GetProject(Projects.StartupHook))
+                .SetProject(Solution.GetProject(Projects.AutoInstrumentationStartupHook))
```
pjanotti, line 178: Perhaps for a separate PR but internally I don't think that we need the prefix `AutoInstrumentation`

nrcventura (approved):

build/nuke/Build.Steps.cs
Kielek: Well, I have a problem with a good name for `OpenTelemetry.AutoIntrumentation.csproj`. To be clear, I have added prefixes everywhere. But I understand your concern. If you have an idea how to call it, I will be happy to update all of them.

build/nuke/Build.Steps.cs
Kielek: fixed 3cf27c63d4204c6c279def22755c849db1ed03e0

pjanotti (approved):

build/nuke/Build.Steps.cs
pjanotti: Let's keep as it is for now. Let's the rename settle a bit then revisit if we have a better solution.

PR #375, opened 2022-02-18
Bump lang version to 10.0

Kielek:
#351 leftovers

Bump supported lang version to the latest (10.0).

10.0 version is also used in .NET OTel SDK repo.

When the previous PR was created there were not support for .NET 6/Lang 10 in our pipeline.


Kielek:
Consider to disable whitespace diff when you are checking diff.

Directory.Build.props
```diff
@@ -3,7 +3,7 @@
     <IsPackable>false</IsPackable>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
-    <LangVersion>9.0</LangVersion>
+    <LangVersion>latest</LangVersion>
```
pellared, line 6: is it not safer/more manageable to specify `10.0` explicitly?

test/OpenTelemetry.AutoInstrumentation.Tests/Configuration/SettingsTests.cs
```diff
@@ -6,108 +6,107 @@
 using OpenTelemetry.Exporter;
 using Xunit;
 
-namespace OpenTelemetry.AutoInstrumentation.Tests.Configuration
+namespace OpenTelemetry.AutoInstrumentation.Tests.Configuration;
```
pellared, line 9: How about reformatting the whole codebase for sake of consistency?

Maybe it is only me, but I favor consistency when I am reading code. Even when new features reduces some noise. 😄

Kielek: Sure, I can do it, in this PR?

pellared: As for me. YES 👍

Directory.Build.props
RassK: If I remember right, there was a change in how "latest" was interpreted by different VS versions. I ran into that issue lately and had to explicitly define the lang version.

Kielek: Fixed, 10.0

pellared:
> 10.0 version is also used in .NET OTel SDK repo.

Can you check how they solved it as they should also support .NET 4.6.1?

pellared:
> 10.0 version is also used in .NET OTel SDK repo.

Can you check how they solved it as they should also support .NET 4.6.1?

Kielek:
The issue is related to using `MSBuild` task (it is taking version from the installed VS) from Nuke instead of `DotnetMsBuild` Task (taking msbuild included into dotnet).

For most cases, I was able to fix it, the only leftover is https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/tree/main/test/test-applications/integrations/dependency-libs .
It is referenced by the .vcxproj folder and cannot be easily converted to the the `DotnetMsBuild` due to following commend
```
// Can't use dotnet msbuild, as needs to use the VS version of MSBuild
```

Previous comment is still valid. We can bump environment to `windows-2022` but it requires eitjer
* Possibility to install TargetingFramework for .NET 4.6.1 on Win2022. https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md
* Drop support for .NET 4.6.1

pellared:
But everything works for fine `windows-2019` so we could merge as it is? If so I am fine merging as it is.

I created a separate issue where we can discuss .NET Framework 4.6.1 support https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/376

pellared (approved):

Kielek:
@pellared, yes, IMO it is ready to be merged. Some improvements can be done later.

test/test-applications/integrations/dependency-libs/Directory.Build.props
```diff
@@ -0,0 +1,6 @@
+<Project>
+  <Import Project="..\Directory.Build.props" />
+  <PropertyGroup>
+    <LangVersion>9.0</LangVersion>
```
nrcventura: Should we include a comment here that explains why we are still using `9.0`?

Kielek: fixed 683acedaa30c29bb2a569b0947997ab72d6c38d9

nrcventura (approved):

Issue #321, opened 2022-01-18
Telemetry loss when application is killed or when UnhandledException occurs


pellared:
W should add a "Telemetry loss" section to the docs? 
Currently, I understand these situations will cause a data loss:
- abnormal termination of the process
- unhandled exception terminates the process

Could reduce the impact (but not avoid):
- use Simple processor
- use local OTel Collector for faster data handover

_Originally posted by @RassK in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/294#discussion_r765653240_

pellared:
Can we use the [`UnhandledException`](https://docs.microsoft.com/en-us/dotnet/api/system.appdomain.unhandledexception) event to handle the "unhandled exception terminates the process" use case?

nrcventura:
> Can we use the [`UnhandledException`](https://docs.microsoft.com/en-us/dotnet/api/system.appdomain.unhandledexception) event to handle the "unhandled exception terminates the process" use case?

New Relic is capturing that event, but it seems like an interesting experiment to try. I don't know if there are gotchas with handling that event that could cause a side-effect for the running application.

pellared:
Notes from SIG meeting

1. Check how upstream cleans up the Tracer (maybe
2. Remember also to handle SIG_INT and SIG_TERM 
3. We should add integration tests for applications throwing unhandled exception and is killed

nrcventura:
Handling the `UnhandledException` event partially works. It can sometimes allow the the exporting of the spans that are sitting in the queue of the batch processor (depending on how quickly the application shuts down). However, if the unhandled exception occurs in code similar to the following  code (when handling the `UnhandledException` event), the "First activity" activity can be exported, but the "Second activity" activity does not get exported.
```csharp
using (var activity = myActivitySource.StartActivity("First activity"))
{
    Console.WriteLine("In activity 1");
}

using (var activity = myActivitySource.StartActivity("Second activity"))
{
    Console.WriteLine("In activity 2");
    throw new Exception("Exception inside of using statement.");
}
```
According to https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/try-finally the behavior for the "Second activity" activity is dependent on how the stack is unwound, so it's possible that the behavior can vary.

pellared:
[...]

Issue #334, opened 2022-01-21
AutoInstrumentation is polluting stdout


pellared:
The auto-instrumentation is making `Console.WriteLine`, it should not make any side-effects. We should work on proper logging to the file or add some DEBUG env var to log to stdout only when explicitly configured.

Also the stdout/Console exporter should not be enabled by default.

lachmatt:
I'd like to work on this issue.

PR #386, opened 2022-03-02
Adds handler for UnhandledException event.

nrcventura:
Relates to #321. 

This change will attempt to flush any pending data held by the batch processor. For more information about the limitations of this approach see #321.

Integration tests and documentation will be handled by separate PRs.

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
@@ -114,17 +114,47 @@ public static class Instrumentation
             return;
         }
 
-        _tracerProvider.Dispose();
+        try
+        {
+            _tracerProvider.Dispose();
 
-        Log("OpenTelemetry tracer exit.");
+            Log("OpenTelemetry tracer exit.");
+        }
+        catch (Exception ex)
+        {
+            try
+            {
+                Log($"An error occured while attempting to exit. {ex}");
+            }
+            catch
+            {
+                // If we encounter an error while logging there is nothing else we can do
+                // with the exception.
+            }
+        }
     }
 
     private static void OnUnhandledException(object sender, UnhandledExceptionEventArgs args)
     {
-        Log("UnhandledException event raised.");
-        if (args.IsTerminating)
+        try
         {
-            OnExit(sender, args);
+            Log("UnhandledException event raised.");
+            if (args.IsTerminating)
+            {
+                OnExit(sender, args);
+            }
+        }
+        catch (Exception ex)
+        {
+            try
+            {
+                Log($"An exception occured while processing an unhandled exception. {ex}");
+            }
+            catch
+            {
+                // If we encounter an error while logging there is nothing else we can do
+                // with the exception.
+            }
         }
     }
 
--- a/src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
@@ -70,6 +70,7 @@ public static class Instrumentation
                 // Register to shutdown events
                 AppDomain.CurrentDomain.ProcessExit += OnExit;
                 AppDomain.CurrentDomain.DomainUnload += OnExit;
+                AppDomain.CurrentDomain.UnhandledException += OnUnhandledException;
             }
         }
         catch (Exception ex)
@@ -116,6 +117,15 @@ public static class Instrumentation
         _tracerProvider.Dispose();
 
         Log("OpenTelemetry tracer exit.");
+    }
+
+    private static void OnUnhandledException(object sender, UnhandledExceptionEventArgs args)
+    {
+        Log("UnhandledException event raised.");
+        if (args.IsTerminating)
+        {
+            OnExit(sender, args);
+        }
     }
 
     private static void Log(string message)
--- a/src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
@@ -138,9 +138,9 @@ public static class Instrumentation
     {
         try
         {
-            Log("UnhandledException event raised.");
             if (args.IsTerminating)
             {
+                Log("UnhandledException event raised with a terminating exception.");
                 OnExit(sender, args);
             }
         }


src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -118,6 +119,15 @@ private static void OnExit(object sender, EventArgs e)
         Log("OpenTelemetry tracer exit.");
     }
 
+    private static void OnUnhandledException(object sender, UnhandledExceptionEventArgs args)
```
nrcventura, line 122: In the future we could consider attempting to report this exception. I do not believe that we can appropriately record the exception as an event on a span, because the necessary span context might not be available.

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -118,6 +119,15 @@ private static void OnExit(object sender, EventArgs e)
...
+    {
+        Log("UnhandledException event raised.");
```
Kielek, line 124: I assume that mostly issue will be not related to our code, but it could be useful to have a reason of the event here. Consider to log the data from `ExceptionObject`.

Kielek (approved):

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
pellared: I would also consider moving this logging inside the `if (args.IsTerminating)` block.

Reference: https://docs.microsoft.com/en-us/dotnet/standard/threading/exceptions-in-managed-threads

pellared (dismissed):

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
pellared: I think it is possible something more or less like:

```csharp
if (Activity.Current != null) { 
  // we can report an even or make set error span status as and add the description from exception message
}
```

WDYT?

pellared: I have just found https://github.com/open-telemetry/opentelemetry-dotnet/blob/d716bb17cd8115a3c6d7afe6da74652000aa8372/docs/trace/reporting-exceptions/README.md#unhandled-exception

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
nrcventura: Write now this log message is really just writing text to stdout. I don't know if it is something that we want to keep. We already have #334 to change the way we handle this logging, but I want to defer that discussion until we work on #334.

As of right now, I think there is value in logging a message outside of the `IsTerminating` check, because we have additional logging that occurs when `OnExit` is called inside of that check (but in the `OnExit` method). This could be particularly useful if we are able to capture the exception and add it to telemetry even if the application is not being terminated. I am purposely being vague about how we can report that exception, because we may not have a span available to report the exception on, so we may need to report the exception via the logging pipeline (when available). @pellared is your concern about logging outside of the `IsTerminating` check related to those messages possibly being noisy or a distraction, or not actionable by someone?

I considered adding the exception details to the current log message, but I wanted to delay that decision. Some people can be sensitive to where exception details get recorded, so I did not want to add the exception details to the log message without a way for someone to control those exception details. I'm open to other thoughts on this

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
nrcventura: Even if we use code like that we would still potentially lose some of the spans that are inflight from parallel work, but it would potentially provide some additional context for resolving the error if the span context is available. At the very least, the code that we have needs to be guarded better. We really should prevent exceptions from being thrown in event handlers.

pellared: This can be addressed in a separate PR if you want

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
pellared: > @pellared is your concern about logging outside of the IsTerminating check related to those messages possibly being noisy or a distraction, or not actionable by someone?

Yes. Not a blocker.

> Some people can be sensitive to where exception details get recorded

Totally agree. I think it would violate OWASP logging guidelines.

pellared (approved):

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
nrcventura: > > @pellared is your concern about logging outside of the IsTerminating check related to those messages possibly being noisy or a distraction, or not actionable by someone?
> 
> Yes. Not a blocker.

I think your concern is valid and it's safer for me to move the logging. If we ever do something outside of that check, it'll be easy for us to add logging later. It's easier for us to add something than to take away something later.

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
pellared: Created https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/390

pellared (approved):

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -113,9 +114,48 @@ private static void OnExit(object sender, EventArgs e)
             return;
         }
 
-        _tracerProvider.Dispose();
+        try
+        {
+            _tracerProvider.Dispose();
```
pjanotti, line 119: Just curious: the try-catch here is more to log some info in the unlikely case of an exception. I mean you didn't actually found a case in which the `_traceProvider.Dispose()` is throwing, right?

pjanotti (approved):
LGTM

Kielek (approved):

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
nrcventura: Correct. I didn't look to see if there's an existing case where that call could throw an exception. I put it in place to be extra defensive. It may not be necessary, because the process is terminating anyways, but I'd rather not take the chance of having an exception from our code potentially hide what happened in the application. Taking on #321 will introduce code that will have more potential for triggering an exception. Similarly, I've seen logging systems lead to exceptions too, or sometimes it's the interpolated string that leads to the exception.

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -70,6 +70,7 @@ public static void Initialize()
                 // Register to shutdown events
                 AppDomain.CurrentDomain.ProcessExit += OnExit;
                 AppDomain.CurrentDomain.DomainUnload += OnExit;
+                AppDomain.CurrentDomain.UnhandledException += OnUnhandledException;
```
noahfalk, line 73: Using this functionality typically improves one set of problems (losing extra telemetry) but creates a new set of problems (apps that hang instead of crash, apps that obscure the error in the user's code with errors from OT's cleanup code). It might still be a very favorable trade if 10% of the time the results are improved vs 0.1% of the time the results are worse, it just something to be aware of and be pragmatic about.

PR #391, opened 2022-03-04
Add health check delay to ASP.NET integration tests

RassK:
### What

Adds health checks and delays if ASP.NET app is not available.

### DIFF
--- a/test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
+++ b/test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Threading;
 using DotNet.Testcontainers.Containers.Builders;
 using DotNet.Testcontainers.Containers.Modules;
 using DotNet.Testcontainers.Containers.OutputConsumers;
@@ -82,9 +83,27 @@ public abstract class TestHelper
 
         PowershellHelper.RunCommand($"docker exec {container.Name} curl -v {healthCheckEndpoint}", Output);
 
-        var (standardOutput, _) = PowershellHelper.RunCommand($"Write-Host (Test-NetConnection -ComputerName 'localhost' -Port {webPort}).TcpTestSucceeded", Output);
+        var healthChecks = 0;
+        var healthChecksUrl = $"http://localhost:{webPort}/health-check";
+        var healthChecksPassed = false;
 
-        standardOutput.Should().Contain("True", $"Test connection to the port number {webPort} failed. On this port container service should be available.");
+        do
+        {
+            var (standardOutput, _) = PowershellHelper.RunCommand($"Write-Host (Invoke-WebRequest -Uri {healthChecksUrl}).StatusCode", Output);
+
+            healthChecksPassed = standardOutput.Trim() == "200";
+
+            if (!healthChecksPassed)
+            {
+                healthChecks++;
+
+                // wait for next health check
+                Thread.Sleep(2000 * healthChecks);
+            }
+        }
+        while (healthChecks < 5 && !healthChecksPassed);
+
+        healthChecksPassed.Should().BeTrue($"Health checks to the url '{healthChecksUrl}' failed. Application didn't respond with given time.");
 
         return new Container(container);
     }
--- a/test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
+++ b/test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
@@ -84,7 +84,7 @@ public abstract class TestHelper
         var healthChecksUrl = $"http://localhost:{webPort}/healthz";
 
         // try healthz a few times
-        HttpStatusCode statusCode = 0;
+        HttpStatusCode statusCode = default;
         HttpClient client = new();
 
         for (int retry = 0; retry < 5; retry++)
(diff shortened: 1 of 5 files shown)


test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
```diff
@@ -82,9 +83,27 @@ public Container StartContainer(int traceAgentPort, int webPort)
...
+            var (standardOutput, _) = PowershellHelper.RunCommand($"Write-Host (Invoke-WebRequest -Uri {healthChecksUrl}).StatusCode", Output);
+
+            healthChecksPassed = standardOutput.Trim() == "200";
+
+            if (!healthChecksPassed)
+            {
+                healthChecks++;
+
+                // wait for next health check
+                Thread.Sleep(2000 * healthChecks);
```
rajkumar-rangaraj, line 101: This value is be less and could cause flaky test results. It better to increase the time window or change our approach.

test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
```diff
@@ -82,9 +83,27 @@ public Container StartContainer(int traceAgentPort, int webPort)
...
 
-        var (standardOutput, _) = PowershellHelper.RunCommand($"Write-Host (Test-NetConnection -ComputerName 'localhost' -Port {webPort}).TcpTestSucceeded", Output);
+        var healthChecks = 0;
+        var healthChecksUrl = $"http://localhost:{webPort}/health-check";
+        var healthChecksPassed = false;
 
-        standardOutput.Should().Contain("True", $"Test connection to the port number {webPort} failed. On this port container service should be available.");
+        do
+        {
+            var (standardOutput, _) = PowershellHelper.RunCommand($"Write-Host (Invoke-WebRequest -Uri {healthChecksUrl}).StatusCode", Output);
```
pjanotti, line 92: Q.: Why not use `HttpClient` (or `Ping`) directly?

test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
pjanotti, line 101: For tests we are not typically looking at exponential back-off: we want the test to proceed as soon as it is ready - of course, we don't want to hog resources while doing that but this should be already an I/O call so this probably is not actually needed.

nrcventura: The current hypothesis for the flaky tests is that there is a delay between when the port is available (the current check in place) and when the application can respond to requests. Usually that delay is minuscule, but when the system is under a heavier load, that delay can be relatively long.

nrcventura: It might be possible that the non-busy wait from the timeout will be good enough, and allow the test app enough cpu time to finish starting up.

pellared:
I am going to rerun the build ~10 times. If they pass then I will merge the PR for sake of stability. The comments can be addressed later.

test/integration-tests/aspnet/IntegrationTests.AspNet/AspNetTests.cs
```diff
@@ -35,6 +35,8 @@ public void SubmitsTraces()
             Output.WriteLine("Sample response:");
             Output.WriteLine(content);
 
+            agent.SpanFilters.Add(x => x.Name != "health-check");
```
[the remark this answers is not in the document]
RassK: healtz seems good, as we can reuse it >
https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-6.0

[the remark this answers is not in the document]
pellared: Let's go for `healthz`. It is also kind of inline with https://github.com/open-telemetry/opentelemetry-specification/blob/main/experimental/trace/zpages.md

test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
```diff
@@ -83,27 +81,26 @@ public Container StartContainer(int traceAgentPort, int webPort)
...
 
-        var healthChecks = 0;
         var healthChecksUrl = $"http://localhost:{webPort}/healthz";
-        var healthChecksPassed = false;
 
-        do
-        {
-            var (standardOutput, _) = PowershellHelper.RunCommand($"Write-Host (Invoke-WebRequest -Uri {healthChecksUrl}).StatusCode", Output);
+        // try healthz a few times
+        HttpStatusCode statusCode = 0;
```
pellared, line 87: nit

```suggestion
        HttpStatusCode statusCode = default;
```

pellared (approved):

nrcventura (approved):

test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
```diff
@@ -82,9 +81,26 @@ public Container StartContainer(int traceAgentPort, int webPort)
...
+        {
+            var response = await client.GetAsync(healthChecksUrl);
+            statusCode = response.StatusCode;
+
+            if (statusCode == HttpStatusCode.OK)
+            {
+                break;
+            }
+
+            Thread.Sleep(TimeSpan.FromSeconds(4));
```
pjanotti, line 100: nit: since the code now already has an `await` use `Task.Delay` here instead of `Thread.Sleep`.

pjanotti (approved):

test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
RassK: Just an initial value, might need some fine-tuning.

Kielek (approved):

rajkumar-rangaraj (approved):

pjanotti:
@pellared it passed 4 times since the last changes, since it is already an improvement of previous I'm merging now. We can keep improving if needed.

Issue #411, opened 2022-03-15
MongoDB Instrumentation keeps generating spans on the background


pjanotti:
**Description**
The MongoDB Instrumentation keeps generating spans triggered by the driver that are not related to any transaction or client call. This seems a regression of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/207

1. Run the Samples.AspNetCoreMvc31 and let it running in the background
2. Check the spans being generated (see image above)

**Expected behavior**
Only spans directed related to the client app requests.

**Screenshots**
![image](https://user-images.githubusercontent.com/5600755/158480502-a10c9f62-a4af-4c52-9de9-33386f990c3f.png)

pellared:
We should consider using https://www.nuget.org/packages/MongoDB.Driver.Core.Extensions.OpenTelemetry/ instead

Issue #315, opened 2022-01-12
Replace all usages of OpenTelemetry.ClrProfiler.Emit with OpenTelemetry.ClrProfiler.DuckTyping


zacharycmontoya:
The original lookup of methods/properties on instrumented types was done via reflection and implemented in the OpenTelemetry.ClrProfiler.Emit namespace. However, the new DuckTyping mechanisms in the OpenTelemetry.ClrProfiler.DuckTyping namespace are faster and should be easier to use, so we should replace the initial reflection-based mechanisms.

The primary benefit is that we have less code to maintain and fewer mechanisms to do the same thing.

Issue #414, opened 2022-03-16
Flaky test: `IntegrationTests.MongoDB.MongoDbTests.SubmitsTraces`


pjanotti:
Test `IntegrationTests.MongoDB.MongoDbTests.SubmitsTraces` failure that passed on the retry.

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/runs/5575541741?check_suite_focus=true#step:5:3035

```terminal
  Error: [xUnit.net 00:05:23.93]     IntegrationTests.MongoDB.MongoDbTests.SubmitsTraces [FAIL]
  19:50:59 [ERR] [xUnit.net 00:05:23.93]     IntegrationTests.MongoDB.MongoDbTests.SubmitsTraces [FAIL]
  19:50:59 [DBG]   Failed IntegrationTests.MongoDB.MongoDbTests.SubmitsTraces [1 ms]
  19:50:59 [DBG]   Error Message:
  19:50:59 [DBG]    System.AggregateException : One or more errors occurred. (One or more errors occurred. (The operation has timed out.)) (The following constructor parameters did not have matching fixture data: MongoDbFixture mongoDb)
  19:50:59 [DBG] ---- System.AggregateException : One or more errors occurred. (The operation has timed out.)
  19:50:59 [DBG] -------- System.TimeoutException : The operation has timed out.
  19:50:59 [DBG] ---- The following constructor parameters did not have matching fixture data: MongoDbFixture mongoDb
  19:50:59 [DBG]   Stack Trace:
  19:50:59 [DBG]   
  19:50:59 [DBG] ----- Inner Stack Trace #1 (System.AggregateException) -----
  19:50:59 [DBG]    at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
  19:50:59 [DBG]    at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
  19:50:59 [DBG]    at System.Threading.Tasks.Task.Wait()
  19:50:59 [DBG]    at IntegrationTests.MongoDB.MongoDbFixture.LaunchMongoContainer(Int32 port) in D:\a\opentelemetry-dotnet-instrumentation\opentelemetry-dotnet-instrumentation\test\integration-tests\IntegrationTests.MongoDB\MongoDbCollection.cs:line 78
[...]

PR #421, opened 2022-03-23
Update MongoDb integration

RassK:
### What

Upgrades MongoDb integration to community's solution (DiagnosticSource package).

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/315
Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/411
Changes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/414
 
### Notes
It's a first POC using community's contributions to OTel.

After some internal discussions we found that we should provide an adapter that does the mongodb internal telemetry conversion to activities package (`MongoDB.Driver.Core.Extensions.DiagnosticSources`) initialization. 

There are 2 conflicting paths:
* It's not good to compete with `MongoDB.Driver.Core.Extensions.DiagnosticSources`
* It's not good to require customers to bootstrap `MongoDB.Driver.Core.Extensions.DiagnosticSources`

In case we do the adapter initialization ourself, we need to be sure if customer does it then we do merge the changes instead of replacing customer changes or vice-versa. Could be another dependency versioning issue.


rajkumar-rangaraj:
Researched to check, if reflection could help. It won't help here as we need to update the settings during `MongoClient` object creation. IL Rewrite would be the best fit for this case.

RassK:
> Researched to check, if reflection could help. It won't help here as we need to update the settings during `MongoClient` object creation. IL Rewrite would be the best fit for this case.

I was thinking about IL Rerewrite + DuckTyping, so there should not be any version conflicts.

RassK:
Unfortunately the IL rewrite part was the easiest one. Due `MongoDB.Driver.Core.Extensions.DiagnosticSources` contains the logic and depends on `MongoDB.Driver.Core` then these references are required. This takes us further towards dependency hell.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/MongoDb/MongoClientIntegration.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// <returns>Calltarget state value</returns>
+        public static CallTargetState OnMethodBegin<TTarget, TMongoClientSettings>(TTarget instance, TMongoClientSettings settings)
+        {
+            // Load driver adapter
+            Type driverAcitivitySubscriber = Type.GetType("MongoDB.Driver.Core.Extensions.DiagnosticSources.DiagnosticsActivityEventSubscriber, MongoDB.Driver.Core.Extensions.DiagnosticSources");
+            if (driverAcitivitySubscriber == null)
+            {
+                string basePath = Environment.GetEnvironmentVariable("DOTNET_SHARED_STORE");
+                string processArch = Environment.Is64BitProcess ? "x64" : "x32";
+                string assemblyPath = Path.Combine(basePath, processArch, @"netcoreapp3.1\mongodb.driver.core.extensions.diagnosticsources\1.1.0\lib\netstandard2.0\MongoDB.Driver.Core.Extensions.DiagnosticSources.dll");
```
RassK, line 57: Any ways to prob this path automatically? So we can only use Assembly.Load?

src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationHelper.cs
```diff
@@ -31,7 +31,8 @@ internal static class EnvironmentConfigurationHelper
     {
         [Instrumentation.HttpClient] = builder => builder.AddHttpClientInstrumentation(),
         [Instrumentation.AspNet] = builder => builder.AddSdkAspNetInstrumentation(),
-        [Instrumentation.SqlClient] = builder => builder.AddSqlClientInstrumentation()
+        [Instrumentation.SqlClient] = builder => builder.AddSqlClientInstrumentation(),
+        [Instrumentation.MongoDb] = builder => builder.AddSource("MongoDB.Driver.Core.Extensions.DiagnosticSources")
```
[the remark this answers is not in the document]
RassK: hmm, wondering what we should add? Instrumentation is source code based but it would not work without bytecode modifications 🤔

nrcventura:
Just a thought about simplifying things... Should we split this into 2 separate efforts? 1. Provide a way to use the community provided instrumentation but require manual bootstrapping, and 2. Provide auto-bootstrapping of the community provided package. By splitting it up this way we could potentially delay needing to solve the dependency management problem. I think this is also what we are currently doing for Redis.

RassK:
> Just a thought about simplifying things... Should we split this into 2 separate efforts? 1. Provide a way to use the community provided instrumentation but require manual bootstrapping, and 2. Provide auto-bootstrapping of the community provided package. By splitting it up this way we could potentially delay needing to solve the dependency management problem. I think this is also what we are currently doing for Redis.

I currently simplified it another way: dynamically loading `MongoDB.Driver.Core.Extensions.DiagnosticSources` if this library is missing. Reflection + Expressions will bridge `MongoDB.Driver.Core.Extensions.DiagnosticSources` with `MongoDB.Driver.Core`, last one should be anyway loaded if MongoDB is used.

src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationHelper.cs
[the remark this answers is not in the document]
nrcventura: I'm leaning towards calling it byte-code instrumentation, because it will only work if the profiling APIs are enabled. The would need to provide their own code to do the bootstrapping for it to be fully considered source-code instrumentation.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/MongoDb/MongoClientIntegration.cs
```diff
@@ -0,0 +1,151 @@
...
+        /// <typeparam name="TMongoClientSettings">Type of the settings</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="settings">The source of the original GraphQL query</param>
+        /// <returns>Calltarget state value</returns>
+        public static CallTargetState OnMethodBegin<TTarget, TMongoClientSettings>(TTarget instance, TMongoClientSettings settings)
+        {
+#if NETFRAMEWORK
+            // Additional deps doesn't support .NET FX
+            // TODO: Find another way how to ship "MongoDB.Driver.Core.Extensions.DiagnosticSources"
+            return CallTargetState.GetDefault();
```
nrcventura, line 58: I think that this is a good compromise for now. We should document this behavior so that we don't forget about it, and to help communicate this limitation and what it means for .net framework applications.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/MongoDb/MongoClientIntegration.cs
```diff
@@ -0,0 +1,151 @@
...
+#endif
+
+#if NETCOREAPP3_1_OR_GREATER
+            Assembly driverExtensions;
+            Type driverAcitivitySubscriber = Type.GetType("MongoDB.Driver.Core.Extensions.DiagnosticSources.DiagnosticsActivityEventSubscriber, MongoDB.Driver.Core.Extensions.DiagnosticSources");
```
zacharycmontoya, line 63: Typo:

```suggestion
            Type driverActivitySubscriber = Type.GetType("MongoDB.Driver.Core.Extensions.DiagnosticSources.DiagnosticsActivityEventSubscriber, MongoDB.Driver.Core.Extensions.DiagnosticSources");
```

src/OpenTelemetry.AutoInstrumentation/Instrumentations/MongoDb/MongoClientIntegration.cs
```diff
@@ -0,0 +1,151 @@
...
+            if (driverAcitivitySubscriber == null)
```
zacharycmontoya, line 64: In your testing, did you make sure that this if statement is only entered once? If not, I think you might need to assign `driverAcitivitySubscriber = driverExtensions.GetType("MongoDB.Driver.Core.Extensions.DiagnosticSources.DiagnosticsActivityEventSubscriber")`

src/OpenTelemetry.AutoInstrumentation/Instrumentations/MongoDb/MongoClientIntegration.cs
```diff
@@ -0,0 +1,151 @@
...
+
+            if (optionsType != null)
+            {
+                var setListenerAction = _setActivityListener ??
+                    (_setActivityListener = GetClusterConfiguratorExpression(driverExtensions, optionsType).Compile());
+
+                settings
+                   .GetType()
+                   .GetProperty("ClusterConfigurator", BindingFlags.Public | BindingFlags.Instance)
+                   .SetValue(settings, setListenerAction);
```
zacharycmontoya, line 93: Should we combine the delegates so we don't overwrite a developer's existing delegate?
```csharp
                var property = settings.GetType().GetProperty("ClusterConfigurator", BindingFlags.Public | BindingFlags.Instance);
                var existingDelegate = property.GetValue(settings) as Delegate;
                property.SetValue(settings, Delegate.Combine(existingDelegate, setListenerAction));
```

zacharycmontoya:
Looking at the diagnostic listener from a higher level, I saw the following [blog post](https://jimmybogard.com/building-end-to-end-diagnostics-activity-and-span-correlation/) that describes how Jimmy Bogard created a diagnostic listener for MongoDB (and he's the author of the extension package). Fundamentally, we just need to create an object that implements the `IEventSubscriber` which exists in the core MongoDb assembly, and then we manage the start/stop of the Activity ourselves.

I don't want to block the existing PR, but with the ducktyping tools we have I think that we could roll our own subscriber for .NET Framework so that we can do instrumentation when the extension package is not available.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/MongoDb/MongoClientIntegration.cs
RassK: Thanks, wanted to keep it later but as you already provided the solution, I'm going to include it.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/MongoDb/MongoClientIntegration.cs
RassK: Seems the else part of the if is never invoked 🤔

src/OpenTelemetry.AutoInstrumentation/Instrumentations/MongoDb/MongoClientIntegration.cs
RassK: @rajkumar-rangaraj thanks for the AdditionalDeps clue. I completely forgot that it works in pair with shared store and actually defines what and where to load. Managed to cleanup all that loading.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/MongoDb/MongoClientIntegration.cs
RassK: solved the original issue. https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/421#pullrequestreview-925911047

src/OpenTelemetry.AutoInstrumentation/Instrumentations/MongoDb/MongoClientIntegration.cs
RassK: ... btw, do you also know if we can only include `MongoDB.Driver.Core.Extensions.DiagnosticSources.dll` in the shared store? Currently it's pulling all of the sub dependencies also 🤔

nrcventura:
@RassK I think that we should address the merge conflicts (from the PR where we removed the previous instrumentation and tests) before we review this again. I think that we are all in agreement on the approach taken in this PR.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/MongoDb/MongoClientIntegration.cs
RassK: @rajkumar-rangaraj any ideas ⬆️ ? It will help to optimize delivery by removing 2 unnecessary packages.

nrcventura (approved):

rajkumar-rangaraj (approved):
LGTM

pjanotti (approved):
LGTM

PR #312, opened 2022-01-11
Remove OTEL_SERVICE, OTEL_VERSION, OTEL_ENV, OTEL_TAGS env vars

pellared:
Fixes #287

Changes proposed in this pull request:
- Remove `OTEL_SERVICE`, `OTEL_VERSION`, `OTEL_ENV`, `OTEL_TAGS` in favor of `OTEL_RESOURCE_ATTRIBUTES` and `OTEL_SERVICE_NAME` which are already supported.

PR #456, opened 2022-04-06
Logging improvements

### LINKED ISSUE #334: AutoInstrumentation is polluting stdout

lachmatt:
## Why
Remove console logging.

Fixes #334 

## What

Simple logging to a file, based on logic from loader:
- uses modified `FileSink` from loader
- bounded log file size
- noop logger is used in case of problems with constructing file-based logger

Possible improvements:
- file rolling
- more efficient logging
- better formatting
- log entries formatting consistent with native logs

## Tests
Smoke tests.


src/OpenTelemetry.AutoInstrumentation/Logging/OtelLogging.cs
```diff
@@ -0,0 +1,133 @@
...
+        Logger = new Logger(sink);
+    }
+
+    internal static ILogger GetLogger() => Logger;
+
+    private static string GetLogFileName()
+    {
+        try
+        {
+            using var process = Process.GetCurrentProcess();
```
lachmatt, line 60: should it be handled similarly to https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/Util/ProcessHelpers.cs ?

Kielek (approved):

src/OpenTelemetry.AutoInstrumentation/Logging/OtelLogging.cs
lachmatt: I think it's not needed for now.

src/OpenTelemetry.AutoInstrumentation/Logging/OtelLogging.cs
```diff
@@ -0,0 +1,133 @@
...
+            return $"dotnet-tracer-managed-{process.ProcessName}-{process.Id}.log";
```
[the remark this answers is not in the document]
pjanotti: See https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/312#discussion_r782478960

pjanotti:
Added a related issue https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/457

src/OpenTelemetry.AutoInstrumentation/Instrumentations/GraphQL/GraphQLCommon.cs
```diff
@@ -43,7 +43,7 @@ internal class GraphQLCommon
     internal static readonly ActivitySource ActivitySource = new ActivitySource(
         "OpenTelemetry.AutoInstrumentation.GraphQL", "0.0.1");
 
-    private static readonly ILogger Log = ConsoleLogger.Create(typeof(GraphQLCommon));
```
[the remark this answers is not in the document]
lachmatt: Previously, every call was returning new instance of `ConsoleLogger`, that was prepending type name to a message.

src/OpenTelemetry.AutoInstrumentation/Logging/OtelLogging.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+        catch
+        {
+            // We can't get the process info
+            return "dotnet-tracer-managed.log";
```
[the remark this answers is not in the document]
pjanotti: you could use a timestamp instead of guid but but then outside of this function you would need to handle the case if the file is already owned by other process.

src/OpenTelemetry.AutoInstrumentation/Logging/OtelLogging.cs
```diff
@@ -0,0 +1,133 @@
...
+                var fileName = GetLogFileName();
+                var logPath = Path.Combine(logDirectory, fileName);
+                sink = new FileSink(logPath);
+            }
+        }
+        catch (Exception)
+        {
+            // ignored
+        }
+
```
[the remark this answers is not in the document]
lachmatt: fixed in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/456/commits/829226dcd1b3a309321395d2367e98cd603106f9

src/OpenTelemetry.AutoInstrumentation/Logging/OtelLogging.cs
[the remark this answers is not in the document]
lachmatt: fixed in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/456/commits/829226dcd1b3a309321395d2367e98cd603106f9

src/OpenTelemetry.AutoInstrumentation/Logging/OtelLogging.cs
```diff
@@ -0,0 +1,133 @@
...
+            if (logDirectory != null)
+            {
+                var fileName = GetLogFileName();
+                var logPath = Path.Combine(logDirectory, fileName);
+                sink = new FileSink(logPath);
+            }
+        }
+        catch (Exception)
+        {
+            // ignored
```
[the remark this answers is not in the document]
lachmatt: fixed in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/456/commits/829226dcd1b3a309321395d2367e98cd603106f9

pjanotti (approved):
LGTM - @lachmatt could you please rebase and rebuild so you fix the change that came with the addition of the EventListener

Kielek (approved):

rajkumar-rangaraj (approved):

src/OpenTelemetry.AutoInstrumentation/Logging/OtelLogging.cs
```diff
@@ -0,0 +1,140 @@
...
+    {
+        string logDirectory;
+
+        try
+        {
+            logDirectory = Environment.GetEnvironmentVariable(ConfigurationKeys.LogDirectory);
+
+            if (logDirectory == null)
+            {
+                var nativeLogFile = Environment.GetEnvironmentVariable(ConfigurationKeys.LogPath);
```
pellared, line 87: how about renaming `nativeLogFile` to something more or less like `envVarValue` ?

docs/config.md
```diff
@@ -58,6 +58,8 @@ Default logs directory paths are:
 - Windows: `%ProgramData%\OpenTelemetry .NET AutoInstrumentation\logs`
 - Linux: `/var/log/opentelemetry/dotnet`
 
+If there is a problem with creation of the default directory, the path of the current user's [temporary folder](https://docs.microsoft.com/en-us/dotnet/api/System.IO.Path.GetTempPath?view=net-6.0) is used.
```
pellared, line 61:
```suggestion
If there is a problem with the creation of the default directory,
the path of the current user's [temporary folder](https://docs.microsoft.com/en-us/dotnet/api/System.IO.Path.GetTempPath?view=net-6.0)
is used.
```

pellared (approved):
little things

zacharycmontoya (commented):
Looks good, however since https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/458 was merged, the following files may need to be updated for tests to pass:
- `src/OpenTelemetry.AutoInstrumentation/Diagnostics/SdkSelfDiagnosticsEventListener.cs`
- `test/OpenTelemetry.AutoInstrumentation.Tests/Diagnostics/SdkSelfDiagnosticsEventListenerTests.cs`

src/OpenTelemetry.AutoInstrumentation/Logging/Logger.cs
```diff
@@ -206,11 +195,14 @@ private void WriteImpl(LogLevel level, Exception exception, string messageTempla
     {
         try
         {
-            Console.WriteLine($"[{_name}] {level}:{string.Format(messageTemplate, args)}");
+            var message =
```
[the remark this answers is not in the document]
lachmatt: fixed in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/456/commits/3bd2baa6343ce26aa9e37e50f7aed7c829d91951

src/OpenTelemetry.AutoInstrumentation/Logging/OtelLogging.cs
lachmatt: fixed in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/456/commits/3bd2baa6343ce26aa9e37e50f7aed7c829d91951

docs/config.md
lachmatt: fixed in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/456/commits/3bd2baa6343ce26aa9e37e50f7aed7c829d91951

Kielek (approved):

src/OpenTelemetry.AutoInstrumentation/Logging/FileSink.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace OpenTelemetry.AutoInstrumentation.Logging
+{
+    internal sealed class FileSink : ISink, IDisposable
+    {
+        readonly TextWriter _output;
+        readonly FileStream _underlyingStream;
+        readonly WriteCountingStream _countingStreamWrapper;
+        readonly object _syncRoot = new object();
+        static readonly long FileSizeLimitBytes = 10 * 1024 * 1024;
```
pellared, line 33: I suggest starting with unbounded file writting
especially that currently the user would not be able to find out the file reached its limit and it is also not configurable
we can enhance (e.g. add file rotation) in next PRs

lachmatt: I assumed having file growth bounded is preferred to unbounded file growth. I implemented it this way and didn't make if configurable because I wanted to keep it simple for now (and assumed we want to add file rotation in the near future).

test/OpenTelemetry.AutoInstrumentation.Tests/Diagnostics/SdkSelfDiagnosticsEventListenerTests.cs
```diff
@@ -16,57 +16,75 @@
...
+        var sinkField = defaultLogger.GetType().GetField("_sink", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
+        var defaultSink = (ISink)sinkField.GetValue(defaultLogger);
+
+        try
+        {
+            var testSink = new TestSink();
+            var listener = new SdkSelfDiagnosticsEventListener(EventLevel.Error);
+
+            // override default logger's sink with test sink
+            sinkField.SetValue(defaultLogger, testSink);
```
lachmatt, line 44: I considered other options, but wanted to stay consistent how logger is obtained (didn't want to inject it into ctor of `SdkSelfDiagnosticEventListener` when other classes obtain it directly from `OtelLogging`), didn't want to add methods to `Logger` or `SdkSelfDiagnosticEventListener` for the sake of these tests. Let me know if find this approach acceptable.

zacharycmontoya (approved):

src/OpenTelemetry.AutoInstrumentation/Logging/OtelLogging.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            var logDirectory = GetLogDirectory();
+            if (logDirectory != null)
+            {
+                var fileName = GetLogFileName();
+                var logPath = Path.Combine(logDirectory, fileName);
+                sink = new FileSink(logPath);
+            }
+        }
+        catch (Exception)
```
pellared, line 46: Do we know the possible exceptions? we do not want to catch exceptions like `OutOfMemoryException` ?
The same comment applies `catch` statements in this PR.
As for me, it can be addressed in another PR. But I guess it should be addressed.

pjanotti (approved):

src/OpenTelemetry.AutoInstrumentation/Logging/OtelLogging.cs
pellared: Created https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/470

src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
```diff
@@ -84,6 +84,21 @@ public class ConfigurationKeys
...
+    /// "/var/log/opentelemetry/dotnet/" on Linux.
+    /// </summary>
+    public const string LogDirectory = "OTEL_DOTNET_AUTO_LOG_DIRECTORY";
+
+    /// <summary>
+    /// Configuration key for setting the path for the profiler's native log file.
+    /// Environment variable OTEL_DOTNET_AUTO_LOG_DIRECTORY takes precedence over this setting, if set.
+    /// </summary>
+    public const string LogPath = "OTEL_DOTNET_AUTO_LOG_PATH";
+
```
nrcventura, line 101: Do we need to have both settings?

pjanotti: Yeah, better to have a single env var and use the same for native and managed.

Another thing that we may want to consider post-beta is: the native log is one file for multiple applications so it is not symmetrical to the managed one. Perhaps it will make the user's life simpler if the logs of a single run were easier to "group".

Issue #242, opened 2021-10-13
Add missing integration tests


pellared:
During the "PoC" effort we have been doing manual tests: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/README.md

We should automate these tests and update the README.

More: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/226

pellared:
We are missing integration tests for following instrumented libraries:
- ASP.NET Core
- System.Net.Http.HttpClient
- System.Net.WebRequest
- SqlClient

pellared:
This should be automated as well https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/385

pellared:
It would be good to cover also `SqlClient` instrumentation

PR #461, opened 2022-04-09
HttpClient and ASP.NET tests

dszmigielski:
Add test for HttpClient and ASP.NET integration.

Resolves part of the https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/242


test/integration-tests/IntegrationTests.Http/HttpTests.cs
```diff
@@ -0,0 +1,82 @@
...
+
+using System.Linq;
+using FluentAssertions.Execution;
+using IntegrationTests.Helpers;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace IntegrationTests.Http
+{
+    [Collection(HttpCollection.Name)]
```
pellared, line 25: Why do we need this?

test/integration-tests/IntegrationTests.Http/HttpTests.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            using var scope = new AssertionScope();
+            Assert.True(spans.Count == expectedSpanCount, $"Expecting {expectedSpanCount} spans, received {spans.Count}");
+
+            var httpClientSpan = spans.FirstOrDefault(span => span.Name.Equals("HTTP GET"));
+            var httpServerSpan = spans.FirstOrDefault(span => span.Name.Equals("/test"));
+            var manualSpan = spans.FirstOrDefault(span => span.Name.Equals("manual span"));
+
+            Assert.NotNull(httpClientSpan);
+            Assert.NotNull(httpServerSpan);
```
pellared, line 58: maybe it would be worth adding comments like `// ASP.NET Core auto-instrumentation is generating spans` (same comment applies to other places where it seems fit for your)

test/integration-tests/IntegrationTests.Http/HttpTests.cs
```diff
@@ -0,0 +1,82 @@
...
+            Assert.NotNull(manualSpan);
+
+            Assert.False(httpClientSpan.ParentId.HasValue);
+            Assert.Equal(httpClientSpan.SpanId, httpServerSpan.ParentId);
+            Assert.Equal(httpServerSpan.SpanId, manualSpan.ParentId);
```
pellared, line 63: maybe worth adding a comment like `checking trace hierarchy`?

test/test-applications/integrations/Samples.Http/Program.cs
```diff
@@ -0,0 +1,44 @@
...
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            var port = TcpPortProvider.GetOpenPort();
+            CreateHostBuilder(args, port).Build().RunAsync();
+
+            using var httpClient = new HttpClient();
+            httpClient.GetAsync($"http://localhost:{port}/test").Wait();
+        }
```
pellared, line 33: should it not be something more or less like

```suggestion
        public static async Task Main(string[] args)
        {
            var port = TcpPortProvider.GetOpenPort();
            var serverTask = CreateHostBuilder(args, port).Build().RunAsync();

            using var httpClient = new HttpClient();
            var clientTask = httpClient.GetAsync($"http://localhost:{port}/test");

            await Task.WhenAll(serverTask, clientTask);
        }
```

?

test/test-applications/integrations/Samples.Http/Directory.Build.props
```diff
@@ -0,0 +1,11 @@
+<Project>
+  <PropertyGroup>
+
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>netcoreapp3.1;net5.0;net6.0</TargetFrameworks>
```
pellared, line 5: Would it be enought to remove this fine and set `<TargetFrameworks>netcoreapp3.1;net5.0;net6.0</TargetFrameworks>` in `Samples.Http.csproj` ?

test/test-applications/integrations/Samples.Http/Program.cs
dszmigielski: serverTask won't complete in this case, do in the integration test it would hang on `RunSampleAndWaitForExit` method.

test/integration-tests/IntegrationTests.Http/HttpTests.cs
dszmigielski: I actually found that quite obvious, but I can add this

test/integration-tests/IntegrationTests.Http/HttpTests.cs
```diff
@@ -0,0 +1,83 @@
...
+        public void SubmitTraces()
+        {
+            var agentPort = TcpPortProvider.GetOpenPort();
+            using var agent = new MockZipkinCollector(Output, agentPort);
+
+            const int expectedSpanCount = 3;
+
+            using var processResult = RunSampleAndWaitForExit(agent.Port, enableStartupHook: true);
+            Assert.True(processResult.ExitCode >= 0, $"Process exited with code {processResult.ExitCode} and exception: {processResult.StandardError}");
+            var spans = agent.WaitForSpans(expectedSpanCount, 500);
```
pjanotti, line 47: I see the same 500 milliseconds in other tests but it seems rather optimistic to me, perhaps 3 seconds?

test/integration-tests/IntegrationTests.Http/HttpTests.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            using var scope = new AssertionScope();
+            Assert.True(spans.Count == expectedSpanCount, $"Expecting {expectedSpanCount} spans, received {spans.Count}");
```
pjanotti, line 50: Should we consider fluent style instead? Not, necessarily on this PR but we have been trying to use more fluent style.

test/test-applications/integrations/Samples.Http/TcpPortProvider.cs
```diff
@@ -0,0 +1,43 @@
+// <copyright file="TcpPortProvider.cs" company="OpenTelemetry Authors">
```
pjanotti, line 1: This file is duplicating test/integration-tests/IntegrationTests.Helpers/TcpPortProvider.cs. Is there any reason why not to add a reference to the project that contains it?

test/test-applications/integrations/Samples.Http/Program.cs
pellared: Is it possible that  `httpClient.GetAsync($"http://localhost:{port}/test")` might be called before the server is listening? I do not want to have another flaky test 😄 If not then it is worth adding a comment e.g. that the server is listening before `RunAsync()` returns.

test/integration-tests/IntegrationTests.Http/HttpTests.cs
pellared: I agree with @pjanotti. We should use `FluentAssertions` API for asserting.

test/test-applications/integrations/Samples.Http/Program.cs
pjanotti: Good point @pellared - I think it should use `Start()` in this case, see https://source.dot.net/#Microsoft.Extensions.Hosting.Abstractions/HostingAbstractionsHostBuilderExtensions.cs,16

The IHost should also be disposed - the program is terminating but it seems reasonable.

test/integration-tests/IntegrationTests.Http/HttpTests.cs
dszmigielski: I prefer fluent assertions as well, but I wanted to be consistent.

test/test-applications/integrations/Samples.Http/Program.cs
pellared: @dszmigielski you can consider adding a `/shutdown` endpoint that triggers application shutdown

test/test-applications/integrations/Samples.Http/TcpPortProvider.cs
dszmigielski: I don't want for a sample to reference any other projects here. I just find it odd - it should be independent

pellared (approved):

pellared (approved):

pellared:
A small note - it also tests the trace propagation.

test/integration-tests/IntegrationTests.Http/HttpTests.cs
```diff
@@ -0,0 +1,86 @@
...
+            using (new AssertionScope())
+            {
+                spans.Count.Should().Be(expectedSpanCount);
+
+                // ASP.NET Core auto-instrumentation is generating spans
+                var httpClientSpan = spans.FirstOrDefault(span => span.Name.Equals("HTTP GET"));
+                var httpServerSpan = spans.FirstOrDefault(span => span.Name.Equals("/test"));
+                var manualSpan = spans.FirstOrDefault(span => span.Name.Equals("manual span"));
+
+                httpClientSpan.Should().NotBeNull();
```
Kielek, line 59: consider grouping assertion by [`AssertionScope`s](https://fluentassertions.com/introduction#assertion-scopes).
If more than one assertion fails it produces message with all issues instead of only first one.

Kielek (approved):

test/integration-tests/IntegrationTests.Http/HttpTests.cs
dszmigielski: There is a single assertion scope used for all of them but one

Kielek: Sorry, I do not know how I can missed that. :)

pjanotti (approved):

PR #475, opened 2022-04-11
Fix plugin to add Redis Instrumentation

### LINKED ISSUE #423: Plugin to add Redis Instrumentation is not working
The sample server app loads a plugin that was intended to load the instrumentation for Redis. However, it doesn't report any error but I didn't see any data for the Redis client. 

Note that the plugin as part of the sample server app is not an ideal scenario, it would be better to move the "functionality" to the Vendor plugin under .the /samples folder.

rajkumar-rangaraj:
Fixes #423 

```
Activity.TraceId:          0e7539868d19cc51efe9f4d794a78c6c
Activity.SpanId:           b3e674a5f7ae1fd3
Activity.TraceFlags:           Recorded
Activity.ParentSpanId:    33d914062120c183
Activity.ActivitySourceName: OpenTelemetry.StackExchange.Redis
Activity.DisplayName: SET
Activity.Kind:        Client
Activity.StartTime:   2022-04-11T22:29:42.6849054Z
Activity.Duration:    00:00:00.0109342
Activity.Tags:
    db.system: redis
    db.redis.flags: DemandMaster
    db.statement: SET
    net.peer.name: localhost
    net.peer.port: 6379
    db.redis.database_index: 0
   StatusCode : UNSET
Activity.Events:
    Enqueued [4/11/2022 10:29:42 PM +00:00]
    Sent [4/11/2022 10:29:42 PM +00:00]
    ResponseReceived [4/11/2022 10:29:42 PM +00:00]
Resource associated with Activity:
    service.name: StartupHook.Self-hosted
```


pjanotti (approved):
LGTM

pjanotti:
The explanation why the caching of the connection works: the plugin is adding the instrumentation to `RedisController.Connection` so if it was instrumenting an instance that actually was not being used by the controller.

Kielek (approved):

pellared (approved):

PR #489, opened 2022-04-14
Adds configuration to control the processing of unhandled exceptions.

nrcventura:
Fixes # 321

Adds a configuration option to control whether or not unhandled exceptions should be processed.

Due to the [concerns about the potential impact on the running application](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/386#discussion_r820063893), the default value for this setting is "false" (or turned off).

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
@@ -93,9 +93,9 @@ public class ConfigurationKeys
     public const string LogDirectory = "OTEL_DOTNET_AUTO_LOG_DIRECTORY";
 
     /// <summary>
-    /// Configuration key for enabling the notification of unhandled exceptions.
+    /// Configuration key for enabling the flushing of telemetry data when an unhandled exception occurs.
     /// </summary>
-    public const string UnhandledExceptionsEnabled = "OTEL_DOTNET_AUTO_UNHANDLEDEXCEPTION_ENABLED";
+    public const string FlushOnUnhandledException = "OTEL_DOTNET_AUTO_FLUSH_ON_UNHANDLEDEXCEPTION";
 
     /// <summary>
     /// String format patterns used to match integration-specific configuration keys.
--- a/src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
@@ -93,6 +93,11 @@ public class ConfigurationKeys
     public const string LogDirectory = "OTEL_DOTNET_AUTO_LOG_DIRECTORY";
 
     /// <summary>
+    /// Configuration key for enabling the notification of unhandled exceptions.
+    /// </summary>
+    public const string UnhandledExceptionsEnabled = "OTEL_DOTNET_AUTO_UNHANDLEDEXCEPTION_ENABLED";
+
+    /// <summary>
     /// String format patterns used to match integration-specific configuration keys.
     /// </summary>
     internal static class Integrations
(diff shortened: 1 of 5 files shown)


src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
```diff
@@ -92,6 +92,11 @@ public class ConfigurationKeys
     /// </summary>
     public const string LogDirectory = "OTEL_DOTNET_AUTO_LOG_DIRECTORY";
 
+    /// <summary>
+    /// Configuration key for enabling the notification of unhandled exceptions.
+    /// </summary>
+    public const string UnhandledExceptionsEnabled = "OTEL_DOTNET_AUTO_UNHANDLEDEXCEPTION_ENABLED";
```
pjanotti, line 98: No constructive proposal at this moment but we should try to came up with a better name for the env var.

src/OpenTelemetry.AutoInstrumentation/Configuration/Settings.cs
```diff
@@ -167,6 +169,12 @@ private Settings(IConfigurationSource source)
     /// </summary>
     public bool Http2UnencryptedSupportEnabled { get; }
 
+    /// <summary>
+    /// Gets a value indicating whether the `AppDomain.UnhandledException` event should be enabled.
```
pjanotti, line 173: nit: this is not markdown :) 
```suggestion
    /// Gets a value indicating whether the <see cref="AppDomain.CurrentDomain.UnhandledException"/> event should be enabled.
```

pjanotti (commented):
It is looking good, 2 very minor comments.

src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
nrcventura: Agreed, I just wanted to get something out there start the conversation.
Some ideas:
- NOTICE_UNHANDLEDEXCEPTIONS
- PROCESS_UNHANDLEDEXCEPTIONS
- TROUBLESHOOT_UNHANDLEDEXCEPTIONS

Some things that I think we should avoid:
- Names that imply that we are capturing and reporting the unhandled exception because I don't think that we can guaranty that it will be added to any piece of telemetry other than our own logs.
- Names that are overloaded and used to represent other OpenTelemetry concepts.

lachmatt (approved):

Kielek (approved):

rajkumar-rangaraj (approved):

src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
pjanotti: Another suggestion: FLUSH_ON_UNHANDLEDEXCEPTION

pjanotti (approved):
LGTM. We just need to close on the env var name, since this is a beta we may change it later, but, as usual would prefer to avoid that.

src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
nrcventura: I thought about something specific like `FLUSH_ON_UNHANDLEDEXCEPTION`. I did not go with something like that initially because I wanted to leave open the possibility of doing other things like actually attempting to associate that exception to a span. However, it might make more sense to control that future possibility with a separate environment variable. Having a separate setting for that type of behavior may also be good to allow people to control how much risk they want to take on with these types of settings.

pjanotti (approved):

Kielek:
linking #321

PR #487, opened 2022-04-14
Dependabot for native code

Kielek:
Fixes #460
Fixes leftovers from #486  

Changes proposed in this pull request:
* Dependabot for native code
* update dependabot config after renaming projects (I have forgot to attach this file in previous PR)

Tested on my fork
image

PR #493, opened 2022-04-15
Execute dependabot daily

Kielek:
Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/487#discussion_r850146854

Execute dependabot daily


Kielek:
It fixes also https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/runs/6028862011

lachmatt (approved):

nrcventura:
Is daily too frequent of a cadence? In the rare case that one of these packages deploy a problematic version, and remove that version quickly, would this daily cadence cause us to unnecessarily depend on the problematic version?

pjanotti:
That's a good Q @nrcventura - the flip side is another unlikely case: we publish a version that misses some security fix by a few days. No a big deal if we don't have NuGet packages but it is the downside of reducing the frequency.

nrcventura:
> That's a good Q @nrcventura - the flip side is another unlikely case: we publish a version that misses some security fix by a few days. No a big deal if we don't have NuGet packages but it is the downside of reducing the frequency.

Would a daily "security" scan potentially cover that gap?

RassK:
> > That's a good Q @nrcventura - the flip side is another unlikely case: we publish a version that misses some security fix by a few days. No a big deal if we don't have NuGet packages but it is the downside of reducing the frequency.
> 
> Would a daily "security" scan potentially cover that gap?

I guess it depends on the release cycle. The latter could be way too slow to benefit if it's not automated patch and release.

Kielek:
I think that we should use dependabot just for notification that the newer version occurred. The decision if the upgrade is needed for current release is always on the dev side. What is more PRs usually take more than 1 day, so the scenario when the version will be shortly removed should not be the case.

I think that, we are discussing preferences, the current settings are (almost) fine, the reason for creating PRs was 2x 👍 in the previous PR :)

pjanotti (approved):

nrcventura (approved):

rajkumar-rangaraj (approved):

pellared (approved):

PR #557, opened 2022-04-19
Pre-beta example fixes

pjanotti:

 1. Remove current MongoDB auto-instrumentation (removed all of it to get a clean CI pass - the MongoDB will be re-added via #421). The reason is that currently it generates a bunch of spans that don't make sense from the user perspective and can keep being added to a trace after the original user transaction was finished.
 2. Add the CoreAppOldReference project to the solution (with proper rename). 
 3. Expose the OTLP gRPC port on dev/docker-compose.yaml to facilitate testing OTLP w/ gRPC protocol


nrcventura (approved):

zacharycmontoya (approved):
LGTM

PR #623, opened 2022-04-21
Update troubleshooting doc

RassK:
### What

Removes block of troubleshooting net core dependencies, as it's not valid.
* There is no way to change compile time direct dependency version.
* Additional deps corrects the version when there is no direct dependency.

@pellared suggested currently not to add details about build time fixing, to better estimate the scope of the issue.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -26,19 +26,6 @@ The example can only run successfully under the instrumentation, as the
 binding redirect makes the application dependent on a version of `System.Diagnostics.DiagnosticSource`
 that is not available at build time.
 
-### .NET Core dependency file
-
-To fix assembly version conflicts in .NET Core, edit the default
-`<application>.deps.json` file generated at build for .NET Core applications.
-Build a .NET Core app with package references using the required version,
-and use the respective `deps.json` file to see what changes are needed.
-
-To test your edits to the `deps.json` file, add a reference to the required
-version of the OpenTelemetry package to the [examples/CoreAppOldReference](./../examples/CoreAppOldReference/)
-example and rebuild the application. Save the generated `deps.json` file,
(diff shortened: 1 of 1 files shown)


rajkumar-rangaraj (approved):

nrcventura (approved):

pjanotti (approved):

Kielek (approved):

pellared:
Shouldn't we remove the line below as well?

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/71886b9befc4ca759529ec77440d35cf8759739a/docs/troubleshooting.md?plain=1#L19

pellared:
I read the section below a few times and I am not sure if this is understandable: 

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/0683f38542d85f671e6a174f2498665a7cf0f21b/docs/troubleshooting.md?plain=1#L22-L25

@nrcventura can you please double-check it and fix it as part of this PR if needed?

pellared (approved):
I left https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/623#issuecomment-1105775020 but it is not related to changes in this PR so it can be addressed separately if needed

nrcventura:
> I read the section below a few times and I am not sure if this is understandable:
> 
> https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/0683f38542d85f671e6a174f2498665a7cf0f21b/docs/troubleshooting.md?plain=1#L22-L25
> 
> @nrcventura can you please double-check it and fix it as part of this PR if needed?

I agree that it is confusing. I'm leaning towards merging this as is, and then following up. I'm still searching to see if there are better explanations that we can link to that explain how/why we are relying on the runtime package store.

PR #653, opened 2022-05-06
[WIP | PoC] Move OTel libs to separate load context

RassK:
### What

Moves all OpenTelemetry libraries and it's dependencies to a separate load context. 
Reduces conflicts between already loaded SDK libs and conflicting subdependencies.

Works best with:
* Small Nuget package that syncs shared state libraries (Eg DiagnosticSource, ILogger etc)
* OR with bridge that syncs two different versions (does not need Nuget any more then)

### Architectural view
dependencies_updated


RassK:
Closing as focus has shifted to other ways, will recheck back later.

Issue #662, opened 2022-05-11
Boostrap meter provider


pellared:
1. Boostrap the meter provider using the SDK.
2. Add OTLP metrics exporter - use `http/protobuf` as default protocol
3. Create an integration test
4. Add/update some example

PR #671, opened 2022-05-12
Add metrics environment variables

rajkumar-rangaraj:
Address part 3 in #662

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
@@ -15,6 +15,7 @@
 // </copyright>
 
 using System.Diagnostics;
+using System.Diagnostics.Metrics;
 
 namespace OpenTelemetry.AutoInstrumentation.Configuration;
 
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
```diff
@@ -15,6 +15,7 @@
 // </copyright>
 
 using System.Diagnostics;
+using System.Diagnostics.Metrics;
```
nrcventura, line 18: Is this using statement necessary? I'm not seeing any references to anything in it.

rajkumar-rangaraj: To support `cref` in comment we need `System.Diagnostics.Metrics` namespace. `System.Diagnostics` namespace in this file is also used for the same reason.

```
       /// <summary>
        /// Configuration key for additional <see cref="Meter"/> names to be added to the meter at the startup.
        /// </summary>
        public const string AdditionalSources = "OTEL_DOTNET_AUTO_METRICS_ADDITIONAL_SOURCES";
```

nrcventura (approved):

pellared (approved):

PR #675, opened 2022-05-13
Refactor Settings and add MeterSettings 

rajkumar-rangaraj:
Fixes part of #662 to include `MeterSettings `

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Configuration/MeterSettings.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Configuration/MeterSettings.cs
@@ -1,0 +1,159 @@
+// <copyright file="MeterSettings.cs" company="OpenTelemetry Authors">
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace OpenTelemetry.AutoInstrumentation.Configuration
+{
+    /// <summary>
+    /// Meter Settings
+    /// </summary>
+    public class MeterSettings : Settings
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MeterSettings"/> class
+        /// using the specified <see cref="IConfigurationSource"/> to initialize values.
+        /// </summary>
+        /// <param name="source">The <see cref="IConfigurationSource"/> to use when retrieving configuration values.</param>
+        private MeterSettings(IConfigurationSource source)
+            : base(source)
+        {
+            MetricExporter = ParseMetricExporter(source);
+            ConsoleExporterEnabled = source.GetBool(ConfigurationKeys.Metrics.ConsoleExporterEnabled) ?? false;
+
+            var instrumentations = new Dictionary<string, MeterInstrumentation>();
+            var enabledInstrumentations = source.GetString(ConfigurationKeys.Metrics.Instrumentations);
+            if (enabledInstrumentations != null)
+            {
+                foreach (var instrumentation in enabledInstrumentations.Split(separator: ','))
+                {
+                    if (Enum.TryParse(instrumentation, out MeterInstrumentation parsedType))
+                    {
+                        instrumentations[instrumentation] = parsedType;
+                    }
+                    else
+                    {
+                        throw new ArgumentException($"The \"{instrumentation}\" is not recognized as supported metrics instrumentation and cannot be enabled");
+                    }
+                }
+            }
+
+            var disabledInstrumentations = source.GetString(ConfigurationKeys.Metrics.DisabledInstrumentations);
+            if (disabledInstrumentations != null)
+            {
+                foreach (var instrumentation in disabledInstrumentations.Split(separator: ','))
+                {
+                    instrumentations.Remove(instrumentation);
+                }
+            }
+
+            EnabledInstrumentation = instrumentations.Values.ToList();
+
+            var providerPlugins = source.GetString(ConfigurationKeys.Metrics.ProviderPlugins);
+            if (providerPlugins != null)
+            {
+                foreach (var pluginAssemblyQualifiedName in providerPlugins.Split(':'))
+                {
+                    MetricPlugins.Add(pluginAssemblyQualifiedName);
+                }
+            }
+
+            var additionalSources = source.GetString(ConfigurationKeys.Metrics.AdditionalSources);
+            if (additionalSources != null)
+            {
+                foreach (var sourceName in additionalSources.Split(separator: ','))
+                {
+                    Meters.Add(sourceName);
+                }
+            }
+
+            MetricsEnabled = source.GetBool(ConfigurationKeys.Metrics.Enabled) ?? true;
+            LoadMetricsAtStartup = source.GetBool(ConfigurationKeys.Metrics.LoadMeterAtStartup) ?? true;
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether metrics is enabled.
+        /// Default is <c>true</c>.
+        /// </summary>
+        /// <seealso cref="ConfigurationKeys.Metrics.Enabled"/>
+        public bool MetricsEnabled { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether the metrics should be loaded by the profiler. Default is true.
+        /// </summary>
+        public bool LoadMetricsAtStartup { get; }
+
+        /// <summary>
+        /// Gets the metrics exporter.
+        /// </summary>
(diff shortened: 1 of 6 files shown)


src/OpenTelemetry.AutoInstrumentation/Configuration/MeterSettings.cs
```diff
@@ -0,0 +1,159 @@
...
+            : base(source)
+        {
+            MetricExporter = ParseMetricExporter(source);
+            ConsoleExporterEnabled = source.GetBool(ConfigurationKeys.Metrics.ConsoleExporterEnabled) ?? false;
+
+            var instrumentations = new Dictionary<string, MeterInstrumentation>();
+            var enabledInstrumentations = source.GetString(ConfigurationKeys.Metrics.Instrumentations);
+            if (enabledInstrumentations != null)
+            {
+                foreach (var instrumentation in enabledInstrumentations.Split(separator: ','))
```
pjanotti, line 43: This pattern doesn't come from this PR, but, now we have the typical separator, comma, and the one for plugins hard coded in many places. It seems reasonable to consolidate them. That said it will be better to do it in a separate PR.

src/OpenTelemetry.AutoInstrumentation/Configuration/MeterSettings.cs
```diff
@@ -0,0 +1,159 @@
...
+
+        /// <summary>
+        /// Gets the list of meters to be added to the MeterProvider at the startup.
+        /// </summary>
+        public IList<string> Meters { get; } = new List<string>();
+
+        internal static MeterSettings FromDefaultSources()
+        {
+            var configurationSource = new CompositeConfigurationSource
+        {
```
pjanotti, line 129: Indentation is broken here.

pjanotti (approved):
LGTM minus the indentation issue.

src/OpenTelemetry.AutoInstrumentation/Configuration/MeterSettings.cs
rajkumar-rangaraj: Created issue #678 to track this effort.

pellared (approved):

src/OpenTelemetry.AutoInstrumentation/Configuration/MeterSettings.cs
```diff
@@ -0,0 +1,159 @@
...
+                        instrumentations[instrumentation] = parsedType;
+                    }
+                    else
+                    {
+                        throw new ArgumentException($"The \"{instrumentation}\" is not recognized as supported metrics instrumentation and cannot be enabled");
+                    }
+                }
+            }
+
+            var disabledInstrumentations = source.GetString(ConfigurationKeys.Metrics.DisabledInstrumentations);
```
Kielek, line 56: I know that we have exactly same config structure for traces, but I have some doubts that we need it.
For now, it can be replaced by just removing the instrumentation from the enabled instrumentations list. It will change if we enable the metrics by default.

Kielek (approved):

src/OpenTelemetry.AutoInstrumentation/Configuration/MeterSettings.cs
pellared: I was thinking about the same as well.
But then I thought that it can be used for `Integrations` as well (in the same way as for traces). Nothing for adding byte-code metrics instrumentation, right?

src/OpenTelemetry.AutoInstrumentation/Configuration/MeterSettings.cs
```diff
@@ -0,0 +1,159 @@
...
+#endif
+            };
+
+            return new MeterSettings(configurationSource);
+        }
+
+        private static MetricsExporter ParseMetricExporter(IConfigurationSource source)
+        {
+            var metricsExporterEnvVar = source.GetString(ConfigurationKeys.Metrics.Exporter) ?? "otlp";
+            switch (metricsExporterEnvVar)
```
zacharycmontoya, line 144: Should we do a `string.ToLower()` here just to be safe?

src/OpenTelemetry.AutoInstrumentation/Configuration/Settings.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        /// <summary>
+        /// Gets the the OTLP transport protocol. Supported values: Grpc and HttpProtobuf.
+        /// </summary>
+        public OtlpExportProtocol? OtlpExportProtocol { get; }
+
+        /// <summary>
+        /// Gets a collection of <see cref="Integrations"/> keyed by integration name.
+        /// </summary>
+        public IntegrationSettingsCollection Integrations { get; }
```
zacharycmontoya, line 53: The backing collection of integrations is sourced from the `IntegrationRegistry`, which is populated by the values from the `TracerInstrumentation` enum. Did you want to combine the metrics integrations with the tracer integrations, or keep them separate?

src/OpenTelemetry.AutoInstrumentation/Configuration/MetricsExporter.cs
```diff
@@ -0,0 +1,38 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+namespace OpenTelemetry.AutoInstrumentation.Configuration;
+
+/// <summary>
+/// Enum representing supported metrics exporters.
+/// </summary>
+public enum MetricsExporter
```
zacharycmontoya, line 22: This enum starts with `Metrics` but other types start with `Meter`. Are there OTEL naming guidelines for this? I'm unfamiliar with OTEL metrics so maybe I'm just unfamiliar with the conventions

zacharycmontoya (approved):
LGTM, just left some questions on smaller details

src/OpenTelemetry.AutoInstrumentation/Configuration/MeterSettings.cs
rajkumar-rangaraj: Correct.

Below is the code from `EnvironmentConfigurationTracerHelper.cs`. Instrumentation list is parsed directly and added to `TracerProviderBuilder`. Incase of metrics addition is required to `MeterProviderBuilder`. Instrumentation libraries supported for traces and metrics as different and this needs to be handled separately. OpenTelemetry SDK treats traces, metrics and logs as 3 different signals and provides and option to enable/disable all 3 signals on a need basis. Design motivation is same here to treat all the signals different.

```
    public static TracerProviderBuilder UseEnvironmentVariables(this TracerProviderBuilder builder, TracerSettings settings)
    {
        var resourceBuilder = ResourceBuilder.CreateDefault();

        builder
            .SetResourceBuilder(resourceBuilder)
            .SetExporter(settings);

        foreach (var enabledInstrumentation in settings.EnabledInstrumentations)
        {
            if (AddInstrumentation.TryGetValue(enabledInstrumentation, out var addInstrumentation))
            {
                addInstrumentation(builder);
            }
        }

        builder.AddSource(settings.ActivitySources.ToArray());
        foreach (var legacySource in settings.LegacySources)
        {
            builder.AddLegacySource(legacySource);
        }

        return builder;
    }

```

src/OpenTelemetry.AutoInstrumentation/Configuration/MeterSettings.cs
rajkumar-rangaraj: Very good recommendation, we have this pattern uses in TracerSettings and other places. I want to get this covered in a different PR. What are your thoughts.

src/OpenTelemetry.AutoInstrumentation/Configuration/MetricsExporter.cs
rajkumar-rangaraj: All metrics API based terminologies uses prefix `Meter` - https://opentelemetry.io/docs/reference/specification/metrics/api/.

Metrics Exporters starts with `Metrics` as per specification - https://opentelemetry.io/docs/reference/specification/metrics/sdk_exporters/otlp/

src/OpenTelemetry.AutoInstrumentation/Configuration/Settings.cs
rajkumar-rangaraj: At this point, no integrations are needed for metrics. I want to keep them separate when use case arises for it.

rajkumar-rangaraj: Moved `IntegrationSettingsCollection` to `TracerSettings`.

src/OpenTelemetry.AutoInstrumentation/Configuration/MeterSettings.cs
zacharycmontoya: Yes another PR is fine with me 😄

src/OpenTelemetry.AutoInstrumentation/Configuration/MetricsExporter.cs
zacharycmontoya: Thanks for the links, that helps a lot 👍🏼

src/OpenTelemetry.AutoInstrumentation/Configuration/MeterSettings.cs
zacharycmontoya: Are there any requested changes here? If not, I think we can merge the PR in its current state

rajkumar-rangaraj: Similar question has come during Environment variables too. It's better to discuss the topic in next SIG meeting. I will get the PR merged to unblock the metrics integration.

PR #690, opened 2022-05-20
Refine pull request template

pellared:
## Why

1. I propose to add a minimal structure which I believe will make the PR descriptions a little more readable.
2. I propose to add the "Testing" section as sometimes testing is not straightforward and it may be good to leave some notes for the reviewers.
3. I propose a checklist that I would us to follow. I prefer to have docs changes and tests as part of PR that brings some "visible" changes - otherwise, it is easy to create gaps. It makes also reviewing easier as docs can be seen also as descriptions of the changes and tests can be used to understand the differences in the behavior. 

## What

Add sections and a checklist to the PR template.

## Tests

N/A

## Checklist

- [x] The code changes are covered by tests.


.github/pull_request_template.md
```diff
@@ -1,4 +1,22 @@
...
+<!-- Describe changes proposed in this pull request. -->
+
+## Tests
+
+<!-- Describe how the change was tested (both manual and automated) for reviewers to find edge cases more easily. -->
+
+## Checklist
+
+<!-- All items should be verified and marked as done.
+     If an item doesn't apply to the introduced changes - check it as done too. -->
```
Kielek, line 18: It is probably about preferences, but if I see checkbox checked I am considering it is done.
I'd prefer to see it strikethrough if not applicable.

pellared: Addressed https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/690/commits/ddb3e9b9b74eb33954d24601e620df4083a86c5c

.github/pull_request_template.md
```diff
@@ -1,4 +1,22 @@
...
+<!-- Describe how the change was tested (both manual and automated) for reviewers to find edge cases more easily. -->
+
+## Checklist
+
+<!-- All items should be verified and marked as done.
+     ~~strikethrough~~ if an item doesn't apply to the introduced changes. -->
 
+- [ ] `CHANGELOG.md` is updated.
+- [ ] Documentation is updated.
+- [ ] New features are covered by tests.
```
nrcventura, line 22: If this step is skipped when a feature is added, I think we should call out why it is skipped. For example, with the metrics work, in order to make the PRs easier to review the work could be split up in such a way to add the test framework/harness code in a separate PR.

pjanotti: Yeah, the way I see it listing those explicitly invite consideration about the items, but, we can always accept a justification to not cover those.

pjanotti (approved):

Kielek (approved):

.github/pull_request_template.md
pellared: 1. I think we should avoid such cases. I think that adding "new signals support" like metrics and logs is unique and we do not have to call it out in this template.
2. If we skip this step then we should call out (in the PR description) why it is skipped and create a GitHub issue for tracking.

nrcventura: @pellared In general I agree with you. While we are still in these early stages and we are still improving our testing strategy we might have more of these larger testing gaps. Do we want to have the exception process you are recommending defined in the template, or do we think it's sufficient for approvers and maintainers to enforce it without including more info the template? I'm ok with either approach.

pjanotti: I agree with @pellared bullets 1 and 2. I would rather not add anything about possible exceptions on the template itself. We want processes and checklists to invite/guide good practices, but, we don't want to dogmatically and blindingly be constrained by them - one needs to be careful to not make processes/checklists useless by having too many exceptions.  Achieving that requires a bit of judgement/trade-offs but I think our SIG is up to the task :)

rajkumar-rangaraj (approved):

nrcventura (approved):

pjanotti:
@pellared:
> I would like to have it approved by all maintainers before merging 😉

@zacharycmontoya I think you are the last one :)

pjanotti (approved):

zacharycmontoya (approved):

PR #669, opened 2022-05-12
SqlClient tests

dszmigielski:
Add test for SqlClient integration.

This should resolve https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/242

PR #694, opened 2022-05-24
Remove hardcoded pwd from SqlClient test

pjanotti:
Removing hardcoded password from integration test. The password is now passed as a parameter to the test application, it is not using a secret store (the ideal) but a guid seems good enough to avoid the hard coded password.

### DIFF
--- a/test/integration-tests/IntegrationTests.SqlClient/SqlClientCollection.cs
+++ b/test/integration-tests/IntegrationTests.SqlClient/SqlClientCollection.cs
@@ -14,6 +14,7 @@
 // limitations under the License.
 // </copyright>
 
+using System;
 using System.Threading.Tasks;
 using DotNet.Testcontainers.Containers.Builders;
 using DotNet.Testcontainers.Containers.Modules;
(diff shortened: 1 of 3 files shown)


pellared (commented):
To be honest, I do not see any issue in having hardcoded creds in test applications.
This will only make a little trouble when one would like to interactively log in to the SQL Server (e.g. when debugging).

test/integration-tests/IntegrationTests.SqlClient/SqlClientCollection.cs
```diff
@@ -33,7 +34,6 @@ public class SqlClientFixture : IAsyncLifetime
     {
         private const int DatabasePort = 1433;
         private const string DatabaseImage = "mcr.microsoft.com/mssql/server:2019-CU15-ubuntu-20.04";
-        private const string DatabasePassword = "@someThingComplicated1234";
```
nrcventura, line 36: Isn't this password still hardcoded in one of the yml files? Perhaps one of the other changes in this PR is overriding that setting, but it's not clear to me?

pjanotti: My VS Code doesn't find `@someThingComplicated1234` in any other place.

nrcventura: It looks like I was remembering a change in an [older commit](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/669/commits/80226cdef371e8fc7a00c9bab9f8235baaf0cd38) that was later removed.

pellared (approved):

rajkumar-rangaraj (approved):

PR #697, opened 2022-05-25
Ignore GraphQL updates for GraphQL test app

nrcventura:
Relates to #679, #680, #681 

Adds an explicit ignore in the dependabot configuration with a comment to both ignore future updates to the GraphQL libraries and explain why they are ignored. I think that ignoring these specific updates in this manner will be more discoverable and maintainable in the future, and it is the [guidance provided by github](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/managing-pull-requests-for-dependency-updates#managing-dependabot-pull-requests-with-comment-commands).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -168,6 +168,9 @@ updates:
     schedule:
       interval: "daily"
     open-pull-requests-limit: 20
+    ignore:
+      # The instrumentation only supports specific versions of GraphQL and the test app needs to test against a specific version
+      - dependency-name: "GraphQL*"
 
   - package-ecosystem: nuget
     directory: /test/test-applications/integrations/dependency-libs/TestApplication.ExampleLibraryTracer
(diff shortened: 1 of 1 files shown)


pjanotti (approved):

.github/dependabot.yml
```diff
@@ -168,6 +168,9 @@ updates:
     schedule:
       interval: "daily"
     open-pull-requests-limit: 20
+    ignore:
+      # The instrumentation only supports specific versions of GraphQL and the test app needs to test against a specific version
+      - dependency-name: "GraphQL*"
```
pellared, line 173: Should it not be for `/test/test-applications/integrations/TestApplication.GraphQL` instead of `/test/test-applications/integrations/dependency-libs/TestApplication.ExampleLibrary`?

pellared: You can change default branch to `IgnoreGraphQLUpdates` on your fork and see what dependabot will do.

pellared (approved):

PR #716, opened 2022-06-03
Add http metrics tests

nrcventura:
Changes:
* Adds OTLP protos to the test helper project
* Adds a mock collector that uses the OTLP metrics protos so that it can receive metrics via OTLP in our tests
* Updates test framework to support independently enabling tracing and/or metrics
* Adds tests for AspNetCore http.server.duration metrics
* Adds tests for Http (client) http.client.duration metrics


test/integration-tests/IntegrationTests.Helpers/MockCollector.cs
```diff
@@ -0,0 +1,258 @@
...
+        while (true)
+        {
+            // seems like we can't reuse a listener if it fails to start,
+            // so create a new listener each time we retry
+            var listener = new HttpListener();
+
+            try
+            {
+                listener.Start();
+                listener.Prefixes.Add($"http://*:{port}/"); // Warning: Requires admin access
```
Kielek, line 54: Is it possible to be more specific to avoid elevated privileges?
Usually, my VS/terminal is running on user permissions level.

Kielek (approved):

rajkumar-rangaraj (approved):
LGTM.

There is an update to [opentelemetry-proto](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto) with C# namespace. I recommend taking the newer version of proto files from the repo.

test/integration-tests/IntegrationTests.Helpers/MockCollector.cs
nrcventura: I think that I got that URL from the mockzipkincollector. I can change it for the mockcollector now, and if I get that working I'll get an issue submitted to do the same for the mockzipkincollector.

nrcventura:
> There is an update to [opentelemetry-proto](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto) with C# namespace. I recommend taking the newer version of proto files from the repo.

I took a closer look at when that namespace was added, and it was added after the 0.18.0 release (which is the latest release). I think that we should wait to update these protos until there is another release. I can submit an issue for it as a reminder.

test/integration-tests/IntegrationTests.Helpers/MockCollector.cs
nrcventura: It looks like I was able to get it to work with normal user permissions by switching from the wildcard to the ipaddress. I'm waiting to see if this works for the macos and linux CI pipelines too.

PR #823, opened 2022-06-07
Remove hacks to set correct version for Microsoft.Extensions.Logging.Abstractions

### LINKED ISSUE #719: Fix 'hacks' for dependencies
Can you double-check if this hack is still needed? See: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/399

Most probably the same comment applies to other places (I remember some `example` app to also reference `Microsoft.Extensions.Logging.Abstractions`).

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/717#discussion_r890084054_

Kielek:
Fixes #719
Closes #558

remove hacks to set correct version for `Microsoft.Extensions.Logging.Abstractions`
It is handled by additional dependency.

References to `System.Diagnostics.DiagnosticSource` cannot be removed. It is compile-time dependency.

### DIFF
--- a/test/test-applications/integrations/TestApplication.StartupHook/TestApplication.StartupHook.csproj
+++ b/test/test-applications/integrations/TestApplication.StartupHook/TestApplication.StartupHook.csproj
@@ -1,6 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.0" />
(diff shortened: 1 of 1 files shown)


pellared (approved):
I think we should also remove the reference in `TestApplication.StartupHook.csproj`

Kielek:
> I think we should also remove the reference in `TestApplication.StartupHook.csproj`

Yes, it is possible. Tests passsed on my machine. Thanks for the fix.

PR #824, opened 2022-06-07
Remove redundant references to System.Diagnostics.DiagnosticSource

Kielek:
Fixes N/A

remove redundant references to System.Diagnostics.DiagnosticSource


pellared:
1. What about `Examples.ConsoleApp.csproj`?
2. Also, what do you think of removing manually created spans in `TestApplication.MongoDB`, `TestApplication.Http`, `TestApplication.SqlClient`? It is enough to test the manual span in `TestApplication.StartupHook`. This can be a separate PR.

Kielek:
All of them tested. There is no possibility to removed it due to compilation/runtime issues.

pellared:
ad. 1
I think we can remove

```csharp
        Console.WriteLine("Ensure right version of DiagnosticSource is installed");
        Console.WriteLine(typeof(ActivitySource).Assembly.FullName);
```

and as a separate PR which can be together with ad.2

PR #851, opened 2022-06-21
ASP.NET Metrics Integration tests using Prometheus exporter

### LINKED ISSUE #838: Integration tests for ASP.NET and HttpClient metrics instrumentation
The code changes looks good, but there is no tests for that. Do you have a plan to create integration tests?

_Originally posted by @Kielek in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/836#pullrequestreview-1002279521_

rajkumar-rangaraj:
## Why

Fixes #838

## Tests

Updated ASP.NET integration tests to test ASPNET, Runtime and custom metrics.
Creating a PR at earlier stage to take feedback on the approach,

## Checklist


test/integration-tests/aspnet/IntegrationTests.AspNet/AspNetTests.cs
```diff
@@ -43,4 +44,36 @@ public async Task SubmitsTraces()
...
+            content = await response.Content.ReadAsStringAsync();
+
+            Output.WriteLine("Raw metrics from Prometheus:");
+            Output.WriteLine(content);
+
+            //PrometheusMetrics
+
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Contains("MyFruitCounter{color=\"red\",name=\"apple\"} 6", content);
+            Assert.Contains("http_server_duration_ms_count{http_method=\"GET\",http_scheme=\"http\",http_status_code=\"200\"}", content);
```
rajkumar-rangaraj, line 76: Want to take feedback on this approach of asserting. If this looks fine, will refactor the PR and mark it for review.

In dotnet OpenTelemetry, asserting is done using similar approach. Ref: https://github.com/open-telemetry/opentelemetry-dotnet/blob/cc5083e37bbe87cbb94aa9a34cbb7bf3a149f7a3/test/OpenTelemetry.Exporter.Prometheus.Tests/PrometheusExporterHttpServerTests.cs#L167

test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
```diff
@@ -103,6 +103,50 @@ public async Task<Container> StartContainerAsync(int traceAgentPort, int webPort
         return new Container(container);
     }
 
+    public async Task<Container> StartContainerForMetricsAsync(int webPort, int prometheusPort)
```
Kielek, line 106: Consider to extract one, common method to start containers. It could us settings parameters, similar to `public ProcessResult RunTestApplicationAndWaitForExit(TestSettings testSettings)`

We avoid duplication code with this approach.

test/integration-tests/aspnet/IntegrationTests.AspNet/AspNetTests.cs
```diff
@@ -43,4 +44,36 @@ public async Task SubmitsTraces()
             Assert.True(spans.Count >= 1, $"Expecting at least 1 span, only received {spans.Count}");
         }
     }
+
+    [Fact]
+    [Trait("Category", "EndToEnd")]
+    [Trait("Containers", "Windows")]
+    public async Task SubmitsMetrics()
```
Kielek, line 51: PR #716 added support to OpenTelemetry Protocol mock collector. I think that it should be our default, proffered method to test all metrics, traces and logs.
It will be great to switch this test to OTLP

The test in current state could be moved to a newly created `IntegrationTests.Exporter.Prometheus` project.

With this we will have, both standardized approach to test instrumentations and separately tested newly added exporter.

test/integration-tests/aspnet/IntegrationTests.AspNet/AspNetTests.cs
pellared: I prefer the way done here: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/716, because:

1. I think we should use OTLP for integration testing of the instrumentations
2. We do not need to make such detailed assertions - we just need to check if the instrumentation from the SDK is working 
3. Uses FluentAssertions

test/integration-tests/aspnet/IntegrationTests.AspNet/AspNetTests.cs
pellared: I agree that it would be better just to have a little sample app for testing different exporters if they work at all. It should not test any instrumentations - just some manual metrics.

Issue #850, opened 2022-06-17
Update mockzipkincollector to not require admin priveledges


nrcventura:
The mockzipkincollector currently uses a wildcard prefix for its httplistener. If we switch it to use the localhost ip address instead (like we did for the mockcollector), the listener should work without requiring elevated permissions.

pjanotti:
Failures on the test PR indicated that we need to expose ports from docker containers as expected, see https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/runs/7010602844?check_suite_focus=true#step:5:2155

PR #872, opened 2022-06-22
[Testing]: MockZipkin: Bind to localhost instead of *

pjanotti:
Per SIG meeting regarding issue #850 checking if using localhost breaks CI - depending on the results will update the description accordingly.


pjanotti:
Failures indicate that we need to expose ports from docker containers, see https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/runs/7010602844?check_suite_focus=true#step:5:2155

I will close this PR for now. Let's track it via #850.

PR #874, opened 2022-06-22
Do not require elevation to run MockZipkinCollector

### LINKED ISSUE #850: Update mockzipkincollector to not require admin priveledges

pjanotti:
## Why

Do not require admin rights by default to run MockZipkinCollector. 

Fixes #850 

## What

Use localhost to avoid hitting firewall and requiring elevation. Admin rights are still required on Windows to run the AspNet integration tests due to issues exposing a localhost port to a Windows container. The test will fail if not running as admin, see justification for that option [here](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/874#issuecomment-1164923083).

Opportunistically, changed the defaults to better match with actual Zipkin server settings.

## Tests

Ran `Workflow` target locally on non-elevated VS prompt.

## Checklist


pjanotti:
Before digging to see if the CI failure below is related to the actual change I'm re-running the failed job.

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/runs/7013294177?check_suite_focus=true#step:4:2494

pjanotti:
Test failure is consistent in CI, likely a test that didn't run locally in my run of `Workflow`. I will look at it later today. 

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/runs/7013489878?check_suite_focus=true#step:4:2494

test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
```diff
@@ -59,7 +59,7 @@ public async Task<Container> StartContainerAsync(int traceAgentPort, int webPort
 
         string agentBaseUrl = $"http://{DockerNetworkHelper.IntegrationTestsGateway}:{traceAgentPort}";
         string agentHealthzUrl = $"{agentBaseUrl}/healthz";
-        string zipkinEndpoint = $"{agentBaseUrl}/api/v2/spans";
+        string zipkinEndpoint = $"http://host.docker.internal:{traceAgentPort}/api/v2/spans";
```
nrcventura, line 62: Maybe the `host.docker.internal` prefix is causing the http listener to not accept the request?

pellared:
I think we can try using the `host` network in the container running IIS instead of `nat` network.

test/integration-tests/IntegrationTests.Helpers/TestHelper.cs
pjanotti: `host.docker.internal` seems busted for Windows containers, see https://github.com/docker/for-win/issues/1976 - I tried some of the scripts described on the issue but no luck so far.

pjanotti:
@nrcventura @pjanotti - later I will do a quick try with the `nost` network. If that fails I am leaning in the direction of adding a helper like `IsWindowsAdministrator` and skipping the test conditionally. The `MockZipkinCollector` will use the same helper to conditionally bind to `*` or `localhost`. This way CI keeps the same coverage, we don't require admin on dev boxes except for ASP.NET test.

nrcventura:
> @nrcventura @pjanotti - later I will do a quick try with the `nost` network. If that fails I am leaning in the direction of adding a helper like `IsWindowsAdministrator` and skipping the test conditionally. The `MockZipkinCollector` will use the same helper to conditionally bind to `*` or `localhost`. This way CI keeps the same coverage, we don't require admin on dev boxes except for ASP.NET test.

It might be helpful to log something out to the console when we're skipping the test. Then at least there's some sort of breadcrumb that could be found to indicate that a test was not run.

pjanotti:
@nrcventura your concern about the visibility is well founded: a local run doesn't make very visible that the test was skipped.

CI also seems stuck... I'm going to merge with other changes and see if that repros.

pjanotti:
I think that instead of skipping the test I'm going to fail it if non-admin run, reasoning:

* If it is CI we want it to fail if it doesn't run since we expect it to run under admin.
* If it is dev box when one sets Windows containers that is *the integration test* that is being targeted.

pjanotti:
@pellared `--network host` is only supported on Linux, see https://docs.docker.com/network/network-tutorial-host/#prerequisites

test/IntegrationTests/Helpers/MockZipkinCollector.cs
```diff
@@ -52,7 +52,10 @@ public MockZipkinCollector(ITestOutputHelper output, int port = 9080, int retrie
             try
             {
                 listener.Start();
-                listener.Prefixes.Add($"http://*:{port}/"); // Warning: Requires admin access
+
+                // For the mock collector to receive data from Windows containers we need to bind to all network interfaces.
+                var host = EnvironmentTools.IsWindowsAdministrator() ? "*" : "localhost";
```
pellared, line 57: I would prefer adding an optional parameter `host` with `localhost` as default to the `MockZipkinCollector` instead of adding such conditional behavior in the test code. It would make the test code more readable, trustworthy, and predictable (add more buzzwords here 😄 ).

Take notice that using `localhost` for most of the tests is fine and less invasive.

pellared (approved):

RassK (approved):

PR #888, opened 2022-06-24
Fix integration tests on Windows when using Linux containers

### LINKED ISSUE #887: [Tests]: Workflow target fails on Windows w/ Linux containers
The issue was introduced via ​#878. Some of the integration test projects didn't include TargetFramework like `net462` and the unified project does. The issue didn't show up in CI since there is no pass against Linux containers on Windows.

```shell
╬══════════════════════
║ Warnings & Errors
╬═════════════
​
[WRN] : Solution C:\src\otel\dotnet\auto\OpenTelemetry.AutoInstrumentation.sln has active build configurations for the build project.
Either enable SuppressBuildProjectCheck on Build.Solution or remove the following entries from the solution file:
  - {4DDEEB10-A6B5-4A46-96C1-1B7276B61E08}.Release|x86.Build.0 = Release|Any CPU
[ERR] RunManagedIntegratio: [xUnit.net 00:00:19.63]     IntegrationTests.MongoDB.MongoDbTests.SubmitsTraces [FAIL]
[ERR] RunManagedIntegratio: [xUnit.net 00:00:23.01]     IntegrationTests.SqlClient.SqlTests.SubmitTraces [FAIL]
[ERR] RunManagedIntegratio: Target RunManagedIntegrationTests has thrown an exception
​```

### PR DESCRIPTION
**Pull request by pjanotti**
## Why

The `Workflow` target should not break on Windows with the default container type Linux, a common dev box case.

Fixes #887

## What

* Don't build tests that aren't supported on .NET Framework

Unfortunately we can't have a Windows runner with docker containers in CI so we have a gap in coverage for .NET Framework apps in which integration tests require a docker container.

## Tests

Tested locally the `Workflow` target.

## Checklist

### CODE REVIEW

File: test/IntegrationTests/MongoDbTests.cs [hunk 1]
```diff
@@ -14,10 +14,8 @@
 // limitations under the License.
 // </copyright>
 
+#if !NETFRAMEWORK
```
pellared, line 17: Can you please document in `config.md` that MongoDB instrumentation is not supported on .NET Framework?

test/IntegrationTests/SqlTests.cs
```diff
@@ -14,6 +14,7 @@
 // limitations under the License.
 // </copyright>
 
+#if !NETFRAMEWORK
```
pellared, line 17: Instead of doing this here and in `SqlClientCollection.cs` we should remove the line:

```csharp
span.Tags["db.statement_type"].Should().Be("Text");
```

The test is failing on .NET Framework, because

> On .NET Framework, "db.statement_type" attribute is no longer set for
  activities created by the instrumentation.

Source: https://github.com/open-telemetry/opentelemetry-dotnet/blob/55260e6858e5ecf691c60161ea25d29bbc24dc1d/src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md?plain=1#L78-L79

Also if possible I suggest refactoring the assertions 

```csharp
span.Tags["db.system"].Should().Be("mssql");
```

into this patttern

```csharp
span.Tags.Should().Contain(new KeyValuePair<string, string>("db.system", "mssql"));
```

It would make the assertion messages a lot clearer.


Reference: https://fluentassertions.com/dictionaries/

test/IntegrationTests/MongoDbCollection.cs
```diff
@@ -14,6 +14,7 @@
 // limitations under the License.
 // </copyright>
 
+#if !NETFRAMEWORK
```
pellared, line 17: I am not sure if it is needed here. AFAIK if it is unused then it is a NOOP.

pellared: Agree

test/IntegrationTests/MongoDbTests.cs
pjanotti: @RassK could you please check if this is correct 17b7df26

test/IntegrationTests/SqlTests.cs
pjanotti: 4b38002

docs/config.md
```diff
@@ -41,7 +41,7 @@ for more details.
 | `AspNet` | ASP.NET Core | * | source |
 | `GraphQL` | [GraphQL](https://www.nuget.org/packages/GraphQL/) | ≥2.3.0 & < 3.0.0 | bytecode |
 | `HttpClient` | [System.Net.Http.HttpClient](https://docs.microsoft.com/dotnet/api/system.net.http.httpclient) and [System.Net.HttpWebRequest](https://docs.microsoft.com/dotnet/api/system.net.httpwebrequest) | * | source |
-| `MongoDb` | [MongoDB.Driver.Core](https://www.nuget.org/packages/MongoDB.Driver.Core) | ≥2.3.0 & < 3.0.0 | source & bytecode |
+| `MongoDB` | [MongoDB.Driver.Core](https://www.nuget.org/packages/MongoDB.Driver.Core) **Not supported on .NET Framework**  | ≥2.3.0 & < 3.0.0 | source & bytecode |
```
pellared, line 44:
```suggestion
| `MongoDb` | [MongoDB.Driver.Core](https://www.nuget.org/packages/MongoDB.Driver.Core) **Not supported on .NET Framework** | ≥2.3.0 & < 3.0.0 | source & bytecode |
```

here is why: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/3d6d2bbfc20a82c9331aa8a16457904cb85b7506/src/OpenTelemetry.AutoInstrumentation/Configuration/TracerInstrumentation.cs#L47

pellared: I agree. Just do not forget to update the examples as well if you decide to do it here 😉

pjanotti: Let's do it as a follow up - I can spare some minutes to do it today.

pellared (approved):

rajkumar-rangaraj (approved):

PR #890, opened 2022-06-27
[DoNotMergeYet] Fix env var seperator in native code

pellared:
## Why

The separator was different than documented.

## What

`OTEL_DOTNET_AUTO_INCLUDE_PROCESSES`, `OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES`,
  `OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS` use `,` as seperator
  instead of `;`.

## Tests

I have no idea how this test is passing: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/ea1f47855d8fc019aace6fea9e9a70a8cd304342/test/IntegrationTests/StartupHookTests.cs#L67-L86

## Checklist

- [x] `CHANGELOG.md` is updated.
- ~~[ ] Documentation is updated.~~
- ~~[ ] New features are covered by tests.~~

PR #892, opened 2022-06-27
Fix and refine env var separators

pellared:
## Why

Refinements before adding missing integration tests for "manual metrics + plugins".

I have found inconsistencies in env var separators for `OTEL_DOTNET_AUTO_INCLUDE_PROCESSES`, `OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES`, `OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS`
- between .NET Fx and .NET Core (`OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES` already used `,` for .NET Core)
- between "actual" vs "documented" (documented `,` - actual `;` in most scenarios)

I decided to improve and document the separator logic for `OTEL_DOTNET_AUTO_INTEGRATIONS_FILE`.

In the next PR I plan to rename `StartupHookTests` and `TestApplication.StartupHook` to `SmokeTests` and `TestApplication.Smoke` (I have no better naming idea 😋 ). The idea behind is that we can use it to test "non-library-instrumentation" features.

I also plan to refactor to use `FluentAssertions` as I hope it will increase the quality of the tests.

## What

- Use `,` as seperator, as documented, instead of `;`, for:
  - `OTEL_DOTNET_AUTO_INCLUDE_PROCESSES`,
  - `OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES`,
  - `OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS`.
- `OTEL_DOTNET_AUTO_INTEGRATIONS_FILE` can accept multiple filepaths
  delimted by the platform-specific path separator
  (`;` on Windows, `:` on Linux and macOS).
- Run `StartupHookTests` for .NET Framework
- Refactor `StartupHookTests` (preparatory refactoring)

## Tests

CI

## Checklist


test/IntegrationTests/StartupHookTests.cs
```diff
@@ -38,78 +37,80 @@ public StartupHookTests(ITestOutputHelper output)
...
+    [Fact]
+    [Trait("Category", "EndToEnd")]
+    public void ApplicationIsExcluded()
     {
-        int agentPort = TcpPortProvider.GetOpenPort();
+#if NETFRAMEWORK
+        SetEnvironmentVariable("OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES", $"dotnet;dotnet.exe;{EnvironmentHelper.FullTestApplicationName};{EnvironmentHelper.FullTestApplicationName}.exe");
+#else
+        SetEnvironmentVariable("OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES", $"dotnet,dotnet.exe,{EnvironmentHelper.FullTestApplicationName},{EnvironmentHelper.FullTestApplicationName}.exe");
+#endif
```
pellared, line 83: I do not understand it...
Reference: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/890

pjanotti: Yeah, this conditional seems strange. Perhaps the intention was `;` for Windows and `,` otherwise. Anyway, we should look and clarify.

IIRC in the past there was a filter for `MockZipkinCollector` this allowed to wait for the correct spans. If the filter was removed and we are simply counting spans (not checking name or attributes) this can pass tests that should fail.

test/IntegrationTests/StartupHookTests.cs
```diff
@@ -131,5 +132,14 @@ private void AssertExpectationsMet(List<WebServerSpanExpectation> expectations,
...
     }
+
+    private IImmutableList<IMockSpan> RunTestApplication(bool enableStartupHook = true)
+    {
+        int agentPort = TcpPortProvider.GetOpenPort();
+        using var agent = new MockZipkinCollector(Output, agentPort);
+        using var processResult = RunTestApplicationAndWaitForExit(agent.Port, enableStartupHook: enableStartupHook);
+
+        Assert.True(processResult.ExitCode >= 0, $"Process exited with code {processResult.ExitCode} and exception: {processResult.StandardError}");
+        return agent.WaitForSpans(2, TimeSpan.FromSeconds(5));
```
pjanotti, line 143: It would be nice to have a method on the mock collector to get all spans already received and bail. This would avoid waiting for spans after the application already finished.

test/IntegrationTests/StartupHookTests.cs
```diff
@@ -131,5 +132,14 @@ private void AssertExpectationsMet(List<WebServerSpanExpectation> expectations,
         string finalMessage = Environment.NewLine + string.Join(Environment.NewLine, failures.Select(f => " - " + f));
         Assert.True(!failures.Any(), finalMessage);
     }
+
+    private IImmutableList<IMockSpan> RunTestApplication(bool enableStartupHook = true)
+    {
+        int agentPort = TcpPortProvider.GetOpenPort();
+        using var agent = new MockZipkinCollector(Output, agentPort);
```
pjanotti, line 139: Not related to the PR but it seems that the call to `TcpPortProvider.GetOpenPort()` should actually be done inside the mock collector and the port parameter be a nullable. Most tests would just do like this one and use `agent.Port` as needed.

test/IntegrationTests/StartupHookTests.cs
pellared: This is a different code path as far as I see

pellared: Here is where the difference comes from https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/3d6d2bbfc20a82c9331aa8a16457904cb85b7506/src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs#L128 and this is different logic for .NET Framework vs .NET (Core)

> Regarding https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/890: the issue is that paths `,` is not a good choice on Windows.

Let's park it for another PR or issue. Here I will make the fixes to make it working as currently described in the docs.

test/IntegrationTests/StartupHookTests.cs
pellared: Issue created https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/894

pjanotti (approved):

PR #898, opened 2022-06-28
Manual metrics test

### LINKED ISSUE #665: Meter provider - integration test
Current state

* [x] ASP.NET metrics - #901
* [x] plugins - #910
* [x] manual metrics - #898

rajkumar-rangaraj:
## Why

Part of #665 Manual Metrics

## Tests

* Added `SubmitMetrics` test case to `StartupHookTests`

## Checklist


test/test-applications/integrations/TestApplication.StartupHook/Program.cs
```diff
@@ -15,25 +15,60 @@
...
-            using (var activity = MyActivitySource.StartActivity("SayHello"))
+            var options = Parser.Default.ParseArguments<Options>(args);
+
+            options.WithParsed<Options>(o =>
             {
-                activity?.SetTag("foo", 1);
-                activity?.SetTag("bar", "Hello, World!");
-                activity?.SetTag("baz", new int[] { 1, 2, 3 });
-            }
+                if (o.Trace)
```
nrcventura, line 32: Should we split the trace and metric blocks into their own methods already?

rajkumar-rangaraj: I did not understand the question. Is the ask to keep the trace and metrics blocks in a different method?

nrcventura: The question is about whether we think it makes sense to have the code look more like the following.
```
if (o.Trace)
{
    GenerateTraceData();
}

if (o.Metrics)
{
    GenerateMetricsData();
}
```

The current code is still small so it may not be worth refactoring it yet, but when we introduce logging, it'll probably make the code easier to read.

rajkumar-rangaraj: Fixed e56c82f650c0bfc23e13240c6e8a8c830e929024

nrcventura (approved):

zacharycmontoya (approved):

test/test-applications/integrations/TestApplication.StartupHook/Program.cs
```diff
@@ -15,17 +15,37 @@
...
 namespace TestApplication.StartupHook
 {
     public class Program
     {
-        private static readonly ActivitySource MyActivitySource = new ActivitySource("TestApplication.StartupHook", "1.0.0");
-
         public static void Main(string[] args)
         {
-            using (var activity = MyActivitySource.StartActivity("SayHello"))
+            var options = Parser.Default.ParseArguments<Options>(args);
```
pellared, line 28: What is the problem if the program would always emit a single span, single metric and single log? 
Do we need to use `CommandLine` and have these `if (o.Trace)` conditionals?
What is the value?

test/IntegrationTests/StartupHookTests.cs
```diff
@@ -101,6 +104,59 @@ public void ApplicationIsNotIncluded()
...
+        const int expectedMetricRequests = 1;
+
+        var testSettings = new TestSettings
+        {
+            MetricsSettings = new MetricsSettings { Port = collectorPort },
+            EnableStartupHook = true,
+            Arguments = "-m"
+        };
+
+        using var processResult = RunTestApplicationAndWaitForExit(testSettings);
```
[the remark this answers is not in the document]
nrcventura: I think that will only result in the tracer provider being configured. The method overloads that do not take a TestSettings object do not configure the metrics-related environment variables.

test/test-applications/integrations/TestApplication.StartupHook/Program.cs
nrcventura: I don't have a strong preference for this either. One benefit of the additional complexity with the `CommandLine` options is that it allows us to test additional scenarios, such as "no traces exported when tracing is enabled but no instrumentation is present" (I don't think it's a useful test though).

pellared: > application in examples folders are used for debugging purpose

Test applications are used for debugging proposes as well (sic!) especially when writing byte-code instrumentation 😉
Examples are intended to be more user-facing - they should serve demonstration purposes.

> It provides a scope for the test to execute what is needed

Currently, the pattern is to create separate test applications. I believe that each "visible" condition makes tests less trustworthy. Mistakes in conditions are often the root cause of bugs. I see the test application is part of our tests. 

Also, I propose to combine the 3 signals into this test application as I would prefer to execute some smoke tests to check if all signals are working as expected. E.g. if one set `*_EXCLUDED_PROCESS` then none of the signals is working.

pellared: > it allows us to test additional scenarios, such as "no traces exported when tracing is enabled but no instrumentation is present" (I don't think it's a useful test though).

I just cannot come up with a good scenario. Otherwise I would agree 😄

test/IntegrationTests/StartupHookTests.cs
```diff
@@ -101,6 +104,48 @@ public void ApplicationIsNotIncluded()
...
+        Assert.True(processResult.ExitCode >= 0, $"Process exited with code {processResult.ExitCode} and exception: {processResult.StandardError}");
+        var metricRequests = collector.WaitForMetrics(expectedMetricRequests, TimeSpan.FromSeconds(5));
+
+        using (new AssertionScope())
+        {
+            metricRequests.Count.Should().Be(expectedMetricRequests);
+
+            var resourceMetrics = metricRequests.Single().ResourceMetrics.Single();
```
pellared, line 132: using `Single()` can cause a crash with an ugly exception. it maybe even hard to know which of the `Single()` has crashed from the stack trace

As far as I remember FluentAssertions has something like `collection.Should().ContainSingle().Which...`

reference: https://fluentassertions.com/collections/

pellared (approved):
I would like to have [this](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/898#discussion_r909611541) addressed but I do not intend to block the PR, because this may be a subjective opinion.

My comments may be also addressed in a separate PR.

Nice work 👍

test/test-applications/integrations/TestApplication.StartupHook/Program.cs
rajkumar-rangaraj: As discussed in SIG removed the conditional argument check from test app.

pellared (approved):

PR #900, opened 2022-06-29
Add Runtime metrics test

### LINKED ISSUE #665: Meter provider - integration test

rajkumar-rangaraj:
## Why

Part of #665 .NET runtime metrics

Metrics are lost during application exit. Especially, in a console app that run and exit metrics are lost. Adding a MeterProvider.ForceFlush resolves an issue. Spent lot of debugging hours in isolating this issue.  

On a side note: We might need a ForceFlush for TracerProvider too, will leave it for discussion and create an issue.

## What

Runtime metrics emitted for .NET Framework and .NET Core are not the same. But, there are few common metrics name asserting on those to prove the runtime metrics is working as expected. 

## Tests

* Added `SubmitMetrics` test case to `RunTimeMetricsTests`

## Checklist


rajkumar-rangaraj:
`ForceFlush` fails `IntegrationTests.Http.HttpTests.SubmitMetrics` as HttpTests started receiving 2 metrics. Will modify `HttpTests` for a fix,

pellared:
> Metrics are lost during application exit. Especially, in a console app that run and exit metrics are lost. Adding a MeterProvider.ForceFlush resolves an issue. Spent lot of debugging hours in isolating this issue.

As far as I understand `MeterProvider.ForceFlush` is forcing the readers to collect metrics (I just quickly looked at the code) which causes .NET Runtime metrics to be recorded and then they are emitted during/before `MeterProvider.ForceFlush`. I do not think we should force metrics collection when the application is closing. At least this is not even shown in none of these examples:

- https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/examples/Console/TestMetrics.cs
- https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/examples/runtime-instrumentation/Program.cs

I have never tried the metrics API and I am not exactly sure how the collection works when using a PUSH-based exporter like OTLP. 

I have found this: https://github.com/open-telemetry/opentelemetry-dotnet/blob/f4dddbe7dac864a6e8ab2be149200b066f2ccc4d/examples/Console/TestMetrics.cs#L73 and I think that the default collection period for OTLP exporter is 1 minute; see: https://github.com/open-telemetry/opentelemetry-dotnet/blob/398d4691908951327c348e07ab12fe78b30e348e/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpMetricExporterExtensions.cs#L28. I have found out that the OTel spec defines [`OTEL_METRIC_EXPORT_INTERVAL`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#periodic-exporting-metricreader) which should be used to configure.

I have created an issue https://github.com/open-telemetry/opentelemetry-dotnet/issues/3417, however, I think we should implement it on our own before it is available. 

> On a side note: We might need a ForceFlush for TracerProvider too, will leave it for discussion and create an issue.

The current implementation gives the processor up to 5 seconds to shut down. I think it is OK.

rajkumar-rangaraj:
Will fix the issue, once after #910 gets merged.

rajkumar-rangaraj:
`LONG_RUNNING` works better here as it is going to execute the same smoke tests.

test/IntegrationTests/RunTimeMetricsTests.cs
```diff
@@ -0,0 +1,67 @@
+// <copyright file="RunTimeMetricsTests.cs" company="OpenTelemetry Authors">
```
pellared, line 1: I suggest renaming to `RuntimeTests`

test/IntegrationTests/RunTimeMetricsTests.cs
```diff
@@ -0,0 +1,67 @@
...
+            assert.Should().NotThrowAfter(
+                waitTime: 30.Seconds(),
+                pollInterval: 1.Seconds());
+        }
+        finally
+        {
+            process.Kill();
+        }
+
+        using (new AssertionScope())
```
pellared, line 63: Leftovers

pellared (approved):

test/IntegrationTests/Helpers/TestHelper.cs
```diff
@@ -121,7 +121,7 @@ public async Task<Container> StartContainerAsync(TestSettings testSettings, int
     /// StartTestApplication starts the test application
     // and returns the Process instance for further interaction.
     /// </summary>
-    public Process StartTestApplication(int traceAgentPort = 0, string arguments = null, string packageVersion = "", int aspNetCorePort = 0, string framework = "", bool enableStartupHook = true)
+    public Process StartTestApplication(int traceAgentPort = 0, int metricsAgentPort = 0, string arguments = null, string packageVersion = "", int aspNetCorePort = 0, string framework = "", bool enableStartupHook = true)
```
nrcventura, line 124: I think that our continued reliance on these optional parameters is going make the tests harder to maintain. On the one hand, it's nice that the test only needs to call `StartTestApplication` with a single named parameter. However, based on my experience with these usages, it gets harder and harder to keep track of which parameters are being used over time. That is why I prefer the overloads that take settings/config objects, which make it a lot easier to track down which specific settings are being used/customized.

PR #901, opened 2022-06-29
ASP.NET metrics tests

### LINKED ISSUE #665: Meter provider - integration test

rajkumar-rangaraj:
## Why

Part of #665 ASP.NET metrics

## Tests

* Added tests to collect ASPNET metrics for ASP.NET app using Prometheus exporter.
* Prometheus exposes the scrape endpoint over localhost, this cannot be mapped outside of container. Hence added a API call from an app to expose the Prometheus metrics. Once OpenTelemetry SDK support Prometheus environment variable we could change our approach to map the ports of docker container and read data. 

## Checklist


pellared:
Can we use the OTLP exporter for ASP.NET tests and have a separate test for testing Prometheus exporter (it can be done with the `TestApplication.StartupHook`?

rajkumar-rangaraj:
> Can we use the OTLP exporter for ASP.NET tests and have a separate test for testing Prometheus exporter (it can be done with the `TestApplication.StartupHook`?

Changed to use OTLP exporter.

rajkumar-rangaraj:
@nrcventura Spent sometime understanding the difference between ASP.NET Core and ASP.NET tests for OTLP. In ASP.NET Core, during application shutdown dispose call is invoked and metrics are collected. Incase of ASP.NET, app is hosted inside a container and we destroy the container before application exit. This does not flush all requests on shutdown. Hence introduced an environment variable `OTEL_AUTO_METRIC_EXPORT_INTERVAL` to control the flush time interval. Once this gets implemented in SDK we could remove it.

pellared:
> we destroy the container before application exit

I see two possible solutions for it:

1. Add a `ShutdownController` that will eventually close the application

```using System.Web.Mvc;

namespace TestApplication.AspNet.Controllers;

public class ShutdownController : Controller
{
    public ActionResult Index()
    {
        Task.Run(async Task () => 
        {
            await Task.Delay(TimeSpan.FromSeconds(1));
            Environment.Exit(0);
        });

        return new HttpStatusCodeResult(200);
    }
}
```

Then you would need to call `await client.GetAsync($"http://localhost:{webPort}/shutdown");`

2. **My personal preference**: `docker stop` correctly `CTRL_SHUTDOWN_EVENT` to the Windows container. We can call ``Environment.Exit(0)` when `CTRL_SHUTDOWN_EVENT` is received. Here I found an example of how it can be handled: https://github.com/StephenClearyExamples/DockerWindowsSignalRepro/blob/master/Program.cs. 

>  introduced an environment variable `OTEL_AUTO_METRIC_EXPORT_INTERVAL`

It should be `OTEL_METRIC_EXPORT_INTERVAL` as defined in the spec.

**Summary**

I propose to rewrite the test so that it stops the container after `await client.GetAsync($"http://localhost:{webPort}");` and the application should handle `CTRL_SHUTDOWN_EVENT`. Of course, we would need to use the ForceFlush on exit.

rajkumar-rangaraj:
> **Summary**
> 
> I propose to rewrite the test so that it stops the container after `await client.GetAsync($"http://localhost:{webPort}");` and the application should handle `CTRL_SHUTDOWN_EVENT`. Of course, we would need to ForceFlush the metrics on the application exit.

My recommendation is to use `OTEL_METRIC_EXPORT_INTERVAL` for this PR. Create a new issue to refine the container to shutdown an app/forceflush with one of the discussed options. Want to keep the scope of this PR to test ASP.NET metric tests using `OTEL_METRIC_EXPORT_INTERVAL`. Please let me know your thoughts.

pellared:
> > **Summary**
> > I propose to rewrite the test so that it stops the container after `await client.GetAsync($"http://localhost:{webPort}");` and the application should handle `CTRL_SHUTDOWN_EVENT`. Of course, we would need to ForceFlush the metrics on the application exit.
> 
> My recommendation is to use `OTEL_METRIC_EXPORT_INTERVAL` for this PR. Create a new issue to refine the container to shutdown an app/forceflush with one of the discussed options. Want to keep the scope of this PR to test ASP.NET metric tests using `OTEL_METRIC_EXPORT_INTERVAL`. Please let me know your thoughts.

I am OK with that 👍  Let's keep the momentum.

rajkumar-rangaraj:
> > > **Summary**
> > > I propose to rewrite the test so that it stops the container after `await client.GetAsync($"http://localhost:{webPort}");` and the application should handle `CTRL_SHUTDOWN_EVENT`. Of course, we would need to ForceFlush the metrics on the application exit.
> > 
> > 
> > My recommendation is to use `OTEL_METRIC_EXPORT_INTERVAL` for this PR. Create a new issue to refine the container to shutdown an app/forceflush with one of the discussed options. Want to keep the scope of this PR to test ASP.NET metric tests using `OTEL_METRIC_EXPORT_INTERVAL`. Please let me know your thoughts.
> 
> I am OK with that 👍 Let's keep the momentum.

Created an issue #914 and fixed OTEL_AUTO_METRIC_EXPORT_INTERVAL to OTEL_METRIC_EXPORT_INTERVAL.

test/IntegrationTests/AspNetTests.cs
```diff
@@ -65,5 +75,51 @@ public async Task SubmitsTraces()
...
+
+        var metricRequests = collector.WaitForMetrics(expectedMetricRequests, TimeSpan.FromSeconds(5));
+
+        using (new AssertionScope())
+        {
+            metricRequests.Count.Should().BeGreaterThanOrEqualTo(expectedMetricRequests);
+            var resourceMetrics = metricRequests.LastOrDefault().ResourceMetrics.Single();
+
+            var expectedServiceNameAttribute = new KeyValue { Key = "service.name", Value = new AnyValue { StringValue = "TestApplication.AspNet" } };
+            resourceMetrics.Resource.Attributes.Should().ContainEquivalentOf(expectedServiceNameAttribute);
```
pellared, line 117: No need to test this here.
This feature is already covered in `SmokeTests`.
We do not want all tests to fail if `"service.name"` is not set.

test/IntegrationTests/AspNetTests.cs
```diff
@@ -65,5 +75,51 @@ public async Task SubmitsTraces()
...
+            var aspnetScope = resourceMetrics.ScopeMetrics.Single(rm => rm.Scope.Name.Equals("OpenTelemetry.Instrumentation.AspNet", StringComparison.OrdinalIgnoreCase));
+
+            var httpServerDurationMetric = aspnetScope.Metrics.FirstOrDefault(m => m.Name.Equals("http.server.duration", StringComparison.OrdinalIgnoreCase));
+            httpServerDurationMetric.Should().NotBeNull();
```
pellared, line 121: Try something like:

```csharp
var aspnetMetrics = resourceMetrics.ScopeMetrics.Should().ContainSingle(x => x.Scope.Name == "OpenTelemetry.Instrumentation.AspNet").Which.Metrics;
aspnetMetrics.Should().ContainSingle(x => x.Name == "http.server.duration");
```


Because

```csharp
var aspnetScope = resourceMetrics.ScopeMetrics.Single(rm => rm.Scope.Name.Equals("OpenTelemetry.Instrumentation.AspNet", StringComparison.OrdinalIgnoreCase));
```

can throw an exception

test/test-applications/integrations/aspnet/TestApplication.AspNet/TestApplication.AspNet.csproj
```diff
@@ -208,4 +208,4 @@
   </Target>
   <Target Name="AfterBuild">
   </Target> -->
-</Project>
+</Project>
```
pellared, line 211: nit: bring back the EOL 😄

pellared (approved):
Good job 👍 

Left a few comments.

docs/config.md
```diff
@@ -147,6 +147,7 @@ Important environment variables include:
 | `OTEL_DOTNET_AUTO_FLUSH_ON_UNHANDLEDEXCEPTION` | Controls whether the telemetry data is flushed when an [AppDomain.UnhandledException](https://docs.microsoft.com/en-us/dotnet/api/system.appdomain.unhandledexception) event is raised. Set to `true` when you suspect that you are experiencing a problem with missing telemetry data and also experiencing unhandled exceptions. | `false` |
 | `OTEL_DOTNET_AUTO_TRACES_PLUGINS` | Colon-separated list of OTel SDK instrumentation tracer plugin types, specified with the [assembly-qualified name](https://docs.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname?view=net-6.0#system-type-assemblyqualifiedname). _Note: This list must be colon-separated because the type names may include commas._ | |
 | `OTEL_DOTNET_AUTO_METRICS_ADDITIONAL_SOURCES` | Comma-separated list of additional `System.Diagnostics.Metrics.Meter` names to be added to the meter at the startup. Use it to capture manually instrumented spans. |  |
+| `OTEL_METRIC_EXPORT_INTERVAL` | The time interval (in milliseconds) between the start of two export attempts. |  |
```
pellared, line 150: `60000` should be the default

test/IntegrationTests/AspNetTests.cs
```diff
@@ -65,5 +75,51 @@ public async Task SubmitsTraces()
...
+
+        Output.WriteLine("Sample response:");
+        Output.WriteLine(content);
+
+        var metricRequests = collector.WaitForMetrics(expectedMetricRequests, TimeSpan.FromSeconds(5));
+
+        using (new AssertionScope())
+        {
+            metricRequests.Count.Should().BeGreaterThanOrEqualTo(expectedMetricRequests);
+            var resourceMetrics = metricRequests.LastOrDefault().ResourceMetrics.Single();
```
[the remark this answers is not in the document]
nrcventura: I think that you are correct and that the majority of our tests will be simple enough and fast enough that we only care about getting 1 metric request, and there will only be 1 metric request. I think that at least for now, if there are multiple metric requests, we may not care which request data that the test looks at especially if they are more of test of whether we successfully "enabled" an instrumentation library that we do not own.

The counter argument to this approach may come up when we start testing trace data via OTLP. The MockZipkinCollector has some helper methods that we take advantage of to ensure we are looking at the expected span data regardless of which request the span was included in. For traces we want to analyze all of the trace data that came in, instead of just looking at the latest request. We can either accept this break in symmetry between traces and metrics in the way we wait for and look at the data, or we try to keep them similar. I have not spent enough time trying out the approaches for testing traces via OTLP to get a feeling for what the approach should look like.

test/IntegrationTests/AspNetTests.cs
rajkumar-rangaraj: Was not aware of this feature. This makes it simpler.

test/IntegrationTests/AspNetTests.cs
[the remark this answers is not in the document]
nrcventura: That seems like an interesting idea to try out.

nrcventura (approved):

test/IntegrationTests/Helpers/TestHelper.cs
```diff
@@ -73,17 +68,35 @@ public async Task<Container> StartContainerAsync(int traceAgentPort, int webPort
...
             .WithBindMount(logPath, "c:/inetpub/wwwroot/logs")
             .WithBindMount(EnvironmentHelper.GetNukeBuildOutput(), "c:/opentelemetry");
 
+        string agentBaseUrl = $"http://{DockerNetworkHelper.IntegrationTestsGateway}:{agentPort}";
+        string agentHealthzUrl = $"{agentBaseUrl}/healthz";
+
+        if (testSettings.TracesSettings != null)
+        {
+            string zipkinEndpoint = $"{agentBaseUrl}/api/v2/spans";
+            Output.WriteLine($"Zipkin Endpoint: {zipkinEndpoint}");
```
pellared, line 88: consider doing the same for OTLP metrics endpoint

pellared (approved):

Issue #388, opened 2022-03-03
Enable all instrumentations by default


pellared:

nrcventura:
I think that this is a good thing for us to aim for to improve the experience for our users. When we do this we may also need to provide a way to "disable" or not load certain instrumentations. For now, to simplify things, I think that this should primarily apply to source-based instrumentation. I think that we need to think further about what that means for byte-code based instrumentation.

pellared:
I just want to note that current behavior is strange and not user-friendly.

- Bytecode Instrumentations are automatically enabled.
- Source Instrumentations have to be enabled using `OTEL_DOTNET_AUTO_ENABLED_INSTRUMENTATIONS`.

I find it is so confusing that we should address it even in beta...

Both bytecode instrumentations and source instrumentations can be disabled via `OTEL_DOTNET_AUTO_DISABLED_INSTRUMENTATIONS`.

pellared:
Let's assume that we would like to enable all source instrumentation by default.
How do we want to handle it if it cannot be applied?
E.g. when a console application tries to add ASP.NET Core source instrumentation.
I see the following possibilities:

1. Fail-fast and require the user to set `OTEL_DOTNET_AUTO_DISABLED_INSTRUMENTATIONS=AspNet`. Or explicitly set only the needed source instrumentation via `OTEL_DOTNET_AUTO_ENABLED_INSTRUMENTATIONS`. This gives the user more control over what is going on, reduces warning errors but would require more time for initial setup.
2. Catch exceptions when source instrumentation cannot be applied and simply log a warning. Faster to adopt but if something goes wrong with something that should work - it may be not noticed - because the more source instrumentation will be added the more warning logs would obfuscate the logs.

Of course, both approaches could be implemented and controlled using e.g. an `OTEL_DOTNET_AUTO_FAILFAST` env var.

Any other ideas?

nrcventura:
I see the biggest blocker to auto-loading all source instrumentations being the dependency management issue. The byte-code instrumentation doesn't have explicit dependencies so if we load those instrumentation libraries it won't cause a problem. If I am remembering correctly, once you load an assembly, and it has some errors, I don't think that there's a good away to unload that assembly. Furthermore, if the user is also using some sort of assembly scanning library (typically to support plugins or some other dynamic loading scenario), the simple act of expecting the types in that assembly can potentially trigger exceptions and crash the app if the required dependencies are not available.

[...]

Issue #875, opened 2022-06-23
[Feature Request] Replace the exception during the instrumentation loading with the warning


yzhoholiev:
# Feature Request

Replace the exception during the instrumentation loading with the logging of the warning.

**Is your feature request related to a problem? Please describe.**

Replace the exception during the instrumentation loading with the logging of the warning, to not break the entire startup of the instrumentation and improve a backward compatibility in case of centralized management of the configuration (when the new instrumentation is added, but older versions of the auto instrumentation are still in use).

**Describe the solution you'd like**

Consider replacing the throwing of the `FormatException` during the configuration loading
https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/0a539f9f6636b14c0f8006dfce6d58a55e4f53f0/src/OpenTelemetry.AutoInstrumentation/Configuration/MeterSettings.cs#L45-L52

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/0a539f9f6636b14c0f8006dfce6d58a55e4f53f0/src/OpenTelemetry.AutoInstrumentation/Configuration/TracerSettings.cs#L46-L53

**Describe alternatives you've considered**
-

**Additional context**
-

pellared:
Our current approach is described here: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/design.md#error-handling

The reason is that the user may notice the misconfiguration when it is too late (e.g. when the application is having problems). We prefer to fail fast. It is in line with https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md 

You can also read more here when we were deciding to choose this behavior: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/231 

> improve a backward compatibility in case of centralized management of the configuration (when the new instrumentation is added, but older versions of the auto instrumentation are still in use)

This is not backward compatibility as you would add new instrumentation 😉

However, we are working on enabling all instrumentations by default in a such way that the auto instrumentations are applied in a "non-breaking" way. This should make users happier. See https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/388

We also cannot safely change the `throw` to logging error as they may be side-effects caused by improper instrumentation loading (e.g. some assemblies may be loaded).

pellared:
[...]

PR #903, opened 2022-06-29
[DoNotMerge] Enable all trace instrumentations by default

RassK:
## Why

Part of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/388

## What

Enables to lazily load trace instrumentations that contain external dependencies (such as ASP.NET Core instrumentation). Trace instrumentations will be created if assembly becomes available.

Lazy load is only necessary for trace instrumentations that depend on external dependencies!

This is first example fixing ASP.NET Core instrumentation. 
This PR also enables ASP.NET Core instrumentation on .NET 4.6.2.

`SdkSelfDiagnosticsEventListenerTests` fix is separated to https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/927. This PR is dependent on that fix.

### Guidelines

* Sources (also legacy) must be added before `TraceProvider` build.
** This requires manual extraction of the sources from instrumentation libraries.
* Instrumentation instance must be added only if dependencies are available and loaded.
**  SDK only tracks the instances and disposes them on SDK dispose call. We can replicate the same behaviour in auto-instrumentation side.

## Schema

This PR includes the blue part.
instrumentation_registration_options
 
## Tests

Removed `OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS` config from all of the integration tests.
Added `LazyInstrumentationLoaderTests` unit tests.

## Checklist


src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
```diff
@@ -56,10 +55,14 @@ public static TracerProviderBuilder UseEnvironmentVariables(this TracerProviderB
     public static TracerProviderBuilder AddSdkAspNetInstrumentation(this TracerProviderBuilder builder)
     {
 #if NET462
-        return builder.AddAspNetInstrumentation();
+        builder.AddAspNetInstrumentation();
 #elif NETCOREAPP3_1_OR_GREATER
-        return builder.AddAspNetCoreInstrumentation();
+        // Instrumentation will be added via assembly load detector
+        builder.AddSource("OpenTelemetry.Instrumentation.AspNetCore");
+        builder.AddLegacySource("Microsoft.AspNetCore.Hosting.HttpRequestIn");
```
pellared, line 62: I think we can call this even for .NET Framework as it is possible to use ASP.NET Core on .NET Framework

Reference: https://www.nuget.org/packages/OpenTelemetry.Instrumentation.AspNetCore#supportedframeworks-body-tab

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        private readonly List<AssemblyLoadDetector> _assemblyLoadDetectors = new();
+        private readonly ConcurrentBag<object> _instrumentations = new();
+        private readonly TracerSettings _tracerSettings;
+
+        public LazyInstrumentationLoader(TracerSettings tracerSettings)
+        {
+            _tracerSettings = tracerSettings;
+
+#if NETCOREAPP
```
pellared, line 37: this if is not needed as mentioned before

RassK:
This PR leaves one question: If user explicitly defines `OTEL_DOTNET_AUTO_ENABLED_INSTRUMENTATIONS`, should we use the SDK builder extensions directly, instead of creating lazy instrumentation bootstrapper?

src/OpenTelemetry.AutoInstrumentation/Loading/AssemblyLoadDetector.cs
```diff
@@ -0,0 +1,68 @@
...
+    internal abstract class AssemblyLoadDetector
+    {
+        private HashSet<string> _requiredAssemblies;
+        private int _loadedCount;
+
+        public AssemblyLoadDetector(IEnumerable<string> requiredAssemblies)
+        {
+            _requiredAssemblies = new HashSet<string>(requiredAssemblies);
+
+            AppDomain.CurrentDomain.AssemblyLoad += CurrentDomain_AssemblyLoad;
```
pellared, line 34: Adding an `AssemblyLoad` handler per each detector would be inefficient. We should have one handler which works against some kind of `Dictionary<string(assemblyName), AssemblyLoadDetector>`.

I think you can change the `LazyInstrumentationLoader` so that `_assemblyLoadDetectors` will be a dictionary initialized when a detector is subscribed. After everything is ready you can add a single `AssemblyLoad` handler. The handler should be also removed when `LazyInstrumentationLoader` is disposed.

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
```diff
@@ -0,0 +1,87 @@
...
+            while (_instrumentations.TryTake(out object instrumentation))
+            {
+                if (instrumentation is IDisposable disposableInstrumentation)
+                {
+                    disposableInstrumentation.Dispose();
+                }
+            }
+        }
+
+        private void SubScribe(TracerInstrumentation instrumentation, Func<AssemblyLoadDetector> builder)
```
pellared, line 53: nit: Subscribe 😉

pellared (commented):
I think you can remove this line to test if it works: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/17de331195ed3d954097011bc434495642dbc1da/test/IntegrationTests/StartupHookTests.cs#L39

pellared:
> This PR leaves one question: If user explicitly defines `OTEL_DOTNET_AUTO_ENABLED_INSTRUMENTATIONS`, should we use the SDK builder extensions directly, instead of creating lazy instrumentation bootstrapper?

We should still use lazy instrumentation bootstrapper -> easier, more consistent, less crashing, would solve e.g. https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/875

src/OpenTelemetry.AutoInstrumentation/Loading/AssemblyLoadDetector.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        private void CurrentDomain_AssemblyLoad(object sender, AssemblyLoadEventArgs args)
+        {
+            string assemblyName = args.LoadedAssembly.FullName.Split(',')[0];
+
+            if (_requiredAssemblies.Contains(assemblyName, StringComparer.InvariantCultureIgnoreCase))
+            {
+                if (Interlocked.Increment(ref _loadedCount) == _requiredAssemblies.Count)
+                {
+                    var loader = GetInstrumentationLoader();
```
pellared, line 49: this should be wrapped in some kind of `try-catch` as we do not want to crash the application at runtime

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
```diff
@@ -0,0 +1,87 @@
...
+using OpenTelemetry.AutoInstrumentation.Logging;
+
+namespace OpenTelemetry.AutoInstrumentation.Loading
+{
+    internal class LazyInstrumentationLoader : IDisposable
+    {
+        private static readonly ILogger Logger = OtelLogging.GetLogger();
+
+        private readonly List<AssemblyLoadDetector> _assemblyLoadDetectors = new();
+        private readonly ConcurrentBag<object> _instrumentations = new();
```
pellared, line 30: maybe `ConcurrentBag<IDisposable>` or even `ConcurrentBag<Action>` ?

src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
RassK: I would move this to the new PR, as it currently replicates the same behaviour we have. Currently we have only one enum member ASP.NET, and based on the framework we either use ASP.NET Core package or ASP.NET framework package.

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetCoreDetector.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            return (ConcurrentBag<object> instrumentations) =>
+            {
+                var instrumentationType = Type.GetType("OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreInstrumentation, OpenTelemetry.Instrumentation.AspNetCore");
+                var httpInListenerType = Type.GetType("OpenTelemetry.Instrumentation.AspNetCore.Implementation.HttpInListener, OpenTelemetry.Instrumentation.AspNetCore");
+
+                object httpInListener = Activator.CreateInstance(httpInListenerType, args: new OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreInstrumentationOptions());
+                object instrumentation = Activator.CreateInstance(instrumentationType, args: httpInListener);
+
+                instrumentations.Add(instrumentation);
```
pellared, line 47: the detector should be able to register some code/objects for cleanup. right now it could even execute `instrumentations.Clear()` 😄

RassK: Should be better now, returning just the created instrumentation 👍

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
RassK: We will need to track every instance, but not all of them are `IDisposable`.

pellared: So why do we need them?

pellared:
Would this PR not solve https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/388?
If so then I suggest renaming the PR title 😉

pellared:
> I think you can remove this line to test if it works:
> 
> https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/17de331195ed3d954097011bc434495642dbc1da/test/IntegrationTests/StartupHookTests.cs#L39


We could remove it in all tests to make sure that the default scenario is working. We can add other tests to check if enabling disabling instrumentations via env var is working (e.g. in `StartupHookTests` which I plan to rename to `SmokeTests`)

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
RassK: Need to be sure that these instances are not garbage collected and live until SDK is disposed.

pellared: please add a comment in code and resolve 😉

src/OpenTelemetry.AutoInstrumentation/Loading/AssemblyLoadDetector.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        internal abstract Func<object> GetInstrumentationBuilder();
+
+        private void CurrentDomain_AssemblyLoad(object sender, AssemblyLoadEventArgs args)
+        {
+            string assemblyName = args.LoadedAssembly.FullName.Split(',')[0];
+
+            if (_requiredAssemblies.Contains(assemblyName, StringComparer.InvariantCultureIgnoreCase))
+            {
+                if (Interlocked.Increment(ref _loadedCount) == _requiredAssemblies.Count)
```
zacharycmontoya, line 46: Can you imagine any scenarios where multiple versions of the same assembly name are loaded? That could mess up the comparison with `loadedCount`

src/OpenTelemetry.AutoInstrumentation/Loading/AssemblyLoadDetector.cs
```diff
@@ -0,0 +1,67 @@
...
+            AppDomain.CurrentDomain.AssemblyLoad += CurrentDomain_AssemblyLoad;
+        }
+
+        public event EventHandler<LoadDetectorReadyEventArgs> OnReady;
+
+        internal abstract Func<object> GetInstrumentationBuilder();
+
+        private void CurrentDomain_AssemblyLoad(object sender, AssemblyLoadEventArgs args)
+        {
+            string assemblyName = args.LoadedAssembly.FullName.Split(',')[0];
```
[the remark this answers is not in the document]
pellared: or  `args.LoadedAssembly.FullName.Split(',', 2)[0];` (I do not have the measurements as well)

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetCoreDetector.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        internal override Func<object> GetInstrumentationBuilder()
+        {
+            return () =>
+            {
+                var instrumentationType = Type.GetType("OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreInstrumentation, OpenTelemetry.Instrumentation.AspNetCore");
+                var httpInListenerType = Type.GetType("OpenTelemetry.Instrumentation.AspNetCore.Implementation.HttpInListener, OpenTelemetry.Instrumentation.AspNetCore");
+
+                object httpInListener = Activator.CreateInstance(httpInListenerType, args: new OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreInstrumentationOptions());
```
zacharycmontoya, line 41: I noticed that you're creating a new `AspNetCoreInstrumentationOptions` without reflection, but the other two types use reflection. Can we do it all without reflection?

pellared:
1. Update the documentation (from memory: `README.md` and `config.md`
2. Update the changelog
3. This PR also adds ASP.NET Core instrumentation for .NET Framework - worth mentioning in PR description and the changelog

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            string detectorName = sender.GetType().Name;
+
+            Logger.Debug("Detector '{0}' executed load", detectorName);
+
+            try
+            {
+                var instrumentation = e.Builder();
+
+                _instrumentations.Add(instrumentation);
```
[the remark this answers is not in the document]
pellared: Nope. TraceProvider only does what we were doing here:

1. keeps the references so that the objects are not disposed
1. during its disposal also disposes the instrumentations if they implement `IDisposable`

This is what the .NET SDK SIG suggested to us in order not to extend the SDK's public API 😉

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetCoreDetector.cs
pellared: They are `internal` 😢

src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
```diff
@@ -56,10 +55,14 @@ public static TracerProviderBuilder UseEnvironmentVariables(this TracerProviderB
...
-        return builder.AddAspNetInstrumentation();
-#elif NETCOREAPP3_1_OR_GREATER
-        return builder.AddAspNetCoreInstrumentation();
+        builder.AddAspNetInstrumentation();
 #endif
+
+        // Instrumentation will be added via assembly load detector
+        builder.AddSource("OpenTelemetry.Instrumentation.AspNetCore");
+        builder.AddLegacySource("Microsoft.AspNetCore.Hosting.HttpRequestIn");
+
```
pellared, line 64: remove blank line 😄

src/OpenTelemetry.AutoInstrumentation/Loading/AssemblyLoadDetector.cs
RassK: That's good point. NET Framework should be safe until we watch only default appdomain, in .NET Core I think I need to fix it a bit, by looking only into default assembly load context.

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetCoreDetector.cs
RassK: Others are internal types in SDK. `AspNetCoreInstrumentationOptions` is actually pretty complicated type with funcs to filter and enrich. In case we would need to reconfigure options, it would be much easier to allocate it directly.

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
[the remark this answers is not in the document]
RassK: No we decided to track these instances ourself, so we can go without modifying SDK.

src/OpenTelemetry.AutoInstrumentation/Loading/AssemblyLoadDetector.cs
RassK: That's already a way bigger issue, we are not generating spans in sibling AppDomains / load contextes as the SDK is loaded into default context.

pellared: Then I suggest only watching the default AppDomain/LoadContext. It would be good to mention it in `troubleshooting.md` that auto-instrumentation is only working for the default AppDomain/LoadContext.

RassK: I would extract an issue from this comment. Need to build some tests apps to verify which parts do work and which not. With source based instrumentation I have some idea and it's already partially tested, but I'm not sure how rejited libraries are selected - it probably doesn't not look for appdomain / load context.

src/OpenTelemetry.AutoInstrumentation/Loading/AssemblyLoadDetector.cs
nrcventura: I'm also concerned about having multiple handlers being registered and that we are not also unregistering the handlers. 

Perhaps this class needs to be its own thing and not part of an inheritance hierarchy. And this class just focuses on listening to assembly load events, and maintaining a collection of lazily loadable instrumentation loaders. This is similar to what @pellared mentioned above except that it assumes that the `LazyInstrumentationLoader` will not also be responsible for handling the assembly load events.

test/test-applications/integrations/TestApplication.GraphQL/Properties/launchSettings.json
```diff
@@ -16,14 +16,21 @@
 
         "CORECLR_ENABLE_PROFILING": "1",
         "CORECLR_PROFILER": "{918728DD-259F-4A6A-AC2B-B85E1B658318}",
-        "CORECLR_PROFILER_PATH": "$(SolutionDir)bin\\tracer-home\\win-x64\\OpenTelemetry.AutoInstrumentation.Native.dll",
-        "DOTNET_STARTUP_HOOKS": "$(SolutionDir)bin\\tracer-home\\netcoreapp3.1\\OpenTelemetry.AutoInstrumentation.StartupHook.dll",
+        "CORECLR_PROFILER_PATH_64": "F:\\Dev\\forks\\rassk\\opentelemetry-dotnet-instrumentation\\bin\\tracer-home\\win-x64\\OpenTelemetry.AutoInstrumentation.Native.dll",
```
nrcventura, line 19: I don't think that you meant to commit most of the changes in in this file.

nrcventura:
Where in this code are we enabling "all" activity sources?

RassK:
> Where in this code are we enabling "all" activity sources?

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/10f9bbcf795c81670fe619ac5a78b21b9fedc86f/src/OpenTelemetry.AutoInstrumentation/Configuration/TracerSettings.cs#L58

test/test-applications/integrations/TestApplication.GraphQL/Properties/launchSettings.json
RassK: Good catch!

CHANGELOG.md
```diff
@@ -45,6 +45,9 @@ This release is built on top of [OpenTelemetry .NET](https://github.com/open-tel
   `OTEL_DOTNET_AUTO_DISABLED_INSTRUMENTATIONS`&#8594;`OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS`
   `OTEL_DOTNET_AUTO_INSTRUMENTATION_PLUGINS`&#8594;`OTEL_DOTNET_AUTO_TRACES_PLUGINS`
   `OTEL_DOTNET_AUTO_ADDITIONAL_SOURCES`&#8594;`OTEL_DOTNET_AUTO_TRACES_ADDITIONAL_SOURCES`
+
+- `OTEL_DOTNET_AUTO_ENABLED_INSTRUMENTATIONS` default value is changed to
```
pellared, line 49: `OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS` 😉 

We should change the title (and description) of this PR that at this moment it is only about "tracing instrumentations"

test/IntegrationTests/SmokeTests.cs
```diff
@@ -39,7 +39,6 @@ public SmokeTests(ITestOutputHelper output)
         : base("Smoke", output)
     {
         SetEnvironmentVariable("OTEL_SERVICE_NAME", ServiceName);
-        SetEnvironmentVariable("OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS", "HttpClient");
```
pellared, line 42: I would leave this one, as probably, we would have no integration test for `OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS` at all. If there is some test that checks `OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS` then ignore this comment.

docs/config.md
```diff
@@ -26,7 +26,7 @@ for more details.
 | Environment variable | Description | Default value |
 |-|-|-|
 | `OTEL_DOTNET_AUTO_INTEGRATIONS_FILE` | List of bytecode instrumentations JSON configuration filepaths, delimited by the platform-specific path separator (`;` on Windows, `:` on Linux and macOS). For example: `%ProfilerDirectory%/integrations.json` | |
-| `OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS` | Comma-separated list of traces source instrumentations you want to enable. |  |
+| `OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS` | Comma-separated list of traces source instrumentations you want to enable. | all available instrumentations |
```
pellared, line 29: Can you remove setting `OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS` from `run-example.sh` and all `launchSettings.json` example files? And test if the examples are still working as expected.

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
```diff
@@ -0,0 +1,90 @@
...
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using OpenTelemetry.AutoInstrumentation.Configuration;
+using OpenTelemetry.AutoInstrumentation.Logging;
+
+namespace OpenTelemetry.AutoInstrumentation.Loading
+{
+    internal class LazyInstrumentationLoader : IDisposable
```
pellared, line 25: How about adding a unit test for this class? As it is quite complex.

src/OpenTelemetry.AutoInstrumentation/Loading/AssemblyLoadDetector.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/10f9bbcf795c81670fe619ac5a78b21b9fedc86f/src/OpenTelemetry.AutoInstrumentation/Loading/AssemblyLoadDetector.cs#L52

Deregistration is actually here.

I'm more concerned if it becomes synchronized. The instrumentation instances need to be created asap after assembly load. Synchronized loop could cause some delays. That could be changed to parallel looping, but I'm not sure if that is actually cheaper.

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
```diff
@@ -0,0 +1,90 @@
...
+                if (instrumentation is IDisposable disposableInstrumentation)
+                {
+                    disposableInstrumentation.Dispose();
+                }
+            }
+        }
+
+        private void Subscribe(TracerInstrumentation instrumentation, Func<AssemblyLoadDetector> builder)
+        {
+            if (_tracerSettings.EnabledInstrumentations.Contains(instrumentation))
```
pellared, line 53: I suggest moving this `if` to the caller. `_tracerSettings` could be removed and this class would be easier to test and also easier to reuse for metrics instrumentations.

RassK: This would then generate multiple ifs instead of one.

pellared: Check if it won't be better if `Subscribe` calls would be moved to `Instrumentation`. Otherwise, I think, it would be hard to unit test the assembly loading + instrumentation loading code.

nrcventura:
> > Where in this code are we enabling "all" activity sources?
> 
> https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/10f9bbcf795c81670fe619ac5a78b21b9fedc86f/src/OpenTelemetry.AutoInstrumentation/Configuration/TracerSettings.cs#L58
> 
> Builder then adds all instrumentations. Problematic instrumentations (currently only ASP.NET Core) have to be separated to sources and instrumentation instance. During the TraceProvider build all added sources are converted to ActivitiListener filter.

I was imagining that we were trying to do something more generic instead of limiting ourselves to the activity sources that correspond to the instrumentations defined in that enum. For example, if someone is using a library that is already instrumented via an `ActivitySource` and does not require a separate instrumentation library in order to make the data OpenTelemetry compatible, this current approach still requires someone to configure the additional sources environment variable in order to collect that telemetry. If we are able to simply enable all `ActivitySource` data to be collected, and then dynamically enable any relevant instrumentation libraries that are needed (so that those instrumentations that require instrumentation libraries to generate OTel compatible telemetry) I think that will be an easier configuration experience for our users.

RassK:
> I was imagining that we were trying to do something more generic instead of limiting ourselves to the activity sources that correspond to the instrumentations defined in that enum. For example, if someone is using a library that is already instrumented via an `ActivitySource` and does not require a separate instrumentation library in order to make the data OpenTelemetry compatible, this current approach still requires someone to configure the additional sources environment variable in order to collect that telemetry. If we are able to simply enable all `ActivitySource` data to be collected, and then dynamically enable any relevant instrumentation libraries that are needed (so that those instrumentations that require instrumentation libraries to generate OTel compatible telemetry) I think that will be an easier configuration experience for our users.

Enabling all ActivitySources could cost quite the performance and adds loads of extra allocations - the bigger the application, the more problems we have. Currently DiagnosticSource is not generating Activities if we are not listening a corresponding source. If we are subscribing to all, then we need to filter activities after creation. Also there are more ActivitySources than auto-instrumentation would be interested in (such as internal library diagnostics, libraries that don't cross process boundary). 

Additional sources must be configured for sources, that we are not directly supporting (such as custom named sources created by manual instrumentation). If the library is supported by us, then there are no extra steps to configure.

src/OpenTelemetry.AutoInstrumentation/Diagnostics/SdkSelfDiagnosticsEventListener.cs
```diff
@@ -31,13 +31,19 @@ internal class SdkSelfDiagnosticsEventListener : EventListener
...
+    private readonly ILogger log;
     private readonly List<EventSource> eventSourcesBeforeConstructor = new();
 
     public SdkSelfDiagnosticsEventListener(EventLevel eventLevel)
+        : this(eventLevel, OtelLogging.GetLogger())
     {
+    }
+
+    internal SdkSelfDiagnosticsEventListener(EventLevel eventLevel, ILogger logger)
+    {
```
pellared, line 45: Can we have a single constructor (just this one)?

References:
- https://blogs.cuttingedge.it/steven/posts/2013/di-anti-pattern-multiple-constructors/
- https://blog.ploeh.dk/2011/05/30/DesignSmellDefaultConstructor/

RassK: I don't have any strong opinion on this as we don't have DI anyway. Default constructor would have been more futureproof if `GetLogger` is changed to `GetLoggerFor` as passing the wrong instance. Currently, as the logger is static all over the app anyway, I'm fine with refactored approach also.

src/OpenTelemetry.AutoInstrumentation/Loading/AssemblyLoadDetector.cs
```diff
@@ -0,0 +1,67 @@
...
+    internal abstract class AssemblyLoadDetector
+    {
+        private HashSet<string> _requiredAssemblies;
+        private int _loadedCount;
+
+        public AssemblyLoadDetector(IEnumerable<string> requiredAssemblies)
+        {
+            _requiredAssemblies = new HashSet<string>(requiredAssemblies);
+
+            AppDomain.CurrentDomain.AssemblyLoad += CurrentDomain_AssemblyLoad;
```
rajkumar-rangaraj, line 33: Forked the code to test out the approach. I commented out this event handler wiring and tried the code with console and asp.net core app it could collect traces. 

I tired it on smoke test application, it could collect http client trace too. Does this event wiring needed?

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -86,7 +88,8 @@ public static void Initialize()
             if (TracerSettings.LoadTracerAtStartup || MeterSettings.LoadMetricsAtStartup)
             {
                 // Initialize SdkSelfDiagnosticsEventListener to create an EventListener for the OpenTelemetry SDK
-                _sdkEventListener = new(EventLevel.Warning);
+                _sdkEventListener = new(EventLevel.Warning, Logger);
+                _lazyInstrumentation = new(new LazyInstrumentationBuilders(TracerSettings));
```
rajkumar-rangaraj, line 92: How lazily does it load? I could see the library load before the AddSource being called.

src/OpenTelemetry.AutoInstrumentation/Loading/AssemblyLoadDetector.cs
RassK: Without `AspNetCoreInstrumentation` instance these traces are invalid. With this approach it's very important to create instrumentation instance ASAP after assembly load. This instance overrides ASP.NET spans and fills with correct data (tags).

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
RassK: Unfortunately I didn't have better naming for this - could be bad choice. 
I'm afraid that ASP.NET apps at least will load the framework dlls quite early in the bootstrapping process. This could not be the same behaviour with other instrumentations - specially the ones that are dynamically loaded.

pellared (commented):
I find dependencies and differences between `AssemblyLoadDetector`, `LazyInstrumentationBuilders`, `LazyInstrumentationLoader`  a little hard to understand/follow. Sorry for "just complaining" without concrete proposals hard to refactor. What makes it hard to understand for me is probably:
- usage of events - e.g. I do not think that `OnReady` has to be an event
- some logic in constructors - e.g. in `AssemblyLoadDetector` and `LazyInstrumentationLoader`
- usage of abstract classes
- naming

I can try finding some time to refactor this next week.

test/OpenTelemetry.AutoInstrumentation.Tests/Loading/LazyInstrumentationLoaderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+        public void LazyInstrumentationLoader_InstrumentationNotEnabled()
+        {
+            var detectors = new Dictionary<TracerInstrumentation, Func<AssemblyLoadDetector>>()
+            {
+                { TracerInstrumentation.HttpClient, () => new DummyDetector() }
+            };
+
+            var config = new NameValueCollection()
+            {
+                { ConfigurationKeys.Traces.Instrumentations, nameof(TracerInstrumentation.AspNet) }
```
pellared, line 58: can it be e.g. `OtherLibrary` instead of  `nameof(TracerInstrumentation.AspNet)`

test/OpenTelemetry.AutoInstrumentation.Tests/Loading/LazyInstrumentationLoaderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+    {
+        [Fact]
+        public void LazyInstrumentationLoader_DefaultLoad()
+        {
+            var detectors = new Dictionary<TracerInstrumentation, Func<AssemblyLoadDetector>>()
+            {
+                { default(TracerInstrumentation), () => new DummyDetector() }
+            };
+
+            var loader = CreateLoader(new NameValueCollection(), detectors);
```
pellared, line 39: loader is not disposed

test/OpenTelemetry.AutoInstrumentation.Tests/Loading/LazyInstrumentationLoaderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            {
+                { TracerInstrumentation.HttpClient, () => new DummyDetector() }
+            };
+
+            var config = new NameValueCollection()
+            {
+                { ConfigurationKeys.Traces.Instrumentations, nameof(TracerInstrumentation.AspNet) }
+            };
+
+            var loader = CreateLoader(config, detectors);
```
pellared, line 61: loader is not disposed

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetCoreDetector.cs
```diff
@@ -0,0 +1,48 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+
+namespace OpenTelemetry.AutoInstrumentation.Loading
+{
+    internal class AspNetCoreDetector : AssemblyLoadDetector
```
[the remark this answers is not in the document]
RassK: I guess we don't want to overexpose? Especially if it's not meant to be used as a public api.

test/OpenTelemetry.AutoInstrumentation.Tests/Loading/LazyInstrumentationLoaderTests.cs
RassK: That would require turning away from enum based values. Perhaps I'll add some comments to the code?

src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
```diff
@@ -56,10 +55,14 @@ public static TracerProviderBuilder UseEnvironmentVariables(this TracerProviderB
     public static TracerProviderBuilder AddSdkAspNetInstrumentation(this TracerProviderBuilder builder)
     {
 #if NET462
-        return builder.AddAspNetInstrumentation();
-#elif NETCOREAPP3_1_OR_GREATER
-        return builder.AddAspNetCoreInstrumentation();
+        builder.AddAspNetInstrumentation();
```
zacharycmontoya, line 58: Does our existing automatic instrumentation for ASP.NET actually work? It [looks like](https://opentelemetry.io/docs/instrumentation/net/netframework/#aspnet-initialization) the user would have to also modify their `web.config` file for automatic instrumentation to occur. I would also expect us to have the same pattern as AspNetCore where we add the sources here and then actually build the instrumentation later

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationBuilders.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace OpenTelemetry.AutoInstrumentation.Loading
+{
+    internal class LazyInstrumentationBuilders
+    {
+        // Add here all of the available detectors
+        private static readonly IReadOnlyDictionary<TracerInstrumentation, Func<AssemblyLoadDetector>> _defaultDetectors
+            = new Dictionary<TracerInstrumentation, Func<AssemblyLoadDetector>>()
+        {
+            { TracerInstrumentation.AspNet, () => new AspNetCoreDetector() }
```
zacharycmontoya, line 29: If we want this to work with ASP.NET on .NET Framework, I think we may also need to add a mapping to a potential `new AspNetDetector()`

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -87,6 +89,7 @@ public static void Initialize()
             {
                 // Initialize SdkSelfDiagnosticsEventListener to create an EventListener for the OpenTelemetry SDK
                 _sdkEventListener = new(EventLevel.Warning, Logger);
+                _lazyInstrumentation = new(new LazyInstrumentationBuilders(TracerSettings));
```
zacharycmontoya, line 92: This looks strange to me. Once you build the `LazyInstrumentationBuilders` and use it in the constructor of the `LazyInstrumentationLoader`, you just ignore it. Could you instead combine all of the logic from `LazyInstrumentationBuilders` into the `LazyInstrumentationLoader` class and only create that object?

src/OpenTelemetry.AutoInstrumentation/Loading/AssemblyLoadDetector.cs
zacharycmontoya: > I'm more concerned if it becomes synchronized

If I'm not mistaken, the event callback for `AppDomain.CurrentDomain.AssemblyLoad` is already synchronized, so if we loop over a list in one handler or invoke all the handlers there shouldn't be a difference. My preference though would be to only have one AssemblyLoad handler that handles all of the "detectors", like what @nrcventura is saying. This could make the design simpler and it could make possible some optimizations (for example, skipping processing of all the handlers if it's a well-known assembly that won't provide integrations)

zacharycmontoya:
This looks to be on the right track and I left some PR comments. I share Robert's concerns about the difficulty in discerning some of the types and the logic, that's the blocker for me right now. I think having only one assembly loader and then refactoring could be a good solution, but I'd like to hear yours and Robert's opinions

src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
RassK: `Web.config` modification is definitely needed, the current asp.net framework instrumentation is covered with integration tests and according to them should work ok. 

The ASP.NET Framework instrumentation actually does not need a detector. Why?
- There are no external types. ASP.NET Framework was built into the core framework. 
- [AspNetInstrumentation.cs](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.AspNet/AspNetInstrumentation.cs)
- [HttpInListener.cs](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.AspNet/Implementation/HttpInListener.cs)
- [OpenTelemetry.Instrumentation.AspNet.csproj](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.AspNet/OpenTelemetry.Instrumentation.AspNet.csproj#L13)

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
RassK: The concrete implementations were separated and moved to `LazyInstrumentationBuilders` to support testing. Source where the modification comes in: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/903#discussion_r911803125

src/OpenTelemetry.AutoInstrumentation/Loading/AssemblyLoadDetector.cs
RassK: Seems you are right. I don't have currently newer reference but expect it's not changed in [AppDomain.cs](https://referencesource.microsoft.com/#mscorlib/system/appdomain.cs,3114).

So my idea would be then also to use one handler only for `AssemblyLoad` event, but the detector then should be changed to async event handling, to make sure that creating the instrumentation instances are not causing delays to each other. What do you think? @zacharycmontoya

pellared:
>  the difficulty in discerning some of the types and the logic, that's the blocker for me right now

It is a blocker for me as well 😢 I try to propose refactorings that would make it easier to read and maintain for me.

>  I think having only one assembly loader

I thought about the same initially, but after reviewing it a few times and my refactoring attempt I think it is OK to have an assembly load per instrumentation library that needs to be initialized in a lazy manner.

src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
zacharycmontoya: Thanks for clarifying, it makes sense that we do not need a detector.

I think what we should do as a new issue is to figure out how to instrument AspNet _without the user changing their web.config_. Currently in our sample application, we don't reference the NuGet package but we do manually reference the HttpModule, which makes the tracing work. Users would ideally not have to touch their web.config to get this working and instead we add the HttpModule automatically for them

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationBuilders.cs
zacharycmontoya: Thanks, resolving

src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
RassK: Yep, that would be the best if possible.

RassK:
Important: Merged @pellared's refactoring. Seems much better to me.

nrcventura:
> **Important: Merged @pellared's refactoring. Seems much better base to me to go forward with.**

The refactor does look clearer to me.

pellared (approved):

pellared:
@rajkumar-rangaraj @zacharycmontoya @nrcventura I think this PR is in pretty good shape for another review.

rajkumar-rangaraj:
Which one of the below would be a convenient approach

1. Changes proposed in this PR to use lazy loader **(or)**
2. Use `Assembly.GetEntryAssembly().GetReferencedAssemblies().Contains("ASP.NET or ASP.NET Core reference")` to add/remove ASPNET instrumentation based on it. 

These approaches are needed only to handle ASPNET instrumentation. If I understand correctly, there are no challenges with other supported tracer instrumentation.

nrcventura:
> Which one of the below would be a convenient approach
> 
> 1. Changes proposed in this PR to use lazy loader **(or)**
> 2. Use `Assembly.GetEntryAssembly().GetReferencedAssemblies().Contains("ASP.NET or ASP.NET Core reference")` to add/remove ASPNET instrumentation based on it.
> 
> These approaches are needed only to handle ASPNET instrumentation. If I understand correctly, there are no challenges with other supported tracer instrumentation.

That is a good question. For ASP.NET and ASP.NET Core in particular, I think that option 2 will work. I don't know how well that approach will scale to other libraries. The approach in this PR is a more generalized approach that comes with some additional complexities, and it may be a good idea to wait for a need for it before we introduce that complexity. While we're still in beta it may be a good idea to experiment with the simpler approach to see just how well it will work in practice.

pellared:
AFAIK only ASP.NET Core. 

>  it may be a good idea to wait for a need for it before we introduce that complexity

I agree 👍 we can always find this PR if we would need the lazy loader approach

RassK:
> Which one of the below would be a convenient approach
> 
> 1. Changes proposed in this PR to use lazy loader **(or)**
> 2. Use `Assembly.GetEntryAssembly().GetReferencedAssemblies().Contains("ASP.NET or ASP.NET Core reference")` to add/remove ASPNET instrumentation based on it.
> 
> These approaches are needed only to handle ASPNET instrumentation. If I understand correctly, there are no challenges with other supported tracer instrumentation.

2. seems also very interesting option. This way we do not need to break (separate) SDK extensions to register particular instrumentation. I see only one drawback, enterprise platforms that are leveraging plugin architecture. This way there is no hard reference to the dependencies. Although I'm pretty sure ASP.NET Core would be never dyn loaded to these platforms. 

I actually would think now similar to @nrcventura. Let's try out the simple approach and fix / change / add if there is a actual need for lazy loader. At least we should have integration tests to verify the actual case.

RassK:
SIG: Closing as the simplified approach seems better candidate. This might be useful in the future.

Issue #893, opened 2022-06-28
MockCollector should expose the allocated Port instead of requiring it


pellared:
It seems that the call to `TcpPortProvider.GetOpenPort()` should actually be done inside the mock collector and the port parameter be a nullable. Most tests would just do like this one and use `agent.Port` as needed.

_Originally posted by @pjanotti in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/892#discussion_r907591976_

PR #910, opened 2022-06-30
Add PluginsTests

pellared:
## Why

Part of #665

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/893

## What

- [Add PluginsTests](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/910/commits/9cf937b2b3c00d9ba0cb1cd60c701d052ddbaac1)
  - SmokeTests additionally verifies `OTEL_DOTNET_AUTO_TRACES_ADDITIONAL_SOURCES`
- [Refactor RunTestApplicationAndWaitForExit](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/910/commits/1ff018ce7afb840b8741a62cc49dcb07e8eda1e2)
- Refactor [MockZipkinCollector](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/910/commits/fe19301e63d87a927f83c5e506ba553f9f6d8cd8) and afterwards [MockCollector](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/910/commits/a29e39acfa16dfb1dafe2504ddeed6611bca1bc2)

## Tests

CI

## Checklist


test/IntegrationTests/Helpers/MockMetricsCollector.cs
```diff
@@ -0,0 +1,174 @@
...
+using System.Text;
+using System.Threading;
+using Google.Protobuf;
+using IntegrationTests.Helpers.Models;
+using Opentelemetry.Proto.Collector.Metrics.V1;
+using Xunit.Abstractions;
+
+namespace IntegrationTests.Helpers;
+
+public class MockMetricsCollector : IDisposable
```
nrcventura, line 32: Right now we use 2 separate collectors for testing traces and metrics. When we start testing traces using OTLP do we think it will be better to configure 2 independent collectors on different ports, or do we think we should use a single listener on the same port? I think that decision affects whether or not we should rename this file. Another alternative, could be using a single OTLP collector but have the code for traces and metrics in separate partial classes.

pellared: I am leaning towards having 2 independent collectors on different ports. I think it will make tests readable. Also, we won't end up with a big "god" collector. It will make also the listeners independent so e.g. a bug in the trace handler will not affect the metrics handler.

pellared:
I will resolve the conflicts tomorrow 😉

test/IntegrationTests/Helpers/MockMetricsCollector.cs
nrcventura: One of the current downsides to having 2 independent collectors is that the OTLP exporter doesn't support telemetry specific environment variable configuration yet. So if we need to support sending the telemetry to 2 different OTLP endpoints we need to implement that support ourselves and then get the SDK to adopt those telemetry specific variables. Since this is something that we know we want to do anyways, it's not a big deal. It's just slightly more scope being added to one of the next 2 releases.

With the refactoring in this PR you already structured the code in such a way that it is easier to share code between the different listeners.

pellared: I think that whatever way we go regarding OTLP mock collector, that this refactoring is still beneficial. It can be used e.g. to create a MockJaegerCollector.

test/IntegrationTests/Helpers/TestHttpListener.cs
```diff
@@ -0,0 +1,120 @@
...
+    private readonly HttpListener _listener;
+    private readonly Thread _listenerThread;
+
+    public TestHttpListener(ITestOutputHelper output, Action<HttpListenerContext> requestHandler, string host = "localhost", string sufix = "/")
+    {
+        _output = output;
+        _requestHandler = requestHandler;
+
+        // try up to 5 consecutive ports before giving up
+        int retries = 4;
```
rajkumar-rangaraj, line 37: Good catch :smile:

rajkumar-rangaraj (approved):
LGTM.

This PR covers lot of changes. To make the review easier, going forward we might need to recommend smaller PRs.

test/IntegrationTests/PluginsTests.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+    }
+
+    [Fact]
+    [Trait("Category", "EndToEnd")]
+    public void SubmitsTraces()
+    {
+        SetEnvironmentVariable("OTEL_DOTNET_AUTO_TRACES_PLUGINS", "TestApplication.Plugins.Plugin, TestApplication.Plugins, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null");
+
+        int agentPort = TcpPortProvider.GetOpenPort();
```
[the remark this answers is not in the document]
pellared: Addressed https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/910/commits/0e181cb1eec48766475b792757bdbc6333eab546

nrcventura (approved):

PR #931, opened 2022-07-12
Add doc for instrumentation library support discussion

nrcventura:
## Why

The goal is to identify a list of libraries that we want to support out of the box so that we can cover a large enough default set of libraries for our users.

## What

Adds a document so that we can discuss and share our instrumentation library support plans. 

I took an initial pass at ordering the libraries based on reported usage from https://insights.stackoverflow.com/survey/2021#most-popular-technologies-database and some internal usage data I could find.

I am especially interested in feedback from other vendors that have access to some library usage data, so that we can either confirm some consistency and/or identify other libraries that are being used across a "large" number of users or customers.

## Checklist


nrcventura:
The libraries in the ordered lists are in an approximate order of usage. 

Some of the libraries are already supported by this project, while others either need to be reviewed to determine if instrumentation already exists or is even necessary or desired.

docs/instrumentation-libraries.md
```diff
@@ -0,0 +1,99 @@
...
+1. Log4net
+1. Microsoft.Extensions.Logging
+1. Serilog
+1. NLog
+1. Common.Logging
+
+### Other
+
+1. AWSSDK
+1. Polly
```
pellared, line 87:
```suggestion
1. Microsoft.Extensions.Caching
```

docs/instrumentation-libraries.md
```diff
@@ -0,0 +1,99 @@
...
+1. Microsoft.extensions.caching
+1. System.Threading.Channels
```
pellared, line 89: I propose to move it to "Libraries we should not support"

Channels are low-level API. For me, it would be similar to instrumenting usage of e.g. AutoResetEvent, locks, async-await.

pellared (commented):
Thanks 👍

docs/instrumentation-libraries.md
```diff
@@ -0,0 +1,99 @@
+# Support for OpenTelemetry Instrumentation Libraries
```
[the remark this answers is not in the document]
nrcventura: `docs/internal` might be a good place for it. I see this document as more of a temporary strategy and roadmap planning document. I'm trying to see if a markdown doc in github makes it easier to collaborate and comment on these things. I think that this document should refer to the other document for libraries that we already support, and I should be able to do that later today.

docs/instrumentation-libraries.md
```diff
@@ -0,0 +1,99 @@
...
+1. MariaDB
+1. Firebase
+1. Elasticsearch
+1. Oracle
+1. DynamoDB
+1. Cassandra
+1. IBM DB2
+1. Couchbase
+
+### External request libraries (REST/RPC)
```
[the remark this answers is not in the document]
nrcventura: If we rename it to IPC, should the message queuing libraries be included there too?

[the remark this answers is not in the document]
zacharycmontoya: IPC is so broad that I would rather break this down into the following categories:

1. Http Clients
1. Messaging
1. RPC

docs/instrumentation-libraries.md
```diff
@@ -0,0 +1,99 @@
...
+
+1. Log4net
+1. Microsoft.Extensions.Logging
+1. Serilog
+1. NLog
+1. Common.Logging
+
+### Other
+
+1. AWSSDK
```
zacharycmontoya, line 86: `AWSSDK` is pretty large that I suggest making it into its own category, and later we could list individual components of the SDK. What do you think?

docs/instrumentation-libraries.md
```diff
@@ -0,0 +1,99 @@
...
+1. Redis
+    * StackExchange.Redis
+1. MariaDB
+1. Firebase
+1. Elasticsearch
+1. Oracle
+1. DynamoDB
+1. Cassandra
+1. IBM DB2
+1. Couchbase
```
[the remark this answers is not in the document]
nrcventura: @zacharycmontoya do you think that there's enough CosmosDb usage to make it on the roadmap?

docs/instrumentation-libraries.md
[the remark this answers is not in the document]
nrcventura: WCF can be used for REST, but I consider it more of a RPC framework. Should it be listed under RPC? I'm also wondering if grpc and WCF should be listed under frameworks because of the potential for running them in a self-hosted fashion instead of relying on ASP.NET or AspNetCore?

docs/instrumentation-libraries.md
nrcventura: It is large, and perhaps it makes sense to have a section related to cloud integrations, where we would put things like cloud-specific resource detectors and cloud-sdk instrumentations. Many of the SDKs will likely be automatically instrumented by either http client or grpc instrumentations. So I think it might make more sense to consider cloud-integrations with a more wholistic approach, especially since there are things like the [AWS distribution of OpenTelemetry](https://aws.amazon.com/otel/?otel-blogs.sort-by=item.additionalFields.createdDate&otel-blogs.sort-order=desc) to consider.

docs/instrumentation-libraries.md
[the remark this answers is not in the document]
rajkumar-rangaraj: cosmosdb has more usage and its better to add it to the roadmap.

docs/instrumentation-libraries.md
```diff
@@ -0,0 +1,99 @@
...
+1. OWIN
+1. Azure Service Fabric
+
+### Databases
+
+#### General database frameworks
+
+1. ADO.NET
+1. EntityFramework
+1. Dapper
```
reyang, line 34: Consider adding links, I assume Dapper here refers to https://github.com/DapperLib/Dapper?

docs/instrumentation-libraries.md
```diff
@@ -0,0 +1,99 @@
...
+1. System.Data.LINQ
```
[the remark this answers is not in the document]
reyang: Should this be `Linq` instead of `LINQ`? https://docs.microsoft.com/en-us/dotnet/api/system.data.linq

docs/instrumentation-libraries.md
nrcventura: Yes, and that's a good idea.

docs/instrumentation-libraries.md
```diff
@@ -0,0 +1,99 @@
...
+1. Microsoft.Practices.EnterpriseLibrary.Data
+1. EntityFrameworkCore
```
reyang, line 37:
```suggestion
1. [Microsoft.EntityFrameworkCore](https://www.nuget.org/packages/Microsoft.EntityFrameworkCore)
```

docs/instrumentation-libraries.md
```diff
@@ -0,0 +1,99 @@
...
+
+#### Specific Databases
+
+1. MySQL
```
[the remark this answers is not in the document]
reyang: It's not clear what does "MySQL" mean. Is it a particular client library, or any libraries that interact with any version of MySQL?

docs/instrumentation-libraries.md
```diff
@@ -0,0 +1,99 @@
+# Support for OpenTelemetry Instrumentation Libraries
+
+OpenTelemetry instrumentation are implemented using a variety of techniques.
+Sometimes the instrumentation is built directly into the library being
+instrumented, sometimes the instrumentation requires the addition of a separate
+instrumentation library, and other times it is a hybrid of those approaches. This
```
reyang, line 6: Add links to the spec where these terms are defined. For example https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md#instrumentation-library.

docs/instrumentation-libraries.md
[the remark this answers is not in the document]
nrcventura: Before we can clarify that, I think we'll need to decide what supporting a particular database means. Does that mean we support specific driver versions, and we have tests focused on validating support of those driver versions? Or, does that mean we have tests, using specific drivers in addition to specific databases? In the short-term it may be easier to focus on just the driver-side of things, but in the long term we may need to also have some emphasis on testing and listing compatibility for databases too.

docs/instrumentation-libraries.md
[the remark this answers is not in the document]
zacharycmontoya: I think so. Even if it's a later phase of our roadmap, I think it should be there

docs/instrumentation-libraries.md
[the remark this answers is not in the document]
zacharycmontoya: Yes I consider both WCF and GRPC and RPC frameworks and consider them separately from ASP.NET and AspNetCore

docs/instrumentation-libraries.md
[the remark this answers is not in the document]
reyang: +1

docs/instrumentation-libraries.md
pellared: I put this comment by mistake (sic!). Sorry

docs/instrumentation-libraries.md
```diff
@@ -0,0 +1,121 @@
...
+## Libraries and our instrumentation support plans
+
+The libraries listed here contain our current thoughts about which libraries should
+include out of the box support, and which libraries we decided to not support. If
+you have a library that you would like to consider adding to this list, please
+submit an issue to request its inclusion.
+
+### Application Frameworks
+
+| Framework | Tracing Support | Metrics Support | Comments |
```
nrcventura, line 24: I think that we can use the Tracing and Metrics Support columns to also include links to the issues that can be used to track the future support of that library/framework.

docs/instrumentation-libraries.md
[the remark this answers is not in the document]
pellared: How about "IPC" and putting the messaging libraries here as well?

docs/instrumentation-libraries.md
```diff
@@ -0,0 +1,99 @@
...
+1. MySQL
+    * MySql.data
+    * MySqlConnector
+1. PostgreSQL
+1. SQLite
+1. MongoDB
+1. MS SQL Server
+1. Redis
+    * StackExchange.Redis
+1. MariaDB
```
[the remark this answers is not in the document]
pellared: We are instrumenting libraries, therefore I think we should say that we instrument an ADO.NET driver library. We can always add notes on what the instrumented library is for. E.g. it may be not obvious that `Npgsql` is a driver for PostgreSQL.

docs/instrumentation-libraries.md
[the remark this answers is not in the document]
pellared: We are instrumenting libraries, therefore I think we should say that we instrument an ADO.NET driver library. We can always add notes on what the instrumented library is for. For example, it may be not obvious that `Npgsql` is a driver for PostgreSQL.

docs/instrumentation-libraries.md
```diff
@@ -0,0 +1,121 @@
...
+The libraries listed here contain our current thoughts about which libraries should
+include out of the box support, and which libraries we decided to not support. If
+you have a library that you would like to consider adding to this list, please
+submit an issue to request its inclusion.
+
+### Application Frameworks
+
+| Framework | Tracing Support | Metrics Support | Comments |
+| --- | :---: | :---: | --- |
+| [ASP.NET](https://docs.microsoft.com/en-us/aspnet/overview) | [Yes](./config.md#instrumented-traces-libraries-and-frameworks) | [Yes](./config.md#instrumented-metrics-libraries-and-frameworks) | |
```
pellared, line 26: I suggest to put the hyperlinks to `config.md#instrumented-traces-libraries-and-frameworks` and `config.md#instrumented-metrics-libraries-and-frameworks` in the "header column" instead

docs/instrumentation-libraries.md
pellared, line 24: Maybe `Notes` instead of `Comments`?
Shorter and this is what OTel spec has often in the last column

docs/instrumentation-libraries.md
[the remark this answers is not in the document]
nrcventura: > We are instrumenting libraries, therefore I think we should say that we instrument an ADO.NET driver library. We can always add notes on what the instrumented library is for. For example, it may be not obvious that `Npgsql` is a driver for PostgreSQL.

You are right, when I convert this section to use a table, I can make the driver/library the focus, but I think it will be important to call out the databases we validate against, and a note about the relevant databases for that driver/library. When supporting instrumentation, I usually get asked the question "do you support this database?" instead of "do you support this driver?". I do not know if that is a result of the original question changing by the time it reaches engineering, or if it is related to New Relic's explain plan support (which requires database and driver specific variations).

docs/instrumentation-libraries.md
nrcventura: See https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/931/commits/413ddefe5975877b141d044b6eff5329c002b1d8

docs/instrumentation-libraries.md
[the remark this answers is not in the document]
nrcventura: Do the changes in [ea07f30](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/931/commits/ea07f30c16de3e510b096e5484d804b4028ed4d9) make things a little clearer?

docs/instrumentation-libraries.md
```diff
@@ -0,0 +1,117 @@
...
+1. Grpc
+1. GraphQL
+1. WCF Core
+
+### Messaging
+
+1. RabbitMQ
+1. MSMQ
+1. Kafka
+1. NServiceBus
```
RassK, line 67: Should we add here the open source competitor also: [MassTransit](https://www.nuget.org/packages/MassTransit/8.0.6-develop.512)

nrcventura: It looks like nuget usage is high enough, and there is an instrumentation library already.

docs/internal/instrumentation-libraries.md
```diff
@@ -0,0 +1,117 @@
...
+instrumentation libraries to support.
+
+### Libraries we should not support
+
+| Library | Notes |
+| --- | --- |
+| [System.IO.Pipelines](https://docs.microsoft.com/en-us/dotnet/standard/io/pipelines) | This library is used for high-throughput scenarios and instrumenting this library could potentially impact the performance of the application in a negative manner. |
+| [Pipelines.Sockets.Unofficial](https://github.com/mgravell/Pipelines.Sockets.Unofficial) | This is a lower-level networking library and instrumenting it can affect the performance of an application in negative ways causing an increase in network traffic (as compared to the application running without instrumentation). |
+| [System.Threading.Channels](https://docs.microsoft.com/en-us/dotnet/api/system.threading.channels) | This is a lower-level library and may not provide enough contextual information on its own. |
+| [Cassandra](https://www.nuget.org/packages/CassandraCSharpDriver) | Lower usage |
```
pellared, line 110: Should it be `Low usage` instead of `Lower usage`? If so then the same applies to other places.

pellared (approved):

docs/instrumentation-libraries.md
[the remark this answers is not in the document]
pellared: Awesome

docs/internal/instrumentation-libraries.md
```diff
@@ -0,0 +1,117 @@
...
+TODO: Determine appropriate resource detectors and GCP SDK related
+instrumentation libraries to support.
+
+### Libraries we should not support
+
+| Library | Notes |
+| --- | --- |
+| [System.IO.Pipelines](https://docs.microsoft.com/en-us/dotnet/standard/io/pipelines) | This library is used for high-throughput scenarios and instrumenting this library could potentially impact the performance of the application in a negative manner. |
+| [Pipelines.Sockets.Unofficial](https://github.com/mgravell/Pipelines.Sockets.Unofficial) | This is a lower-level networking library and instrumenting it can affect the performance of an application in negative ways causing an increase in network traffic (as compared to the application running without instrumentation). |
+| [System.Threading.Channels](https://docs.microsoft.com/en-us/dotnet/api/system.threading.channels) | This is a lower-level library and may not provide enough contextual information on its own. |
```
reyang, line 109: nit:
```suggestion
| [System.Threading.Channels](https://docs.microsoft.com/dotnet/api/system.threading.channels) | This is a lower-level library and may not provide enough contextual information on its own. |
```

reyang (approved):

rajkumar-rangaraj (approved):
nit: There are several places in the document where link has `/en-us` which needs to be removed.. This breaks the localization read for users.

pellared:
> nit: There are several places in the document where link has `/en-us` which needs to be removed.. This breaks the localization read for users.

Personally, I never use the localization (at least the Polish one). I know literally nobody from Poland who really uses it.

docs/internal/instrumentation-libraries.md
```diff
@@ -0,0 +1,117 @@
...
+| --- | :---: | :---: | --- | --- |
+| [Entity Framework](https://docs.microsoft.com/ef/ef6/) | | | | Needs investigation. It might be implicitly supported based on the configured [database provider](https://docs.microsoft.com/ef/ef6/fundamentals/providers/). |
+| [Entity Framework Core](https://docs.microsoft.com/ef/core/) | | | | Needs investigation. It might be implicitly supported based on the configured [database provider](https://docs.microsoft.com/ef/core/providers/). |
+| [Dapper](https://github.com/DapperLib/Dapper) | | | | Needs investigation. It might be implicitly supported based on the configured [database provider](https://github.com/DapperLib/Dapper#will-dapper-work-with-my-db-provider). |
+| [Microsoft.Data.SqlClient](https://www.nuget.org/packages/Microsoft.Data.SqlClient) | Yes | | [MS SQL Server](https://www.microsoft.com/sql-server/sql-server-2019) | |
+| [System.Data.SqlClient](https://www.nuget.org/packages/System.Data.SqlClient) | Yes | | | |
+| [MySql.Data](https://dev.mysql.com/doc/connector-net/en/connector-net-introduction.html) | | | | This is the official [MySQL](https://dev.mysql.com/) library. |
```
nrcventura, line 40: I did not figure out how to get a non-localized url for this link.

docs/internal/instrumentation-libraries.md
```diff
@@ -0,0 +1,117 @@
...
+| [MySqlConnector](https://mysqlconnector.net/) | | | | Seems to be the [recommended library for MariaDB](https://mariadb.com/kb/en/mysqlconnector-for-adonet/). |
```
nrcventura, line 41: I did not figure out how to get a non-localized url for this link.

pellared (approved):

PR #932, opened 2022-07-13
Enable all trace instrumentations by default (simplified)

RassK:
## Why

Part of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/388
Simplification of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/903

## What

* Enables all trace instrumentations by default
* Checks for ASP.NET Core framework presence

## Tests

Removed `OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS` config from all of the integration tests (except smoke test).

## Checklist


src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
```diff
@@ -56,10 +57,14 @@ public static TracerProviderBuilder UseEnvironmentVariables(this TracerProviderB
     public static TracerProviderBuilder AddSdkAspNetInstrumentation(this TracerProviderBuilder builder)
     {
 #if NET462
-        return builder.AddAspNetInstrumentation();
-#elif NETCOREAPP3_1_OR_GREATER
-        return builder.AddAspNetCoreInstrumentation();
+        builder.AddAspNetInstrumentation();
 #endif
+        if (Assembly.GetEntryAssembly().GetReferencedAssemblies().Any(x => x.Name == "Microsoft.AspNetCore"))
```
RassK, line 62: Notice: We cannot detect `Microsoft.AspNetCore.Http` dll directly as it's probably a subdependency of a dependency. I think it's reasonable to assume, if `Microsoft.AspNetCore` is present then `Microsoft.AspNetCore.Http` is also present.

src/OpenTelemetry.AutoInstrumentation/Configuration/TracerSettings.cs
```diff
@@ -31,7 +31,7 @@ public class TracerSettings : Settings
     /// using the specified <see cref="IConfigurationSource"/> to initialize values.
     /// </summary>
     /// <param name="source">The <see cref="IConfigurationSource"/> to use when retrieving configuration values.</param>
-    private TracerSettings(IConfigurationSource source)
+    public TracerSettings(IConfigurationSource source)
```
pellared, line 34: why has it changed to `public`?

pellared (approved):
Can you describe what are the next steps?

Guessing:

- support `none` as value to explicitly disable source instrumentations
- do the same for metrics instrumentations

src/OpenTelemetry.AutoInstrumentation/Configuration/TracerSettings.cs
RassK: Seems leftover from #903

src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
RassK: Another important one: `Assembly.GetEntryAssembly()` returns null if application is created in unmanaged process eg IIS. Wondering if the same happens when ASP.NET Core is inproc hosted in IIS.

RassK: One way is to cut off the feature: run ASP.NET Core with .NET Framework.

rajkumar-rangaraj: To identify ASP.NET, we could use `Assembly.GetAssembly()` API and check for `System.Web.dll`. Similar check is needed for ASP.NET MVC/Web API.

src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
```diff
@@ -56,10 +56,15 @@ public static TracerProviderBuilder UseEnvironmentVariables(this TracerProviderB
     public static TracerProviderBuilder AddSdkAspNetInstrumentation(this TracerProviderBuilder builder)
     {
 #if NET462
-        return builder.AddAspNetInstrumentation();
+        builder.AddAspNetInstrumentation();
```
nrcventura, line 59: Is this safe to do without some sort of dependency check?

src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
RassK: > To identify ASP.NET, we could use `Assembly.GetAssembly()` API and check for `System.Web.dll`. Similar check is needed for ASP.NET MVC/Web API.

As far as I understand Assembly.GetAssembly() etc requires assembly to be loaded. We need to know before if the assembly exists / will be loaded / is loadable, because the StartupHook runs early.

rajkumar-rangaraj: ASP.NET does not use StartupHook, by the time we reach this code all the ASP.NET libraries are loaded and we could use this approach for .NET Framework. 

But, I agree to your comment - https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/932#discussion_r920326457 we may not need to do anything special here.

nrcventura (approved):

rajkumar-rangaraj (approved):

PR #941, opened 2022-07-15
Remove OTEL_DOTNET_AUTO_DOMAIN_NEUTRAL_INSTRUMENTATION

pellared:
## Why

1. From the description it looked like it was needed prior .NET Framework 4.5.2
1. Looks like the code was NOP anyway

## What

Remove `OTEL_DOTNET_AUTO_DOMAIN_NEUTRAL_INSTRUMENTATION` configuration.

## Tests

CI

## Checklist


pellared:
@zacharycmontoya PTAL as I saw you were originally implementing it. Maybe it would be possible to remove more code?

Kielek:
I like the idea of removing redundant stuff, the question is if it is easier to do it now or we should wait for updating native code from the DD-upstraem.

pellared:
> I like the idea of removing redundant stuff

It is not even redundant, but not needed/used.

> the question is if it is easier to do it now or we should wait for updating native code from the DD-upstraem

Waiting is always easier 😆 But I do not know when the update will happen. Additionally, I think that updating native code would be a lot harder than this PR.

rajkumar-rangaraj (approved):

Kielek (approved):

zacharycmontoya:
All that being said, we can likely also remove the implementation of `CorProfiler::GetAssemblyReferences`

PR #950, opened 2022-07-19
MySql.Data intrumentation (.NET Core 3.1+)

Kielek:
## Why

Fixes #907

## What

MySql.Data instrumentation for .NET Core 3.1+.
It utilize `OpenTelemetry.Instrumentation.MySqlData`.
Lazy loading based on: #903.

Support for .NET Framework has similar issues like MongoDB. It is candidate for separate task.

## Tests

New test created.
CI
Tested locally also with `MySql.Data` v. 8.0.29.

## Checklist

- [x] `CHANGELOG.md` is updated.
- [x] Documentation is updated.
- [x] New features are covered by tests.

## Code review

Please double check the way how `OpenTelemetry.Instrumentation.MySqlData` is referenced.

PR #967, opened 2022-07-21
[LazyInstrumentationLoader] Refactor OnAssemblyLoadInitializer

Kielek:
## Why

Follow up to #950 

Fixes N/A

## What

Change the way how we check if all required libraries were loaded.

## Tests

CI

## Checklist


src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
```diff
@@ -63,8 +63,8 @@ private class OnAssemblyLoadInitializer
         private static readonly ILogger Logger = OtelLogging.GetLogger();
         private readonly InstrumentationInitializer _instrumentationInitializer;
         private readonly LazyInstrumentationLoader _manager;
+        private readonly object _requiredAssembliesLocker = new();
```
pellared, line 66: It would be fine to lock on `_requiredAssemblies`.
We would gain a little memory + it would be more clear that `_requiredAssemblies` is the shared resource.

Reference: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/statements/lock#guidelines

> lock on [...] instance that is unlikely to be used as a lock object by unrelated parts of the code

Kielek: Fixed.

pellared (approved):

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
```diff
@@ -80,21 +79,29 @@ private void CurrentDomain_AssemblyLoad(object sender, AssemblyLoadEventArgs arg
 
             if (_requiredAssemblies.Contains(assemblyName))
             {
-                if (Interlocked.Increment(ref _loadedCount) == _requiredAssemblies.Count)
+                lock (_requiredAssemblies)
```
rajkumar-rangaraj, line 82: Is there a way to avoid this lock? In practice, we should avoid locks in startup code. If there are no options we will consider this approach.

nrcventura: Switching to either a ConcurrentBag or ConcurrentDictionary might help, but we probably we need to run a microbenchmark to really understand how the performance compares. But if you're concern is about using a lock in general during startup, even those collections have some locking behind the scenes (they just try to minimize lock usage).

rajkumar-rangaraj: I'm worried about some kind of contention. .NET Assembly loader uses a loader lock. We acquire lock, on top of it when loading an assembly another lock is applied in the background. This should not cause an issue. Want to be cautious to avoid future issues.

nrcventura:
Given the concerns about introducing a lock, what is driving the need for this change? Was a particular problem discovered with the assembly counting approach, or is that problem still hypothetical?

Kielek:
> Given the concerns about introducing a lock, what is driving the need for this change? Was a particular problem discovered with the assembly counting approach, or is that problem still hypothetical?

PR is based on conversation with @pellared.

I have found 3 reasons to do this
* it should cover https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/903#discussion_r910189324
* [personal preferences] easier debugging, if needed,
* [personal preferences] more readable code.

If we are talking about lock. It is inside the if statement `_requiredAssemblies.Contains(assemblyName)`.
As I understand the problem, in typical scenario we will not touch the lock. We will just return from the method without any impact.

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
nrcventura: @rajkumar-rangaraj I don't have a lot of knowledge about the assembly loader lock, so I don't know if it applies to the assembly load event, or if it means that we don't need the extra lock, because we're already in a "locked" context. Do you know more about that loader lock?

nrcventura: We might be able to use a lock-less design if we change `_requiredAssemblies` from a `HashSet` to a custom type similar to the following (that get's passed in to the LazyInstrumentationLoader constructor).

```csharp
public class ExampleRequiredAssembliesImplementation : IRequiredAssemblies
{
    private volatile bool _dependency1Loaded = false;
    private volatile bool _dependency2Loaded = false;
    
    public bool AllRequiredAssembliesLoaded
    {
        return _dependency1Loaded && _dependency2Loaded;
    }
    
    public bool NotifyAssemblyLoaded(string assemblyName)
    {
        if (assemblyName == "Dependency1")
        {
            _dependency1Loaded = true;
            return true;
        }
        
        if (assemblyName == "Dependency2")
        {
            _dependency2Loaded = true;
            return true;
        }
        
        return false;
    }
}
```

We would need to define a new type for each set of required assemblies, but it would prevent us from requiring another lock to synchronize access to what should be a small collection. If the volatile read operation is not enough we can also change the private fields to integers and use something like `return Interlocked.Exchange(_dependency1Loaded, 1) == 0` to replace the separate assignment and return so that `true` is only returned when we actually mark that dependency as being loaded.

Kielek: I really like this idea, even if a lot of boilerplate code is needed.

5f529aada1ae7d5e07e63d8d7d43cc97788c4a5a

src/OpenTelemetry.AutoInstrumentation/Loading/IRequiredAssemblies.cs
```diff
@@ -0,0 +1,26 @@
...
+// </copyright>
+
+#if NETCOREAPP3_1_OR_GREATER
+namespace OpenTelemetry.AutoInstrumentation.Loading;
+
+internal interface IRequiredAssemblies
+{
+    bool AllRequiredAssembliesLoaded { get; }
+
+    bool NotifyAssemblyLoaded(string assemblyName);
```
pellared, line 24: how about simply

```csharp
    internal interface IRequiredAssemblies
    {
        bool AllAssemblyLoaded(string currentAssemblyName);
     }

    private class MySqlDataRequiredAssemblies : IRequiredAssemblies
    {
        private volatile bool _mySqlDataLoaded;

        public bool AllAssemblyLoaded(string currentAssemblyName)
        {
            if (currentAssemblyName == "MySql.Data")
            {
                _mySqlDataLoaded = true;
            }

            return _mySqlDataLoaded;
        }
    }
```

or even

    private class SingleAssemblyRequired: IRequiredAssemblies
    {
        private volatile bool _mySqlDataLoaded;
        private readonly string _requiredAssemblyName;

        SingleAssemblyRequired(string requiredAssemblyName)
        {
            _requiredAssemblyName = requiredAssemblyName;
        }

        public bool AllAssemblyLoaded(string currentAssemblyName)
        {
            if (currentAssemblyName == _requiredAssemblyName)
            {
                _mySqlDataLoaded = true;
            }

            return _mySqlDataLoaded;
        }
    }
```

File: src/OpenTelemetry.AutoInstrumentation/Loading/MySqlDataInitializer.cs [hunk 4]
```diff
@@ -37,5 +37,26 @@ public override void Initialize(ILifespanManager lifespanManager)
 
         lifespanManager.Track(instrumentation);
     }
+
+    private class MySqlDataRequiredAssemblies : IRequiredAssemblies
```
pellared, line 41: 1. From the reader's perspective is less safe. Are we sure that it is not possible to load two times the same assembly in the same domain  (e.g. if one loads `"MySql.Data"` via reflection)? If so then we do not even need any state -> we just load the instrumentation when `assemblyName == "MySql.Data"` and then remove the handler. BTW both implementation using `lock` and `Interlocked.Increment` were protected from such a scenario.
1. Do we expect that we will require more assemblies than one?

Maybe we could simply change `public IRequiredAssemblies RequiredAssemblies` to `public string RequiredAssemblyName` in`InstrumentationInitializer`. This would allow us the get rid of `MySqlDataRequiredAssemblies` and `IRequiredAssemblies`. It should reduce the complexity and improve performance a little (interfaces bring a _little_ overhead and we would have fewer objects).

nrcventura: > 1. Are we sure that it is not possible to load two times the same assembly in the same domain  (e.g. if one loads `"MySql.Data"` via reflection)?

According to https://docs.microsoft.com/en-us/dotnet/framework/deployment/best-practices-for-assembly-loading it is possible to load an assembly more than once. I think those scenarios are edge cases, but they are possible.

> 2\. Do we expect that we will require more assemblies than one?

I don't know if we have a need for that yet. If we do not have a need for that yet, then we should probably use one of the simpler approaches that you recommended.

pellared (approved):

nrcventura (approved):

PR #972, opened 2022-07-21
Rename Tests Collection and Fixture types

Kielek:
## Why

Fixes N/A
Better namings for test classes.

## What

`NpgqslCollection` ~> `PostgresCollection`.
Test itself should be called NpgqslTests. We are testing library, but we are using Postgres as a docker.

## Tests

CI

## Checklist

### DIFF
--- a/test/IntegrationTests/MySqlCollection.cs
+++ b/test/IntegrationTests/MySqlCollection.cs
@@ -1,4 +1,4 @@
-// <copyright file="MySqlDataCollection.cs" company="OpenTelemetry Authors">
+// <copyright file="MySqlCollection.cs" company="OpenTelemetry Authors">
 // Copyright The OpenTelemetry Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


pellared (commented):
From a logical perspective, this PR makes sense. But I think that using `NpgsqlCollection` makes it more maintainable b/c it is easier to find as it follows the pattern `xyzTests` and `xyzCollection`.

I would prefer to keep the existing naming but I am not blocking this PR as well 😉

Kielek:
You can expect more similar differences: `StackExchangeRedisTests` and `RedisCollection`, etc.

I think, in this PR we should agree on the convention.

Each test contains reference to the collection, so it is pretty easy to map it.

BTW I have similar doubts related to this naming.

pellared:
If we change `SqlClientCollection` to `SqlServerCollection` then I am fine with this PR 😉

pellared:
I think we could consider also renaming the `xyzCollection.cs` filenames to `xyzCollectionFixture.cs` as 

1. they contain both the `ICollectionFixture` and `Fixture` implementation
2. it would indicate that it is a "CollectionFixture" :smile:

pellared (approved):

Issue #969, opened 2022-07-21
Testing via dependabot


pellared:
> Well, for now we are using the oldest dependency. In general we should have at least two cases, the oldest, the newest (and probably some other minor/major releases).
> 
> I would stay with the oldest supported version. For each new, dependabot will create PR and verify CI.

👍 I think this "testing via dependabot strategy" could be documented in our docs.

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/962#issuecomment-1191120782_

pellared:
We could add `do-not-merge` label in PRs created under `/test-applications`

PR #979, opened 2022-07-25
StackExchange.Redis instrumentation for .NET Core 3.1+

### LINKED ISSUE #947: Add StackExchange.Redis instrumentation
Add `StackExchange.Redis` source&bytecode instrumentation using https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.StackExchangeRedis

Kielek:
## Why

Fixes #947 

## What

StackExchange.Redis instrumentation for .NET Core 3.1+

## Tests

CI test.
Tested also locally on [StackExchange.Redis 2.6.48](https://www.nuget.org/packages/StackExchange.Redis/2.6.48) and [StackExchange.Redis 2.0.495](https://www.nuget.org/packages/StackExchange.Redis/2.0.495)

## Checklist


integrations.json
```diff
@@ -105,5 +105,58 @@
         }
       }
     ]
+  },
```
Kielek, line 108: I do not found any script to refresh this file in case of changes in binary instrumentation. Do we have anything like this?

src/OpenTelemetry.AutoInstrumentation/Loading/ILifespanManager.cs
```diff
@@ -25,6 +25,6 @@ internal interface ILifespanManager
     /// it will be disposed together with the manager.
     /// </summary>
     /// <param name="instance">Trackable object</param>
-    void Track(object instance);
+    public void Track(object instance);
```
pellared, line 28: I think that `public` is the only access modifier acceptable inside an `interface`.
I do not recall ever setting it explicitly (at least in this repo).

pellared (commented):
I think it is worth to update the dependabot config. More: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/969

src/OpenTelemetry.AutoInstrumentation/Loading/ILifespanManager.cs
Kielek: d331a754cc93a64088f40a690b911744384ae2f6

Kielek:
> I think it is worth to update the dependabot config. More: #969

d331a754cc93a64088f40a690b911744384ae2f6

src/OpenTelemetry.AutoInstrumentation/Instrumentations/StackExchangeRedis/StackExchangeRedisConstants.cs
```diff
@@ -0,0 +1,37 @@
...
+
+#if NETCOREAPP3_1_OR_GREATER
+
+namespace OpenTelemetry.AutoInstrumentation.Instrumentations.StackExchangeRedis;
+
+internal static class StackExchangeRedisConstants
+{
+    public const string AssemblyName = "StackExchange.Redis";
+
+    public const string MinimumVersion = "2.0.0"; // 2.1.58, but dll uses 2.0.0
```
pellared, line 25: Have you tested the behavior when `StackExchange.Redis` version is `< 2.1.58`?

pellared: I think you can also consider creating an issue in StackExchange.Redis`https://github.com/StackExchange/StackExchange.Redis
 See: https://github.com/StackExchange/StackExchange.Redis/blob/74f4c9a0d39b822f67e50e851d764f68d540f304/tests/ConsoleTest/Program.cs#L47

dszmigielski (approved):
LGTM

src/OpenTelemetry.AutoInstrumentation/Instrumentations/StackExchangeRedis/StackExchangeRedisConstants.cs
Kielek: https://github.com/StackExchange/StackExchange.Redis/pull/2197

mgravell: To echo my comment on the issue:

>This is not accidental, and is a pattern pretty common among a wide range of libraries. The reality is: having a granular assembly-version places a very large burden on consumers, in particular (but not quite exclusively) on .NET Framework - mostly due to assembly binding redirect problems. Having this break constantly creates unnecessary friction, which is often avoided by pinning the assembly version to majors. Or at a push: major+minor.
> 
> My suggestion here would be for OpenTelemetry AutoIntrumention should be tweaked to be a little sympathetic to this intent, perhaps considering other attributes.

nrcventura: That assembly version pattern is common, and is actually listed in some [.net versioning guidance documentation](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/versioning).

integrations.json
```diff
@@ -105,5 +105,58 @@
...
+          "type": "StackExchange.Redis.ConnectionMultiplexer",
+          "method": "Connect",
+          "signature_types": [
+            "StackExchange.Redis.ConnectionMultiplexer",
+            "StackExchange.Redis.ConfigurationOptions",
+            "System.IO.TextWriter"
+          ],
+          "minimum_major": 2,
+          "minimum_minor": 0,
+          "minimum_patch": 0,
```
nrcventura, line 125: Do we know if this could actually be set to the minimum version that we care about? Does the native code see a different version number than the managed code?

test/test-applications/integrations/TestApplication.StackExchangeRedis/Properties/launchSettings.json
```diff
@@ -0,0 +1,23 @@
+{
+  "profiles": {
+    "TestApplication.StackExchangeRedis": {
+      "commandName": "Project",
+      "commandLineArgs": "--mongodb 5295",
```
nrcventura, line 5: We can probably clean up all of the mongodb references from this file, including the additional sources variable.

nrcventura:
One thing that we should consider before enabling this instrumentation library out of the box, is to see how it affects performance of the application. A problem that I have seen with other Redis instrumentation is that when the instrumentation is enabled, it can cause the application to have a significant increase in network usage as compared to when no instrumentation is enabled. This can happen if too much code from the instrumentation is running between the different calls to Redis. If the performance impact is not acceptable, we should consider not enabling this instrumentation by default, or to wait until a later time to support Redis (until we determine a better approach).

test/test-applications/integrations/TestApplication.StackExchangeRedis/Properties/launchSettings.json
Kielek: The file was included by mistake. Removed.

integrations.json
Kielek: As I understand current implementation, we are using `AssemblyVersion` to define ranges.
I do not see any option to narrow this versions with current implementation. We should be able to add validation on another attribute.

Kielek:
> Have you tested the behavior when `StackExchange.Redis` version is `2.0.0 <= ver < 2.1.58`?

It was really good question. It was working only partially. It should be fully supported after https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/979/commits/aad6c78f8f5bcd2bb2a9f6f123f5260ed9cd0848

Some important information
 * [OpenTelemetry.Intrumentation.StackExchangeRedis](https://www.nuget.org/packages/OpenTelemetry.Instrumentation.StackExchangeRedis/1.0.0-rc9.7#dependencies-body-tab) references `StackExchange.Redis` v. 2.1.58+
 * All 2+ StackExchange.Redis` releases are released with AssemblyVersion = 2.0.0.0
 * If the AsemblyVersion is not changed, it can be replaced without any issues - https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/versioning#assembly-version
 
Previously I have tried to instrument public API, but it is much harder. For some versions public methods calls other publics method, so there were a chance to register instrumentation twice.

Follow up candidate - as AssemblyVersion is stable for this library, we should be able to add support also for .NET Framework without any changes (until 3.0 version is released).

Kielek:
> One thing that we should consider before enabling this instrumentation library out of the box, is to see how it affects performance of the application. [A problem](https://github.com/newrelic/newrelic-dotnet-agent/issues/282) that I have seen with other Redis instrumentation is that when the instrumentation is enabled, it can cause the application to have a significant increase in network usage as compared to when no instrumentation is enabled. This can happen if too much code from the instrumentation is running between the different calls to Redis. If the performance impact is not acceptable, we should consider not enabling this instrumentation by default, or to wait until a later time to support Redis (until we determine a better approach).

@nrcventura, could you please share the reproduction app or verify this instrumentation yourself?

I have executed following code in console application with "StackExchange.Redis" Version="2.6.48, redis:7.0.4 hosted on docker on the same machine as console app.

```
        var connectionString = $@"127.0.0.1:{redisPort}";

        using (var connection = await ConnectionMultiplexer.ConnectAsync(connectionString))
        {
            for (int i = 0; i < 500; i++)
            {
                var db = connection.GetDatabase();
                await db.StringGetAsync("zzzzz");
            }
        }
```

Number of captured packets send/received to/from redis port is around 2050. Both for instrumented and normal application.

nrcventura:
> could you please share the reproduction app or verify this instrumentation yourself

The reproduction app was just an aspnetcore app that used the code snippet in the linked issue. To reproduce the issue, we just called the endpoint multiple times, and checked the number of packets. 

> Number of captured packets send/received to/from redis port is around 2050. Both for instrumented and normal application.

It's good that we're not seeing a difference in packets with the code that you used. However, your code is not really attempting to execute these requests in parallel. I think that the easiest way to test the situation, with enough of a natural delay (so that not everything is executed in parallel immediately), is to create an aspnetcore app with an endpoint containing the following code. Then you can just hit the endpoint in a browser and refresh a few times.

```csharp
private static StackExchange.Redis.ConnectionMultiplexer _connection = StackExchange.Redis.ConnectionMultiplexer.Connect(connectionString);

public async Task<string> GetValueAsync(string key)
{
    if (!_connection.IsConnected)
    {
        return null;
    }

    var db = _connection.GetDatabase();
    var val = await db.StringGetAsync(key, flags: CommandFlags.PreferSlave);

    if (val.HasValue)
    {
        return val;
    }

    return null;
}
```

Kielek:
@nrcventura, for the application based on your code, the number packets is on the same level.

BTW application even without any activity is producing a lot of packets related to Keeping connection alive.

pellared (approved):

nrcventura (approved):

rajkumar-rangaraj (approved):

pjanotti (approved):

PR #1022, opened 2022-08-02
Grpc.Net.Client instrumentation

### LINKED ISSUE #477: Add Grpc.Net.Client instrumentation
Add Grpc.Net.Client "source" instrumentation using https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Instrumentation.GrpcNetClient

Kielek:
## Why

Fixes #477

## What

`Grpc.Net.Client` instrumentation.

## Tests

Added to CI.
Tested locally also with 2.47.0

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
@@ -36,7 +36,8 @@ internal static class EnvironmentConfigurationTracerHelper
         [TracerInstrumentation.MySqlData] = builder => builder.AddSource("OpenTelemetry.Instrumentation.MySqlData"),
         [TracerInstrumentation.StackExchangeRedis] = builder => builder.AddSource("OpenTelemetry.Instrumentation.StackExchangeRedis"),
 #endif
-        [TracerInstrumentation.Npgsql] = builder => builder.AddSource("Npgsql")
+        [TracerInstrumentation.Npgsql] = builder => builder.AddSource("Npgsql"),
+        [TracerInstrumentation.GrpcNetClient] = builder => builder.AddGrpcClientInstrumentation(options => options.SuppressDownstreamInstrumentation = !Instrumentation.TracerSettings.EnabledInstrumentations.Contains(TracerInstrumentation.HttpClient))
     };
 
     public static TracerProviderBuilder UseEnvironmentVariables(this TracerProviderBuilder builder, TracerSettings settings)
--- a/test/IntegrationTests/GrpcNetClientTests.cs
+++ b/test/IntegrationTests/GrpcNetClientTests.cs
@@ -39,7 +39,7 @@ public class GrpcNetClientTests : TestHelper
 
         // Grpc.Net.Client is using various version of http communication under the hood.
         // Disabling HttpClient instrumentation to have consistent set of spans.
-        EnvironmentHelper.CustomEnvironmentVariables.Add("OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS", "HttpClient");
+        SetEnvironmentVariable("OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS", "HttpClient");
 
         RunTestApplication(agent.Port);
 
--- a/test/IntegrationTests/GrpcNetClientTests.cs
+++ b/test/IntegrationTests/GrpcNetClientTests.cs
@@ -1,0 +1,56 @@
+// <copyright file="GrpcNetClientTests.cs" company="OpenTelemetry Authors">
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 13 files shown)


test/IntegrationTests/GrpcNetClientTests.cs
```diff
@@ -0,0 +1,56 @@
...
+
+    [Fact]
+    [Trait("Category", "EndToEnd")]
+    public void SubmitsTraces()
+    {
+        using var agent = new MockZipkinCollector(Output);
+
+        // Grpc.Net.Client is using various version of http communication under the hood.
+        // Disabling HttpClient instrumentation to have consistent set of spans.
+        EnvironmentHelper.CustomEnvironmentVariables.Add("OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS", "HttpClient");
```
RassK, line 42:
```suggestion
        SetEnvironmentVariable("OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS", "HttpClient");
```

RassK (approved):

test/test-applications/integrations/TestApplication.GrpcNetClient/Program.cs
```diff
@@ -0,0 +1,58 @@
...
+            HttpHandler = new GrpcWebHandler(new HttpClientHandler())
+        });
+#else
+        var channel = GrpcChannel.ForAddress(uri);
+#endif
+
+        try
+        {
+            var greeterClient = new Greeter.GreeterClient(channel);
+            await greeterClient.SayHelloAsync(new HelloRequest());
```
pjanotti, line 51: Did I miss the server somewhere or are we expecting this to fail?

pjanotti (approved):

src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
```diff
@@ -36,7 +36,8 @@ internal static class EnvironmentConfigurationTracerHelper
         [TracerInstrumentation.MySqlData] = builder => builder.AddSource("OpenTelemetry.Instrumentation.MySqlData"),
         [TracerInstrumentation.StackExchangeRedis] = builder => builder.AddSource("OpenTelemetry.Instrumentation.StackExchangeRedis"),
 #endif
-        [TracerInstrumentation.Npgsql] = builder => builder.AddSource("Npgsql")
+        [TracerInstrumentation.Npgsql] = builder => builder.AddSource("Npgsql"),
+        [TracerInstrumentation.GrpcNetClient] = builder => builder.AddGrpcClientInstrumentation(options => options.SuppressDownstreamInstrumentation = !Instrumentation.TracerSettings.EnabledInstrumentations.Contains(TracerInstrumentation.HttpClient))
```
nrcventura, line 40: I think that this is a good default to use, but it is different than the default used when manually adding the instrumentation library. We should probably add a setting to control this value.

Kielek: I fully agree that it should be configurable, but I would like to avoid making configuration on our side.
It will be great to have it on the library side.

I think that configuration is less important than the instyrumentation itself so I postoped working on this.
When we will more or less ready with tasks for 0.3.0 we should consider focues also on extending instrumentation libraries on sd/contrib repos.

For now, the whole progress in this context is covered by 
 https://github.com/open-telemetry/opentelemetry-dotnet/pull/3457
and the task for configuration on MySql side https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/519.

test/test-applications/integrations/TestApplication.GrpcNetClient/Program.cs
Kielek: Span is created even if it fails. As we only smoke tests if the instrumenation is started I decided to not start real server.
Do you think that additional comment is needed or uri defined as `dummyserver` is enough?

src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
nrcventura: I agree that ideally, the config will be in the instrumentation library implementation. I'm also asking the question to see if we need an interim solution as well, especially if the next release of the instrumentation library or sdk requires S.D.DS. version 7 (which is in prerelease). If we decide that we should have an interim solution, then I'm fine with implementing it in a separate PR; I just want us to make a decision.

test/test-applications/integrations/TestApplication.GrpcNetClient/Program.cs
pjanotti: SGTM, just wanted to confirm.

src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationTracerHelper.cs
pjanotti: > I'm also asking the question to see if we need an interim solution

The default position should be to treat it as a limitation of current implementation unless there is a compelling enough case to make our own configuration. I think the case with the gRPC setup (that required an extra setting) was a good example of that.

pjanotti: In this specific case I don't think we need an interim solution.

pjanotti: >  if the next release of the instrumentation library or sdk requires S.D.DS. version 7 (which is in prerelease)

That's a very good point!

nrcventura (approved):

PR #1068, opened 2022-08-10
Fix flaky test PrometheusExporter

### LINKED ISSUE #978: IntegrationTests.SmokeTests.PrometheusExporter [FAIL] Did not expect any exceptions after 30s, but found System.Net.Http.HttpRequestException with message "Connection refused"
Potentially flaky test

- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/runs/7500128904?check_suite_focus=true
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/runs/7513623116?check_suite_focus=true

pjanotti:
## Why

To improve reliability of PrometheusExporter smoke test, see #978

Fixes #978

## What

The test application uses `Thread.Yield()` when `LONG_RUNNING` environment variable is set on the test. However, `Thread.Yield()` is non-deterministic and the application may terminate before the test can successfully retrieve the data. There is also no guarantee that the expected counter is going to be there on the first read of the endpoint. The test application now won't terminate when `LONG_RUNNING` is set to true. The tests using this setting are expected to kill the process when they finish.

## Tests

Ran the tests multiple times on dev box.

## Checklist

### DIFF
--- a/test/test-applications/integrations/TestApplication.Smoke/Program.cs
+++ b/test/test-applications/integrations/TestApplication.Smoke/Program.cs
@@ -35,10 +35,17 @@ namespace TestApplication.Smoke
             EmitTraces();
             EmitMetrics();
 
+            // The "LONG_RUNNING" environment variable is used by tests that access/receive
+            // data that takes time to be produced.
             var longRunning = Environment.GetEnvironmentVariable("LONG_RUNNING");
             while (longRunning == "true")
             {
-                Thread.Yield();
+                // In this case it is necessary to ensure that the test has a chance to read the
+                // expected data, only by keeping the application alive for some time that can
+                // be ensured. Anyway, tests that set "LONG_RUNNING" env var to true are expected
+                // to kill the process directly.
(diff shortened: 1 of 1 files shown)


Kielek (approved):

test/test-applications/integrations/TestApplication.Smoke/Program.cs
```diff
@@ -35,10 +35,17 @@ public static void Main(string[] args)
             EmitTraces();
             EmitMetrics();
 
+            // The "LONG_RUNNING" environment variable is used by tests that access/receive
+            // data that takes time to be produced.
             var longRunning = Environment.GetEnvironmentVariable("LONG_RUNNING");
             while (longRunning == "true")
```
rajkumar-rangaraj, line 41: Should we change while to if?

pjanotti: Sorry missed your comment @rajkumar-rangaraj, yeah, it makes more sense to have an `if` I will change opportunistically in another PR.

Issue #707, opened 2022-05-31
Consistent name for metric settings


pjanotti:
Review and make the property names for metric settings consistent, eg.:

- `MetricExporter` is consistent with the spec name for the metrics exporter interface, but, some other properties use metrics (plural)
- `Meter` is used as symmetrical to `Tracer` in certain locations but it is more like a logic grouping of somehow related metrics
- `EnabledInstrumentation` should be plural like the respective value on the tracer

PR #1071, opened 2022-08-10
Rename meter to metric in several places

dszmigielski:
## Why

Names are not consistent.

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/707

## What

From the issue description:
- `MetricExporter` is consistent with the spec name for the metrics exporter interface, but, some other properties use metrics (plural) <- I haven't found an example of that besides tests code, I have not changed those.
- `Meter` is used as symmetrical to `Tracer` in certain locations but it is more like a logic grouping of somehow related metrics <- I have renamed Meter to Metric in several places.
- `EnabledInstrumentation` should be plural like the respective value on the tracer <- this one is already named `EnabledInstrumentations` in `MeterSettings`.

## Tests

N/A

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationMetricHelper.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationMetricHelper.cs
@@ -23,14 +23,14 @@ namespace OpenTelemetry.AutoInstrumentation.Configuration;
 
 internal static class EnvironmentConfigurationMetricHelper
 {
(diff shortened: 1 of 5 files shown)


dszmigielski:
@pjanotti please confirm if that is what you meant in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/707

nrcventura (approved):
I think that these changes make sense, and seem to be consistent with the configuration options that I saw available in the java agent and autoconfigure modules, where Metric was used instead of Meter.

rajkumar-rangaraj (approved):

pjanotti (approved):

Issue #1081, opened 2022-08-17
Implement or remove OTEL_DOTNET_AUTO_METRICS_ENABLED


pjanotti:
There is a config key for `OTEL_DOTNET_AUTO_METRICS_ENABLED`, however, the key is not used. Either it should be removed or implemented.

Kielek:
SIG meeting: remove it for now, we can bring it back when needed.

PR #1088, opened 2022-08-18
Remove unused env variables

### LINKED ISSUE #1080: Implement or remove OTEL_DOTNET_AUTO_TRACES_ENABLED
`OTEL_DOTNET_AUTO_TRACES_ENABLED` was incorrectly disabling the CLR profiler. After that is [fixed](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1082) there is still a config key for this variable, however, the key is not used. Either it should be removed or implemented.

Kielek:
## Why

Fixes #1080 and #1081

## What

Remove unused env. variables
`OTEL_DOTNET_AUTO_METRICS_ENABLED` and `OTEL_DOTNET_AUTO_TRACES_ENABLED`

## Tests

CI

## Checklist

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -5,8 +5,7 @@
 | Environment variable | Description | Default value |
 |-|-|-|
 | `OTEL_DOTNET_AUTO_HOME` | Installation location. |  |
-| `OTEL_DOTNET_AUTO_TRACES_ENABLED` | Enables the tracer. | `true` |
-| `OTEL_DOTNET_AUTO_METRICS_ENABLED` | Enables the meter. | `true` |
+| `OTEL_DOTNET_AUTO_ENABLED` | Enables CLR profiler needed by Tracer. | `true` |
 | `OTEL_DOTNET_AUTO_INCLUDE_PROCESSES` | Names of the executable files that the profiler can instrument. Supports multiple comma-separated values, for example: `MyApp.exe,dotnet.exe`. If unset, the profiler attaches to all processes by default. |  |
 | `OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES` | Names of the executable files that the profiler cannot instrument. Supports multiple comma-separated values, for example: `ReservedProcess.exe,powershell.exe`. The list is processed after `OTEL_DOTNET_AUTO_INCLUDE_PROCESSES`. If unset, the profiler attaches to all processes by default. |  |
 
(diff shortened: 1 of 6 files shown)


docs/config.md
```diff
@@ -5,8 +5,7 @@
 | Environment variable | Description | Default value |
 |-|-|-|
 | `OTEL_DOTNET_AUTO_HOME` | Installation location. |  |
-| `OTEL_DOTNET_AUTO_TRACES_ENABLED` | Enables the tracer. | `true` |
-| `OTEL_DOTNET_AUTO_METRICS_ENABLED` | Enables the meter. | `true` |
+| `OTEL_DOTNET_AUTO_ENABLED` | Enables CLR profiler needed by Tracer. | `true` |
```
pjanotti, line 8: 👍

pjanotti (approved):
Still a draft but LGTM.

docs/config.md
nrcventura: I'm wondering if this setting should also affect either the startup hook or the code that attempts to configure the SDK? The name implies that it should affect everything, but the description states that it only affects the profiler (leading to a difference in behavior between .net framework and .net core 3.1+).

pjanotti: The setting only affects the CLR profiler: it makes it return E_FAIL on the initialize call, which means no further calls to the CLR profiler.

nrcventura: Right. so what I'm referring to is that for .net framework apps, that will lead to no auto instrumentation running at all. However, for a .net 6 app, because we don't rely on the clr profiler, auto instrumentation will still be available (for source code instrumentation). We may want to at least call out that difference in behavior.

Kielek: @nrcventura, what do you think about following statement? If you are fine please please apply it to PR.

```suggestion
| `OTEL_DOTNET_AUTO_ENABLED` | Enables CLR profiler. If `false`, it disables Automatic Instrumentation for .NET Framework and it disables binary instrumentation for .NET (Core). | `true` |
```

RassK (approved):

docs/config.md
pjanotti: I find `binary instrumentation` a bit confusing, elsewhere we used `bytecode instrumentation` iirc.

nrcventura: I like the combination of suggestions from @Kielek and @pjanotti.

pjanotti (approved):

nrcventura (approved):

PR #1121, opened 2022-08-23
dotnet-format check step

Kielek:
## Why

Follow up to #375

## What

Apply file scoped namespacaces once more time (to the new code and leftwovers).
Adding dotnet-format check step taken from contrib repository to keep code formating consistent.this. 

## Tests

CI

## Checklist


rajkumar-rangaraj (approved):

zacharycmontoya (approved):

Kielek:
> There are a couple of files changed within the native code projects that look like they contain vendorized code. Should we exclude those files from the format rules?

Good catch. I think that I have forced it manually. dotnet-format does not change files with auto-generated header.

Kielek:
> There are a couple of files changed within the native code projects that look like they contain vendorized code. Should we exclude those files from the format rules?

Reverted changes both in native folder and coreclr.
auto-generated header prevent code from analyzing by the tool.

nrcventura (approved):

pjanotti (approved):

pellared:
What do you think of running `dotnet-format` via Nuke target instead?

Kielek:
> What do you think of running `dotnet-format` via Nuke target instead?

The dedicated step is better option in my opinion. It is executed only on one platform, but it can be converted to the nuke task if needed.

Issue #970, opened 2022-07-21
IntegrationTests.SqlClientTests.SubmitTraces [FAIL] Microsoft.Data.SqlClient.SqlException (0x80131904): Login failed for user 'sa'.


pellared:
https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/runs/7444225072?check_suite_focus=true

Flaky? Is it a problem with the generated password?

pjanotti:
Any other reports of this one?

nrcventura:
> Any other reports of this one?

I've only seen it once. During the SIG meeting we decided to keep this issue around and tentatively scheduling looking into it after we wrap up our 0.3.0-beta milestone.

pjanotti:
Same failure at https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/runs/7924735948?check_suite_focus=true#step:5:2188

pjanotti:
Another one https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/runs/7924836032?check_suite_focus=true#step:5:2193

Now it looks more than just flakiness. Bumping this to 0.3.0 until we are sure.

cc @open-telemetry/dotnet-instrumentation-approvers

pjanotti:
It passed on retries, but, there were a relatively high number of failures. Likely this is a test issue but we should investigate and fix it if appropriate before 0.3.0.

pjanotti:
I couldn't repro the exact same error (tried multiple runs delaying start of the docker, incorrect password, invalid sa password, etc). I will move this back to 0.4.0 since it really seems something related to the CI environment.

Kielek:
> Flaky? Is it a problem with the generated password?

In my opinion it is not related to generated password. I was able to take generated password from failing execution and execute test on my env. It passed.

Kielek:
All failures are related to exit code 134.

Is it `SIGABORT`?

Kielek:
Still occurring after @1122 merge https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/runs/7988227878?check_suite_focus=true

PR #1122, opened 2022-08-23
SqlTest improvements

Kielek:
## Why

Related to #970

## Tests

Simplify assertions.
Change connection string: `localhost` --> `127.0.0.1`
Cleanup code.
Execute unit tests before integration tests.

## Checklist

### DIFF
--- a/test/IntegrationTests/SqlClientTests.cs
+++ b/test/IntegrationTests/SqlClientTests.cs
@@ -55,12 +55,7 @@ namespace IntegrationTests
 
                 foreach (var span in spans)
                 {
-                    span.Service.Should().Be(ServiceName);
-                    span.Name.Should().Be("master");
-                    span.Tags.Should().Contain(new KeyValuePair<string, string>("db.system", "mssql"));
-                    span.Tags.Should().Contain(new KeyValuePair<string, string>("db.name", "master"));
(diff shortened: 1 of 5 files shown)


Kielek:
On my fork, similar changes passed 10/10 (with changed localhost to 127.0.0.1).

rajkumar-rangaraj (approved):
LGTM

test/IntegrationTests/SqlClientTests.cs
```diff
@@ -55,12 +55,7 @@ public void SubmitTraces()
 
                 foreach (var span in spans)
                 {
-                    span.Service.Should().Be(ServiceName);
-                    span.Name.Should().Be("master");
-                    span.Tags.Should().Contain(new KeyValuePair<string, string>("db.system", "mssql"));
-                    span.Tags.Should().Contain(new KeyValuePair<string, string>("db.name", "master"));
-                    span.Tags.Should().Contain(new KeyValuePair<string, string>("peer.service", $"localhost,{_sqlServerFixture.Port}"));
```
zacharycmontoya, line 62: Was there a reason for removing these? These seem like good tests to have to ensure the spans have the right information

Kielek: I do not think that it is AutoInstrumentation responsibility to test it.
It should be covered by library tests. Here we should only make a sanity check that instrumentation is _somehow working_.,

PR #1147, opened 2022-09-01
Change mock agent to async

### LINKED ISSUE #894: MockCollector should finish when the instrumented application exited
It would be nice to have a method on the mock collector to get all spans already received and bail. This would avoid waiting for spans after the application already finished.

_Originally posted by @pjanotti in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/892#discussion_r907588279_

RassK:
## Why

Partially Fixes #894

I'm not sure if we had any major changes in mock Zipkin collector but to me it seems there is generally no blocking when application has exited already and proper number of spans are generated.

Additionally the current logic prevents flackiness:
* When application is exited and OS network stack causes minor delay, so not all expected spans are returned. I guess it's very unlikely but not impossible.

## What

Currently proposing that we lose all thread blockers and convert to async pattern.

## Tests

Existing.


pellared (approved):
It is also worth noting that `Task.Delay` can accept a `CancellationToken` so that the call can be canceled 😉

Kielek (approved):

rajkumar-rangaraj (approved):

rajkumar-rangaraj:
@RassK `CompileManagedTests` is failing in CI, could you please take a look?

PR #1233, opened 2022-09-15
Stabilize Smoke and Plugins tests

Kielek:
## Why

Fixes N/A.

`http://httpstat.us/200` is not always responding on the GitHub pipeline. It cause crashing application and leads to failing tests.

## What

Just log exception when there is no possibility to call `http://httpstat.us/200` instead of crashing application.

## Tests

Improving CI tests.
Tests passed locally with some dummy (always failing) endpoint.

## Checklist

### DIFF
--- a/test/test-applications/integrations/TestApplication.Smoke/Program.cs
+++ b/test/test-applications/integrations/TestApplication.Smoke/Program.cs
@@ -59,8 +59,15 @@ public class Program
             activity?.SetTag("baz", new int[] { 1, 2, 3 });
         }
 
-        var client = new HttpClient();
-        client.GetStringAsync("http://httpstat.us/200").Wait();
+        try
+        {
+            var client = new HttpClient();
+            client.GetStringAsync("http://httpstat.us/200").Wait();
+        }
+        catch (Exception e)
+        {
+            Console.WriteLine(e);
+        }
     }
 
     private static void EmitMetrics()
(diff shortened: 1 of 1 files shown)


test/test-applications/integrations/TestApplication.Smoke/Program.cs
```diff
@@ -59,8 +59,15 @@ private static void EmitTraces()
...
-        client.GetStringAsync("http://httpstat.us/200").Wait();
+        try
+        {
+            var client = new HttpClient();
+            client.GetStringAsync("http://httpstat.us/200").Wait();
+        }
+        catch (Exception e)
+        {
+            Console.WriteLine(e);
+        }
```
pellared, line 70:
```suggestion
        var client = new HttpClient();
        try
        {
            client.GetStringAsync("http://httpstat.us/200").Wait();
        }
        catch (Exception ex)
        {
            Console.WriteLine(ex);
        }
```

pellared (approved):

pellared:
Can you check if setting `OTEL_DOTNET_AUTO_FLUSH_ON_UNHANDLEDEXCEPTION=true` is helping? I had the same problem when preparing for the meetup 😄

Kielek:
> Side note: Can you check if setting `OTEL_DOTNET_AUTO_FLUSH_ON_UNHANDLEDEXCEPTION=true` is helping? I had the same problem when preparing for the meetup 😄

Not in this case.
In tests we are checking if the ExitCode is 0.

PR #1237, opened 2022-09-16
Add waiting for server in WCF tests

dszmigielski:
## Why

WCF tests fail randomly.

## What

Wait for WCF Server to open TCP port before starting test application.

## Tests

N/A

## Checklist

### DIFF
--- a/test/IntegrationTests/WcfTestsBase.cs
+++ b/test/IntegrationTests/WcfTestsBase.cs
@@ -15,6 +15,7 @@
 // </copyright>
 
 using System;
+using System.Net.Sockets;
 using System.Threading.Tasks;
 using FluentAssertions;
 using FluentAssertions.Execution;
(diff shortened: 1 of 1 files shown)


test/IntegrationTests/WcfTestsBase.cs
```diff
@@ -72,4 +74,25 @@ public void Dispose()
...
+            try
+            {
+                await tcpClient.ConnectAsync("127.0.0.1", tcpPort);
+                Output.WriteLine("WCF Server is running.");
+                return;
+            }
+            catch (Exception)
+            {
+                Output.WriteLine("Waiting for WCF Server to open ports.");
+                await Task.Delay(500);
```
Kielek, line 94: What if server fully failed? We will be waiting here forever.
We need to have some timeout/attempts count then fail.

Kielek (approved):

pjanotti (approved):

PR #1243, opened 2022-09-19
Add a convenient Bash script for installation

pellared:
## Why

Facilitate installation. Increase ease of use.

## What

Add a convenient Bash script for installation.

## Tests

Manual and added new `install-script` job in `ci` GH workflow.

## Checklist


docs/README.md
```diff
@@ -92,7 +92,27 @@ See [config.md#instrumented-libraries-and-frameworks](config.md#instrumented-lib
...
+  default: `https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/releases`
+- `TMPDIR` - temporary directory used when downloading the files;
+  default: `$(mktemp -d)`  
+- `VERSION` - version to download;
+  required
+
+```sh
+( set -o pipefail
+curl -sSfL https://raw.githubusercontent.com/open-telemetry/opentelemetry-dotnet-instrumentation/main/download.sh |
+  VERSION=v0.3.1-beta.1 FLAVOUR=linux-glibc bash -s )
```
Kielek, line 114: Extend release process to update this line.

pellared: PTAL 83e999a3b

docs/README.md
```diff
@@ -92,7 +92,27 @@ See [config.md#instrumented-libraries-and-frameworks](config.md#instrumented-lib
...
+
+- `INSTALL_DIR` - location where binaries are to be installed;
+  default: `./otel-dotnet-auto`
+- `FLAVOUR` - possible values: `linux-glibc`, `linux-musl`, `macos`, `windows`;
+  required
+- `RELEASES_URL` - GitHub releases URL;
+  default: `https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/releases`
+- `TMPDIR` - temporary directory used when downloading the files;
+  default: `$(mktemp -d)`  
+- `VERSION` - version to download;
```
pjanotti, line 108: nit: I guess that it should only be used for current version and above (just because it won't work for musl for the old ones and I'm not sure if it was tested against the old ones).

docs/README.md
```diff
@@ -104,32 +124,32 @@ COR_ENABLE_PROFILING=1
...
 ```
 
 On **Windows** you need to additionally set:
 
 ```env
-COR_PROFILER_PATH_64=%InstallationLocation%/win-x64/OpenTelemetry.AutoInstrumentation.Native.dll
-COR_PROFILER_PATH_32=%InstallationLocation%/win-x86/OpenTelemetry.AutoInstrumentation.Native.dll
-CORECLR_PROFILER_PATH_64=%InstallationLocation%/win-x64/OpenTelemetry.AutoInstrumentation.Native.dll
-CORECLR_PROFILER_PATH_32=%InstallationLocation%/win-x86/OpenTelemetry.AutoInstrumentation.Native.dll
+COR_PROFILER_PATH_64=$INSTALL_DIR/win-x64/OpenTelemetry.AutoInstrumentation.Native.dll
```
pjanotti, line 137: On Windows this is not the correct way to expand an env var.

docs/config.md
```diff
@@ -197,9 +197,9 @@ for more information.
...
-| `COR_PROFILER_PATH` | `CORECLR_PROFILER_PATH` | Path to the profiler. | `%InstallationLocation%/OpenTelemetry.AutoInstrumentation.Native.dll` for Windows, `%InstallationLocation%/OpenTelemetry.AutoInstrumentation.Native.so` for Linux, `%InstallationLocation%/OpenTelemetry.AutoInstrumentation.Native.dylib` for macOS |
-| `COR_PROFILER_PATH_32` | `CORECLR_PROFILER_PATH_32` | Path to the 32-bit profiler. Bitness-specific paths take precedence over generic paths. | `%InstallationLocation%/win-x86/OpenTelemetry.AutoInstrumentation.Native.dll` for Windows |
-| `COR_PROFILER_PATH_64` | `CORECLR_PROFILER_PATH_64` | Path to the 64-bit profiler. Bitness-specific paths take precedence over generic paths. | `%InstallationLocation%/win-x64/OpenTelemetry.AutoInstrumentation.Native.dll` for Windows |
+| `COR_PROFILER_PATH` | `CORECLR_PROFILER_PATH` | Path to the profiler. | `$INSTALL_DIR/OpenTelemetry.AutoInstrumentation.Native.dll` for Windows, `$INSTALL_DIR/OpenTelemetry.AutoInstrumentation.Native.so` for Linux, `$INSTALL_DIR/OpenTelemetry.AutoInstrumentation.Native.dylib` for macOS |
```
pjanotti, line 200: Correct env var for Windows. Same below.

docs/config.md
```diff
@@ -212,7 +212,7 @@ are used to mitigate assembly version conflicts in .NET Core.
 
 | Environment variable | Required value |
 |-|-|
-| `DOTNET_STARTUP_HOOKS` | `%InstallationLocation%/netcoreapp3.1/OpenTelemetry.AutoInstrumentation.StartupHook.dll` |
-| `DOTNET_ADDITIONAL_DEPS` | `%InstallationLocation%/AdditionalDeps` |
-| `DOTNET_SHARED_STORE` | `%InstallationLocation%/store` |
+| `DOTNET_STARTUP_HOOKS` | `$INSTALL_DIR/netcoreapp3.1/OpenTelemetry.AutoInstrumentation.StartupHook.dll` |
```
pjanotti, line 215: Bummer, it will get verbose when adding Windows syntax... perhaps resolve all of these with a note on a central place? E.g.: "using unix env var syntax on the docs unless otherwise noted blah blah"

@theletterf any suggestions on how to handle the diff syntaxes and keep it short? Or do you think it is better to go with unix and Windows syntaxes?

download.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
+set -e
+
+test -z "$VERSION" && {
```
pjanotti, line 4: Why not put the latest as a fallback for VERSION like the others and update it on every release?

pjanotti (approved):
Script looks good just want for us to be careful about presenting env vars on Windows with proper syntax.

download.sh
```diff
@@ -0,0 +1,31 @@
...
+	echo "Please specify the version by setting the VERSION env var." >&2
+	exit 1
+}
+test -z "$FLAVOUR" && {
```
zacharycmontoya, line 8: Two very small naming suggestions, but I don't feel too strongly about them:
1. Renaming `flavour` to `distribution`
2. Renaming folder `otel-dotnet-auto` to `otel-dotnet-auto-installation`

download.sh
zacharycmontoya: +1, if you put the version in this file, then this is the only file that needs to be updated on every version bump. All other uses of `download.sh` don't need to have a version number unless they want to override it

zacharycmontoya (approved):
LGTM with some small suggestions

rajkumar-rangaraj (approved):

docs/README.md
```diff
@@ -92,7 +92,27 @@ See [config.md#instrumented-libraries-and-frameworks](config.md#instrumented-lib
 Download and extract the appropriate binaries from
 [the latest release](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/releases/latest).
 
-> The path where you put the binaries is referenced as `%InstallationLocation%`
+> The path where you put the binaries is referenced as `$INSTALL_DIR`
+
+You can use a convenient Bash script to do it.
+Parameters:
```
theletterf, line 98:
```suggestion
You can also use the `download.sh` script with the following parameters:
```

docs/README.md
theletterf: I'd turn this list into a table with columns `Parameter`, `Required`, and `Description`

docs/config.md
theletterf: Actually, doesn't PowerShell support the dollar sign as a var mark? Do we have to wrap the statements in double quotes to make it safe? 

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_variables?view=powershell-7.2#using-variables-in-commands-and-expressions

If those can be used for both Bash and PSh, I'd rename the column heading "Requires value" to "Required value (Bash and Powershell)

theletterf (changes requested):

docs/README.md
pellared: What do you mean by "in Windows"? E.g. in PowerShell it is the correct way 😉

download.sh
pellared: `latest` (see [here](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/releases/latest)) is not working right now as we do not have any release (with a green label, right now we have only pre-releases 😉 )

We can do a separate issue for it.

download.sh
pellared: I like first suggestion 😉

download.sh
pellared: I created https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1250

docs/README.md
```diff
@@ -94,24 +94,20 @@ Download and extract the appropriate binaries from
...
-  required
-- `RELEASES_URL` - GitHub releases URL;
-  default: `https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/releases`
-- `TMPDIR` - temporary directory used when downloading the files;
-  default: `$(mktemp -d)`  
-- `VERSION` - version to download;
-  required
+You can also use the [download.sh](../download.sh) script with the following parameters:
+
+| Parameter | Description | Required | Default value |
```
Kielek, line 99: nit: https://marketplace.visualstudio.com/items?itemName=darkriszty.markdown-table-prettify :)

Kielek (approved):

pellared:
@pjanotti PTAL. Feel free to add edits and merge 😉

download.sh
pjanotti: I was thinking that in putting `v0.3.1-beta.1` until #1250 is done and update it on every release.

docs/config.md
pjanotti: PowerShell would be `$Env:ENV_VAR_NAME` - the worst `$ENV_VAR_NAME` is treated as a variable and since it was not defined empty ...

That said I'm fine with using PowerShell syntax as long as it is something that works on Windows without a bash shell 😬

nrcventura (approved):

Kielek (approved):

PR #1253, opened 2022-09-21
Do not load autoinstrumentation assembly on splashscreen

pjanotti:
## Why

Loading OTel.AutoInstrumentation assembly on the integration tests application via the splash screen is not a real test scenario.

## What

Removed calls to test code that ends up loading OTel.AutoInstrumentation assembly to make the tests more realistic. Also removed some duplication of the code used to retrieve env vars - the example app still has its own copy to avoid it referencing the actual test code.

## Tests

N/A

## Checklist


test/test-applications/integrations/dependency-libs/TestApplication.Shared/ConsoleHelper.cs
```diff
@@ -23,8 +26,12 @@ public static class ConsoleHelper
     public static void WriteSplashScreen(string[] args)
     {
         Console.WriteLine($"Command line: {string.Join(" ", args)}");
-        var (isProfilerAttached, additionalMessage) = ProfilerHelper.IsProfilerAttached();
```
Kielek, line 26: When I was working on otel-operator this was really helpful.
Do you see any option to put that information in the logs?
At the end of application lifecycle?

pellared (approved):

test/test-applications/integrations/dependency-libs/TestApplication.Shared/ConsoleHelper.cs
pjanotti: What information was useful? That OpenTelemetry.AutoInstrumentation was loaded or that the CLR Profiler was attached to the process?

For the former we could list the loaded assemblies and check for specific assemblies. For the second there is already a native API, but don't use the wrapper from OpenTelemetry.AutoInstrumentation (otherwise that will end up loading the assembly again).

Kielek (approved):

zacharycmontoya (approved):

nrcventura (approved):

PR #1254, opened 2022-09-21
Fix race condition in mock collectors

pellared:
## Why

Make the tests stable. **Exceptions:** `PrometheusExporter` and `StrongNamedTests`. These are very flaky.

The tests have started to be unstable most probably after https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1147/. I believe it happened because of changing `Thread.Sleep` (which always makes a memory barrier) to `Task.Delay` (which does not make a memory barrier). Because of that, the thread running the test may never see the "replaced" immutable collection. Besides this bug might be not so likely to happen on "our" machines b/c of arch differences in CPUs on our laptops vs GitHub Actions runners.

## What

Fix race conditions in mock collectors.

I could also change back to `Thread.Sleep` or use `Thread.MemoryBarrier`, but it would be an unmaintainable hack.

I also removed the usage of `timeout` when calling `WaitFor*` to ensure that the default timeout, 20 sec, is used in all tests. It looks like 5 sec for some tests might have been not enough.

## Tests

Execute CI 5 times.

Run 1: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3097103461/attempts/1
Run 2: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3097103461/attempts/2
Run 3: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3097103461/attempts/3
Run 4: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3097103461/attempts/4
Run 5: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3097103461/attempts/5


RassK (approved):
👍

Kielek (approved):

pjanotti (approved):

rajkumar-rangaraj (approved):

Kielek:
> image
> 
> Might also help with this one now...

It will not help on this. Line before is assertion that we expect one tag. It passes. The issue is with the received span.

Issue #1123, opened 2022-08-23
Prometheus flaky test


rajkumar-rangaraj:
Creating this issue to track failure from test `IntegrationTests.SmokeTests.PrometheusExporter`. Yesterday there was a different exception. 

 Error: [xUnit.net 00:00:34.85]     IntegrationTests.SmokeTests.PrometheusExporter [FAIL]
  16:22:12 [ERR] [xUnit.net 00:00:34.85]     IntegrationTests.SmokeTests.PrometheusExporter [FAIL]
  16:22:12 [DBG]   Failed IntegrationTests.SmokeTests.PrometheusExporter [31 s]
  16:22:12 [DBG]   Error Message:
  16:22:12 [DBG]    Did not expect any exceptions after 30s, but found Xunit.Sdk.XunitException with message "Expected string "# HELP process_runtime_dotnet_gc_collections_count Number of garbage collections that have occurred since process start.
  16:22:12 [DBG] # TYPE process_runtime_dotnet_gc_collections_count counter
  16:22:12 [DBG] process_runtime_dotnet_gc_collections_count{generation="gen2"} 0 1661271707095
  16:22:12 [DBG] process_runtime_dotnet_gc_collections_count{generation="gen1"} 0 1661271707095
  16:22:12 [DBG] process_runtime_dotnet_gc_collections_count{generation="gen0"} 0 1661271707095
  16:22:12 [DBG] 
  16:22:12 [DBG] # HELP process_runtime_dotnet_assemblies_count The number of .NET assemblies that are currently loaded.
  16:22:12 [DBG] # TYPE process_runtime_dotnet_assemblies_count gauge
  16:22:12 [DBG] process_runtime_dotnet_assemblies_count 32 1661271707095
  16:22:12 [DBG] " to contain "TYPE MyFruitCounter counter"."
  16:22:12 [DBG]      at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message)
  16:22:12 [DBG]      at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc)
  16:22:12 [DBG]      at FluentAssertions.Primitives.StringAssertions`1.Contain(String expected, String because, Object[] becauseArgs)
[...]

PR #1259, opened 2022-09-21
Stabilize PrometheusExporter test

pellared:
## Why

It looks like there is some bug in PrometheusExporter: https://github.com/open-telemetry/opentelemetry-dotnet/issues/3679

I think we should not test the SDK. I find that the test is "overspecified".

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1123

Still, we should explore what is/was going on...

## What

Assert that any metric is exported.

## Tests

CI

## Checklist


pjanotti (approved):
Since we are suspecting that this is a bug on the SDK it seems reasonable to just check that the exporter is functional.

pjanotti:
The failure fixed by this PR seems to be the most common for this test, but, there is also a failure mode in which no output is captured from the exporter, e.g.: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3100295202/jobs/5020411194#step:4:3382

PR #1262, opened 2022-09-22
Add ILogger bytecode instrumentation

rajkumar-rangaraj:
## Why

Only `ILogger` logs from ASP.NET Core are supported. Changes proposed will enable log collection from all types of .NET Core apps using CLR profiler integration.

Fixes #

## What

* Added CLR Profiler based `ILogger` collection.
* Both Hosting Startup and CLR profiler uses the same common method for wire-up.
* Avoided duplicate integration of `AddOpenTelemetry` on `LoggingBuilder`.
* Updated tests to validate both Hosting Startup and CLR Profiler.
* Rename existing `Logger` to `CustomLogger` as `Logger` is most apt for the current proposed integration.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - `OTEL_DOTNET_AUTO_LOGS_CONSOLE_EXPORTER_ENABLED`,
   - `OTEL_DOTNET_AUTO_LOGS_PARSE_STATE_VALUES`,
   - `OTEL_DOTNET_AUTO_LOGS_INCLUDE_FORMATTED_MESSAGE`.
-- Adds CLR Profiler based OpenTelemetry Logs support from .NET Core 3.1+.  
+- Adds CLR Profiler based OpenTelemetry Logs support for .NET Core 3.1+.  
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,12 +11,15 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Add WCF traces instrumentation (server-side for .NET Framework, client-side
   for both .NET Core and .NET Framework).
-- Support for ASP.NET Core OpenTelemetry Log exporter related environment variables:
+- Support ASP.NET Core OpenTelemetry Log exporter related environment variables:
   - `OTEL_LOGS_EXPORTER`,
   - `OTEL_DOTNET_AUTO_LOGS_CONSOLE_EXPORTER_ENABLED`,
   - `OTEL_DOTNET_AUTO_LOGS_PARSE_STATE_VALUES`,
   - `OTEL_DOTNET_AUTO_LOGS_INCLUDE_FORMATTED_MESSAGE`.
-- Adds CLR Profiler based OpenTelemetry Logs support for .NET Core 3.1+.  
+- Support `OTEL_DOTNET_AUTO_GRAPHQL_SET_DOCUMENT` (default value: `false`)
+  environment variable which controls whether `graphql.document` attribute
+  is set.
+- Add ILogger logging instrumentation for .NET Core 3.1+.  
 
 ### Changed
 
@@ -101,7 +104,7 @@ You can find all OpenTelemetry references in
 
 ### Added
 
-- Adds MongoDB instrumentation support from .NET Core 3.1+.
+- Add MongoDB instrumentation support from .NET Core 3.1+.
 - Support for OpenTelemetry metric exporter related environment variables:
   - `OTEL_DOTNET_AUTO_METRICS_ENABLED`,
   - `OTEL_DOTNET_AUTO_LOAD_METER_AT_STARTUP`,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - `OTEL_DOTNET_AUTO_LOGS_CONSOLE_EXPORTER_ENABLED`,
   - `OTEL_DOTNET_AUTO_LOGS_PARSE_STATE_VALUES`,
   - `OTEL_DOTNET_AUTO_LOGS_INCLUDE_FORMATTED_MESSAGE`,
+- Adds CLR Profiler based OpenTelemetry Logs support from .NET Core 3.1+.  
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - `OTEL_LOGS_EXPORTER`,
   - `OTEL_DOTNET_AUTO_LOGS_CONSOLE_EXPORTER_ENABLED`,
   - `OTEL_DOTNET_AUTO_LOGS_PARSE_STATE_VALUES`,
-  - `OTEL_DOTNET_AUTO_LOGS_INCLUDE_FORMATTED_MESSAGE`,
+  - `OTEL_DOTNET_AUTO_LOGS_INCLUDE_FORMATTED_MESSAGE`.
 - Adds CLR Profiler based OpenTelemetry Logs support from .NET Core 3.1+.  
 
 ### Changed
--- a/docs/config.md
+++ b/docs/config.md
@@ -63,7 +63,7 @@ Instrumentation. Recommended option for ASP.NET Core is to use HostingStartup.
 For other .NET Core apps logs could be collected only through .NET CLR Profiler
 Instrumentation.
 
-### Enable Logs using HostingStartup 
+### Enable Logs using HostingStartup
 
 | Environment variable | Required value |
 |-|-|
@@ -71,7 +71,7 @@ Instrumentation.
 
 ### Enable Logs using .NET CLR Profiler
 
-Please follow [Instructions](config.md#net-clr-profiler) here to collect logs
+Please follow [instructions](config.md#net-clr-profiler) here to collect logs
 using .NET CLR Profiler.
 
 ### Logs Parse and Format Options
--- a/docs/config.md
+++ b/docs/config.md
@@ -182,6 +182,12 @@ Important environment variables include:
 |-|-|-|
 | `OTEL_EXPORTER_ZIPKIN_ENDPOINT` | Zipkin URL. | `http://localhost:8126` |
 
+## Instrumentation options
+
+| Environment variable | Description | Default value |
+|-|-|-|
+| `OTEL_DOTNET_AUTO_GRAPHQL_SET_DOCUMENT` | Whether GraphQL instrumentation can pass raw queries as `graphql.document` attribute. This may contain sensitive information and therefore is disabled by default. | `false` |
+
 ## Additional settings
 
 | Environment variable | Description | Default value |
--- a/docs/config.md
+++ b/docs/config.md
@@ -74,7 +74,7 @@ Instrumentation.
 Please follow [instructions](config.md#net-clr-profiler) here to collect logs
 using .NET CLR Profiler.
 
-### Logs Parse and Format Options
+### Parsing and Format Options for Logs
 
 These options could be used with both HostingStartup and .NET CLR Profiler logs
 enablement.
--- a/docs/config.md
+++ b/docs/config.md
@@ -55,6 +55,35 @@ for more details.
 | `HttpClient` | [System.Net.Http.HttpClient](https://docs.microsoft.com/dotnet/api/system.net.http.httpclient) and [System.Net.HttpWebRequest](https://docs.microsoft.com/dotnet/api/system.net.httpwebrequest) | * | source |
 | `NetRuntime` | [OpenTelemetry.Instrumentation.Runtime](https://www.nuget.org/packages/OpenTelemetry.Instrumentation.Runtime) | * | source |
 
+## Logs
+
+Logs from ASP.NET Core apps could be captured using two options. First option is
+to use HostingStartup and the second option is to rely on CLR Profiler
+Instrumentation. Recommended option for ASP.NET Core is to use HostingStartup.
+For other .NET Core apps logs could be collected only through .NET CLR Profiler
+Instrumentation.
+
+### Enable Logs using HostingStartup 
+
+| Environment variable | Required value |
+|-|-|
+| `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES` | `OpenTelemetry.AutoInstrumentation.AspNetCoreBootstrapper` |
+
+### Enable Logs using .NET CLR Profiler
+
+Please follow [Instructions](config.md#net-clr-profiler) here to collect logs
+using .NET CLR Profiler.
+
+### Logs Parse and Format Options
+
+These options could be used with both HostingStartup and .NET CLR Profiler logs
+enablement.
+
+| Environment variable | Description | Default value |
+|-|-|-|
+| `OTEL_DOTNET_AUTO_LOGS_PARSE_STATE_VALUES` | Whether the log state should be parsed. | `false` |
+| `OTEL_DOTNET_AUTO_LOGS_INCLUDE_FORMATTED_MESSAGE` | Whether the log state should be formatted. | `false` |
+
 ## Logging
 
 The default log directory paths are:
@@ -179,7 +208,7 @@ public OpenTelemetry.Metrics.MeterProviderBuilder ConfigureMeterProvider(OpenTel
 ```
 
 ```csharp
-public OpenTelemetryLoggerOptions ConfigureLoggerOptions(OpenTelemetryLoggerOptions builder)
+public OpenTelemetry.Logs.OpenTelemetryLoggerOptions ConfigureLoggerOptions(OpenTelemetry.Logs.OpenTelemetryLoggerOptions builder)
 ```
 
 The plugin must use the same version of the `OpenTelemetry` as the
--- a/docs/config.md
+++ b/docs/config.md
@@ -57,11 +57,10 @@ for more details.
 
 ## Logs
 
-Logs from ASP.NET Core apps could be captured using two options. First option is
-to use HostingStartup and the second option is to rely on CLR Profiler
-Instrumentation. Recommended option for ASP.NET Core is to use HostingStartup.
-For other .NET Core apps logs could be collected only through .NET CLR Profiler
-Instrumentation.
+Logs from ASP.NET Core apps can be captured in two ways: The first option is to
+use `HostingStartup`, and the second is to rely on CLR Profiler Instrumentation.
+For ASP.NET Core, use `HostingStartup`. For other .NET Core apps, you can only
+collect logs through .NET CLR Profiler Instrumentation.
 
 ### Enable Logs using HostingStartup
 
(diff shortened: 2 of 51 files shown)


pellared:
Can you update the docs? E.g. by adding an "Instrumented logging libraries and frameworks" section in `config.md`. It would be good to specify that this instrumentation is available only for .NET (Core).

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - `OTEL_DOTNET_AUTO_LOGS_CONSOLE_EXPORTER_ENABLED`,
   - `OTEL_DOTNET_AUTO_LOGS_PARSE_STATE_VALUES`,
   - `OTEL_DOTNET_AUTO_LOGS_INCLUDE_FORMATTED_MESSAGE`,
```
Kielek, line 18:
```suggestion
  - `OTEL_DOTNET_AUTO_LOGS_INCLUDE_FORMATTED_MESSAGE`.
```

src/OpenTelemetry.AutoInstrumentation/Instrumentations/Logger/LoggingBuilderIntegration.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace OpenTelemetry.AutoInstrumentation.Instrumentations.Logger;
+
+/// <summary>
+/// Microsoft.Extensions.Logging.LoggingBuilder calltarget instrumentation
+/// </summary>
+[InstrumentMethod(
+    AssemblyName = "Microsoft.Extensions.Logging",
+    TypeName = "Microsoft.Extensions.Logging.LoggingBuilder",
+    MethodName = ".ctor",
+    ReturnTypeName = ClrNames.Void,
```
Kielek, line 29: Returned type is tricky for me. I suppose that it should be `LoggingBuilder` as it is a constructor.
As I remember the returned type is nowhere verified while rejiting/building bytecode.

Same should be applied for `instrumentation.json`

src/OpenTelemetry.AutoInstrumentation/AutoInstrumentationEventSource.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace OpenTelemetry.AutoInstrumentation;
+
+/// <summary>
+/// EventSource implementation for OpenTelemetry AutoInstrumentation implementation.
+/// </summary>
+[EventSource(Name = "OpenTelemetry-AutoInstrumentation")]
+internal class AutoInstrumentationEventSource : EventSource
+{
+#pragma warning disable SA1401 // Fields should be private
+    public static AutoInstrumentationEventSource Log = new AutoInstrumentationEventSource();
```
Kielek, line 28: Consider Lazy field/property.

src/OpenTelemetry.AutoInstrumentation/AutoInstrumentationEventSource.cs
```diff
@@ -0,0 +1,46 @@
...
+
+using System.Diagnostics.Tracing;
+
+namespace OpenTelemetry.AutoInstrumentation;
+
+/// <summary>
+/// EventSource implementation for OpenTelemetry AutoInstrumentation implementation.
+/// </summary>
+[EventSource(Name = "OpenTelemetry-AutoInstrumentation")]
+internal class AutoInstrumentationEventSource : EventSource
```
Kielek, line 25: Please add private constructor, to avoid multiple instances of this class.
We have had CI issue related to this stuff: #1230

src/OpenTelemetry.AutoInstrumentation/AutoInstrumentationEventSource.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+#pragma warning disable SA1401 // Fields should be private
+    public static AutoInstrumentationEventSource Log = new AutoInstrumentationEventSource();
+#pragma warning restore SA1401 // Fields should be private
+
+    /// <summary>Logs as Trace level message.</summary>
```
Kielek, line 31: It can be a dummy question, but Tracer level is the same as Informational (two lines later)?

src/OpenTelemetry.AutoInstrumentation/OpenTelemetry.AutoInstrumentation.csproj
```diff
@@ -34,6 +34,7 @@
     <PackageReference Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.0.0-rc9.4" />
     <PackageReference Include="OpenTelemetry.Instrumentation.MySqlData" Version="1.0.0-beta.3" />
     <PackageReference Include="OpenTelemetry.Instrumentation.StackExchangeRedis" Version="1.0.0-rc9.7" />
+    <PackageReference Include="OpenTelemetry.Exporter.OpenTelemetryProtocol.Logs" Version="1.3.0-rc.2" />
```
Kielek, line 37: Can we remove this reference from OpenTelemetry.AutoInstrumentation.AspNetCoreBootstrapper.csproj?
It is transitive dependency (same for Exporter.Console).

src/OpenTelemetry.AutoInstrumentation/Instrumentations/Logger/LoggingBuilderIntegration.cs
rajkumar-rangaraj: I had the similar thought, but constructor has a return type as void in both our repo and in DataDog.

Reference: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/2072410667c15e1804fe71e36d4501bca3057082/src/OpenTelemetry.AutoInstrumentation/Instrumentations/MongoDb/MongoClientIntegration.cs#L35

src/OpenTelemetry.AutoInstrumentation/AutoInstrumentationEventSource.cs
rajkumar-rangaraj: This is for EventSource data collection, we won't be able to mark this as lazy. This is being utilized by framework itself and initialized by us. 

Only with EventListener  we might run into multiple instance issue.

src/OpenTelemetry.AutoInstrumentation/AutoInstrumentationEventSource.cs
pjanotti: Putting together with the private constructor the static field becomes the only way to access the instance methods. In light of that it is not necessary to lazy initialize it: the runtime takes care of that (if not used never initialized).

Add `readonly` too.

rajkumar-rangaraj: This is for EventSource data collection, we won't be able to mark this as lazy. This is being utilized by framework itself and not initialized in auto-instrumentation.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/Logger/LoggingBuilderIntegration.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// <param name="state">Calltarget state value</param>
+    /// <returns>A default CallTargetReturn to satisfy the CallTarget contract</returns>
+    public static CallTargetReturn OnMethodEnd<TTarget>(TTarget instance, Exception exception, CallTargetState state)
+    {
+#if !NETFRAMEWORK
+        if (instance is not null)
+        {
+            var logBuilderExtensionsType = Type.GetType("OpenTelemetry.AutoInstrumentation.Logger.LogBuilderExtensions, OpenTelemetry.AutoInstrumentation");
+            var methodInfo = logBuilderExtensionsType?.GetMethod("AddOpenTelemetryLogs");
+            methodInfo?.Invoke(null, new[] { (object)instance });
```
pellared, line 51: @zacharycmontoya Would it be better to use `OpenTelemetry.AutoInstrumentation.DuckTyping` here instead (I never used it so I am not sure 😅) ?

src/OpenTelemetry.AutoInstrumentation/OpenTelemetry.AutoInstrumentation.csproj
rajkumar-rangaraj: Yes, it is already removed and changes are part of this PR.

src/OpenTelemetry.AutoInstrumentation/AutoInstrumentationEventSource.cs
rajkumar-rangaraj: Reference from OpenTelemetry SDK - https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Internal/OpenTelemetrySdkEventSource.cs

src/OpenTelemetry.AutoInstrumentation/AutoInstrumentationEventSource.cs
rajkumar-rangaraj: Yes, that is correct. We have the same implementation here - https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHookEventSource.cs

src/OpenTelemetry.AutoInstrumentation/OpenTelemetry.AutoInstrumentation.csproj
Kielek: I have missed that.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/Logger/LoggingBuilderIntegration.cs
pjanotti: Remember with have access to the instance inside the constructor. The instance is built with `newobj` and then that instance to call the actual constructor.

src/OpenTelemetry.AutoInstrumentation/AutoInstrumentationEventSource.cs
pjanotti: @rajkumar-rangaraj `readonly` seems a good addition to the static field (I think you may have missed my comment because I published it around the same time as your comments above).

CHANGELOG.md
rajkumar-rangaraj: Fixed - 02500098e5714212260ab4c78e89942bcc509842

src/OpenTelemetry.AutoInstrumentation/AutoInstrumentationEventSource.cs
rajkumar-rangaraj: Changed to `readonly` - 02500098e5714212260ab4c78e89942bcc509842

src/OpenTelemetry.AutoInstrumentation/Instrumentations/Logger/LoggingBuilderIntegration.cs
rajkumar-rangaraj: I tried `OpenTelemetry.AutoInstrumentation.DuckTyping` during my earlier implementation but it created an object with different type which could not be used to invoke the method on current object instance.

CHANGELOG.md
```diff
@@ -15,7 +15,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - `OTEL_LOGS_EXPORTER`,
   - `OTEL_DOTNET_AUTO_LOGS_CONSOLE_EXPORTER_ENABLED`,
   - `OTEL_DOTNET_AUTO_LOGS_PARSE_STATE_VALUES`,
-  - `OTEL_DOTNET_AUTO_LOGS_INCLUDE_FORMATTED_MESSAGE`,
+  - `OTEL_DOTNET_AUTO_LOGS_INCLUDE_FORMATTED_MESSAGE`.
+- Adds CLR Profiler based OpenTelemetry Logs support from .NET Core 3.1+.
```
nrcventura, line 19: Changing `from` to `for`.
```suggestion
- Adds CLR Profiler based OpenTelemetry Logs support for .NET Core 3.1+.  
```

docs/config.md
```diff
@@ -55,6 +55,35 @@ for more details.
...
+| Environment variable | Required value |
+|-|-|
+| `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES` | `OpenTelemetry.AutoInstrumentation.AspNetCoreBootstrapper` |
+
+### Enable Logs using .NET CLR Profiler
+
+Please follow [instructions](config.md#net-clr-profiler) here to collect logs
+using .NET CLR Profiler.
+
+### Logs Parse and Format Options
```
nrcventura, line 77: This heading reads a little awkwardly for me. Maybe it should be something like the following.

```suggestion
### Parsing and Format Options for Logs
```

src/OpenTelemetry.AutoInstrumentation/Instrumentations/Logger/LoggingBuilderIntegration.cs
RassK: I also didn't find any good ways to use ducktyping if the method accepts custom types. Built in / framework types are fine.

rajkumar-rangaraj:
> Can you update the docs? E.g. by adding an "Instrumented logging libraries and frameworks" section in `config.md`. It would be good to specify that this instrumentation is available only for .NET (Core).
> 
> I think we could even have two rows for `ILogger`:
> 
> 1. One for source instrumentation + info that it requires setting the env var and works only for ASP.NET Core apps.
> 2. Second for bytecode instrumentation and info that it works for all .NET (Core) apps.
> 

Done!

> What will happen if someone on .NET (Core) would set both `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES` and enable the CLR profiler? If some strange behavior then we might need to add a note in `troubleshooting.md`.

If both are enabled, there are no issues, `AddOpenTelemetry()` is called once. I will add a test scenario in Integration tests.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/Logger/LoggingBuilderIntegration.cs
pjanotti: It seems that we need to dig a bit deeper on our understanding about when/how to correctly use DuckTyping. It seems fine to me to go with the current approach at this time.

nrcventura (approved):

Kielek (approved):

pellared (approved):

pellared:
@theletterf PTAL

docs/config.md
```diff
@@ -55,6 +55,35 @@ for more details.
...
+### Enable Logs using HostingStartup
+
+| Environment variable | Required value |
+|-|-|
+| `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES` | `OpenTelemetry.AutoInstrumentation.AspNetCoreBootstrapper` |
+
+### Enable Logs using .NET CLR Profiler
+
+Please follow [instructions](config.md#net-clr-profiler) here to collect logs
+using .NET CLR Profiler.
```
pellared, line 75: I see this PR as adding "bytecode instrumentation" for `ILogger` 

Maybe we could try to describe it as such in a `Instrumented logging libraries and frameworks` section?

docs/config.md
```diff
@@ -55,6 +55,35 @@ for more details.
...
+to use HostingStartup and the second option is to rely on CLR Profiler
+Instrumentation. Recommended option for ASP.NET Core is to use HostingStartup.
+For other .NET Core apps logs could be collected only through .NET CLR Profiler
+Instrumentation.
+
+### Enable Logs using HostingStartup
+
+| Environment variable | Required value |
+|-|-|
+| `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES` | `OpenTelemetry.AutoInstrumentation.AspNetCoreBootstrapper` |
```
pellared, line 70: I see this as  "source instrumentation" for `ILogger` that requires additionally setting `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES`.

Maybe we could try to describe it as such in a `Instrumented logging libraries and frameworks` section?

docs/config.md
```diff
@@ -55,6 +55,35 @@ for more details.
 | `HttpClient` | [System.Net.Http.HttpClient](https://docs.microsoft.com/dotnet/api/system.net.http.httpclient) and [System.Net.HttpWebRequest](https://docs.microsoft.com/dotnet/api/system.net.httpwebrequest) | * | source |
 | `NetRuntime` | [OpenTelemetry.Instrumentation.Runtime](https://www.nuget.org/packages/OpenTelemetry.Instrumentation.Runtime) | * | source |
 
+## Logs
+
+Logs from ASP.NET Core apps could be captured using two options. First option is
+to use HostingStartup and the second option is to rely on CLR Profiler
+Instrumentation. Recommended option for ASP.NET Core is to use HostingStartup.
+For other .NET Core apps logs could be collected only through .NET CLR Profiler
+Instrumentation.
```
pellared, line 64: This could be a note in the "Instrumented logging libraries and frameworks" section.

docs/config.md
```diff
@@ -55,6 +55,35 @@ for more details.
...
+
+### Parsing and Format Options for Logs
+
+These options could be used with both HostingStartup and .NET CLR Profiler logs
+enablement.
+
+| Environment variable | Description | Default value |
+|-|-|-|
+| `OTEL_DOTNET_AUTO_LOGS_PARSE_STATE_VALUES` | Whether the log state should be parsed. | `false` |
+| `OTEL_DOTNET_AUTO_LOGS_INCLUDE_FORMATTED_MESSAGE` | Whether the log state should be formatted. | `false` |
```
pellared, line 85: How about moving these to the `Exporters` section as this looks like a common config for the log exporters?

CHANGELOG.md
```diff
@@ -15,7 +15,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - `OTEL_LOGS_EXPORTER`,
   - `OTEL_DOTNET_AUTO_LOGS_CONSOLE_EXPORTER_ENABLED`,
   - `OTEL_DOTNET_AUTO_LOGS_PARSE_STATE_VALUES`,
-  - `OTEL_DOTNET_AUTO_LOGS_INCLUDE_FORMATTED_MESSAGE`,
+  - `OTEL_DOTNET_AUTO_LOGS_INCLUDE_FORMATTED_MESSAGE`.
+- Adds CLR Profiler based OpenTelemetry Logs support for .NET Core 3.1+.
```
pellared, line 19:
```suggestion
- Add ILogger logging instrumentation for .NET Core 3.1+.
```

pellared (approved):
I just added some comments to the docs. These can be addressed in a separate PR.

docs/config.md
rajkumar-rangaraj: There are no additional instrumentation options like traces or metrics, hence placed in a different section. Even if we add `Instrumented logging libraries and frameworks`, there will be only `AspNet` option. 

If I follow you correct, `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES` should get to source instrumentation and  .NET CLR Profiler part will move to `Instrumented logging frameworks` section. Is that correct?

docs/config.md
theletterf, line 64:
```suggestion
Logs from ASP.NET Core apps can be captured in two ways: The first option is
to use `HostingStartup`, and the second is to rely on CLR Profiler Instrumentation. 
For ASP.NET Core, use `HostingStartup`. For other .NET Core apps, you can only
collect logs through .NET CLR Profiler Instrumentation.
```

docs/config.md
pellared: I think it may be easier if we merge this and I will try to create a PR 😋 At least I think it will be easier and faster for me 😉

pellared (approved):
I am OK with resolving the conflicts and merging. The docs can be refined as a separate PR.

docs/config.md
rajkumar-rangaraj: If both are present `HostingStartup` adds the logger integration. There is no harm in having both of them.

Issue #1242, opened 2022-09-16
ReJIT race: 1st time execution may not be instrumented - detected via StrongNamedTests


pjanotti:
[Edit 01: updated the description to reflect the actual underlying issue]

The bytecode instrumentations are added via [requests to ReJIT](https://learn.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/icorprofilerinfo4-requestrejit-method) the targeted methods. Those requests are made when a module finishes loading at the callback [MethodLoadedFinished](https://learn.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/icorprofilercallback-moduleloadfinished-method). The actual ReJIT request happens in a dedicated thread (per docs recommendation) and until such request is completed there is a time window in which the method can be JIT compiled and executed without the expected bytecode instrumentation. The instrumentation is going to happen only on the first method invocation after the ReJIT request is completed.

4 instances of test failures showing this issue:

- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3069538719/jobs/4958307009#step:4:2590
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3069538719/jobs/4958306796#step:4:2682
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3069538719/jobs/4958306876#step:4:3392
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3062693118/jobs/4943951142#step:4:3392

pjanotti:
Confirming that is actually flakyness: macos and alpine passed on retry on PR #1234 (Windows did not complete yet)

pellared:
Some span is captured but it does not have the expected attribute: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3097103461/jobs/5013412328

pjanotti:
It seems that the issue is related to the ReJIT happening asynchronous. In a failed pass of the test we have:
```
[2022-09-22T02:47:31.803708579Z] [3580|3580] [debug] Enqueue for ReJIT [ModuleId=140150729282504, MethodDef=06000001, AppDomainId=94699831388896, IsDomainNeutral=0, Assembly=TestLibrary.InstrumentationTarget, Type=TestLibrary.InstrumentationTarget.Command, Method=Execute, Signature=200001]
[...]

PR #1264, opened 2022-09-22
Skip StrongNamedTests until reJITing is fixed

pellared:
## Why

It is extremly flaky and @pjanotti is already working on fixing the issue #1242

## What

Ignore the test so that we do not waste time and electricity.

### DIFF
--- a/test/IntegrationTests/StrongNamedTests.cs
+++ b/test/IntegrationTests/StrongNamedTests.cs
@@ -33,7 +33,7 @@ public class StrongNamedTests : TestHelper
     {
     }
 
-    [Fact]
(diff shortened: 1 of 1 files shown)


test/IntegrationTests/StrongNamedTests.cs
```diff
@@ -33,7 +33,7 @@ public StrongNamedTests(ITestOutputHelper output)
     {
     }
 
-    [Fact]
+    [Fact(Skip = "Very flaky. See: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1242")]
```
Kielek, line 36:
```suggestion
    [Fact(Skip = "Bug in product. See: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1242")]
```

Kielek (approved):

PR #1265, opened 2022-09-22
Add support for OTEL_DOTNET_AUTO_GRAPHQL_SET_DOCUMENT env var

### LINKED ISSUE #1256: GraphQL - disable sending document by default
We should give an abilty to explicitly enable/disable the addition of the `graphql.document` attribute by using an env var. IMO it should be disabled by default to avoid processing PII. It is fine for me if we just create an issue for it, but I would love to have it addressed in a timely manner 😉

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1222#discussion_r975160043_

Lets follow the pattern from SQL documentation.

RassK:
## Why

Fixes #1256 

## What

Does not send GraphQL Document tag (query) when disabled (the default behavior).

## Next

Proposing general key for PII protection to turn on or off all such's variables. `OTEL_DOTNET_AUTO_DISABLE_PII_PROTECTION`
While all subsets could be overridden separately, as all booleans are 3 state values.

## Tests

* Full tests (enabled).
* Safe tests (disabled).
* Default settings tests.

~Missing disabled integration tests. Not sure if we want to cover it now (to add second set of integration tests)?~

## Checklist

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -153,6 +153,12 @@ Important environment variables include:
 |-|-|-|
 | `OTEL_EXPORTER_ZIPKIN_ENDPOINT` | Zipkin URL. | `http://localhost:8126` |
 
+## Instrumentation options
+
+| Environment variable | Description | Default value |
+|-|-|-|
+| `OTEL_DOTNET_AUTO_GRAPHQL_SET_DOCUMENT` | Whether GraphQL instrumentation can pass raw queries as `graphql.document` attribute. This may contain sensitive information and therefore is disabled by default. | `false` |
+
 ## Additional settings
 
 | Environment variable | Description | Default value |
--- a/src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
@@ -107,7 +107,7 @@ public static class ConfigurationKeys
             /// Configuration key for GraphQL instrumentation to enable passing query as a document attribute.
             /// See <see cref="GraphQLTags.Document"/>.
             /// </summary>
-            public const string GraphQLEnableDocumentTag = "OTEL_DOTNET_AUTO_GRAPHQL_ENABLE_DOCUMENT_TAG";
+            public const string GraphQLSetDocument = "OTEL_DOTNET_AUTO_GRAPHQL_SET_DOCUMENT";
         }
     }
 
--- a/src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
@@ -16,6 +16,7 @@
 
 using System.Diagnostics;
 using System.Diagnostics.Metrics;
+using OpenTelemetry.AutoInstrumentation.Instrumentations.GraphQL;
 using OpenTelemetry.Logs;
 
 namespace OpenTelemetry.AutoInstrumentation.Configuration;
@@ -23,7 +24,7 @@ namespace OpenTelemetry.AutoInstrumentation.Configuration;
 /// <summary>
 /// Configuration keys
 /// </summary>
-public class ConfigurationKeys
+public static class ConfigurationKeys
 {
     /// <summary>
     /// Configuration key for the OTLP protocol to be used.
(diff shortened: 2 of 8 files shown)


src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
```diff
@@ -96,6 +97,18 @@ public static class Traces
...
         /// </summary>
         public const string LegacySources = "OTEL_DOTNET_AUTO_LEGACY_SOURCES";
+
+        /// <summary>
+        /// Configuration keys for instrumentation options.
+        /// </summary>
+        public static class InstrumentationOptions
+        {
+            /// <summary>
+            /// Configuration key for GraphQL instrumentation to enable passing query as a document tag.
```
pellared, line 107: nit: in OTel these are called attributes

```suggestion
            /// Configuration key for GraphQL instrumentation to enable passing query as a document attribute.
```

src/OpenTelemetry.AutoInstrumentation/Configuration/ConfigurationKeys.cs
```diff
@@ -96,6 +97,18 @@ public static class Traces
...
+            /// See <see cref="GraphQLTags.Document"/>.
+            /// </summary>
+            public const string GraphQLEnableDocumentTag = "OTEL_DOTNET_AUTO_GRAPHQL_ENABLE_DOCUMENT_TAG";
```
pellared, line 110: how about:

```csharp
public const string GraphQLSetDocument = "OTEL_DOTNET_AUTO_GRAPHQL_SET_DOCUMENT";
```

based on: https://github.com/open-telemetry/opentelemetry-dotnet/blob/84762870ddb18405d07a6cbf6a256fe831434c23/src/OpenTelemetry.Instrumentation.SqlClient/SqlClientInstrumentationOptions.cs#L83

pellared (commented):
Missing info in `config.md` 😉

pellared:
> Proposing general key for PII protection to turn on or off all such's variables. `OTEL_DOTNET_AUTO_DISABLE_PII_PROTECTION`
While all subsets could be overridden separately, as all booleans are 3 state values.

I really like the idea. It is possible that even the OTel spec would like something like that. I am just not sure about the name... Maybe ``OTEL_DOTNET_AUTO_SET_USERCONTROLLED_ATTRIBUTES` (I do not like this name as well). In general, this is about attributes whose value we cannot control (come from user input). In theory (and practice) even `http.url` can have PII data - https://support.google.com/adsense/answer/6156630?hl=en#:~:text=The%20most%20common%20pages%20with,associated%20with%20the%20user's%20information (I would say that it is bad that the spec says that it is a required attribute).

test/IntegrationTests/GraphQLTests.cs
```diff
@@ -73,6 +73,7 @@ public GraphQLTests(ITestOutputHelper output)
     public async Task SubmitsTraces()
     {
         SetEnvironmentVariable("OTEL_SERVICE_NAME", ServiceName);
+        SetEnvironmentVariable("OTEL_DOTNET_AUTO_GRAPHQL_SET_DOCUMENT", bool.TrueString);
```
Kielek, line 76: As it is our instrumentation, can we execute test twice with both values for the env. variable?

RassK: I was just between doing it now or doing it later after tests are stabilized. I don't have a strong opinion on it, so that's why the open question.

RassK: well added the missing set... let's see how it behaves.

pellared:
@theletterf PTAL

pellared (approved):

pjanotti (approved):

rajkumar-rangaraj (approved):

Kielek (approved):

nrcventura (approved):

docs/config.md
```diff
@@ -153,6 +153,12 @@ Important environment variables include:
 |-|-|-|
 | `OTEL_EXPORTER_ZIPKIN_ENDPOINT` | Zipkin URL. | `http://localhost:8126` |
 
+## Instrumentation options
+
+| Environment variable | Description | Default value |
+|-|-|-|
+| `OTEL_DOTNET_AUTO_GRAPHQL_SET_DOCUMENT` | Whether GraphQL instrumentation can pass raw queries as `graphql.document` attribute. This may contain sensitive information and therefore is disabled by default. | `false` |
```
pellared, line 160: It would not hurt to mention it in changlog especially that the default behavior is different

pellared: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1265/commits/55c0c0084f9d4540ce97f9ca3217071a155a9044

pellared:
@theletterf PTAL

Issue #1232, opened 2022-09-15
Dump all spans in the failing tests


RassK:
## What

Got a new idea that we can reuse our own solution to improve tests debugging. Currently we filter out all spans that we are interested in asserting on. If there are errors (flackyness) we just lose the information. 

I propose we dump all spans from the mock collector to the output if the test fails.

pjanotti:
Temporary workaround to investigate tests: enable the console exporter, on failure we dump the stdout anyway.

pellared:
Moved to next milestone as we have a workaround. I will try to solve this when working on https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/915

I want to create to create API that could be used in tests more-or-less like:

```csharp
public async Task SubmitTraces()
{
    using var collector = new MockSpanCollector();
    collector.Expect(attributes: new[] { K = "otel.library.name", V = "OpenTelemetry.SqlClient" });

    await RunTestApplication(collector, arguments: $"{_sqlServerFixture.Password} {_sqlServerFixture.Port}", enableClrProfiler: !IsCoreClr());
}
```

If it fails it would dump:

- missing expectation
- expectation met
- additional spans

Issue #915, opened 2022-06-30
Refactor assertions in integration tests


pellared:
Refactor the span and metrics assertions to make them more readable and maintainable.

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/901#discussion_r911136873_

PR #1269, opened 2022-09-22
Refine MockLogsCollector and LogTests

pellared:
## Why

An idea of how to address both https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/915 and https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1232. I took logs as an example because it has only one test class.

I think this might lead to more readable and maintainable code in the tests (the test code is simpler). Also, I think it will result in better test failure messages.

## What

Refine MockLogsCollector and LogTest. When the test fails it prints the missing expectations (user-defined messages) and what was collected. I was inspired by @RassK "expectations" in `GraphQLTests`.

If you like it then the same pattern could be used for other signals.

Also, it should be "faster" as we are not using `Thread.Sleep` nor `Task.Delay`. Here we are using a `BlockingCollection` for synchronization.

### DIFF
--- a/test/IntegrationTests/Helpers/MockLogsCollector.cs
+++ b/test/IntegrationTests/Helpers/MockLogsCollector.cs
@@ -31,9 +31,10 @@ public class MockLogsCollector : IDisposable
 {
     private static readonly TimeSpan DefaultWaitTimeout = TimeSpan.FromMinutes(1);
 
-    private readonly BlockingCollection<string> _logs = new(100); // bounded to avoid memory leak
     private readonly ITestOutputHelper _output;
     private readonly TestHttpListener _listener;
+    private readonly BlockingCollection<global::OpenTelemetry.Proto.Logs.V1.LogRecord> _logs = new(100); // bounded to avoid memory leak
+    private List<Expectation> _expectations = new();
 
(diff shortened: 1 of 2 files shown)


nrcventura (approved):
I think that this makes the simpler tests easier to read. 

This approach may not work as well for more complex instrumentation, where we need to validate that our code is setting data correctly on more than one thing in the otlp object structure. For example, validating scope attributes in addition to attributes on the data object (span, metric, log record). 

This might not work well either if we add metrics support to the byte-code instrumentation that we have for graphql, where we will probably want to validate both the metric and individual data points.

pellared:
> This approach may not work as well for more complex instrumentation, where we need to validate that our code is setting data correctly on more than one thing in the otlp object structure. For example, validating scope attributes in addition to attributes on the data object (span, metric, log record).

We can add additional methods like `ExpectScope()` and `AssertScopeExpectations()`.

> This might not work well either if we add metrics support to the byte-code instrumentation that we have for graphql, where we will probably want to validate both the metric and individual data points.

I am not sure why yet. But maybe let's wait until the problem arrives 😉

Kielek (approved):

PR #1279, opened 2022-09-26
Force ReJIT before 1st execution of a method

### LINKED ISSUE #1242: ReJIT race: 1st time execution may not be instrumented - detected via StrongNamedTests

pjanotti:
## Why

Enqueueing requests to ReJIT can cause the first executions of a method targeted for bytecode instrumentation to not be instrumented.

Fixes #1242

## What

Make the request to ReJIT a method targeted for bytecode instrumentation to be made on the same thread finishing the module load targeted for instrumentation. This ensures that any execution of the targeted methods is going to use the ReJIT-ed version of the method. 

## Tests

Manual runs on mac and Windows without hitting the error - these errors have low frequency on more powerful machines but runs in ci and verify-test workflows were also good.

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation.Native/rejit_handler.cpp
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/rejit_handler.cpp
@@ -118,7 +118,8 @@ void RejitHandlerModuleMethod::RequestRejitForInlinersInModule(ModuleID moduleId
             methodEnum = nullptr;
             if (total > 0)
             {
-                handler->EnqueueForRejit(modules, methods);
+                // handler->EnqueueForRejit(modules, methods);
+                handler->RequestRejit(modules, methods);
                 Logger::Info("NGEN:: Processed with ", total, " inliners [ModuleId=", currentModuleId,
                              ",MethodDef=", currentMethodDef, "]");
             }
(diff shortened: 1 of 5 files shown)


rajkumar-rangaraj (approved):

src/OpenTelemetry.AutoInstrumentation.Native/rejit_handler.cpp
```diff
@@ -370,35 +288,41 @@ void RejitHandler::AddNGenModule(ModuleID moduleId)
...
-}
+    HRESULT hr;
+    auto profilerInfo = m_profilerInfo;
+    auto profilerInfo10 = m_profilerInfo10;
 
-void RejitHandler::Shutdown()
-{
-    m_rejit_queue->push(RejitItem::CreateEndRejitThread());
-    if (m_rejit_queue_thread->joinable())
+    if (profilerInfo10 != nullptr)
```
Kielek, line 305: What is the reason to have this condition here?
`m_profilerInfo` is `ICorProfilerInfo4` which is interface containing `RequestReJIT` fuction https://learn.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/icorprofilerinfo4-requestrejit-method.

Do you want to use here `RequestReJITWithInliners ` https://learn.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/icorprofilerinfo10-requestrejitwithinliners-method or you have some other reason?

RassK:
Any ideas what was the original idea about queuing rejit requests? Any possible skeletons when changing from async to sync?
(Just about to gain some knowledge on this)

pjanotti:
@RassK 
> Any ideas what was the original idea about queuing rejit requests? Any possible skeletons when changing from async to sync?

The docs state that requesting a ReJIT can cause a GC plus if requesting for everything (or almost everything) in a module it could be an extensive list. In sync mode that can delay finishing the ModuleLoadFinished thread and no code from that module will be executed until all requests are made. In our case, we typically instrument a small number of methods of a module, and more importantly, we need to be consistent and not miss a single execution of the method, i.e.: I don't think we can allow that to happen in name of a perhaps slightly quicker time from module load to first method execution.

nrcventura (approved):

src/OpenTelemetry.AutoInstrumentation.Native/rejit_handler.cpp
pjanotti: Good point @Kielek, ICorProfilerInfo4 is .NET Framework 4.5 so we don't need to worry about that version anymore, we can consider that we have ICorProfilerInfo6 and above. Cleaning it up...

pjanotti: We can remove this multiple profiler interface members but it is better to do in a separate PR. I will do it as soon as this one is merged.

PR #1295, opened 2022-09-28
Stabilize and speedup LogTests

pellared:
## Why

The tests sometimes failed. I think that the pattern used in `HttpTests` seems to be more stable (and faster). Also, it makes the test code simpler.

## What

Change the `TestApplication.Logs` to call itself and exit.

## Tests

CI


rajkumar-rangaraj (approved):
LGTM.

Issue #1296, opened 2022-09-28
IntegrationTests on Linux are sometimes hanging


pellared:
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3140353105/jobs/5101662658
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3141964977/jobs/5105009137
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3143906850/jobs/5109224240

It might be connected with changes introduced here: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1276

pjanotti:
@pellared I'm not seeing this happening, but, caveat: querying GH actions doesn't get me an automated count. I looked at the last 5 days failures and didn't see this one. I did see some hanging on Windows, many failures of a Prometheus test (most on Linux), and one or another that seem "random". I will open issues for the mentioned failures, unless we have a repro for this one it seems time to close it - we can always reopen if needed.

PR #1298, opened 2022-09-28
Add diagnostic messages and hang dumps for IntegrationTests

pellared:
## Why

Add more verbose output when running integration tests so that we can have more info when troubleshooting e.g. deadlocks in tests. Part of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1296

## What

Applies for `IntegrationTests`

- Add diagnostic messages when starting and finishing the test (output inspired by `go test`)
- Add diagnostics messages if a single test is taking longer than 1 minute (a message is printed each 1 minute)
- Crash and dump if a single test is taking more than 3 minutes. The dumps are stored under `build_data\results\IntegrationTests` and are stored as artifacts in `ci` GH workflow
- Configure xunit to run sequentially (more readable output, should be more pleasant for GH runners) - may take a little longer (~ 3 minutes right now)  - **I CAN REVERT THIS CHANGE**

Sample output: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3147118594/jobs/5116310304#step:4:2635

References:
- https://github.com/dotnet/docs/blob/main/docs/core/tools/dotnet-test.md
- https://xunit.net/docs/configuration-files#diagnosticMessages
- https://andrewlock.net/tracking-down-a-hanging-xunit-test-in-ci-building-a-custom-test-framework/

### DIFF
--- a/test/IntegrationTests/IntegrationTests.csproj
+++ b/test/IntegrationTests/IntegrationTests.csproj
@@ -14,7 +14,7 @@
 
   <ItemGroup>
     <PackageReference Include="Google.Protobuf" Version="3.21.6" />
-    <PackageReference Include="Grpc.Tools" Version="2.49.0" PrivateAssets="all" />
+    <PackageReference Include="Grpc.Tools" Version="2.49.1" PrivateAssets="all" />
     <PackageReference Include="Microsoft.Data.SqlClient" Version="5.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageReference Include="StrongNamer" Version="0.2.5" Condition="$(TargetFramework.StartsWith('net4'))" />
--- a/test/IntegrationTests/IntegrationTests.csproj
+++ b/test/IntegrationTests/IntegrationTests.csproj
@@ -27,6 +27,10 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Content Include="xunit.runner.json" CopyToOutputDirectory="PreserveNewest" />
+  </ItemGroup>
+
+  <ItemGroup>
     <ProjectReference Include="..\..\src\OpenTelemetry.AutoInstrumentation\OpenTelemetry.AutoInstrumentation.csproj" />
   </ItemGroup>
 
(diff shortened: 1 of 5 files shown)


test/IntegrationTests/IntegrationTests.csproj
```diff
@@ -26,6 +26,10 @@
     <PackageReference Include="xunit.abstractions" Version="2.0.3" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Content Include="xunit.runner.json" CopyToOutputDirectory="PreserveNewest" />
+  </ItemGroup>
```
pellared, line 31: We can consider adding it to `Directory.Build.props` in `/test`

pjanotti:
> Add diagnostic messages when starting and finishing the test (output inspired by go test)

I assume that `--logger` doesn't offer something similar, i.e.: it is either too terse or too verbose, right?

pellared:
> > Add diagnostic messages when starting and finishing the test (output inspired by go test)
> 
> I assume that `--logger` doesn't offer something similar, i.e.: it is either too terse or too verbose, right?

I think you are right that `--logger "console;verbosity=detailed"` should be good enough 👍

test/IntegrationTests/VerboseTestFramework.cs
```diff
@@ -0,0 +1,138 @@
...
+            }
+
+            var test = $"{TestMethod.TestClass.Class.Name}.{TestMethod.Method.Name}({parameters})";
+            _diagnosticMessageSink.OnMessage(new DiagnosticMessage($"=== RUN   {test}"));
+
+            try
+            {
+                var result = await base.RunTestCaseAsync(testCase);
+                var status = result.Failed > 0 ? "FAIL" : "PASS";
+                _diagnosticMessageSink.OnMessage(new DiagnosticMessage($"--- {status}: {test} ({result.Time:0.00}s)"));
```
RassK, line 128: just an idea > 
"--- PASS: Test"
"xxx FAIL: Test"

(use xxx if fails instead of ---, makes it visually more catchable)

RassK (approved):

nrcventura (approved):

test/IntegrationTests/VerboseTestFramework.cs
pjanotti: I like this idea.

pjanotti (approved):

rajkumar-rangaraj (approved):

PR #1309, opened 2022-09-29
Fix message in MockMetricsCollector

dszmigielski:
## Why

Wrong type name in `MockMetricsCollector`

## What

Fix type name to match reality.


pellared (approved):
🎉

RassK (approved):

Issue #1282, opened 2022-09-27
Null Pointer Exception when using bytecode instrumentation


muhaook:
# Bug Report

## Symptom

**Describe the bug**

I am implementing our own bytecode instrumentation and sometimes hit NPE:

```
[2022-09-27T13:52:14.9181394Z] [Error] Object reference not set to an instance of an object. 
Exception: Object reference not set to an instance of an object.
System.NullReferenceException: Object reference not set to an instance of an object.
   at OpenTelemetry.AutoInstrumentation.Util.ActivityHelper.StartActivityWithTags(ActivitySource activitySource, String operationName, ActivityKind kind, ITags tags)
   at OpenTelemetry.AutoInstrumentation.Instrumentations.Odp.OdpCommon.OnMethodBeginConnection[TTarget](TTarget instance, String spanNamePrefix)
   at OdpOpenConnectionIntegration.OnMethodBegin(OracleConnection )
   at OpenTelemetry.AutoInstrumentation.CallTarget.Handlers.BeginMethodHandler`2.Invoke(TTarget instance)
   at OpenTelemetry.AutoInstrumentation.CallTarget.CallTargetInvoker.BeginMethod[TIntegration,TTarget](TTarget instance)
   at Oracle.ManagedDataAccess.Client.OracleConnection.Open()
```

Looked into source code and I found [this line](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/de45afdb36733468fa2e78a7ba7197d54abc0c73/src/OpenTelemetry.AutoInstrumentation/Util/ActivityHelper.cs#L42) . It is possible that activitySource.StartActivity() returns null. So we'd better check for it in [the following logic](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/de45afdb36733468fa2e78a7ba7197d54abc0c73/src/OpenTelemetry.AutoInstrumentation/Util/ActivityHelper.cs#L49).

**Runtime environment:**

- OpenTelemetry Automatic Instrumentation version: 0.3.1
- OS: Windows 10
- .NET version: .NET Framework 4.6.2

RassK:
@muhaook The ActivitySource.StartActivity() returns null if there is no listener. The null check is going to hide the actual problem.

muhaook:
BTW, I did not get notified by email whenever I got @-ed or my git issue had new comment. Does anyone know how to subscribe for that? Thanks

[...]

PR #1321, opened 2022-09-30
Replace span.kind tag by ActivityKind

### LINKED ISSUE #1283: ActivityKind is always 'internal' for bytecode intrumentation
# Bug Report

## Symptom

**Describe the bug**
[activitySource.StartActivity(operationName)](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/de45afdb36733468fa2e78a7ba7197d54abc0c73/src/OpenTelemetry.AutoInstrumentation/Util/ActivityHelper.cs#L42) will use `internal` as default ActivityKind and it can not be changed after activity is started. 

I tried to add tag with ActivityKind = client, it turned out that there will be 2 `kind` tags in the final activity: one is `internal`; the other is `client`.

So suggest to add input argument ActivityKind to [activitySource.StartActivity()](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/de45afdb36733468fa2e78a7ba7197d54abc0c73/src/OpenTelemetry.AutoInstrumentation/Util/ActivityHelper.cs#L42)

**Runtime environment:**

- OpenTelemetry Automatic Instrumentation version: 0.3.1
- OS: Windows 10
- .NET version: .NET Framework 4.6.2

Kielek:
## Why

Fixes #1283

## What

Replace `span.kind` tag by `ActivityKind`
It is serialized by as span `kind` property.

## Tests

CI + extended tests.

## Checklist

## Notes
There were no assertions on `WebServerSpanExpectation` so I have removed this class.


Kielek:
@muhaook, thanks for reporting #1283. Could you please verify proposed changes?

src/OpenTelemetry.AutoInstrumentation/Util/ActivityHelper.cs
```diff
@@ -70,4 +59,20 @@ internal static void DisposeWithException(this Activity activity, Exception exce
             }
         }
     }
+
+    private static Activity StartActivityWithTags(this ActivitySource activitySource, string operationName, ActivityKind kind, ITags tags)
+    {
+        var activity = activitySource.StartActivity(operationName, kind);
+
```
pellared, line 66: How about adding

```suggestion
        var activity = activitySource.StartActivity(operationName, kind);
        if (activity == null)
        {
            return null;
        }
```

to reduce the possibility of throwing a NullPointerException? 

It should fix https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1282

I think we should be very careful and make sure we never throw an exception at runtime.

pellared (approved):

nrcventura (approved):

muhaook:
looks good to me[👍](https://whatemoji.org/blog/thumbs-up-emoji-guide/#:~:text=The%20%F0%9F%91%8D%20Thumbs%20Up%20emoji%20is%20optimistic.%20Everything,that%20you%20like%20something%20and%20approve%20of%20it.) Thanks

src/OpenTelemetry.AutoInstrumentation/Instrumentations/GraphQL/GraphQLCommon.cs
```diff
@@ -61,7 +61,7 @@ internal static Activity CreateActivityFromExecuteAsync(IExecutionContext execut
                 tags.Document = query;
             }
 
-            activity = ActivitySource.StartActivityWithTags(operation, tags);
+            activity = ActivitySource.StartServerActivityWithTags(operation, tags);
```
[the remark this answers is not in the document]
pjanotti: In principle I think that we should keep it as a server span: the instrumentation should be independent of each other. Anyway, let's handle that separately. We can talk about it at the next SIG meeting.

src/OpenTelemetry.AutoInstrumentation/Util/ActivityHelper.cs
```diff
@@ -70,4 +59,20 @@ internal static void DisposeWithException(this Activity activity, Exception exce
             }
         }
     }
+
+    private static Activity StartActivityWithTags(this ActivitySource activitySource, string operationName, ActivityKind kind, ITags tags)
```
pjanotti, line 63: Why not make this one public and remove `StartServerActivityWithTags`?

pjanotti (approved):
LGTM. I have a suggestion but it is fine either way.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/GraphQL/GraphQLCommon.cs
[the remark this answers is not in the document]
pjanotti: Note that this slightly differs from the grpc client case suppressing the HttpClient spans under it.

src/OpenTelemetry.AutoInstrumentation/Util/ActivityHelper.cs
Kielek: Sure, lets put it into separate PR. I know that it will be small, but maybe there will be a chance to somehow test it.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/GraphQL/GraphQLCommon.cs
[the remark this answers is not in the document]
Kielek: #1322

src/OpenTelemetry.AutoInstrumentation/Util/ActivityHelper.cs
Kielek: I do not have strong opinion about api shape. Fixed in df41937244b183b81854f6758ea7f3f4c00cd2ba

Issue #1310, opened 2022-09-29
Add the possiiblity to disable Microsoft.Extensions.Logging instrumentation when CLR Profiler is enabled


pellared:
I think we should add support for `OTEL_DOTNET_AUTO_LOGS_DISABLED_INSTRUMENTATIONS` and  `OTEL_DOTNET_AUTO_LOGS_ENABLED_INSTRUMENTATIONS`. It would require changes in native code and `integrations.json` to define the "type" of bytecode instrumentation.

The same could be done to enable metrics bytecode instrumentation.

Side note: Is it possible to have two bytecode instrumentations for the same API? So that we can have e.g. separate bytecode GraphQL instrumentation for metrics and spans.

Moreover, I am not sure if `LoggingBuilder` is a good/"unique" name/id. Maybe `MicrosoftExtensionsLogging` or `ILogger`. The second one is not so unique but I think more obvious than `LoggingBuilder`.

Related changes: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1324

**EDITED** (kudos for @RassK)

pellared:
CC @rajkumar-rangaraj

RassK:
There's actually no way to disable it.

RassK:
> Right now, the "hack" to disable Microsoft.Extensions.Logging instrumentation when CLR Profiler is enabled is to set OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS=LoggingBuilder.

There is actually no way to disable it. Native layer is fine to have  `LoggingBuilder` in `OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS`  but managed layer crashes with unsupported option, due `TracerInstrumentation` enum does not contain the value.

rajkumar-rangaraj:
> Moreover, I am not sure if LoggingBuilder is a good/"unique" name/id. Maybe MicrosoftExtensionsLogging or ILogger. The second one is not so unique but I think more obvious than LoggingBuilder.

`LoggingBuilder` is an internal implementation, dotnet users know this feature as `ILogger`. Better apt name would be `ILogger`.

PR #1324, opened 2022-10-03
Native code supports OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS

### LINKED ISSUE #1323: Native code supports OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS
Update:
It may be actually bug. 
Managed layer implements `OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS` and `OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS`
Native layer is missing `OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS`

Native layer instead has only `OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS`. So when you explicitly don't enable instrumentation, it will be disabled in managed layer... but native layer doesn't know it.

_Originally posted by @RassK in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1282#issuecomment-1261913253_

Kielek:
## Why

Fixes #1323

## What

Native code supports `OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS`.

## Tests

Extended tests

## Checklist


src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.cpp
```diff
@@ -127,31 +127,37 @@ namespace
             return;
         }
 
-        if (!enabledIntegrationNames.empty())
+        // ignore LoggingBuilder, it is technically non-instrumentation
+        // it is using only the same functionality
+        if (name != WStr("LoggingBuilder"))
```
pellared, line 132: Related: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1310 

Consider leaving a comment referring to the issue.

Kielek: fix 46a0800753ad076f8b78c17594f98aa09bb1986b

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.cpp
```diff
@@ -114,12 +127,37 @@ namespace
...
+                }
+            }
+
+            // check if the integration is disabled
+            for (const WSTRING& disabledName : disabledIntegrationNames)
+            {
+                if (name == disabledName)
+                {
+                    return;
+                }
```
pellared, line 161: consider extracting this code to a method that returns `true` if the instrumentation is enabled

src/OpenTelemetry.AutoInstrumentation.Native/environment_variables.h
```diff
@@ -34,9 +34,15 @@ const WSTRING include_process_names = WStr("OTEL_DOTNET_AUTO_INCLUDE_PROCESSES")
 // "MyApp.exe,dotnet.exe"
 const WSTRING exclude_process_names = WStr("OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES");
 
-// Sets a list of integrations to disable. All other integrations will remain
-// enabled. If not set (default), all integrations are enabled. Supports
-// multiple values separated with comma, for example:
+// Sets a list of integrations to enable. If not set (default), all integrations are enabled.
+// Supports multiple values separated with comma, for example:
+// "ElasticsearchNet,AspNetWebApi2"
+const WSTRING enabled_integrations =
```
pellared, line 40: add this here: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/6c20a738c2ffd92ba5a3e374d0c571552d3c8bd5/src/OpenTelemetry.AutoInstrumentation.Native/otel_profiler_constants.h#L12

src/OpenTelemetry.AutoInstrumentation.Native/otel_profiler_constants.h
```diff
@@ -15,6 +15,7 @@ const WSTRING env_vars_to_display[]{environment::clr_profiler_enabled,
                                     environment::integrations_path,
                                     environment::include_process_names,
                                     environment::exclude_process_names,
+                                    environment::enabled_integrations,
```
pellared, line 18: It would be better to log all env vars having following prefixes:
- `COR_`
- `DOTNET_`
- `OTEL_`

pellared: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1325

pellared (approved):
👍

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.cpp
Kielek: ce70090976a5528456e467ee52058bdc486e1531

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.cpp
```diff
@@ -108,6 +108,39 @@ void LoadIntegrationsFromStream(
...
+        {
+            return true;
+        }
+
+        // check if the integration is disabled
+        for (const WSTRING& disabledName : disabledIntegrationNames)
+        {
+          if (name == disabledName) {
+            return false;
+          }
```
pellared, line 127: nit: "fix" formatting 😄

Kielek: 697b5a3170cf30d1667e4fb2248d356d5ba93b44

pellared (approved):

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.cpp
```diff
@@ -98,8 +108,45 @@ void LoadIntegrationsFromStream(std::istream& stream, std::vector<IntegrationMet
...
+          {
+              return false;
+          }
+        }
+
+        if (enabledIntegrationNames.empty())
+        {
+            return true;
+        }
+        // check if the integration is enabled
```
rajkumar-rangaraj, line 135: Should this be other way, first check enabled Integration and then check for disabled ones? 
In the managed layer we check enabled instrumentations first - https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/d6c9ffad4de5686ec856a3c6ca1e243928e9f44f/src/OpenTelemetry.AutoInstrumentation/Util/ConfigurationSourceExtensions.cs#L26

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.cpp
```diff
@@ -55,8 +63,10 @@ void LoadIntegrationsFromFile(const WSTRING& file_path, std::vector<IntegrationM
     }
 }
 
-void LoadIntegrationsFromStream(std::istream& stream, std::vector<IntegrationMethod>& integrationMethods, const std::vector<WSTRING>& disabledIntegrationNames)
-{
+void LoadIntegrationsFromStream(
+    std::istream& stream, std::vector<IntegrationMethod>& integrationMethods,
```
pjanotti, line 67: Since the previous method you put all parameters in their own line do the same here.
```suggestion
    std::istream& stream,
    std::vector<IntegrationMethod>& integrationMethods,
```

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.cpp
```diff
@@ -98,8 +108,45 @@ void LoadIntegrationsFromStream(std::istream& stream, std::vector<IntegrationMet
 
 namespace
 {
+    bool InstrumentationEnabled(const WSTRING name, const std::vector<WSTRING>& enabledIntegrationNames, const std::vector<WSTRING>& disabledIntegrationNames)
```
pjanotti, line 111: This line is pretty long, suggestion: let's keep putting each parameter on its own line.

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.cpp
```diff
@@ -98,8 +108,45 @@ void LoadIntegrationsFromStream(std::istream& stream, std::vector<IntegrationMet
...
+    {
+        // LoggingBuilder has to be always enabled/
```
pjanotti, line 113:
```suggestion
        // LoggingBuilder has to be always enabled.
```

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.cpp
```diff
@@ -98,8 +108,45 @@ void LoadIntegrationsFromStream(std::istream& stream, std::vector<IntegrationMet
...
+        // Technically it is  non-instrumentation but
```
pjanotti, line 114:
```suggestion
        // Technically it is not an instrumentation but
```

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.cpp
```diff
@@ -98,8 +108,45 @@ void LoadIntegrationsFromStream(std::istream& stream, std::vector<IntegrationMet
...
+        if (name == WStr("LoggingBuilder"))
+        {
+            return true;
+        }
 
-    void IntegrationFromJson(const json::value_type& src, std::vector<IntegrationMethod>& integrationMethods, const std::vector<WSTRING>& disabledIntegrationNames)
+        // check if the integration is disabled
+        for (const WSTRING& disabledName : disabledIntegrationNames)
+        {
+          if (name == disabledName)
```
pjanotti, line 125: The indentation here is using 2 spaces.

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.h
```diff
@@ -17,18 +17,31 @@ using json = nlohmann::json;
...
+void LoadIntegrationsFromFile(
+    const WSTRING& file_path,
+    std::vector<IntegrationMethod>& integrationMethods,
+    const std::vector<WSTRING>& enabledIntegrationNames,
+    const std::vector<WSTRING>& disabledIntegrationNames);
 
 // LoadIntegrationsFromFile loads the integrations from a stream
-void LoadIntegrationsFromStream(std::istream& stream, std::vector<IntegrationMethod>& integrationMethods, const std::vector<WSTRING>& disabledIntegrationNames);
+void LoadIntegrationsFromStream(
+    std::istream& stream, std::vector<IntegrationMethod>& integrationMethods,
```
pjanotti, line 34:
```suggestion
    std::istream& stream,
    std::vector<IntegrationMethod>& integrationMethods,
```

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.cpp
```diff
@@ -98,8 +108,45 @@ void LoadIntegrationsFromStream(std::istream& stream, std::vector<IntegrationMet
...
+          if (name == disabledName) 
+          {
+              return false;
+          }
+        }
+
+        if (enabledIntegrationNames.empty())
+        {
+            return true;
+        }
```
pjanotti, line 134: nit: add blank line.

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.cpp
pellared: I think that the output of both "algorithms" is the same 🤔

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.h
```diff
@@ -17,18 +17,31 @@ using json = nlohmann::json;
...
+    const std::vector<WSTRING>& enabledIntegrationNames,
+    const std::vector<WSTRING>& disabledIntegrationNames);
 
 namespace
 {
 
-    void IntegrationFromJson(const json::value_type& src, std::vector<IntegrationMethod>& integrationMethods, const std::vector<WSTRING>& disabledIntegrationNames);
+    void IntegrationFromJson(const json::value_type& src,
+                         std::vector<IntegrationMethod>& integrationMethods,
```
pjanotti, line 42: nit: it seems that there is an extra space here.

src/OpenTelemetry.AutoInstrumentation.Native/environment_variables.h
```diff
@@ -34,9 +34,15 @@ const WSTRING include_process_names = WStr("OTEL_DOTNET_AUTO_INCLUDE_PROCESSES")
...
+    WStr("OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS");
+
+// Sets a list of integrations to disable. Status of other integrations will remain
+// unchanged. Calculation order: OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS then OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS.
```
pjanotti, line 44: The code checks first for disabled, so it seems more concise to state that disabled takes precedence.

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.cpp
rajkumar-rangaraj: May not be same. If I have same entries in both enabled and disabled instrumentation, it will give different results. If we have entry in both places we should disable the instrumentation.

pjanotti (commented):
Looks good a bunch of nits though.

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.cpp
pellared: Agree and this is what this code does as well 😉 If `return false` and even does not check if it is enabled.

pjanotti:
I didn't look at the managed, but, I agree with Raj: in general the "disabled" should take precedence. The case that matters is when both enabled and disabled are specified. In this case treating the disabled as more "relevant" is typical - to do different we need a reason to break that expectation.

src/OpenTelemetry.AutoInstrumentation.Native/integration_loader.cpp
rajkumar-rangaraj: Thanks for correcting me. I completely missed we are just checking based on single entry `name`.

rajkumar-rangaraj (approved):

zacharycmontoya (approved):
Looks good!

Kielek:
@pjanotti, I think that I resolved all formatting issues. If there is something else we should think about separate issue to reformat code and introduce strict rules for VS/VS Code autoformater.

src/OpenTelemetry.AutoInstrumentation.Native/environment_variables.h
Kielek: Fixed: 012ad709d14f7b0fea7262ffc84414c9c9ac4136

Kielek:
> I didn't look at the managed, but, I agree with Raj: in general the "disabled" should take precedence. The case that matters is when both enabled and disabled are specified. In this case treating the disabled as more "relevant" is typical - to do different we need a reason to break that expectation.

Managed code produces same result (disabled is always disabled, enabled can be overwritten by disabled). There is just different  order of calculation.

pellared:
I created https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1356 to track down the formatting issue.

pellared (approved):

PR #1354, opened 2022-10-03
Add instrument.sh script

pellared:
## Why

Create a script to help set up the auto-instrumentation. It should reduce the possibility of human errors and make it faster+easier to try out.

## What

- Add validation to `download.sh`
- Add `instrument.sh` script to facilitate instrumenting .NET apps (for *nix)

## Tests

- CI (Ubuntu, macOS, Windows) 🎉 
- Locally Ubuntu in WSL

## Checklist


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -35,7 +35,7 @@ Steps to reproduce the behavior:
  3. etc
 
 <!--
-NOTE: If possible, also include the tracer log files found (by default) in:
-Windows: `%ProgramData%\OpenTelemetry .NET AutoInstrumentation\logs\`
-Linux: /var/log/opentelemetry/dotnet/
+NOTE: If possible, also include the internal log files found (by default) in:
```
pjanotti, line 38:
```suggestion
NOTE: If possible, also include the log files found (by default) in:
```

.github/workflows/ci.yml
```diff
@@ -77,26 +77,43 @@ jobs:
         include:
           - machine: windows-2022
             distribution: windows
+            log-dir: "/c/ProgramData/OpenTelemetry .NET AutoInstrumentation/logs"
```
pjanotti, line 80: I was surprised by the unix-like syntax here, but if it works 🤷‍♂️

.github/workflows/ci.yml
```diff
@@ -77,26 +77,43 @@ jobs:
...
           - machine: ubuntu-20.04
             distribution: linux-glibc
+            log-dir: "/var/log/opentelemetry/dotnet"
           - machine: ubuntu-20.04
             distribution: linux-musl
+            test-instrument: "false"
```
pjanotti, line 86: Why is this false here?

instrument.sh
```diff
@@ -0,0 +1,99 @@
...
+    echo "Invalid ENABLE_PROFILING env var. Supported values: true, false." >&2
+    return 2
+    ;;
+esac
+
+# set defaults
+test -z "$INSTALL_DIR" && INSTALL_DIR="./otel-dotnet-auto"
+
+
+test -z "$ENABLE_PROFILING" && ENABLE_PROFILING="1"
```
pjanotti, line 29: I think it is worth a comment why this is necessary.

instrument.sh
```diff
@@ -0,0 +1,99 @@
...
+      SUFIX="dll"
+      ;;
+    *)
+      echo "BUG: Unknown distribution. Please submit an issue in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation." >&2
+      return 1
+      ;;
+  esac
+
+  # Enable .NET Framework Profiling API
+  export COR_ENABLE_PROFILING="$ENABLE_PROFILING"
```
pjanotti, line 68: These exports here are windows only.

instrument.sh
pellared: it is not :) removing

pellared: U

.github/workflows/ci.yml
pellared: I think that `linux-musl` should be tested on Alpine. I will try to test the Alpine in next PR by using `build/nuke/docker/Alpine.dockerfile`

docs/README.md
```diff
@@ -172,6 +174,23 @@ CORECLR_PROFILER_PATH_64
...
+| Parameter          | Description                                                            | Required | Default value        |
+|--------------------|------------------------------------------------------------------------|----------|----------------------|
+| `DISTRIBUTION`     | Possible values: `linux-glibc`, `linux-musl`, `macos`, `windows`       | Yes      |                      |
+| `ENABLE_PROFILING` | Whether to set the .NET CLR Profiler, possible values: `true`, `false` | No       | `true`               |
+| `INSTALL_DIR`      | Location where binaries are to be installed                            | No       | `./otel-dotnet-auto` |
+
+> On macOS [`coreutils`](https://formulae.brew.sh/formula/coreutils) is required.
+
+```sh
+curl -fL https://raw.githubusercontent.com/open-telemetry/opentelemetry-dotnet-instrumentation/main/instrument.sh -O
```
RassK, line 189: I think this should point to the release artifact instead of to the actual source. Specially when main is not a release tree.

docs/README.md
```diff
@@ -172,6 +174,23 @@ CORECLR_PROFILER_PATH_64
 OTEL_DOTNET_AUTO_INTEGRATIONS_FILE
 ```
 
+You can also use the [instrument.sh](../instrument.sh) script which uses following
+environment variables as parameters:
+
+| Parameter          | Description                                                            | Required | Default value        |
+|--------------------|------------------------------------------------------------------------|----------|----------------------|
+| `DISTRIBUTION`     | Possible values: `linux-glibc`, `linux-musl`, `macos`, `windows`       | Yes      |                      |
+| `ENABLE_PROFILING` | Whether to set the .NET CLR Profiler, possible values: `true`, `false` | No       | `true`               |
```
pjanotti, line 183: I'm not sure if it is going to get too verbose to add it here, but, the important fact in this case is that bytecode instrumentation requires profiling.

pjanotti: We can improve the text here later.

pjanotti (approved):

docs/README.md
pellared: It is described ~20 lines above. I know that it is not ideal 😉

docs/README.md
pellared: I am not sure as any `Writer` (approver) can change/remove/add release artifacts. The `main` branch is extremely protected - requires GPG signatures, also only the maintainers can merge/push to `main. We can consider changing it to a "release tag" during the release process. WDYT?

docs/README.md
```diff
@@ -172,6 +174,23 @@ CORECLR_PROFILER_PATH_64
 OTEL_DOTNET_AUTO_INTEGRATIONS_FILE
 ```
 
+You can also use the [instrument.sh](../instrument.sh) script which uses following
```
zacharycmontoya, line 177: You might consider breaking the "Instrument a .NET application" section into two smaller sections: 1) Environment variables and 2) Setup script. Based on my reading, that would help make sure someone who already sets the environment variables at the top doesn't try to run the instrument script when they don't have to

zacharycmontoya (approved):
LGTM with one suggestion

docs/README.md
pellared: Tomorrow I will try to create a separate PR to address it 👍

docs/README.md
pjanotti: Let's work to improve it later.

docs/README.md
pellared: @RassK PTAL https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1357

PR #1355, opened 2022-10-04
Reduce smoke tests duration

pjanotti:
## Why

Reduce smoke tests by 2 minutes.

## What

Negative tests are waiting 2 minutes without the need for that. This change simply leverages the fact that for the smoke tests the instrumented app is already finished when calling WaitForSpans.

## Tests

- Multiple runs of smoke tests: https://github.com/pjanotti/opentelemetry-dotnet-instrumentation/actions/runs/3178622774
- Clean CI pass: https://github.com/pjanotti/opentelemetry-dotnet-instrumentation/actions/runs/3178579056

## Checklist

### DIFF
--- a/test/IntegrationTests/Helpers/MockZipkinCollector.cs
+++ b/test/IntegrationTests/Helpers/MockZipkinCollector.cs
@@ -104,7 +104,10 @@ public class MockZipkinCollector : IDisposable
 
         IImmutableList<IMockSpan> relevantSpans = ImmutableList<IMockSpan>.Empty;
 
-        while (DateTime.Now < deadline)
+        // Use a do-while to ensure at least one attempt at reading the data already
+        // received. This is helpful for negative tests, ie.: no spans generated, when
+        // the process emitting the spans already finished.
+        do
         {
             lock (_syncRoot)
             {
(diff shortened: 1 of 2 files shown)


test/IntegrationTests/Helpers/MockZipkinCollector.cs
```diff
@@ -122,6 +125,7 @@ public static async Task<MockZipkinCollector> Start(ITestOutputHelper output, st
 
             await Task.Delay(500);
         }
+        while (DateTime.Now < deadline);
```
Kielek, line 128: I know that it is test code and not related to the PR, but it is good practice to work on UTC time when compare it with something else.

Similar change is needed for deadline definition.

```suggestion
        while (DateTime.UtcNow < deadline);
```

Kielek (approved):

test/IntegrationTests/Helpers/MockZipkinCollector.cs
pellared: This would have to be also changed: `var deadline = DateTime.Now.Add(timeout.Value);`

pellared (approved):
Merging as it is. I plan to work on span-related integration tests when I am done with the shell scripts and some docs refinements😉

test/IntegrationTests/Helpers/MockZipkinCollector.cs
pjanotti: Yeah, I missed that.

Issue #633, opened 2022-04-27
DiagnosticSource / OpenTelemetry Assembly Resolution Design


rajkumar-rangaraj:
<H2>DiagnosticSource Assembly Load Design</H2>

<H4>Scope</H4>

- .NET 6.0
- .NET 7.0-Preview

<H4>Out of scope</H4>

- .NET Core 3.1 - End of life on Dec 3rd, 2022
- .NET 5.0 - End of life on May 8th, 2022
- .NET Framework 

<H3>DiagnosticSource Usage in An Application</H3>

<H4>.NET 6.0</H4>
There are two possible ways to use System.Diagnostics.DiagnosticSource with .NET 6.0 apps.

1. App with no package reference to DiagnosticSource. 
In this scenario, if there is a need DiagnosticSource 6.0.0 version from the framework will be utilized.

2. App with direct or indirect package reference to DiagnosticSource >= 6.0.0. 

<H4>.NET 7.0 - Preview</H4>
 
For .NET 7.0 apps, only DiagnosticSource version of 7.0.0 will be utilized. 

1. App with no reference to DiagnosticSource. 
In this scenario, if there is a need DiagnosticSource 7.0.0 version from the framework will be utilized.

2. App with direct or indirect have reference to DiagnosticSource 7.0.0.  

**Note**: Loading DiagnosticSource version less than .NET version is not possible. For example, In .NET 6.0 loading DiagnosticSource version less than 6.0.0 is not possible. Even if an app has reference to use 5.0.0 version of DiagnosticSource, framework will auto-upgrade DiagnosticSource to 6.0.0 (framework version) during the build. 

<H3>DiagnosticSource Assembly Load Issue</H3>

An application has a package reference to V1 version of DiagnosticSource. An attempt to dynamically load V2 (higher) version of DiagnosticSource in default assembly context either from StartupHook or CLR Profiler will crash the application.

In case of OpenTelemetry Auto-Instrumentation, if it tries to load DiagnosticSource version higher than the DiagnosticSource version used by an application we will run into this issue. The workaround for this issue is to take a build time package reference to the same version of DiagnosticSource or use additional deps with runtime store.
 
[...]

PR #1368, opened 2022-10-05
OpenTelemetry.AutoInstrumentation NuGet package

### LINKED ISSUE #316: Build a NuGet package that adjusts all of the dependency references
Build an nuget package that adjusts all of the dependency references so that it can be used to mitigate the issue when the application uses a different version of OTel .NET SDK,  `System.Diagnostics.DiagnosticSource` or any other dependency that can cause problems.

_Originally posted by @RassK in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/307#issuecomment-1012176089_

rajkumar-rangaraj:
## Why

Fixes #316, part of #633 

## What

Created a NuGet package with Auto-Instrumentation dependencies which does have reference to external libraries apart from OpenTelemetry SDK/API.

Removed below dependencies from package as these got reference to `MySql.Data` and `StackExchange.Redis`.

```
<PackageReference Include="OpenTelemetry.Instrumentation.MySqlData" Version="1.0.0-beta.3" />
<PackageReference Include="OpenTelemetry.Instrumentation.StackExchangeRedis" Version="1.0.0-rc9.7" />
```

Attaching the NuGet created from my machine for reference. (Remove zip from the file extension)

[OpenTelemetry.AutoInstrumentation.0.3.1-beta.1.nupkg.zip](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/files/9718254/OpenTelemetry.AutoInstrumentation.0.3.1-beta.1.nupkg.zip)

## Checklist


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   is set.
 - Add ILogger logging instrumentation for .NET Core 3.1+.  
 - Add [telemetry resource attributes](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.13.0/specification/resource/semantic_conventions#telemetry-sdk).
+- Add OpenTelemetry.AutoInstrumentation Nuget package.
```
pellared, line 23:
```suggestion
- Add `OpenTelemetry.AutoInstrumentation` NuGet package.
```

OpenTelemetry.AutoInstrumentation.nuspec
```diff
@@ -0,0 +1,66 @@
...
+        <dependency id="OpenTelemetry.Extensions.Propagators" version="1.3.1" />
+        <dependency id="OpenTelemetry.Instrumentation.GrpcNetClient" version="1.0.0-rc9.4" />
+        <dependency id="OpenTelemetry.Instrumentation.Http" version="1.0.0-rc9.4" />
+        <dependency id="OpenTelemetry.Instrumentation.Runtime" version="1.0.0" />
+        <dependency id="OpenTelemetry.Instrumentation.SqlClient" version="1.0.0-rc9.4" />
+        <dependency id="OpenTelemetry.Instrumentation.Wcf" version="1.0.0-rc.7" />
+        <dependency id="OpenTelemetry.Shims.OpenTracing" version="1.0.0-rc9.4" />
+        <dependency id="OpenTelemetry.Instrumentation.AspNet" version="1.0.0-rc9.5" />
+      </group>
+    </dependencies>
```
pellared, line 50: Could you please elaborate on why this is needed? I thought that `dotnet pack` handles such stuff based on `csproj` content 🤷

build/nuke/Build.Steps.cs
```diff
@@ -214,6 +214,19 @@ partial class Build
...
+
+            CopyFileToDirectory(source, dest, FileExistsPolicy.Overwrite);
+
+            DotNetPack(s => s
+                .SetProject(Solution.GetProject(Projects.AutoInstrumentation))
+                .SetConfiguration(BuildConfiguration)
+                .SetProperty("NuspecFile", dest / "OpenTelemetry.AutoInstrumentation.nuspec")
+                .EnableNoBuild()
+                .EnableNoRestore()
+                .SetOutputDirectory(dest));
```
pellared, line 229: I think it should be a new build step that should be executed after `PublishNativeProfiler`, `PublishManagedProfiler`, ``CopyIntegrationsJson`. Otherwise, it can be created without native binaries and `integrations.json`.

OpenTelemetry.AutoInstrumentation.nuspec
rajkumar-rangaraj: As explained in the problem description, want to keep few dependencies out of the Nuget package. In this way we may not spam the customer application and we will have only OpenTelemetry related assemblies.

rajkumar-rangaraj: Also, if we use nuspec it won't look for dependencies list from csproj file.

Directory.Build.props
```diff
@@ -5,5 +5,6 @@
     <IsPackable>false</IsPackable>
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
     <LangVersion>10.0</LangVersion>
+	<NoWarn>NU5100,NU5104,NU5123,NU5128</NoWarn>
```
pjanotti, line 8: nit: indentation
```suggestion
	  <NoWarn>NU5100,NU5104,NU5123,NU5128</NoWarn>
```

OpenTelemetry.AutoInstrumentation.nuspec
```diff
@@ -0,0 +1,66 @@
+﻿<?xml version="1.0"?>
```
pjanotti, line 1: Can we have it in a folder instead of the repo root?

OpenTelemetry.AutoInstrumentation.nuspec
pellared: > As explained in the problem description, 

You are right. My brain "has not connected the dots" 😆

OpenTelemetry.AutoInstrumentation.nuspec
```diff
@@ -0,0 +1,66 @@
...
+    <authors>OpenTelemetry Authors</authors>
+    <owners>OpenTelemetry Authors</owners>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <license type="expression">Apache-2.0</license>
+    <projectUrl>https://opentelemetry.io/</projectUrl>
+    <tags>Observability OpenTelemetry Monitoring Telemetry Tracing Auto Automatic Instrumentation</tags>
+    <repository type="git" url="https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation" />
+    <dependencies>
+      <group targetFramework="netcoreapp3.1">
+        <dependency id="OpenTelemetry" version="1.3.1" />
```
pjanotti, line 16: Not for the initial version but any idea to ensure that the list stays up-to-date?

pjanotti (commented):
Nice @rajkumar-rangaraj - is the package already copied as an artifact by CI?

Directory.Build.props
pjanotti: In case someone wants to check what are those:

- https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5100
- https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5104
- https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5123
- https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5128

OpenTelemetry.AutoInstrumentation.nuspec
```diff
@@ -0,0 +1,66 @@
...
+      <frameworkAssembly assemblyName="System.Web" targetFramework=".NETFramework4.6.2" />
+      <frameworkAssembly assemblyName="System.ServiceModel" targetFramework=".NETFramework4.6.2" exclude="Build,Analyzers" />
+    </frameworkAssemblies>
+  </metadata>
+  <files>
+    <file src="..\tracer-home\AdditionalDeps\**\*.*" target="AdditionalDeps" />
+    <file src="..\tracer-home\net462\**\*.*" target="net462" />
+    <file src="..\tracer-home\netcoreapp3.1\**\*.*" target="netcoreapp3.1" />
+    <file src="..\tracer-home\store\**\*.*" target="store" />
+    <file src="..\tracer-home\win-x64\**\*.*" target="win-x64" />
```
pjanotti, line 62: Either the win-* folders are conditional or we just build the package on Windows

pellared: We would have to build on ALL OSes and have a job that merges it into one...

rajkumar-rangaraj: @pellared Do you have any thoughts on how to get it done?

pellared: For sure something like this will be needed: https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#passing-data-between-jobs-in-a-workflow

Also, I suggest putting any logic in a Nuke target. No more thoughts at the moment.

Directory.Build.props
Kielek: Can we for now, remove NU5104 and mark our nuget as prerelease also?
If needed it can be reintroduced later.

OpenTelemetry.AutoInstrumentation.nuspec
```diff
@@ -0,0 +1,66 @@
+﻿<?xml version="1.0"?>
+<package>
+  <metadata>
+    <id>OpenTelemetry.AutoInstrumentation</id>
+    <version>0.3.1-beta.1</version>
```
Kielek, line 5: Extend releasing.md to include this version.

OpenTelemetry.AutoInstrumentation.nuspec
Kielek: The most tricky part will be probably pack together alpine and linux native dependencies.

OpenTelemetry.AutoInstrumentation.nuspec
RassK: We really should port auto update target 😄

OpenTelemetry.AutoInstrumentation.nuspec
RassK: I guess we can generate these using custom nuke target.

RassK (commented):
Smoke test would be also good to have? - since it's a quite a bit more complicated packing.

Directory.Build.props
rajkumar-rangaraj: Removed NU5104 in new implementation.

OpenTelemetry.AutoInstrumentation.nuspec
rajkumar-rangaraj: Thanks @pellared. Your inputs has helped to build a package based on git workflow. Now, NuGet has files from all Windows, Ubuntu and Mac-OS.

image

@pjanotti FYI.

rajkumar-rangaraj:
> is the package already copied as an artifact by CI?

Yes, the package is a part of CI. I validated the Nuget that run on my fork.

rajkumar-rangaraj:
> Smoke test would be also good to have? - since it's a quite a bit more complicated packing.

@RassK Good point, we could create a separate issue to track this test.

rajkumar-rangaraj:
Nuget package is generated in CI at https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/suites/8691872351/artifacts/392163225

nuget/OpenTelemetry.AutoInstrumentation.nuspec
```diff
@@ -0,0 +1,68 @@
...
+        <dependency id="OpenTelemetry.Instrumentation.AspNet" version="1.0.0-rc9.5" />
+      </group>
+    </dependencies>
+    <frameworkAssemblies>
+      <frameworkAssembly assemblyName="System.Configuration" targetFramework=".NETFramework4.6.2" />
+      <frameworkAssembly assemblyName="System.Web" targetFramework=".NETFramework4.6.2" />
+      <frameworkAssembly assemblyName="System.ServiceModel" targetFramework=".NETFramework4.6.2" />
+    </frameworkAssemblies>
+  </metadata>
+  <files>
```
Kielek, line 57: What's about alpine stuff?

nuget/OpenTelemetry.AutoInstrumentation.nuspec
```diff
@@ -0,0 +1,68 @@
...
+    <file src="bin-ubuntu-20.04/AdditionalDeps/**/*.*" target="AdditionalDeps" />
+    <file src="bin-ubuntu-20.04/netcoreapp3.1/**/*.*" target="netcoreapp3.1" />
+    <file src="bin-ubuntu-20.04/store/**/*.*" target="store" />
```
Kielek, line 60: What if Windows-store!=ubuntu-store!=macos-store!=alpine-store?

nuget/OpenTelemetry.AutoInstrumentation.nuspec
rajkumar-rangaraj: Both Linux variation share the same name for native code, we could accommodate alpine only if we create a separate folder in the NuGet package to hold `OpenTelemetry.AutoInstrumentation.Native.so`.

nuget/OpenTelemetry.AutoInstrumentation.nuspec
rajkumar-rangaraj: This is a very good question. Our current design does not allow the different stores for these platforms, so we may not run into this issue. If the store logic changes we might end up with this issue. Currently, runtime store is built based on .NET version and not platform.

Kielek: It could be separate issue, but it will be great to have test if all of them are the same.

RassK:
Unless it's too late to think. N packages vs everyones sum in one package.
From that choice I might choose N Packages >
OpenTelemetry.AutoInstrumentation.Windows.0.3.1-beta.1.nupkg
OpenTelemetry.AutoInstrumentation.MacOS.0.3.1-beta.1.nupkg
OpenTelemetry.AutoInstrumentation.Linux.0.3.1-beta.1.nupkg
OpenTelemetry.AutoInstrumentation.Alpine.0.3.1-beta.1.nupkg

Why? - Cause the summary package obfuscates the build output tree. While N packages keep it clean.

nuget/OpenTelemetry.AutoInstrumentation.nuspec
RassK: They should be the same in package level, as AFAIK you cannot publish different packages in the same name.
You would just want to include **every** difference.

Also this comment is addressing the same issue: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1368#issuecomment-1273432658

rajkumar-rangaraj:
> Unless it's too late to think. N packages vs everyones sum in one package. From that choice I might choose N Packages > OpenTelemetry.AutoInstrumentation.Windows.0.3.1-beta.1.nupkg OpenTelemetry.AutoInstrumentation.MacOS.0.3.1-beta.1.nupkg OpenTelemetry.AutoInstrumentation.Linux.0.3.1-beta.1.nupkg OpenTelemetry.AutoInstrumentation.Alpine.0.3.1-beta.1.nupkg
> 
> Why? - Cause the summary package obfuscates the build output tree. While N packages keep it clean - and it's not harded to maintain also?

As .NET Core is platform independent, my preference would be to create a single package. In this way customer has an option to host his app in any platform with our Nuget packages.

nuget/OpenTelemetry.AutoInstrumentation.nuspec
rajkumar-rangaraj: We could scope this PR to solve Windows, Ubuntu and Mac-OS and follow up alpine in a different issue as we are not offering this Nuget as release bits for beta 4.

I feel the best option to accommodate Alpine is through renaming our  native libraries like `OpenTelemetry.AutoInstrumentation.Linux.Native.so`, `OpenTelemetry.AutoInstrumentation.Alpine.Native.so`, etc., Better to consider this in a separate issue.

pellared: > I feel the best option to accommodate Alpine is through renaming our native libraries like OpenTelemetry.AutoInstrumentation.Linux.Native.so, OpenTelemetry.AutoInstrumentation.Alpine.Native.so, etc., Better to consider this in a separate issue.

It could be `OpenTelemetry.AutoInstrumentation.Native.musl.so` and the "regular Linux" could be `OpenTelemetry.AutoInstrumentation.Alpine.Native.glibc.so.

pellared: Or separate folders `/linux-musl` and `/linux-glibc` like we have `win-x86` and `win-x64` for Windows.

pjanotti: Having separate folders instead of file names seems better in case we have more native files in the future, anyway, to be decided later. Follow-up on #1397

pjanotti (approved):

pellared:
This PR misses documentation - it would help me in reviewing and testing 😉 

My proposal is to add some description in:
- `design.md` [here](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/design.md#assembly-conflict-resolution) 
- `troublshooting.md` [here](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/troubleshooting.md#assembly-version-conflicts)

The docs do not need to be perfect. They can be improved later. I really dislike adding something without any docs and tests. Having any docs and tests helps also in retrospection.

.github/workflows/ci.yml
```diff
@@ -106,3 +106,39 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3.1.0
+      - uses: nuget/setup-nuget@v1.0.7
+
+      - name: Download Windows Artifacts from build job
+        uses: actions/download-artifact@v3
+        with:
+          name: bin-windows-2022
+          path: nuget/bin-windows-2022
```
pellared, line 122: I suggest changing the path to `nuget/bin-windows` for forward compatibility.

.github/workflows/ci.yml
```diff
@@ -106,3 +106,39 @@ jobs:
...
+
+      - name: Download Ubuntu Artifacts from build job
+        uses: actions/download-artifact@v3
+        with:
+          name: bin-ubuntu-20.04
+          path: nuget/bin-ubuntu-20.04
```
pellared, line 128: I suggest changing the path to `nuget/bin-linux-glibc` for forward compatibility.

.github/workflows/ci.yml
```diff
@@ -106,3 +106,39 @@ jobs:
...
+        uses: actions/download-artifact@v3
+        with:
+          name: bin-ubuntu-20.04
+          path: nuget/bin-ubuntu-20.04 
+
+      - name: Download Mac-OS Artifacts from build job
+        uses: actions/download-artifact@v3
+        with:
+          name: bin-macos-11
+          path: nuget/bin-macos-11
```
pellared, line 134: I suggest changing the path to `nuget/bin-macos` for forward compatibility.

rajkumar-rangaraj:
> My proposal is to add some description in:
> 
> * `design.md` [here](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/design.md#assembly-conflict-resolution) - _here we could also describe some design decisions and reasoning_
> * `troublshooting.md` [here](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/troubleshooting.md#assembly-version-conflicts)

If we all agree to release the Nuget as a part of 0.4.0-beta I will get these document updated. The purpose of this PR is to solve #633. In the previous SIG we discussed about keeping this Nuget in internal myget. Public release of this PR will even make it simple for integration. Please share your thoughts.

.github/workflows/ci.yml
rajkumar-rangaraj: Fixed!

rajkumar-rangaraj:
SIG Discussion Update: This PR does not push the NuGet package to NuGet or MyGet. We could get the document update when the NuGet update gets as a part of the Release workflow.

Resolved all the PR feedback and it is ready for review.

.github/workflows/ci.yml
```diff
@@ -118,3 +118,39 @@ jobs:
...
+  pack:
+    needs: build
+
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3.1.0
+      - uses: nuget/setup-nuget@v1.0.7
+
+      - name: Download Windows Artifacts from build job
+        uses: actions/download-artifact@v3
```
pellared, line 131: Can you please pin the minor and patch versions of the GH Actions to have (in theory) more stable builds?

pellared (approved):
Just one little comment

Kielek (approved):

Issue #1363, opened 2022-10-05
[Flaky test]: IntegrationTests.SmokeTests.PrometheusExporter


pjanotti:
Opening a new issue for this test after some of the previous fixes/changes eliminated other types of errors in this same test. Some instances:

- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3179645369/jobs/5182352567#step:4:2833
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3179826208/jobs/5182713102#step:4:2843
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3157923186/jobs/5139361403#step:4:2761
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3181053874/jobs/5185353993#step:4:2835
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3180252282/jobs/5183604090#step:4:2829
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3153338982/jobs/5129694829#step:4:2637

There are a few more since the test changes.

```
2022-10-04T06:22:13.7527452Z ##[error][xUnit.net 00:01:45.72]     IntegrationTests.SmokeTests.PrometheusExporter [FAIL]
2022-10-04T06:22:13.7529023Z [38;5;247m06:22:13[0m[38;5;247m [[0m[38;5;196;1mERR[0m[38;5;247m] [0m[39m[xUnit.net 00:01:45.72]     IntegrationTests.SmokeTests.PrometheusExporter [FAIL][0m
2022-10-04T06:22:13.7711669Z [38;5;247m06:22:13[0m[38;5;247m [[0m[39;1mDBG[0m[38;5;247m] [0m[39m  Failed IntegrationTests.SmokeTests.PrometheusExporter [1 m][0m
2022-10-04T06:22:13.7715661Z [38;5;247m06:22:13[0m[38;5;247m [[0m[39;1mDBG[0m[38;5;247m] [0m[39m  Error Message:[0m
2022-10-04T06:22:13.7716938Z [38;5;247m06:22:13[0m[38;5;247m [[0m[39;1mDBG[0m[38;5;247m] [0m[39m   Did not expect any exceptions after 1m, but found System.Net.Http.HttpRequestException with message "Connection refused (localhost:9464)"[0m
[...]

PR #1369, opened 2022-10-05
Prometheus exporter is not stable on Linux and macOS

pellared:
## Why

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1363

## What

- Run PrometheusExporter test only on Windows (.NET Framework). I think it is good enought and we do not need to run the test for .NET (Core). To do it we would need to create a custom `Fact` attribute that will skip tests on OSes other than Windows. I think it is not worth the effort.
- I also looked quickly add other smoke tests. Added comment and one missing test which I find quite important.

## Checklist

### DIFF
--- a/test/IntegrationTests/SmokeTests.cs
+++ b/test/IntegrationTests/SmokeTests.cs
@@ -69,6 +69,20 @@ public class SmokeTests : TestHelper
 
     [Fact]
     [Trait("Category", "EndToEnd")]
+    public async Task WhenClrProfilerIsNotEnabled()
+    {
+        var spans = await RunTestApplicationAsync(enableClrProfiler: false);
+
+#if NETFRAMEWORK
+        // on .NET Framework it is required to set the CLR .NET Profiler
+        AssertNoSpansReceived(spans);
+#else
+        AssertAllSpansReceived(spans);
+#endif
+    }
+
+    [Fact]
+    [Trait("Category", "EndToEnd")]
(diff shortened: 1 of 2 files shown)


test/IntegrationTests/SmokeTests.cs
```diff
@@ -67,6 +67,20 @@ public async Task WhenStartupHookIsNotEnabled()
...
     }
 
+    [Fact]
+    [Trait("Category", "EndToEnd")]
+    public async Task WhenClrProfilerIsNotEnabled()
+    {
+        var spans = await RunTestApplicationAsync(enableClrProfiler: false);
+
+#if NETFRAMEWORK
+        // on .NET Framework it is required to set the CLR .NET Profiler
```
pjanotti, line 77: This comment doesn't match with the `enableClrProfiler: false` from line 74

pjanotti (commented):
Looks good just a comment that seems reversed

test/IntegrationTests/SmokeTests.cs
pjanotti: I mean it would be better to say that it doesn't work/instrument on Fx when the profiler is disabled

pjanotti: on second reading it may be me...

pjanotti (approved):

pellared:
Merging as I want the build to be trustworthy 😉

Issue #1362, opened 2022-10-05
CI sometimes hanging on Windows (net462?)


pjanotti:
Some CI runs on Windows seems to be deadlocking on `IntegrationTests.AspNetTests.SubmitMetrics`:

- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3180848449/jobs/5184906531#step:4:3458
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3175272472/jobs/5173106622#step:4:3472
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3173786946/jobs/5169789503#step:4:3442

```
  Error: The active test run was aborted. Reason: Test host process crashed
  15:29:52 [ERR] The active test run was aborted. Reason: Test host process crashed
  15:29:52 [DBG] Data collector 'Blame' message: The specified inactivity time of 3 minutes has elapsed. Collecting hang dumps from testhost and its child processes.
  15:29:52 [DBG] Data collector 'Blame' message: Dumping 4824 - testhost.net462.
  15:29:52 [DBG] Results File: D:\a\opentelemetry-dotnet-instrumentation\opentelemetry-dotnet-instrumentation\build_data\results\IntegrationTests\runneradmin_fv-az448-3_2022-10-03_15_22_24.trx
  15:29:52 [DBG] 
  Error: The active Test Run was aborted because the host process exited unexpectedly. Please inspect the call stack above, if available, to get more information about where the exception originated from.
  15:29:52 [ERR] The active Test Run was aborted because the host process exited unexpectedly. Please inspect the call stack above, if available, to get more information about where the exception originated from.
  Error: The test running when the crash occurred: 
  15:29:52 [ERR] The test running when the crash occurred: 
  Error: IntegrationTests.AspNetTests.SubmitMetrics
  15:29:52 [ERR] IntegrationTests.AspNetTests.SubmitMetrics
  15:29:52 [ERR] 
  Error: This test may, or may not be the source of the crash.
  15:29:52 [ERR] This test may, or may not be the source of the crash.
  15:29:52 [DBG] 
  15:29:52 [DBG] Attachments:
[...]

PR #1371, opened 2022-10-05
Refactor AspNetTests

pellared:
## Why

When analyzing https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1362

## What

- Use `await using` per design. When I looked at the implementation of `DiposeAwait()` it should finish in a timely manner anyway...
- Give 5 min per test before the crash dump. I saw there were some 5 min intervals so probably 3 min can be too short. E.g. [here](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/3189892623/jobs/5204254694#step:4:3477) it succeeded after 2 ~minutes (it was my first find).

## Tests

- Run the test 50 times (sic!): https://github.com/pellared/opentelemetry-dotnet-instrumentation/actions/runs/3193095691


pellared:
CC @rajkumar-rangaraj I tried to reorder the usings but it is not anything simple... I start to consider making a separate job in GH workflow that will only be only responsible for running this one test class 🤦

rajkumar-rangaraj (approved):

pellared:
The test passed 50 times. Merging. Let's see what will happen. I will not be surprised if it is an "environmental issue" connected with Windows Docker containers.

PR #1372, opened 2022-10-05
[WIP | Experimental] Kestrel Mock Zipkin Collector

RassK:
## Why

Fun thing to try out to stabilize tests.
_Source issue discovered by @pellared  https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1363#issuecomment-1268412906_

## What

Switches `HttpListener` to AspNetCore Kestrel minimal server.

## Tests

Existing.

### DIFF
--- a/test/IntegrationTests/Helpers/MockZipkinCollector.AspNetCore.cs
+++ b/test/IntegrationTests/Helpers/MockZipkinCollector.AspNetCore.cs
@@ -1,4 +1,4 @@
-// <copyright file="MockZipkinCollector.cs" company="OpenTelemetry Authors">
+// <copyright file="MockZipkinCollector.AspNetCore.cs" company="OpenTelemetry Authors">
 // Copyright The OpenTelemetry Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -14,7 +14,7 @@
 // limitations under the License.
 // </copyright>
 
-#if NETFRAMEWORK
+#if NETCOREAPP3_1_OR_GREATER
 
 using System;
 using System.Collections.Generic;
@@ -23,9 +23,12 @@ using System.Collections.Specialized;
 using System.IO;
 using System.Linq;
 using System.Net;
-using System.Text;
 using System.Threading.Tasks;
 using IntegrationTests.Helpers.Mocks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Hosting.Server.Features;
+using Microsoft.AspNetCore.Http;
 using Newtonsoft.Json;
 using Xunit.Abstractions;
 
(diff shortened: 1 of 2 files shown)


pellared:
I realized that probably it is no longer an issue for us because of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1270 .

pellared:
Still this implementation would be a lot better and we could get rid of "guessing" the free port.

test/IntegrationTests/Helpers/MockZipkinCollector.cs
```diff
@@ -14,6 +14,8 @@
 // limitations under the License.
 // </copyright>
 
+#if NETFRAMEWORK
```
pellared, line 17: We could run the tests in .NET 6.0 and start the test applications in .NET Framework, .NET Core 3.1 etc...

pellared: The info in which version the test app should be run could be passed via env vars.

RassK: Sounds perfect, tests should run faster as net6.0 is more performant. Let's see what CI thinks about it.

RassK: Unfortunately breaks pre-processor flags in `IntegrationTests`.
Makes it hard to run directly in Visual Studio Test Explorer, if target is other than 6.0.

RassK: Reverting... might need bigger rebuilds / planning. Could cause loads of discomfort in integrations tests development if done wrong.

RassK (commented):
If anyone sees value in this, we can take it as a futureproofing.
The biggest instability addressed by this issue is already fixed by @Kielek.

Will just close it if there is no interest.

RassK:
> Still this implementation would be a lot better and we could get rid of "guessing" the free port.

I think the current `TcpPortProvider.GetOpenPort()` is doing very similar thing as the Kestrel. 
See both are providing 0 as an initial port and they get it after the server starts.

pellared:
> > Still this implementation would be a lot better and we could get rid of "guessing" the free port.
> 
> I think the current `TcpPortProvider.GetOpenPort()` is doing very similar thing as the Kestrel. See both are providing 0 as an initial port and they get it after the server starts.

Kind of. There can be a race that it is no longer available. And we trying to find next free port. It was mitigated when I setup to run all test sequentially.

RassK:
> > > Still this implementation would be a lot better and we could get rid of "guessing" the free port.
> > 
> > 
> > I think the current `TcpPortProvider.GetOpenPort()` is doing very similar thing as the Kestrel. See both are providing 0 as an initial port and they get it after the server starts.
> 
> Kind of. There can be a race that it is no longer available. And we trying to find next free port. It was mitigated when I setup to run all test sequentially.

Race after the port is released is a good point! Great pointing that out.

test/IntegrationTests/Helpers/MockZipkinCollector.AspNetCore.cs
```diff
@@ -0,0 +1,226 @@
...
+            {
+                x.Map("/api/v2/spans", x =>
+                {
+                    x.Run(HandleHttpRequests);
+                });
+
+                x.Map("/healthz", x =>
+                {
+                    x.Run(HandleHealtz);
+                });
```
pellared, line 62: not needed

pellared (commented):
The first option could be to not test code in .NET Framework. 

The second would be to have two implementations of a `TestHttpServer` (for .NET Framework based on the `TestHttpListener` + healthz handling; for ASP.NET Core based on Kestrel what you have done here) and use it in `MockZipkinCollector` + `MockLogsCollector` + `MockMetricsCollector` if a few `#if` statements limited to the differences of the HttpListener and ASP.NET Core API differences.

The third one is to leave it as it is.

Probably right now we could leave it as it is unless `HttpListener` causes some instability in our tests.

test/IntegrationTests/Helpers/MockZipkinCollector.AspNetCore.cs
RassK: Removed.

Issue #1373, opened 2022-10-06
Create convenient Shell scripts


pellared:
Create Shell scripts that will facilitate installation and instrumentation.

They should primarily target non-Windows OSes.

PR #1375, opened 2022-10-06
Change INSTALL_DIR to OTEL_DOTNET_AUTO_HOME in scripts

pellared:
## Why

Part of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1373

1. User can define a global machine-wide (or user-wide) `OTEL_DOTNET_AUTO_HOME` so that `INSTALL_DIR` (which is not a unique name) does not have to be provided when a custom installation path is used
   1. At some point `install.sh` could even set OTEL_DOTNET_AUTO_HOME as a global env var to facilitate using `./instrument.sh`. Not now for sure - personally I want to postpone setting global env vars until we see that it is beneficial)
1. Takes advantage of existing env var name which is required anyway

Also, change the default so that, by default, the `./instrument.sh` could be used from a different location.

## What

- Change `INSTALL_DIR` env var script param to `OTEL_DOTNET_AUTO_HOME`
- Change the default installation path to `$HOME/.otel-auto-dotnet`
  - Based on what `dotnet-install.sh`
  - It is not dependent on `$PWD`
  - Does not require admin rights
- Remove `RELEASES_URL` script param (YAGNI)

## Tests

CI

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,7 +101,7 @@ jobs:
           export OS_TYPE=${{ matrix.os_type }} 
           curl -sSfL https://raw.githubusercontent.com/${{ github.repository }}/${{ github.sha }}/download.sh -O
           sh ./download.sh
-            test "$(ls -A ./otel-dotnet-auto)"
+            test "$(ls -A "$HOME/.otel-dotnet-auto")"
           curl -sSfL https://raw.githubusercontent.com/${{ github.repository }}/${{ github.sha }}/instrument.sh -O
           . ./instrument.sh
           dotnet help
--- a/docs/README.md
+++ b/docs/README.md
@@ -101,23 +101,23 @@ When running your application, make sure to:
 1. Set the [resources](config.md#resources).
 1. Set the environment variables from the table below.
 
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -101,7 +101,7 @@ jobs:
           export OS_TYPE=${{ matrix.os_type }} 
           curl -sSfL https://raw.githubusercontent.com/${{ github.repository }}/${{ github.sha }}/download.sh -O
           sh ./download.sh
-            test "$(ls -A ./otel-dotnet-auto)"
+            test "$(ls -A "$HOME/.otel-dotnet-auto")"
```
rajkumar-rangaraj, line 104: Want to understand this better, do we need `/.` or `/` would be sufficient here?

pellared: dotnet-install.sh does the same. It makes the dir "hidden". `ls` will not list it, but `ls -a` will list it.

Kielek (approved):

RassK (approved):

nrcventura (approved):

PR #1386, opened 2022-10-10
Log warn when Prometheus exporter is configured

### LINKED ISSUE #1377: Mark Prometheus exporter as dev. only
Mark Prometheus exporter as dev. only.

Consider update documentation and log warning when Prometheus exporter is used.

Please check https://github.com/open-telemetry/opentelemetry-dotnet/pull/3737 and previously merged https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1369

Kielek:
## Why

Fixes #1377

## What

Log warn when Prometheus exporter is configured
Update documentation to reflect changes in OTel SDK.

## Tests

CI

## Checklist


docs/config.md
```diff
@@ -145,7 +145,8 @@ environment variables for the OTLP exporter:
 
 > ⚠️ **Do NOT use in production.**
 >
-> Prometheus exporter uses [an unreliable, unmaintained legacy .NET component](https://github.com/dotnet/runtime/issues/28658#issuecomment-462062760).
+> Prometheus exporter [is intended for dev inner-loop,
+> there is no plan to make it production ready](https://github.com/open-telemetry/opentelemetry-dotnet/blob/622583c6ffaea2e3d83d4fcb8935c6a8dbc6f81f/src/OpenTelemetry.Exporter.Prometheus.HttpListener/README.md?plain=1#L10-L16).
```
pellared, line 149: Should we call out that this is the "HttpListener" version? 

After reading [this](https://github.com/open-telemetry/opentelemetry-dotnet/blob/622583c6ffaea2e3d83d4fcb8935c6a8dbc6f81f/src/OpenTelemetry.Exporter.Prometheus.HttpListener/README.md?plain=1#L10-L16) the user may ask how to use `OpenTelemetry.Exporter.Prometheus.AspNetCore`?

Personally, I think it would be easier if we simply:

```suggestion
> Prometheus exporter is intended for the inner dev loop.
> Production environments can use a combination of OTLP exporter
> with [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector-releases)
> having [`otlp` receiver](https://github.com/open-telemetry/opentelemetry-collector/tree/v0.61.0/receiver/otlpreceiver) and [`prometheus` exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.61.0/exporter/prometheusexporter).
```

I believe that we might add support for the AspNetCore version of the Prometheus exporter in the future.

src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationMetricHelper.cs
```diff
@@ -74,6 +76,7 @@ private static MeterProviderBuilder SetExporter(this MeterProviderBuilder builde
         switch (settings.MetricExporter)
         {
             case MetricsExporter.Prometheus:
+                Logger.Warning("Prometheus exporter is configured. It is intended for dev inner-loop, there is no plan to make it production ready.");
```
pellared, line 79:
```suggestion
                Logger.Warning("Prometheus exporter is configured. It is intended for the inner dev loop. Do NOT use in production");
```

I believe that we might add support for the AspNetCore version of the Prometheus exporter in the future.

docs/config.md
theletterf, line 149: What is the `dev inner-loop` and why are we featuring this exporter here and not in a dev doc if it's not for production?

pellared: I think the wording is not perfect. That is why I suggested to change it in my review

Reference: https://www.getambassador.io/docs/telepresence/latest/concepts/devloop/#what-is-the-inner-dev-loop

Kielek: Taken from OTel SDK side. I think it is useful for developing/debugging scnearios.

pellared: > Why are we featuring this exporter here and not in a dev doc if it's not for production 

Good point. Personally, I think it may still be usable... But we may consider removing it completely. I think we considered it during the SIG meeting, but first we agreed to just mark it as "non-production".

pellared (approved):

PR #1403, opened 2022-10-11
Calulate default OS_TYPE

pellared:
## Why

Better UX. No need to provide `OS_TYPE`

Part of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1373

## What

Calculate `OS_TYPE` using `uname -s` and  `ldd /bin/ls`

## Tests

CI and locally

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,8 +40,7 @@ jobs:
           name: bin-${{ matrix.machine }}
           path: bin/tracer-home
 
-  container-build:
-    name: Container Build 
+  build-container:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


pellared:
> is it worth to put the code to select the OS type in a shareable way

Not sure how to do it

1. `download.sh` is the "first" script that the user downloads to setup. the `instrument.sh` is not available until it is extracted
2. I assume that once the user downloads (installs) then he/she only needs `instrument.sh` and `download.sh` is no longer needed (also getting the path to `download.sh` would be probably close to impossible)

dszmigielski (approved):

Kielek (approved):

PR #1417, opened 2022-10-13
Refine TestHttpListener

pellared:
## Why

Fixes #

## Checklist


test/IntegrationTests/Helpers/TestHttpListener.cs
```diff
@@ -79,11 +79,37 @@ public TestHttpListener(ITestOutputHelper output, Action<HttpListenerContext> re
...
 
-        return HealthzHelper.TestHealtzAsync(healhtzEndpoint, nameof(MockLogsCollector), _output);
+        set
+        {
+            lock (_requestHandlerLocker)
+            {
+                _requestHandler = value;
+            }
+        }
+    }
```
pellared, line 99: This can be improved (even to avoid locking).
However, I do not want to make the code harder to read (I do not think it would ever cause performance issues).

PR #1421, opened 2022-10-14
Remove LoggingBuilder instrumentation from .NET Framework

Kielek:
## Why

Fixes N/A

## What

Remove LoggingBuilder instrumentation from .NET Framework
It does not have any impact on execution and we can avoid one bytecode operation.

We have the same approach for StackExchange.Redis and others.

## Tests

CI

## Checklist


pjanotti:
It seems good, although, it make me want to revisit what happens in the ReJIT in the case of not finding the type listed in the integrations.json. Likely, an error is logged reporting the missing code, but, it seems worth to double check.

pjanotti:
Alternatively: should we consider a way to not even attempt to instrument in case a given runtime is not supported?

pjanotti:
Closing per SIG discussion. We will do when able to target specific frameworks.

PR #1466, opened 2022-10-19
Integrations json file generator

### LINKED ISSUE #1005: Auto-regenerate intergrations.json file
I do not found any script to refresh this file in case of changes in binary instrumentation. Do we have anything like this?

_Originally posted by @Kielek in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/979#discussion_r929519750_

Based on SIG discussion, the script is not available right now. It was use on upstream side on 1.* branch. See https://github.com/DataDog/dd-trace-dotnet/blob/v1.31.2/tracer/build/_build/PrepareRelease/GenerateIntegrationDefinitions.cs

Kielek:
## Why

It is hard to modify our integrations.json files manually.
(Partially) Fixes #1005

## What

Tool to generate integrations.json files.
Integrations.json files regenerated by new tool.
Fix on the C# code for StackExchangeRedis (manually tested for old version).
Fix for tests.json file to sync with C# code.

It was created as a separate tool, as I expect that we need to modify it and it is the most convenient way to tests it e2e.
Further work:
 - split instrumentation based on type - logs, traces and possibly metrics.
 - generate integration files based on the framework. 

Designed to execute under human supervision. Manual verification for json files is required.

## Tests

Manual tests.

## Checklist

### DIFF
--- a/OpenTelemetry.AutoInstrumentation.sln
+++ b/OpenTelemetry.AutoInstrumentation.sln
@@ -151,6 +151,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "TestApplication.DomainNeutr
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "TestApplication.MassTransit", "test\test-applications\integrations\TestApplication.MassTransit\TestApplication.MassTransit.csproj", "{06664166-35D6-484D-9171-DCB99D1D310A}"
 EndProject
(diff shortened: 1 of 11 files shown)


RassK:
Do you plan to add Nuke target?

pellared:
> Do you plan to add Nuke target?

Or maybe even better. Why not implementing it as Nuke target and getting rid of another `csproj`? 😉

Kielek:
I think that we will have to load both .NET Framework and .NET (Core) assembly to fully cover expected features.
Based on my testing, it could be tricky. Creating new bytecode instrumentation is not our daily basis, so current shape IMO is good enough.

We can extend functionality in the following PRs if the current process will be too painful.

pellared (approved):

PR #1484, opened 2022-10-21
Use MockSpansCollector in GraphQLTests

pellared:
## Why

Part of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/915

## What

Use MockSpansCollector in GraphQLTests.


pellared:
@RassK Can you take a quick look? I am considering refactoring it to use `SpanExpectation` however maybe it would be easier to use e.g. https://github.com/FluentValidation/FluentValidation instead?

RassK:
> @RassK Can you take a quick look? I am considering refactoring it to use `SpanExpectation`. Maybe it would be easier to use e.g. https://github.com/FluentValidation/FluentValidation instead. Right now I am too tired to have any opinon 😉 

Just need to be sure that assert details remain, due the instrumentation is ours.
Definitely good to move the logic away from ctor, adds lot of more flexibility 👍

RassK (approved):
👍

pellared:
> Just need to be sure that assert details remain, due the instrumentation is ours.

I removed the comment. It is hard to add details as the assertion is executed for all spans. Which failure message would you like to print in such case? For all the spans? It would be not readable.

PR #1494, opened 2022-10-24
Remove LegacyMockZipkinCollector and cleanup

pellared:
## Why

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/915

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1232

## What

- Change existing tests to use LegacySpanCollector
- Remove LegacyMockZipkinCollector and related types
- Opportunistic cleanup in IntegrationTests


test/IntegrationTests/StackExchangeRedisTests.cs
```diff
@@ -29,38 +26,29 @@ namespace IntegrationTests;
...
-        RunTestApplication(agent.Port, arguments: $"--redis {_redis.Port}");
-
-        const int expectedSpansCount = 8;
-
-        var spans = await agent.WaitForSpansAsync(expectedSpansCount);
-
-        using (new AssertionScope())
+        using var collector = await MockSpansCollector.Start(Output);
+        const int spanCount = 8;
+        for (int i = 0; i < spanCount; i++)
```
RassK, line 44: seems a candidate for later refactoring.

test/IntegrationTests/Helpers/TimeExtensions.cs
```diff
@@ -15,20 +15,23 @@
 // </copyright>
 
 using System;
-using IntegrationTests.Helpers.Constants;
 
 namespace IntegrationTests.Helpers;
 
 internal static class TimeExtensions
 {
+    private const long NanoSecondsPerTick = 1000000 / TimeSpan.TicksPerMillisecond;
```
pjanotti, line 23: nit: this is just moved from a deleted file but given that it is a constant it seems better to just say `100` instead of the division.

pjanotti (approved):
LGTM

test/IntegrationTests/Helpers/TimeExtensions.cs
pellared: TBH for me the current code is more readable. Otherwise I would not know from where this knowledge comes from. The compiler converts it to `100` anyway.

test/IntegrationTests/StackExchangeRedisTests.cs
pellared: I was also considering to remove the "span count" and just wait for the first span. But I had no strong feeling what would be better and I did not want to do too much in this PR.

test/IntegrationTests/Helpers/TimeExtensions.cs
pjanotti: perhaps I have stamped in my brain that it is 100ns ;) - fine for me.

Issue #1469, opened 2022-10-20
Application crash - try to instrument by non-existing method


Kielek:
# Bug Report

## Symptom

**Describe the bug**
Instrumented application is crashing when you try to use non-existing bytecode instrumentation.

```log

Xunit.Sdk.XunitException
Expected process.ExitCode to be 0 because Test application exited with non-zero exit code, but found -532462766 (difference of -532462766).
   at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message)
   at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc)
   at FluentAssertions.Numeric.NumericAssertions`2.Be(T expected, String because, Object[] becauseArgs)
   at IntegrationTests.Helpers.TestHelper.RunTestApplication(TestSettings testSettings) in C:\GitHub\opentelemetry-dotnet-instrumentation\test\IntegrationTests\Helpers\TestHelper.cs:line 170
   at IntegrationTests.Helpers.TestHelper.RunTestApplication(Int32 traceAgentPort, Int32 otlpTraceCollectorPort, Int32 metricsAgentPort, Int32 logsAgentPort, String arguments, String packageVersion, String framework, Int32 aspNetCorePort, Boolean enableStartupHook, Boolean enableClrProfiler) in C:\GitHub\opentelemetry-dotnet-instrumentation\test\IntegrationTests\Helpers\TestHelper.cs:line 140
   at IntegrationTests.SmokeTests.<SubmitMetrics>d__9.MoveNext() in C:\GitHub\opentelemetry-dotnet-instrumentation\test\IntegrationTests\SmokeTests.cs:line 146
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Xunit.Sdk.TestInvoker`1.<>c__DisplayClass48_0.<<InvokeTestMethodAsync>b__1>d.MoveNext() in /_/src/xunit.execution/Sdk/Frameworks/Runners/TestInvoker.cs:line 264
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #1498, opened 2022-10-25
Avoid crash if bytecode instrumentation method is missing

pjanotti:
This is similar to #1469 but in this case the instrumentation finds the type but not the expected method.


pjanotti:
This was intended to be a follow up issue for the PR that is not ready yet 🤦‍♂️ closing this PR.

PR #1507, opened 2022-10-27
Avoid GitHub Trottling

Kielek:
Based on changes from https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/737


pellared (approved):

pellared:
Should we include `.github/workflows/validate-documentation.yml` under `paths:` in the same file; like "run me when I change"? 😉

Issue #1505, opened 2022-10-26
Loading unnecessary exporter modules when instrumenting app


pjanotti:
When running a simple test like StrongNamedTests the logs show that many unexpected exporter modules are loaded, e.g.: Jaeger. This is issue is to investigate why and fix, if appropriate, any exporter module that is being loaded by not actually used.

List for the modules loaded for the StrongNamedTests when running on net462:
```terminal
Anonymously Hosted DynamicMethods Assembly
CrowdStrike.Sensor.ScriptControl
Google.Protobuf
Grpc.Core
Grpc.Core.Api
Microsoft.Extensions.Logging.Abstractions
mscorlib
netstandard
OpenTelemetry
OpenTelemetry.Api
OpenTelemetry.AutoInstrumentation
OpenTelemetry.AutoInstrumentation.Loader
OpenTelemetry.Exporter.Console
OpenTelemetry.Exporter.Jaeger
OpenTelemetry.Exporter.OpenTelemetryProtocol
OpenTelemetry.Exporter.Prometheus
OpenTelemetry.Exporter.Zipkin
OpenTelemetry.Instrumentation.AspNet
OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule
OpenTelemetry.Instrumentation.GrpcNetClient
OpenTelemetry.Instrumentation.Http
OpenTelemetry.Instrumentation.Process
OpenTelemetry.Instrumentation.Runtime
OpenTelemetry.Instrumentation.SqlClient
OpenTelemetry.Instrumentation.Wcf
OpenTelemetry.Shims.OpenTracing
OpenTracing
System
System.Buffers
System.Configuration
System.Core
System.Diagnostics.DiagnosticSource
System.Memory
System.Net.Http
System.Runtime.CompilerServices.Unsafe
System.ServiceModel
System.Web
System.Xml
TestApplication.StrongNamed
TestLibrary.InstrumentationTarget
```

pjanotti:
Per SIG meeting adding it to the 0.5.0-beta milestone.

pjanotti:
Same test on `net60` (Linux):
```terminal
Google.Protobuf
Microsoft.Extensions.Logging.Abstractions
Microsoft.Win32.Primitives
netstandard
OpenTelemetry
OpenTelemetry.Api
OpenTelemetry.AutoInstrumentation
OpenTelemetry.AutoInstrumentation.Loader
[...]

PR #1515, opened 2022-10-27
Verify loaded managed modules

RassK:
## Why

Part of #1505 

## What

Verifies the current state of .Net Core app loaded modules (only `OpenTelemetry.*`)

## Tests

Integration tests reusing Smoke tests.


RassK:
Prob need to switch from Smoke tests base to some ASP.NET Core base, otherwise not all modules will even try to load.

test/IntegrationTests/BuildTests.cs
```diff
@@ -46,6 +46,7 @@ public Task DistributionStructure()
 
         return Verifier.Verify(relativesPaths)
             .UseTextForParameters(systemName)
+            .UseDirectory("./snapshots")
```
Kielek, line 49: VS after changes:
image

VS before changes:
image


I am not sure if I want all files under snapshots folder as all of them are grouped by IDE.

test/IntegrationTests/ModuleTests.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using IntegrationTests.Helpers;
+using VerifyXunit;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace IntegrationTests;
+
+[UsesVerify]
```
Kielek, line 30: Can we execute it on all machines types/framework families?
I suppose that results can be a bit different.

RassK: Somehow net472 and linux and mac only report single native library. Just needs more research why.

test/IntegrationTests/BuildTests.cs
RassK: Hmm, my first preference is always to keep source dir clean and tidy.  Currently it will become a real mess. 
Wondering if this kind of grouping is configurable, so it will work anyway but based on snapshots dir.

Anyway highly subjective topic, I'll keep the comment open for other opinions.

pellared: I agree with @Kielek . I definitely prefer the "original" behavior so that the expectations are closer to the test.

pjanotti: +1 for @Kielek's suggestion

Kielek (approved):

test/IntegrationTests/ModuleTests.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        var json = File.ReadAllText(tempPath);
+        var modules = JsonConvert.DeserializeObject<string[]>(json);
+
+        // Cleanup
+        File.Delete(tempPath);
+
+        await Verifier.Verify(modules)
+           .UniqueForOSPlatform()
+           .UniqueForRuntime()
```
pellared, line 55: We miss the distinction between `netcoreapp3.1` and `net6.0`. Right now we have the same output. Is it OK? Right now we have the exact same output for all configuration so in theory we could even have on `verified.txt` file. When I look at the test I think having a single file would be good enough...

test/IntegrationTests/ModuleTests.cs
```diff
@@ -0,0 +1,58 @@
...
+        if (!File.Exists(tempPath))
+        {
+            Assert.Fail("Could not find modules report file.");
+        }
+
+        var json = File.ReadAllText(tempPath);
+        var modules = JsonConvert.DeserializeObject<string[]>(json);
+
+        // Cleanup
+        File.Delete(tempPath);
```
pellared, line 51: nit: it is safer to do cleanup in `finally` e.g. if `File.ReadAllText` or `JsonConvert.DeserializeObject` crashes for whatever reason

test/test-applications/integrations/TestApplication.Modules/Program.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.IO;
+using System.Linq;
+using Newtonsoft.Json;
+using TestApplication.Shared;
+
+namespace TestApplication.Smoke;
+
+public class Program
+{
+    public const string SourceName = "MyCompany.MyProduct.MyLibrary";
```
pellared, line 27: not used

test/test-applications/integrations/TestApplication.Modules/TestApplication.Modules.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <ItemGroup Condition="$(TargetFramework.StartsWith('net4'))">
+    <Reference Include="System.Net.Http" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="6.0.0" />
```
pellared, line 9: is it needed?

test/IntegrationTests/ModuleTests.cs
RassK: net3.1 and net6.0 are same, NET Fx is different.

test/test-applications/integrations/TestApplication.Modules/TestApplication.Modules.csproj
RassK: Removed. https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1515/commits/42aebf63e7b2b5c18f66335bfb97729405d00203

test/test-applications/integrations/TestApplication.Modules/Program.cs
RassK: Removed. https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1515/commits/42aebf63e7b2b5c18f66335bfb97729405d00203

test/IntegrationTests/ModuleTests.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1515/commits/42aebf63e7b2b5c18f66335bfb97729405d00203

test/IntegrationTests/ModuleTests.cs
RassK:
```csharp
            await Verifier.Verify(modules)
#if NETFRAMEWORK
                 .UniqueForRuntime()
#endif
                 .DisableDiff();
```

I would have preferred this, but .NET Core somehow gets into conflict with existing .NET Fx verify file 🙄

pellared: I think we can simply have ` await Verifier.Verify(modules).DisableDiff();`.

RassK: Not really, then .NET Fx will pick up .NET Core verify files.

test/IntegrationTests/ModuleTests.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1515/commits/138029393eb0b44f4c4a837d3d5e5e8eaec85484

test/IntegrationTests/ModuleTests.cs
pellared: I have not noticed the difference 😂 how about 

```csharp
            await Verifier.Verify(modules)
                 .UniqueForRuntime()
                 .DisableDiff();
```

?

pellared: > net3.1 and net6.0 are same, NET Fx is different.

I have missed this comment...

RassK: Testing another option that reduces verify files to only 2.

```csharp
            await Verifier.Verify(modules)
#if NETFRAMEWORK
                .UseFileName("ModuleTests.RunApplication.NetFx")
#else
                .UseFileName("ModuleTests.RunApplication.NetCore")
#endif
                .DisableDiff();
```

pellared (approved):

pjanotti (commented):
LGTM

PR #1518, opened 2022-10-28
Rename _AT_STARTUP env vars to _ENABLED

### LINKED ISSUE #1478: Rename `AT_STARTUP` env vars
I propose to rename the following env vars accordingly:

- `OTEL_DOTNET_AUTO_LOAD_TRACER_AT_STARTUP` -> `OTEL_DOTNET_AUTO_TRACES_ENABLED`
- `OTEL_DOTNET_AUTO_LOAD_METER_AT_STARTUP` -> `OTEL_DOTNET_AUTO_METRICS_ENABLED`

The reasons are:
- It not only sets the OpenTelemetry .NET SDK's signal providers, but also add source instrumentations
- I think that the names would be more clear 😉 

The only problem that would remain is that the CLR Profiler can do bytecode instrumentation. But I would leave it as it is as it should be a NOOP from the "user perspective" and it could be useful if one would like to use OTel .NET Auto only to get bytecode instrumentations. _I think this use case could be documented._

Kielek:
## Why

Fixes #1478 

## What

Replace `OTEL_DOTNET_AUTO_LOAD_TRACER_AT_STARTUP` by `OTEL_DOTNET_AUTO_TRACES_ENABLED`  and `OTEL_DOTNET_AUTO_LOAD_METER_AT_STARTUP` by `OTEL_DOTNET_AUTO_METRICS_ENABLED`.

## Tests

CI

## Checklist


theletterf (approved):

docs/config.md
```diff
@@ -188,8 +188,8 @@ Important environment variables include:
...
-| `OTEL_DOTNET_AUTO_LOAD_TRACER_AT_STARTUP`      | Whether the tracer is created by the automatic instrumentation library or not. Set to `false` when the application initializes the OpenTelemetry .NET SDK Tracer on its own. This configuration can be used, for example, to retrieve the bytecode instrumentations.                                                               | `true`        |
-| `OTEL_DOTNET_AUTO_LOAD_METER_AT_STARTUP`       | Whether the meter is created by the automatic instrumentation library or not. Set to `false` when the application initializes the OpenTelemetry .NET SDK Meter on its own.                                                                                                                                                         | `true`        |
+| `OTEL_DOTNET_AUTO_TRACES_ENABLED`              | Whether the tracer is created by the automatic instrumentation library or not. Set to `false` when the application initializes the OpenTelemetry .NET SDK Tracer on its own. This configuration can be used, for example, to retrieve the bytecode instrumentations.                                                               | `true`        |
```
pellared, line 191: I think it would be better if just say here that it disables/enables signal providers and source instrumentations. We could add a footer and explain when this could be handy

pellared, line 191: It should be "tracer provider" instead of "tracer"? Same for "meter".

pellared (approved):

Issue #960, opened 2022-07-20
Use OTel SDK's self-diagnostics - native code


pellared:
We should take advantage of [OTel SDK's self-diagnostics](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/README.md#self-diagnostics) for internal logging and try to adopt it in the native code.

We could do it by emitting EventSource events from the C++ code. Then we would probably need to create guidelines/tools to be able to catch these events in case the process crashes before OTel SDK's self-diagnostic is initialized (as well as to catch events that are emitted before it is initialized). Or we should have a separate mechanism for logging before OTel SDK's self-diagnostic is initialized.

nrcventura:
The native code has the ability to run before any managed code is loaded/executed. In particular the initialization code of the profiler runs before any managed code is loaded, and those initialization related logs are very important, so the solution that we choose needs to ensure that we can capture those logs.

pjanotti:
@nrcventura Q is very relevant. Initially, we could simply have a point at which the native code switches from file to EventSource. Anyway, let's record that the runtime provides support to write EventSource events from a CLR profiler. `ICorProfilerInfo12` introduced such support, however, it requires .NET 5.0 and above, so it is not available on any supported .NET Framework version.

PR #1520, opened 2022-10-29
Log error if missing bytecode instrumentation method

### LINKED ISSUE #1499: Silent failure if bytecode instrumentation is missing all expected methods
This is similar to https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1469 but in this case the instrumentation finds the type but not the expected method.

pjanotti:
## Why

If the bytecode instrumentation type doesn't have any of the expected instrumentation methods the bytecode instrumentation is injected during ReJIT but since there are no instrumentation methods it has no practical effect and no error/warning message is logged in any of the logs.

Fixes #1499 

## What

- Logs an error msg if none of the expected bytecode instrumentation methods are present on the instrumentation type.

## Tests

- Added a bytecode instrumentation target that exercises the new log statement to the StrongNamedTests.
- I considered inspecting the native log to ensure the message is there but it seemed non-reliable to depend on the process id (that can be reused) and time window to open the file. I opted to not do it now and see if we can add that validation if we implement #960 (this will be a subscription to the event which should be more reliable).

## Checklist

~~- [] Documentation is updated.~~


src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -2494,14 +2494,47 @@ HRESULT CorProfiler::CallTarget_RewriterCallback(RejitHandlerModule* moduleHandl
...
+        // CallTarget instrumentation doesn't inject calls to the instrumentation methods via IL rewrite.
+        // It injects the OpenTelemetry.AutoInstrumentation.CallTarget.CallTargetInvoker, written in managed code,
+        // that uses reflection to find the expected instrumentation methods on the instrumentation wrapper type.
+        // If the wrapper type doesn't have any of the expected instrumentation methods "nothing happens", but,
+        // the JIT code of the targeted method is modified anyway. To avoid injecting instrumentation that does
+        // nothing and give a clear error message the code below ensures that at least one of the expected methods is
+        // implemented on the wrapper type.
+        static const LPCWSTR expected_wrapper_methods[] = { WStr("OnMethodBegin"), WStr("OnMethodEnd"), WStr("OnAsyncMethodEnd") };
+        bool found_wrapper_method = false;
+        for (auto i = 0; i < sizeof(expected_wrapper_methods) / sizeof(LPCWSTR); i++)
```
Kielek, line 2512: Any chance to use for each loops to avoid calculation of array length?
https://www.w3schools.com/cpp/trycpp.asp?filename=demo_array_loop_foreach

test/IntegrationTests/StrongNamedTestsIntegrations.json
```diff
@@ -46,6 +46,28 @@
           "type": "OpenTelemetry.AutoInstrumentation.Instrumentations.Validations.MissingInstrumentationType",
           "action": "CallTargetModification"
         }
+      },
+      {
+        "caller": {},
+        "target": {
```
Kielek, line 52: Could you please extend
https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/b0d4457610231e5e3a8af32f8856a9d0b0fbdf8d/tools/IntegrationsJsonGenerator/Program.cs#L68-L69
to programmatically add this instrumentation?
It will be removed if you execute the tool.

pjanotti: [73f67ae](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1520/commits/73f67ae9744b3a2d54a39630c809a122f7d513e2)

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
pjanotti: [73f67ae](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1520/commits/73f67ae9744b3a2d54a39630c809a122f7d513e2)

test/IntegrationTests/StrongNamedTestsIntegrations.json
pjanotti: For now hard coding into the tool the integration tests that require mock instrumentations. Perhaps revisiting a more robust way of doing this.

Kielek (approved):

theletterf (approved):

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -2494,14 +2494,47 @@ HRESULT CorProfiler::CallTarget_RewriterCallback(RejitHandlerModule* moduleHandl
             &wrapper_type_def);
         if (FAILED(hr) || wrapper_type_def == mdTypeDefNil)
         {
-            Logger::Warn("*** CallTarget_RewriterCallback() Failed for: ", caller->type.name, ".", caller->name,
+            Logger::Error("*** CallTarget_RewriterCallback() Failed for: ", caller->type.name, ".", caller->name,
```
pellared, line 2497: Should it not be a warning? It is an oddity, but the application and instrumentation would continue to work without any bad side-effects? Should the user react when he notices such an error?

Maybe we should even "accept" such behavior so that we could use e.g. `#if NETFRAMEWORK` on a whole instrumentation class? Then we could even change this level to e.g. debug.

Leaving it for a discussion + potentially separate PR.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -2494,14 +2494,47 @@ HRESULT CorProfiler::CallTarget_RewriterCallback(RejitHandlerModule* moduleHandl
...
+            if (hr == S_OK && cTokens > 0)
+            {
+                found_wrapper_method = true;
+                break;
+            }
+        }
+
+        if (!found_wrapper_method)
+        {
+            Logger::Error("*** CallTarget_RewriterCallback() Failed for: ", caller->type.name, ".", caller->name,
```
pellared, line 2533: Some comment as above.

pellared (approved):

PR #1531, opened 2022-11-02
Fixes RuntimeInformation for .NET 4.6.2 ASP.NET test app

RassK:
## Why

Probably got broken with changing .NET Framework to lower version. 

`RuntimeInformation` is available from .NET Framework 4.7.1.
https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.frameworkdescription?view=net-6.0

## What

Fixes resolving 2 types in index.cshtml

```cshtml
<td>@(System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription)</td>
<td>@(System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture)</td>
```


pellared (approved):
nice catch

PR #1533, opened 2022-11-02
PS module default to ProgramFiles

### LINKED ISSUE #1527: Default location of AppData causes issues
# Bug Report

## Symptom

**Describe the bug**
The application fails to load the module with "Could not load file or assembly 'OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule' or one of its dependencies" because it installed into the users AppData instead of ProgramData. This is because the default -InstallDir is AppData.

**Expected behavior**
It installs to ProgramData as a global installation, with AppData as optional.

> This still grants the ability to limit the application to be within a users scope if required, however it feels more likely to want it to be installed as a general program than just a user. More context would be great if I am mislead.

**Screenshots**
NA

**Runtime environment (please complete the following information):**

Runtime environment (please complete the following information):

- OpenTelemetry Automatic Instrumentation version: [e.g. 1.0.0]
   v0.4.0-beta1

- OS: [e.g. Windows Server 2012 R2 ]
   Windows Server 2016

- .NET version: [e.g. .NET Framework 4.6.2, .NET Core 2.1]
   4.8

**Additional context**
N/A

## Reproduce

 1. Install with default module options, on a different user to that of the application
 2. Application announces an exception that the module could not be found
 3. Reinstall using -InstallDir ProgramData to resolve
 4. See telemetry data coming through as expected

## Notes

I am rather new to gathering telemtry data from IIS and ASP.NET applications and so any context on why this defaults to AppData would be great, additionally you are welcome to close it down if you feel it is not a valid modification.

RassK:
## Why

Fixes #1527 

## What

* Defaults to the ProgramFiles dir
* Adds warning when install path is custom and `Register-OpenTelemetryForIIS` is invoked

### DIFF
--- a/OpenTelemetry.DotNet.Auto.psm1
+++ b/OpenTelemetry.DotNet.Auto.psm1
@@ -23,12 +23,9 @@ function Get-Current-InstallDir() {
 function Get-CLIInstallDir-From-InstallDir([string]$InstallDir) {
     $dir = "OpenTelemetry .NET AutoInstrumentation"
     
-    if ($InstallDir -eq "<auto>" -or $installDir -eq "AppData") {
-        return (Join-Path $env:LOCALAPPDATA "Programs" | Join-Path -ChildPath $dir)
-    }
-    elseif ($InstallDir -eq "ProgramFiles") {
+    if ($InstallDir -eq "<auto>") {
         return (Join-Path $Env:ProgramFiles $dir)
-    } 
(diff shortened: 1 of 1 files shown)


pellared (approved):

pellared (approved):

pellared:
Maybe worth mentioning in the changelog.

PR #1539, opened 2022-11-03
Disable OpenTracing by default

### LINKED ISSUE #1534: Disable OpenTracing by default
Currently OpenTracing is always loaded and setup. We should control it with env and disable by default.
This is also saving startup time by not loading unnecessary OT dlls.

Related to https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1505

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs#L176

RassK:
## Why

Fixes #1534 

## What

Disables OpenTracing by default.

## Tests

Module load tests.

## Checklist


test/IntegrationTests/ModuleTests.cs
```diff
@@ -56,6 +56,7 @@ public async Task Minimal()
         SetEnvironmentVariable(ConfigurationKeys.Metrics.Exporter, Constants.ConfigurationValues.None);
         SetEnvironmentVariable(ConfigurationKeys.Traces.ConsoleExporterEnabled, bool.FalseString);
         SetEnvironmentVariable(ConfigurationKeys.Metrics.ConsoleExporterEnabled, bool.FalseString);
+        SetEnvironmentVariable(ConfigurationKeys.Traces.OpenTracingEnabled, bool.FalseString);
```
Kielek, line 59: As False is default, this probably can be removed.

Kielek (approved):

test/IntegrationTests/ModuleTests.cs
RassK: This was explicitly set to collect all vars that could load something. But can be removed if you see no additional value.

Kielek: I like the idea. We can keep it here.

pellared (approved):

PR #1542, opened 2022-11-03
Support options change in plugins

RassK:
## Why

Adds more control to distros via plugins.

## What

* Tracer instrumentation options and exporter options can be updated in plugins
* Metrics instrumentation options and exporter options can be updated in plugins
* Logs options and exporter options can be updated in plugins
* Unified settings initialization logic
* Plugins live through the application lifetime
    * Enables tracking of lazy instrumentation instances created in plugins (it must be separate, due logic is in auto side and we don't have integration layer).

## Tests

Updated: 
* Plugin tests
* Settings tests

## Checklist


src/OpenTelemetry.AutoInstrumentation/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        return ConfigureBuilder(builder, "ConfigureTracerProvider");
+    }
+
+    public MeterProviderBuilder ConfigureMeterProviderBuilder(MeterProviderBuilder builder)
+    {
+        return ConfigureBuilder(builder, "ConfigureMeterProvider");
+    }
+
+    public T ConfigureOptions<T>(T options)
```
pellared, line 54: I think there is no need to return the `T` as AFAIK the options are always a reference type so they can be simply modified

That is why the SDK accepts `Action<Option>` and not `Func<Option, Option>` e.g. here: https://github.com/open-telemetry/opentelemetry-dotnet/blob/91a9b7c549d8e7149bc3df27a907d5e14070a232/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpMetricExporterExtensions.cs#L57

RassK: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1542/commits/c18fa0c9b02bdaf311d23c04fdc5726e91359ee9

pjanotti (approved):
LGTM. One thing that is crossing my mind is if another model is also desirable. Currently to customize the SDK we have:

1. Plugins;
2. Deferring SDK initialization to the app;

A "plugin" that takes care of all SDK bootstrap is also desirable. The use case is further customization of the SDK that we can't anticipate. In other words, our current bootstrapper becomes the default but we allow the user to specify others. It seems a relatively low-cost and high-flexibility feature.

nrcventura (approved):

Kielek (approved):

pellared:
> LGTM. One thing that is crossing my mind is if another model is also desirable. Currently to customize the SDK we have:
> 
> 1. Plugins;
> 2. Deferring SDK initialization to the app;
> 
> A "plugin" that takes care of all SDK bootstrap is also desirable. The use case is further customization of the SDK that we can't anticipate. In other words, our current bootstrapper becomes the default but we allow the user to specify others. It seems a relatively low-cost and high-flexibility feature.

This idea was also coined here: https://docs.google.com/document/d/1R0_9T3LTslkmrJqvAaDutYz3R3_wHFQftwz9RdGS7iw/edit#heading=h.3609i26mjduw

dszmigielski (approved):

src/OpenTelemetry.AutoInstrumentation/Plugins/PluginManager.cs
```diff
@@ -0,0 +1,79 @@
...
+    {
+        var plugins = new List<(Type, object)>();
+
+        foreach (var assemblyQualifiedName in settings.Plugins)
+        {
+            var type = Type.GetType(assemblyQualifiedName, throwOnError: true);
+            var instance = Activator.CreateInstance(type);
+
+            plugins.Add((type, instance));
+        }
```
pellared, line 39: May be worth adding a comment that would explain why we are doing this in the constructor

As far as I understand it gives us the benefit that we will "fail-fast" in case of wrong plugin configuration.

The downside is that we have a memory footprint as the `_plugins` field is kept for the whole application lifecycle.

Probably we could consider just "validate" the configuration using reflection in the constructor (to have the "fail-fast") and call `Activator.CreateInstance` on demand in the `Configure` methods.

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetCoreInitializer.cs
```diff
@@ -16,22 +16,29 @@
...
     }
 
     public override void Initialize(ILifespanManager lifespanManager)
     {
         var instrumentationType = Type.GetType("OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreInstrumentation, OpenTelemetry.Instrumentation.AspNetCore");
         var httpInListenerType = Type.GetType("OpenTelemetry.Instrumentation.AspNetCore.Implementation.HttpInListener, OpenTelemetry.Instrumentation.AspNetCore");
 
-        var httpInListener = Activator.CreateInstance(httpInListenerType, args: new OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreInstrumentationOptions());
+        var options = new OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreInstrumentationOptions();
+        _pluginManager.ConfigureOptions(options);
```
pellared, line 39: This could crash the application at runtime if the plugin throws an exception. Right?

src/OpenTelemetry.AutoInstrumentation/Loading/MySqlDataInitializer.cs
```diff
@@ -17,22 +17,27 @@
...
     }
 
     public override void Initialize(ILifespanManager lifespanManager)
     {
         var instrumentationType = Type.GetType("OpenTelemetry.Instrumentation.MySqlData.MySqlDataInstrumentation, OpenTelemetry.Instrumentation.MySqlData");
-        var optionsInstrumentationType = Type.GetType("OpenTelemetry.Instrumentation.MySqlData.MySqlDataInstrumentationOptions, OpenTelemetry.Instrumentation.MySqlData");
 
-        var options = Activator.CreateInstance(optionsInstrumentationType);
+        var options = new OpenTelemetry.Instrumentation.MySqlData.MySqlDataInstrumentationOptions();
+        _pluginManager.ConfigureOptions(options);
```
pellared, line 39: This could crash the application at runtime if the plugin throws an exception. Right?

pellared (approved):
This PR is awesome 😉

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetCoreInitializer.cs
RassK: Seems something is handling the exception. Somehow the ASP.NET Core app reports exception (in debug mode), continues and produces spans 🤔

src/OpenTelemetry.AutoInstrumentation/Plugins/PluginManager.cs
pellared: we could consider also "disposing" the plugins if they implement `IDisposable`

Issue #925, opened 2022-07-06
Refactor the usage of optional parameters


pellared:
I think that our continued reliance on these optional parameters is going make the tests harder to maintain. On the one hand, it's nice that the test only needs to call `StartTestApplication` with a single named parameter. However, based on my experience with these usages, it gets harder and harder to keep track of which parameters are being used over time. That is why I prefer the overloads that take settings/config objects, which make it a lot easier to track down which specific settings are being used/customized.

_Originally posted by @nrcventura in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/900#discussion_r915037182 and https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/918#discussion_r915030492_

pellared:
> based on my experience with these usages, it gets harder and harder to keep track of which parameters are being used over time

Do you mean that it is harder to find "unused" named parameters? I mean a case when a non-default value is needed? I cannot imagine a scenario when the same could not happen if we use a "config object" instead of optional parameters. I think I need an example 😬

nrcventura:
What I've run into is something similar to the following. Consider the following code.
```
MyMethod(bool setting1 = true, bool setting2 = false)
MyMethod(Settings settings)

class Settings
{
    bool Setting1 { get; set; } = true
    bool Setting2 { get; set; } = false
}
```
Now assume that the `MyMethod` methods are invoked in a hundred locations, and we need to track down the locations where `setting2` is `true`. In Visual Studio this is really is to do with a property of class (right-click and find usages), but with optional method parameters this is not as easy. You have to find every usage of the method and then look closely to see if the named or ordered parameter is used and then evaluate if the value we care about is being set. The difficulty for tracking down the usage becomes even harder as the number of optional method parameters increase.

RassK:
I would prefer optional parameters, but they must be always named. 
It's much easier to search for `MyMethod.*setting2:\strue` than track the reference trail.

nrcventura:
> I would prefer optional parameters, but they must be always named. It's much easier to search for `MyMethod.*setting2:\strue` than track the reference trail.

[...]

PR #1543, opened 2022-11-04
Refine tests to reduce usage of optional args and optimize defaults

pellared:
## Why

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/925

## What

- Introduce `SetExporter` for setting the OTLP exporter env vars (it is type safe)
- Call `SetEnvironmentVariable("CORECLR_ENABLE_PROFILING", "1");` explicitly when bytecode instrumentation is needed for .NET Core
- Simplify `TestHelper` and `EnvironmentHelper`
- Simplify `TestSettings` and remove some types and fields
- Refactor `InstrumentedProcessHelper`
- Simplify `WcfServerTestHelper`
- Refactor `AspNetTests` (no need to inherit from `TestHelper`)
- Remove reduce noisy output in `WcfTestsBase`

## Tests

CI


RassK:
I see the pattern is now very close to how it should be used in real 👍

test/IntegrationTests/StrongNamedTests.cs
```diff
@@ -34,13 +34,15 @@ public StrongNamedTests(ITestOutputHelper output)
...
         using var collector = await MockSpansCollector.Start(Output);
+        SetExporter(collector);
         collector.Expect("TestApplication.StrongNamedValidation");
 
         var assemblyPath = GetTestAssemblyPath();
         var integrationsFile = Path.Combine(assemblyPath, "StrongNamedTestsIntegrations.json");
         File.Exists(integrationsFile).Should().BeTrue();
         SetEnvironmentVariable("OTEL_DOTNET_AUTO_INTEGRATIONS_FILE", integrationsFile);
-        RunTestApplication(otlpTraceCollectorPort: collector.Port);
+        SetEnvironmentVariable("CORECLR_ENABLE_PROFILING", "1"); // uses bytecode instrumentation
```
RassK, line 44: nit: As it's mostly constant, maybe do it similarly like `SetExporter()`. For example `EnableByteCodeInstrumentations()`?

RassK (approved):

test/IntegrationTests/StrongNamedTests.cs
pellared: I was considering it, but I was not sure if it would not be "over abstraction" 😉 I can address it in a separate PR.

test/IntegrationTests/DomainNeutralTests.cs
```diff
@@ -38,6 +38,7 @@ public async Task SubmitsTraces()
         EnvironmentTools.IsWindowsAdministrator().Should().BeTrue();
 
         using var collector = await MockSpansCollector.Start(Output);
+        SetExporter(collector);
```
pjanotti, line 41: Seeing that this pattern is going to repeat a lot it may be the case that a template pattern applies here. However, because of test method attributes it seems that this wouldn't be that smooth: there will be two method on the test class, the template itself plus a public one just to add the test attributes and call the base class method 🤔 

Not sure if it is worth... any let's leave it outside this PR.

pellared: I would rather add 3 methods (for each MockCollector) in `TestHelper` that starts a collector and sets it as an exporter.
I can do it in a separate PR.

Then this:

```csharp
        using var collector = await MockSpansCollector.Start(Output);
        SetExporter(collector);
```

would become e.g.

```
        using var collector = await StartSpansCollector();
```

test/IntegrationTests/SmokeTests.cs
```diff
@@ -55,23 +55,26 @@ public async Task SubmitsTraces()
...
+        await VerifyTestApplicationNotInstrumented();
 #endif
     }
 
     [Fact]
     [Trait("Category", "EndToEnd")]
     public async Task WhenClrProfilerIsNotEnabled()
     {
+        SetEnvironmentVariable("COR_ENABLE_PROFILING", "0");
+        SetEnvironmentVariable("CORECLR_ENABLE_PROFILING", "0");
```
pjanotti, line 72: ok to make it explicit it here given the test goal.

test/IntegrationTests/DomainNeutralTests.cs
pjanotti: 100% a separate PR.

test/IntegrationTests/SmokeTests.cs
pellared: Should I do something? I am asking because I am not sure 😉

pjanotti: no, it is good as it is ;)

test/IntegrationTests/GraphQLTests.cs
```diff
@@ -44,6 +44,7 @@ public async Task SubmitsTraces(bool setDocument)
     {
         var requests = new List<RequestInfo>();
         using var collector = await MockSpansCollector.Start(Output);
+        SetExporter(collector);
```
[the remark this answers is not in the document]
pellared: I started to hesite as we also have `SetEnviromentalVariable`... Is it a strong opinion? Take notice that most of the usages are going to be refactored in next PR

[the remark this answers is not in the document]
pjanotti: Not a strong opinion, it also can be delayed if most usages are going away, at that time we can look at the name again.

test/IntegrationTests/LogTests.cs
```diff
@@ -52,13 +53,17 @@ public async Task SubmitLogs(bool enableClrProfiler, bool includeFormattedMessag
             collector.Expect(logRecord => Convert.ToString(logRecord).Contains("TestApplication.Logs.Controllers.TestController"));
         }
 
-        if (!enableClrProfiler)
+        if (enableClrProfiler)
+        {
+            SetEnvironmentVariable("CORECLR_ENABLE_PROFILING", "1");
```
nrcventura, line 58: Since this is something that is done in multiple tests, we can probably move this to its own method so that we don't have that environment variable name copied everywhere. I think that this can be done in a separate PR.

test/IntegrationTests/StrongNamedTests.cs
nrcventura: I think that this usage is common enough that it's worth having this abstraction so that we can minimize the chance of someone typing the environment variable name incorrectly. I agree that making such a change should be done in a separate PR.

nrcventura (approved):

Kielek (approved):

test/IntegrationTests/Helpers/EnvironmentHelper.cs
```diff
@@ -365,4 +295,28 @@ private static string GetAdditionalDepsPath()
...
+    private void SetDefaultEnvironmentVariables()
+    {
+        string profilerPath = GetProfilerPath();
+
+        CustomEnvironmentVariables["DOTNET_STARTUP_HOOKS"] = GetStartupHookOutputPath();
+        CustomEnvironmentVariables["DOTNET_SHARED_STORE"] = GetSharedStorePath();
+        CustomEnvironmentVariables["DOTNET_ADDITIONAL_DEPS"] = GetAdditionalDepsPath();
+
+        // when bytecode instrumentation is needed
+        // CoreCLR Profiler must be expliclitly enabled in test in needed by setting CORECLR_ENABLE_PROFILING=1
```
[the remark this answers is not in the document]
pellared: I will address it in a separate PR so that everyone could review the improved comment 😉

PR #1545, opened 2022-11-08
Refactor TestHttpServer and fix HttpListener leakage

pellared:
## Why

Move some logic from MockCollectors to the TestHttpServer in order to remove some redundant lines of code.

image

## What

- Refactor HttpListener and related code.
- Fix leakage of `HttpListener` instances

### DIFF
--- a/test/IntegrationTests/Helpers/TestHttpServer.NetFramework.cs
+++ b/test/IntegrationTests/Helpers/TestHttpServer.NetFramework.cs
@@ -63,7 +63,7 @@ public class TestHttpServer : IDisposable
 
                 // call healthz (wait until HttpListener is operational)
                 var healhtzEndpoint = $"{_prefix.Replace("*", "localhost")}/healthz";
-                var healthz = HealthzHelper.TestHealtzAsync(healhtzEndpoint, nameof(MockLogsCollector), _output);
+                var healthz = HealthzHelper.TestHealtzAsync(healhtzEndpoint, nameof(TestHttpServer), _output);
                 if (!healthz.Result)
                 {
                     _listener.Close(); // always close listener if exception is thrown, whether it was caught or not


Kielek (approved):

test/IntegrationTests/Helpers/TestHttpServer.NetFramework.cs
```diff
@@ -61,18 +61,25 @@ public TestHttpServer(ITestOutputHelper output, Action<HttpListenerContext> requ
                 _listenerThread.Start();
                 WriteOutput($"Listening on '{_prefix}'");
 
+                // call healthz (wait until HttpListener is operational)
+                var healhtzEndpoint = $"{_prefix.Replace("*", "localhost")}/healthz";
+                var healthz = HealthzHelper.TestHealtzAsync(healhtzEndpoint, nameof(MockLogsCollector), _output);
```
RassK, line 66:
```suggestion
                var healthz = HealthzHelper.TestHealtzAsync(healhtzEndpoint, nameof(TestHttpServer), _output);
```

test/IntegrationTests/Helpers/TestHttpServer.NetFramework.cs
```diff
@@ -61,18 +61,25 @@ public TestHttpServer(ITestOutputHelper output, Action<HttpListenerContext> requ
                 _listenerThread.Start();
                 WriteOutput($"Listening on '{_prefix}'");
 
+                // call healthz (wait until HttpListener is operational)
+                var healhtzEndpoint = $"{_prefix.Replace("*", "localhost")}/healthz";
```
RassK, line 65: Might not be needed any more at all. Since the bug happens with .NET Core, but .NET Core has now different HttpServer base.

test/IntegrationTests/Helpers/MockMetricsCollector.cs
```diff
@@ -46,7 +44,7 @@ public class MockMetricsCollector : IDisposable
     private readonly List<Expectation> _expectations = new();
     private readonly BlockingCollection<List<Collected>> _metricsSnapshots = new(10); // bounded to avoid memory leak; contains protobuf type
 
-    private MockMetricsCollector(ITestOutputHelper output, string host = "localhost")
+    public MockMetricsCollector(ITestOutputHelper output, string host = "localhost")
```
RassK, line 47: There is an issue left by me, the host is only available for .NET Fx
If not now, then maybe for some future refactor.

RassK (approved):

test/IntegrationTests/Helpers/MockMetricsCollector.cs
pellared: Yeah I saw it and I was even thinking if I should address it... But I decided that is not needed now so omitted it. AFAIK it is needed only for `AspNetTests`

test/IntegrationTests/Helpers/TestHttpServer.NetFramework.cs
pellared: I will make a PR to remove it and run all tests a lot of times to make sure it is stable.

PR #1546, opened 2022-11-08
Remove GraphQL error tags

### LINKED ISSUE #1485: GraphQL instrumentation adds non-OTel compliant error attributes
## Actual

GraphQL (and possibly other bytecode instrumentations) use https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/da42322b088b6d490456e524eaa9ca61b3c5bf9d/src/OpenTelemetry.AutoInstrumentation/Tagging/Tags.cs#L26-L44

## Expected

For setting the status we probably should use: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.setstatus?view=net-6.0#system-diagnostics-activity-setstatus(system-diagnostics-activitystatuscode-system-string)

For marking the exception we should rather use OTel API: https://github.com/open-telemetry/opentelemetry-dotnet/blob/16e0758c148c6bac26cf4a31a6b405482a798d84/src/OpenTelemetry.Api/Trace/ActivityExtensions.cs#L89

References in spec:
- https://github.com/open-telemetry/opentelemetry-specification/blob/8ed127a403c59e1baa5203c6cdf472ea21df0332/specification/common/mapping-to-non-otlp.md#span-status
- https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md
Most probably https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/Trace/ActivityExtensions.cs should be used

See also:
- https://github.com/open-telemetry/opentelemetry-dotnet/blob/ded2eefcc20da3b27974a313cc52fc1d0f63e997/src/OpenTelemetry.Api/Internal/SemanticConventions.cs#L106-L109
- https://github.com/open-telemetry/opentelemetry-dotnet/blob/ded2eefcc20da3b27974a313cc52fc1d0f63e997/src/OpenTelemetry.Api/Internal/SpanAttributeConstants.cs#L24-L25

RassK:
## Why

Fixes #1485 

## What

Removes manual error tags from GraphQL instrumentation.

## Tests

Existing.

## Checklist [TODO]

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Instrumentations/GraphQL/ExecuteAsyncIntegration.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Instrumentations/GraphQL/ExecuteAsyncIntegration.cs
@@ -76,10 +76,6 @@ public class ExecuteAsyncIntegration
             {
                 activity?.SetException(exception);
             }
-            else if (state.State is IExecutionContext context)
-            {
-                GraphQLCommon.RecordExecutionErrorsIfPresent(activity, ErrorType, context.Errors);
-            }
         }
         finally
         {
--- a/src/OpenTelemetry.AutoInstrumentation/Instrumentations/GraphQL/GraphQLCommon.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Instrumentations/GraphQL/GraphQLCommon.cs
@@ -16,13 +16,9 @@
 
 using System;
 using System.Diagnostics;
-using System.Text;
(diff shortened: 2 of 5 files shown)


RassK:
In general, I think, the whole GraphQL implementation is quite weird, it's a mix of client and server side instrumentations without a clear border.

RassK:
There seems to be something else wrong.

PR #1547, opened 2022-11-08
Remove-httplistener-healthz

pellared:
## Why

Follows https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1545

## What

- Remove `healthz` from HttpListener as it was a workaround for a bug present on non-Windows implementation
- Refine HealthzHelper based on `GraphQLTests` and reuse it

### DIFF
--- a/test/IntegrationTests/Helpers/HealthzHelper.cs
+++ b/test/IntegrationTests/Helpers/HealthzHelper.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Net;
 using System.Net.Http;
+using System.Threading;
 using System.Threading.Tasks;
 using Xunit.Abstractions;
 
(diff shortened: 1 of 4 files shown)


test/IntegrationTests/Helpers/HealthzHelper.cs
```diff
@@ -17,40 +17,42 @@
...
+                output.WriteLine($"Healthz endpoint retured HTTP status code: {response.StatusCode}");
+            }
+            catch (Exception ex)
             {
-                return true;
+                output.WriteLine($"Healthz endpoint call failed: {ex.Message}");
             }
 
-            output.WriteLine($"{logPrefix} healthz failed {retry + 1}/5");
-            await Task.Delay(TimeSpan.FromSeconds(4));
```
pellared, line 51: I remember this could be problematic (discussed with @RassK  some time ago).

nrcventura (approved):

Kielek (approved):

test/IntegrationTests/Helpers/HealthzHelper.cs
pjanotti: `Thread.Sleep` does block the thread, so you are not competing with other async/awaits to run it. However, the possible problem for both, i.e.: failing to wake up at the deadline, is due to CPU oversubscription. This is not going to be fixed by `Thread.Sleep`. In general, async/wait is much preferable to blocking a thread. That said I don't think it will make a material difference here.

pjanotti (approved):

Issue #1437, opened 2022-10-17
Only inject instrumentation assemblies that are going to be used


pjanotti:
We have the goal of automatically enabling all available instrumentations. However, some of the instrumentations have extension methods that perform initialization directly accessing the `TraceProviderSdk` directly using these methods at initialization will inject dependencies on the process that are not used in practice. If the extension method only enables an activity source the cost is expected to be small. That said some of the extension methods can use types and code that comes from the actual dependencies and they may cause the process to load more assemblies and using code that is not needed in practice. This will affect start up time.

This is related to a previous discussion when enabling all instrumentations, see #388 and an issue on the SDK, see https://github.com/open-telemetry/opentelemetry-dotnet/issues/3253

At first it seems to me that we should pursue again having the capability of calling the "add" extension methods after the initial creation of the provider.

Extra exporter or SDK related modules are addressed via #1505.

pellared:
I think we already have this possibility via `LazyInstrumentationLoader` (unless I am missing something). It is used for `MySqlData` and `AspNetCore` instrumentations.

@RassK PTAL

pjanotti:
@pellared and @RassK I want to avoid loading the instrumentations like WCF. The `LazyInstrumentationLoader` is useful for instrumentations that don't require the `builder`, however, some of the instrumentations require that. I didn't look if such instrumentations really need access to the builder or if it is just a bad pattern.

I'm planning to look into this later and if there is a real need to access the builder for some instrumentations we need to go back to the SDK and ask for the capability of adding instrumentations after `Build()`.

RassK:
@pjanotti  I see most of the builder extensions are just for creating the instrumentation instance and subscribing to some source. `LazyInstrumentationLoader` basically does the same thing but separates these 2 parts and invokes those parts in different time frame.

Wcf is a bit exceptional. Instead of creating instrumentation instance, it just creates and assigns global options. It's again nothing that `LazyInstrumentationLoader` can't handle.

pjanotti:
yes, that is the preferred pattern. Although we can use reflection to access the internals of instrumentation packages and use the `LazyInstrumentationLoader` we should move such packages to make related APIs public. This will ensure that they won't break automatic instrumentation in minor releases.

We should review all instrumentations that are being added via extension methods and check which ones can and can't be switched to `LazyInstrumentationLoader`

PR #1605, opened 2022-11-09
Make ASP.NET instrumentation lazy

RassK:
## Why

Towards #1437 

## What

* Makes ASP.NET loading lazy (traces + metrics)

## Tests

Existing

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Loading/AspNetInitializer.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Loading/AspNetInitializer.cs
@@ -1,0 +1,56 @@
+// <copyright file="AspNetInitializer.cs" company="OpenTelemetry Authors">
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+#if NETFRAMEWORK
+
+using System;
+using System.Threading;
+using OpenTelemetry.AutoInstrumentation.Plugins;
+
+namespace OpenTelemetry.AutoInstrumentation.Loading;
+
+internal class AspNetInitializer
+{
+    private readonly PluginManager _pluginManager;
+
+    private int _firstInitialization = 1;
+
+    public AspNetInitializer(LazyInstrumentationLoader lazyInstrumentationLoader, PluginManager pluginManager)
+    {
+        _pluginManager = pluginManager;
+        lazyInstrumentationLoader.Add(new AspNetMvcInitializer(InitializeOnFirstCall));
+        lazyInstrumentationLoader.Add(new AspNetWebApiInitializer(InitializeOnFirstCall));
+    }
+
+    private void InitializeOnFirstCall(ILifespanManager lifespanManager)
+    {
+        if (Interlocked.Exchange(ref _firstInitialization, value: 0) != 1)
+        {
+            // InitializeOnFirstCall() was already called before
+            return;
+        }
+
+        var instrumentationType = Type.GetType("OpenTelemetry.Instrumentation.AspNet.AspNetInstrumentation, OpenTelemetry.Instrumentation.AspNet");
+
+        var options = new OpenTelemetry.Instrumentation.AspNet.AspNetInstrumentationOptions();
+        _pluginManager.ConfigureOptions(options);
+
+        var instrumentation = Activator.CreateInstance(instrumentationType, args: options);
+
+        lifespanManager.Track(instrumentation);
+    }
+}
+#endif
--- a/src/OpenTelemetry.AutoInstrumentation/Loading/AspNetInitializer.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Loading/AspNetInitializer.cs
@@ -37,7 +37,7 @@ internal class AspNetInitializer
 
     private void InitializeOnFirstCall(ILifespanManager lifespanManager)
     {
-        if (Interlocked.Exchange(ref _initialized, value: 1) != 0)
+        if (Interlocked.Exchange(ref _initialized, value: 1) != default)
         {
             // InitializeOnFirstCall() was already called before
             return;
--- a/src/OpenTelemetry.AutoInstrumentation/Loading/AspNetInitializer.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Loading/AspNetInitializer.cs
@@ -26,7 +26,7 @@ internal class AspNetInitializer
 {
     private readonly PluginManager _pluginManager;
 
-    private int _firstInitialization = 1;
+    private int _initialized;
 
     public AspNetInitializer(LazyInstrumentationLoader lazyInstrumentationLoader, PluginManager pluginManager)
     {
@@ -37,7 +37,7 @@ internal class AspNetInitializer
 
     private void InitializeOnFirstCall(ILifespanManager lifespanManager)
     {
-        if (Interlocked.Exchange(ref _firstInitialization, value: 0) != 1)
+        if (Interlocked.Exchange(ref _initialized, value: 1) != 0)
         {
             // InitializeOnFirstCall() was already called before
             return;
--- a/src/OpenTelemetry.AutoInstrumentation/Loading/AspNetMetricsInitializer.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Loading/AspNetMetricsInitializer.cs
@@ -1,0 +1,49 @@
+// <copyright file="AspNetMetricsInitializer.cs" company="OpenTelemetry Authors">
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+#if NETFRAMEWORK
+
+using System;
+using System.Threading;
+
+namespace OpenTelemetry.AutoInstrumentation.Loading;
+
+internal class AspNetMetricsInitializer
+{
+    private int _firstInitialization = 1;
+
+    public AspNetMetricsInitializer(LazyInstrumentationLoader lazyInstrumentationLoader)
+    {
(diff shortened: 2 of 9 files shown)


src/OpenTelemetry.AutoInstrumentation/Loading/AspNetInitializer.cs
```diff
@@ -0,0 +1,56 @@
...
+        lazyInstrumentationLoader.Add(new AspNetWebApiInitializer(InitializeOnFirstCall));
+    }
+
+    private void InitializeOnFirstCall(ILifespanManager lifespanManager)
+    {
+        if (Interlocked.Exchange(ref _firstInitialization, value: 0) != 1)
+        {
+            // InitializeOnFirstCall() was already called before
+            return;
+        }
```
pellared, line 44: personal preference (maybe a little more performant?)

```suggestion
    private int _initialized;

    public AspNetInitializer(LazyInstrumentationLoader lazyInstrumentationLoader, PluginManager pluginManager)
    {
        _pluginManager = pluginManager;
        lazyInstrumentationLoader.Add(new AspNetMvcInitializer(InitializeOnFirstCall));
        lazyInstrumentationLoader.Add(new AspNetWebApiInitializer(InitializeOnFirstCall));
    }

    private void InitializeOnFirstCall(ILifespanManager lifespanManager)
    {
        if (Interlocked.Exchange(ref _initialized, value: 1) != 0)
        {
            // InitializeOnFirstCall() was already called before
            return;
        }
```

maybe `default` instead of `0` would be more readable (not sure if it will compile TBH)

let me know what you prefer

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
```diff
@@ -47,34 +50,67 @@ private class OnAssemblyLoadInitializer
         private readonly ILifespanManager _lifespanManager;
         private readonly string _requiredAssemblyName;
 
+#if NETFRAMEWORK
+        private int _firstInitialization = 1;
```
pellared, line 54: can you elaborate (preferably as comments in code) why these changes are needed?

Can  `AppDomain.CurrentDomain.AssemblyLoad` can be called from multiple threads in for the same assembly on .NET Framework?

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
```diff
@@ -47,34 +50,67 @@ private class OnAssemblyLoadInitializer
...
             AppDomain.CurrentDomain.AssemblyLoad += CurrentDomain_AssemblyLoad;
+
+#if NETFRAMEWORK
+            var isRequiredAssemblyLoaded = AppDomain.CurrentDomain
+                .GetAssemblies()
+                .Any(x => GetAssemblyName(x) == _requiredAssemblyName);
+            if (isRequiredAssemblyLoaded)
+            {
+                OnRequiredAssemblyDetected();
+            }
```
pellared, line 72: I understand that this code is to check if the assembly was already loaded... Maybe it should be not only for .NET Framework? I guess it could happen for .NET as well.

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
```diff
@@ -47,34 +50,67 @@ private class OnAssemblyLoadInitializer
...
-
-                AppDomain.CurrentDomain.AssemblyLoad -= CurrentDomain_AssemblyLoad;
+                OnRequiredAssemblyDetected();
+            }
+        }
+
+        private void OnRequiredAssemblyDetected()
+        {
+#if NETFRAMEWORK
+            if (Interlocked.Exchange(ref _firstInitialization, value: 0) != 1)
```
pellared, line 89: Why this is needed?

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
RassK: There are multiple important points: 
* In .NET Framework we don't have startup hook that executes before the main content. So the startup is injected to presumably to the beginning of the main method. 
* There is a race condition between checking if the assembly is already loaded and loading event. 
* There is a race condition after the determination logic, so checking logic is doubled with the event and pre-check logic

So basically, the check is doubled not to have any slight chance where `isRequiredAssemblyLoaded = false` and assembly load event is not subscribed.

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetInitializer.cs
RassK: no preferences at all, can take yours if you like it more.

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1605#discussion_r1019553855 first bullet.

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1605#discussion_r1019553855 third bullet.

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
pellared: But for .NET it could also happen if one uses "another startup hook" which is executed before "OTel startup hook", right? I just think if he should not get rid of `#if NETFRAMEWORK`

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetInitializer.cs
pellared: for me it is more readable as it uses "zero value" as default/initial value 😉 also it should be a lot faster (kidding)

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
pellared: Also I suggest adding the important points as comments in code 😉

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetInitializer.cs
RassK: Just in case mentioning that I got it from the `Instrument.cs`, I have no idea, why the default is 1 there 😄

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
RassK: Do you think that 1:1 is good enough or you would like to adjust the wording there?

RassK: > But for .NET it could also happen if one uses "another startup hook" which is executed before "OTel startup hook", right? I just think if he should not get rid of #if NETFRAMEWORK

If you want to go there, I guess it gets deeper. I'm not sure if some would use different load context for eg, what happens then.

pellared: > Do you think that 1:1 is good enough or you would like to adjust the wording there?

That is good enough for me 😉

pellared (approved):

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetInitializer.cs
```diff
@@ -0,0 +1,56 @@
...
+
+    private void InitializeOnFirstCall(ILifespanManager lifespanManager)
+    {
+        if (Interlocked.Exchange(ref _initialized, value: 1) != default)
+        {
+            // InitializeOnFirstCall() was already called before
+            return;
+        }
+
+        var instrumentationType = Type.GetType("OpenTelemetry.Instrumentation.AspNet.AspNetInstrumentation, OpenTelemetry.Instrumentation.AspNet");
```
pjanotti, line 46: Because of the dependencies issue, especially with S.D.DS, it seems that is possible at the moment to get null here. The code should check for that and log as appropriate.

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetInitializer.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        var options = new OpenTelemetry.Instrumentation.AspNet.AspNetInstrumentationOptions();
```
pjanotti, line 48: I'm uncertain why to get the `instrumentationType` reflection is used and here it is ok to reference the type directly. Could you please clarify?

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetMetricsInitializer.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    private void InitializeOnFirstCall(ILifespanManager lifespanManager)
+    {
+        if (Interlocked.Exchange(ref _initialized, value: 1) != default)
+        {
+            // InitializeOnFirstCall() was already called before
+            return;
+        }
+
+        var instrumentationType = Type.GetType("OpenTelemetry.Instrumentation.AspNet.AspNetMetrics, OpenTelemetry.Instrumentation.AspNet");
```
pjanotti, line 42: `instrumentationType` can be null

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetInitializer.cs
RassK: The only public api in instrumentation packages is options type, everything else is internal.

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetMetricsInitializer.cs
RassK: That's our shipped library, I guess it could be null only when installation corrupts? - but then we have loads of other places unprotected.

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetInitializer.cs
RassK: That happens so late in the runtime, that we are probably crashed way earlier if that happens.

pellared: Even if it throws then it should be handled here https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/abeb491aabac80b62588fdb87f7536f7520e1b71/src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs#L69, right?

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetMetricsInitializer.cs
pjanotti: Indirect dependencies are the scenario that this may fail. Anyway, it is the same case as the other comment.

src/OpenTelemetry.AutoInstrumentation/Loading/AspNetInitializer.cs
pjanotti: makes sense

pjanotti (approved):

PR #1613, opened 2022-11-13
Add tests for enabling and disabling instrumentations

### LINKED ISSUE #1480: Add tests for `OTEL_DOTNET_AUTO_*_DISABLED_INSTRUMENTATIONS=[id]`
Can you please add a test in `LogTests` that would set `OTEL_DOTNET_AUTO_LOGS_DISABLED_INSTRUMENTATIONS=ILogger` and assert that none logs are send using `collector.AssertEmpty(5.Seconds());`? It can be named  e.g. `Disabled`.

It would test the instrumentation can be disabled using `ILogger` ID. Otherwise we would have not tests that check the instrumentation IDs. 

**Similar tests should be added for ALL ~~instrumentations~~ signals (traces and metrics)** (as a separate PR).

If you want you we can address even this comment in a separate PR.

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1475#discussion_r1001539610_

### LINKED ISSUE #1481: Add tests for `OTEL_DOTNET_AUTO_*_ENABLED_INSTRUMENTATIONS=none`
Can you please add a test in `SmokeTests` similar to `SubmitLogs` that would set `OTEL_DOTNET_AUTO_LOGS_ENABLED_INSTRUMENTATIONS=none` and assert that none logs are send using `collector.AssertEmpty(5.Seconds());`? 

It would test the `OTEL_DOTNET_AUTO_LOGS_ENABLED_INSTRUMENTATIONS` env var. 

Similar tests should be added for traces and metrics (as a separate PR).

I want to have this smoke test added as part of this PR to have any integration tests verifying the new functionality.

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1475#discussion_r1001556498_

birojnayak:
## Why

Fixes #1480
Fixes  #1481

## What

Ensure test coverage when `OTEL_DOTNET_AUTO_*_ENABLED_INSTRUMENTATIONS` is `none`

Ensure test coverage when `OTEL_DOTNET_AUTO_*_DISABLED_INSTRUMENTATIONS` is set

## Tests

New SmokeTests.

## Checklist


test/IntegrationTests/SmokeTests.cs
```diff
@@ -295,6 +295,38 @@ public void LogsNoneInstrumentations()
 
         collector.AssertEmpty(5.Seconds());
     }
+
+    [Fact]
+    [Trait("Category", "EndToEnd")]
+    public void TracesNoneInstrumentations()
```
pellared, line 301: Why this test is inside `#if !NETFRAMEWORK // The feature is not supported on .NET Framework`?

test/IntegrationTests/SmokeTests.cs
```diff
@@ -295,6 +295,38 @@ public void LogsNoneInstrumentations()
...
+    {
+        using var collector = new MockSpansCollector(Output);
+        SetEnvironmentVariable("OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS", "none");
+        RunTestApplication();
+        collector.AssertEmpty(5.Seconds());
+    }
+
+    [Fact]
+    [Trait("Category", "EndToEnd")]
+    public void MetricsNoneInstrumentations()
```
pellared, line 311: Why this test is inside `#if !NETFRAMEWORK // The feature is not supported on .NET Framework`?

test/IntegrationTests/SmokeTests.cs
```diff
@@ -295,6 +295,38 @@ public void LogsNoneInstrumentations()
...
+    {
+        using var collector = new MockMetricsCollector(Output);
+        SetEnvironmentVariable("OTEL_DOTNET_AUTO_METRICS_ENABLED_INSTRUMENTATIONS", "none");
+        RunTestApplication();
+        collector.AssertEpmty();
+    }
+
+    [Fact]
+    [Trait("Category", "EndToEnd")]
+    public void  AutoLogsDisabledInstrumentationsForILogger()
```
pellared, line 321: How about renaming to

```suggestion
    public void LogsDisabledInstrumentation()
```

?

pellared, line 321: Why this test is inside `#if !NETFRAMEWORK // The feature is not supported on .NET Framework`?

pellared, line 321: There are missing similar tests for `OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS` and `OTEL_DOTNET_AUTO_METRICS_DISABLED_INSTRUMENTATIONS`

pellared (commented):
Nice. This is the correct direction 👍 A few comments have to be addressed.

test/IntegrationTests/SmokeTests.cs
birojnayak: added 2 more tests... it was bit hard to find how to configure though..

test/IntegrationTests/Helpers/MockMetricsCollector.cs
```diff
@@ -145,6 +145,41 @@ public void AssertExpectations(TimeSpan? timeout = null)
         }
     }
 
+    internal bool CheckInstrumentationScopeIsAbsent(List<string> instrumentationScopes, TimeSpan? timeout = null)
+    {
+        timeout ??= DefaultWaitTimeout;
+        var cts = new CancellationTokenSource();
```
Kielek, line 151: As I remember it is disposable object.

test/IntegrationTests/SmokeTests.cs
```diff
@@ -271,8 +273,84 @@ public void LogsNoneInstrumentations()
...
+        Assert.False(
+            collector.CheckInstrumentationScopeIsAbsent(
+            new List<string>() { "OpenTelemetry.Instrumentation.Process" },
+            1.Seconds()),
+            "Didn't find metrics instrumentation while it should be present.");
+        Assert.True(
+            collector.CheckInstrumentationScopeIsAbsent(
+            new List<string>() { "OpenTelemetry.Instrumentation.NetRuntime", "OpenTelemetry.Instrumentation.HttpClient" },
+            1.Seconds()),
+            "Found Metrics instrumentation while it should be disabled.");
```
pellared, line 333: This can lead to flakiness.

Each `CheckInstrumentationScopeIsAbsent` is using `_metricsSnapshots.GetConsumingEnumerable()`. Take notice that whatever the first call of `CheckInstrumentationScopeIsAbsent` will process then it won't be processed by the second call of `CheckInstrumentationScopeIsAbsent`. Because of that we can have false positives (passing tests which should fail) if there is a bug in production code.

Personally, I think it would be good enough to have a test like this:

```csharp
    [Fact]
    public void MetricsDisabledInstrumentation()
    {
        using var collector = new MockMetricsCollector(Output);
        SetEnvironmentVariable("OTEL_DOTNET_AUTO_METRICS_ENABLED_INSTRUMENTATIONS", "HttpClient");
        SetEnvironmentVariable("OTEL_DOTNET_AUTO_METRICS_DISABLED_INSTRUMENTATIONS", "HttpClient");
        EnableBytecodeInstrumentation();
        RunTestApplication();
        collector.AssertEmpty(1.Seconds());
    }
```

Other tests are checking if `OTEL_DOTNET_AUTO_METRICS_ENABLED_INSTRUMENTATIONS` is working. This tests would test if `OTEL_DOTNET_AUTO_METRICS_DISABLED_INSTRUMENTATIONS` is working and also if it has higher priority than `OTEL_DOTNET_AUTO_METRICS_ENABLED_INSTRUMENTATIONS`. We would also not need to implement something like `CheckInstrumentationScopeIsAbsent` 😉

test/IntegrationTests/SmokeTests.cs
```diff
@@ -271,8 +273,84 @@ public void LogsNoneInstrumentations()
...
+
+        StringBuilder sbToDisableMetrics = new StringBuilder();
+        sbToDisableMetrics.Append(TracerInstrumentation.HttpClient.ToString())
+        .Append(",").Append(TracerInstrumentation.AspNet.ToString());
+
+        SetEnvironmentVariable("OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS", sbToDisableMetrics.ToString());
+        RunTestApplication();
+        Assert.False(
+            collector.ResourceExpector.CheckForAnyResourceAttributes(1.Seconds()),
+            "There shouldn't be any tracing.");
```
pellared, line 351: Why not 
```suggestion
        collector.AssertEmpty(1.Seconds());
```
?

test/IntegrationTests/SmokeTests.cs
```diff
@@ -271,8 +273,84 @@ public void LogsNoneInstrumentations()
...
+    public void TracesDisabledInstrumentation()
+    {
+        using var collector = new MockSpansCollector(Output);
+        SetExporter(collector);
+
+        StringBuilder sbToDisableMetrics = new StringBuilder();
+        sbToDisableMetrics.Append(TracerInstrumentation.HttpClient.ToString())
+        .Append(",").Append(TracerInstrumentation.AspNet.ToString());
+
+        SetEnvironmentVariable("OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS", sbToDisableMetrics.ToString());
```
pellared, line 347: Clever (really!). However, I would prefer to have it simple as I think it would be more readable to have

```suggestion
        SetEnvironmentVariable("OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS", "AspNet,HttpClient");
```

This is an integration test which is more on the "user level". The env vars changes should not be threated lightly as this is our user facing interface.

Also if the env var value would change the enum would change anyway 😉

test/IntegrationTests/SmokeTests.cs
birojnayak: thanks ! I would say this is because my lack of knowledge... I thought all Metrics Instrumentation are enabled by default and I need to set and check that's not present... didn't know if you set "OTEL_DOTNET_AUTO_METRICS_ENABLED_INSTRUMENTATIONS", "HttpClient" , only that metrics would be selected rather than all.

pellared (approved):

test/IntegrationTests/SmokeTests.cs
```diff
@@ -17,9 +17,11 @@
 using System.Net;
 using System.Net.Http;
 using System.Reflection;
+using System.Text;
```
pellared, line 20: is this still needed?

test/IntegrationTests/SmokeTests.cs
```diff
@@ -17,9 +17,11 @@
...
 using FluentAssertions;
 using FluentAssertions.Extensions;
 using IntegrationTests.Helpers;
+using OpenTelemetry.AutoInstrumentation.Configuration;
```
pellared, line 24: is this still needed?

Issue #1638, opened 2022-11-18
Refactor initializers with multiple dependency checks


RassK:
_First mentioned by @pellared at https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1628#discussion_r1025422022_

### What

Initializers with multiple sub dependency trackers follow different pattern.

Preferred pattern vs invalid.

```csharp
lazyInstrumentationLoader.Add(new WcfInitializer(pluginManager));
```

```csharp
new HttpClientInitializer(lazyInstrumentationLoader, pluginManager);
```

Affected scope:
* `HttpClientInitializer.cs`
* `AspNetInitializer.cs`
* `AspNetMetricsInitializer.cs` 

### Todo

Refactor is needed.

RassK:
not relevant any more.

PR #1639, opened 2022-11-18
Make HttpClient instrumentation lazy (Metrics)

### LINKED ISSUE #1437: Only inject instrumentation assemblies that are going to be used

RassK:
## Why

Fixes #1437 

_I think this is the last possible optimization, since Runtime and Process instrumentations will be loaded anyway and can be controlled via enabled instrumentations._

## What

* Makes Http instrumentation loading lazy (metrics)
* Add no exporter tests 
    * Covers scenario when only console exporter is in use.
    * Futureproofs if there will be other exporters without http required.
    * Catches escaping optimizations where our exporters could obfuscate result.
* **Fixes Http instrumentation loading for tracing (bug)** 
* Fixes bug on .NET Core where exporter causes dependency load, but initializer is not yet setup.

## Tests

Existing + Fixes + DefaultNoExporter verification

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationMetricHelper.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationMetricHelper.cs
@@ -35,21 +35,23 @@ internal static class EnvironmentConfigurationMetricHelper
         MetricSettings settings,
         PluginManager pluginManager)
     {
-        builder
-            .SetExporter(settings, pluginManager)
-            .AddMeter(settings.Meters.ToArray());
-
         foreach (var enabledMeter in settings.EnabledInstrumentations)
         {
             _ = enabledMeter switch
             {
                 MetricInstrumentation.AspNet => Wrappers.AddAspNetInstrumentation(builder, lazyInstrumentationLoader),
-                MetricInstrumentation.HttpClient => Wrappers.AddHttpClientInstrumentation(builder),
+                MetricInstrumentation.HttpClient => Wrappers.AddHttpClientInstrumentation(builder, lazyInstrumentationLoader),
                 MetricInstrumentation.NetRuntime => Wrappers.AddRuntimeInstrumentation(builder, pluginManager),
                 MetricInstrumentation.Process => Wrappers.AddProcessInstrumentation(builder, pluginManager),
                 _ => null,
             };
         }
+
+        // Exporters can cause dependency loads.
+        // Should be called later if dependency listeners are already setup.
+        builder
+            .SetExporter(settings, pluginManager)
+            .AddMeter(settings.Meters.ToArray());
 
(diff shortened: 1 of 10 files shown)


src/OpenTelemetry.AutoInstrumentation/Configuration/EnvironmentConfigurationMetricHelper.cs
```diff
@@ -86,16 +88,20 @@ public static MeterProviderBuilder AddAspNetInstrumentation(MeterProviderBuilder
...
 
             return builder;
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        public static MeterProviderBuilder AddHttpClientInstrumentation(MeterProviderBuilder builder)
+        public static MeterProviderBuilder AddHttpClientInstrumentation(MeterProviderBuilder builder, LazyInstrumentationLoader lazyInstrumentationLoader)
         {
-            return builder.AddHttpClientInstrumentation();
+            new HttpClientMetricsInitializer(lazyInstrumentationLoader);
```
pjanotti, line 102: Linking back to issue #1638

test/IntegrationTests/ModuleTests.cs
```diff
@@ -47,13 +50,34 @@ public async Task Default()
...
+
+        SetEnvironmentVariable(ConfigurationKeys.Traces.Exporter, Constants.ConfigurationValues.None);
+        SetEnvironmentVariable(ConfigurationKeys.Metrics.Exporter, Constants.ConfigurationValues.None);
+        SetEnvironmentVariable(ConfigurationKeys.Logs.Exporter, Constants.ConfigurationValues.None);
+
+        string verifyTestName =
+#if NETFRAMEWORK
+        $"{nameof(ModuleTests)}.{nameof(DefaultNoExporters)}.NetFx";
+#else
+$"{nameof(ModuleTests)}.{nameof(DefaultNoExporters)}.NetCore";
```
pjanotti, line 67: Indentation

pjanotti (approved):
LGTM minus the indentation issue.

pellared (approved):

Issue #1596, opened 2022-11-09
Dependabot security advisory vs test applications


pellared:
What should we do if there is a CVE reported in test application?

For example https://github.com/dotnet/announcements/issues/239 was reported.

What should we do with this:

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/772997f739767eac7fe7c0e2c2f43d351b271052/test/test-applications/integrations/TestApplication.SqlClient/TestApplication.SqlClient.csproj#L8

All feedback is welcome.

##  Bump

Reasons:

1. Static analysis tools would not report issues.
1. We may say that we do not support non-secure libraries.

## Ignore

Reasons:

1. We should ensure that the instrumentation is working for as many library versions as possible
2. Not supporting an older version of a library may be seen as a "breaking?" change

## Hack

We could try to "hack" or configure dependabot to not report issues under "test-applications". However, the hack may not work for all tools and may stop working after some time.

Reasons:

1. Static analysis tools would not report issues.
3. We should ensure that the instrumentation is working for as many library versions as possible

pellared:
@RassK Where to put the warning text? Somewhere globally like [here](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/developing.md#integration-tests)? Or maybe some other/better place?

This was my preference as well. 

However, I remember that @nrcventura reported that some customers are scanning the repos before they are used in the wild. I am not sure if it is something that happens often and if it should care about it at the point of time.

pellared:
SIG meeting:

We decide to bump the dependencies to not worry the users that our repo introduces security vulnerabilities.

PR #1657, opened 2022-11-23
Bump Microsoft.Data.SqlClient to 1.1.4

pellared:
## Why

https://github.com/advisories/GHSA-8g2p-5pqh-5jmc


pellared:
Per https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1596#issuecomment-1309113124

Kielek (approved):

pellared:
> In theory when multi-version testing support is built, we should test these versions anyway.

But we will not run them by default to mitigate the possibility that some static analysis tools would claim that OTel .NET Auto-Instrumentation has vulnerable dependencies.

PR #1658, opened 2022-11-23
Downgrade StartupHook to netcoreapp3.1

### LINKED ISSUE #1644: .NET Core 3.1/.NET 5.0 app crashes with auto-instrumentation
Usage of auto-instrumentation with .NET Core 3.1/.NET 5.0 crashes the app on start-up with below exception.

```
Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly 'System.Runtime, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
File name: 'System.Runtime, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
   at System.Reflection.RuntimeAssembly.GetType(QCallAssembly assembly, String name, Boolean throwOnError, Boolean ignoreCase, ObjectHandleOnStack type, ObjectHandleOnStack keepAlive, ObjectHandleOnStack assemblyLoadContext)
   at System.Reflection.RuntimeAssembly.GetType(String name, Boolean throwOnError, Boolean ignoreCase)
   at System.Reflection.Assembly.GetType(String name, Boolean throwOnError)
   at System.StartupHookProvider.CallStartupHook(StartupHookNameOrPath startupHook)
   at System.StartupHookProvider.ProcessStartupHooks()
```

.NET Core 3.1 and .NET 5.0 has reached end of life, but using auto-instrumentation should not crash the process. Rather it should backoff the instrumentation,

Only option left for us is to target the `OpenTelemetry.AutoInstrumentation.StartupHook` project version as `netcoreapp3.1`. We could modify a folder in `tracer-home` as `netcore` or `net` instead of `netcoreapp3.1` or `net6.0`.

rajkumar-rangaraj:
## Why

Fixes #1644 

## What

* Downgraded target framework version for `OpenTelemetry.AutoInstrumentation.StartupHook` from `net6.0` to `netcoreapp3.1`
* Added a check in `StartupHook.cs` to log unsupported message and backoff the instrumentation. 

## Tests

Manually tested with `Examples.AspNetCoreMvc` project for target framework `netcoreapp3.1`, `net5.0`, `net6.0` and `net7.0`.

## Checklist

### DIFF
--- a/build/TargetFramework.cs
+++ b/build/TargetFramework.cs
@@ -10,6 +10,7 @@ using Nuke.Common.Tooling;
 public class TargetFramework : Enumeration
 {
     public static readonly TargetFramework NET462 = new() { Value = "net462" };
+    public static readonly TargetFramework NETCore3_1 = new() { Value = "netcoreapp3.1" };
     public static readonly TargetFramework NET6_0 = new() { Value = "net6.0" };
     public static readonly TargetFramework NET7_0 = new() { Value = "net7.0" };
 
--- a/src/OpenTelemetry.AutoInstrumentation.StartupHook/OpenTelemetry.AutoInstrumentation.StartupHook.csproj
+++ b/src/OpenTelemetry.AutoInstrumentation.StartupHook/OpenTelemetry.AutoInstrumentation.StartupHook.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/OpenTelemetry.AutoInstrumentation.StartupHook/OpenTelemetry.AutoInstrumentation.StartupHook.csproj
+++ b/src/OpenTelemetry.AutoInstrumentation.StartupHook/OpenTelemetry.AutoInstrumentation.StartupHook.csproj
@@ -1,6 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <!-- To avoid application crash with .Net Core 3.1 and .NET 5.0 apps 
+         this project should target netcoreapp3.1.
+         Reference: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1644 -->
     <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
   </PropertyGroup>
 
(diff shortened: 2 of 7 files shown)


src/OpenTelemetry.AutoInstrumentation.StartupHook/OpenTelemetry.AutoInstrumentation.StartupHook.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
```
pjanotti, line 4: Suggestion: add a comment similar to the one in Build.Steps.cs

build/TargetFramework.cs
```diff
@@ -10,6 +10,7 @@
 public class TargetFramework : Enumeration
 {
     public static readonly TargetFramework NET462 = new() { Value = "net462" };
+    public static readonly TargetFramework NETCore3_1 = new() { Value = "netcoreapp3.1" };
```
pjanotti, line 13: Double-checking: the fact that this targeted was added to the enumeration doesn't mean that we will build any other sources or tests for netcoreapp3.1, right?

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
```diff
@@ -32,6 +33,17 @@ internal class StartupHook
     /// </summary>
     public static void Initialize()
     {
+        var minSupportedFramework = new FrameworkName(".NETCoreApp,Version=v6.0");
```
Kielek, line 36: I think that we should do similiar check on the native code side.

It could be separate PR.

Kielek (approved):

pjanotti (commented):
@rajkumar-rangaraj this covers the scenario for StartupHook, however, if the CLR profiler is enabled it will assume that the startup hook was initialized. I *think* that the CLR profiler will attempt to inject bytecode instrumentation which may crash the application (lack of loaded dependencies). We probably need to check for the same condition on `CorProfiler::Initialize`.

build/TargetFramework.cs
rajkumar-rangaraj: Correct. It is not used with any other source or tests.

Kielek (approved):

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
```diff
@@ -32,6 +33,22 @@ internal class StartupHook
...
```
pellared, line 36: nit: This could be moved inside the `if` to avoid unnecessary allocation

pellared (approved):

PR #1661, opened 2022-11-24
Remove framework version from distribution structure

Kielek:
## Why

Towards #1644

## What

Remove framework version from distribution structure

## Tests

CI

## Checklist


pellared (approved):

RassK (approved):
Hopefully we don't need second .net core distro.

Kielek:
> Hopefully we don't need second .net core distro.

If we will need it, I think that we should create some other structure.
.net Loader and StartupHook should pobably stay here :)

Issue #1388, opened 2022-10-10
Recreate /examples


pellared:
I think we should change the `examples` to just contain a client and server app, Dockerfiles, and a docker-compose to run it.

Previously we used it for testing purposes. Right now I think we should make it more .NET devs (users) friendly.

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1383#pullrequestreview-1135727555_

pjanotti:
Right now there are two `instrument.sh` scripts on the repo. While refactoring the examples we should look into keeping a single one, see https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1681#issuecomment-1329908775

pellared:
TODO: Bring back the `AspNetCoreMvc` example.

PR #1681, opened 2022-11-28
Add eval at the end of the instrument script

dszmigielski:
## Why

Instrument script only sets env variables, but does not execute instrumented app.

## What

Add eval at the end of the script

### DIFF
--- a/instrument.sh
+++ b/instrument.sh
@@ -142,3 +142,5 @@ if [ "$ENABLE_PROFILING" = "true" ]; then
   # Configure the bytecode instrumentation configuration file
   export OTEL_DOTNET_AUTO_INTEGRATIONS_FILE="$OTEL_DOTNET_AUTO_HOME/integrations.json"
 fi
+
+eval "$@"


dszmigielski:
> Can you describe the use case? Currently the docs instructs to use it to export the needed env vars. Will `. instrument.sh` still work?

It seems I was using that in a way it was not supposed to be used. Should I just close it or can it be considered a good change if it doesn't break the previous behaviour (`. instrument.sh`)?

pellared:
I prefer closing it. I think "eval" should be avoided. People can always create "subshells" e.g.

```bash
dotnet publish -o out
(
  source . $HOME/.otel-dotnet-auto/instrument.sh
  export OTEL_SERVICE_NAME=myapp
  export OTEL_RESOURCE_ATTRIBUTES=deployment.environment=staging,service.version=1.0.0
  ./out/myapp
)

pjanotti (approved):

pjanotti:
We also have a script with the same name under `dev/`. The `run-example.sh` script uses the one under `dev/` perhaps something to review when doing #1388

PR #1682, opened 2022-11-29
Update design doc: focus on load mechanism

### LINKED ISSUE #1313: Update design.md with new mechanisms using DOTNET_* env vars
https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/design.md#architecture should be updated after we introduced the usage of `DOTNET_STARTUP_HOOKS`, `DOTNET_ADDITIONAL_DEPS`, `DOTNET_SHARED_STORE`.

One of the most important goals of `design.md` is to tell the users what mechanisms are used to achieve automatic instrumentation.

pjanotti:
## Why

The design docs are slightly out-of-date and don't mention the host startup hook used on .NET 6.0 and above.

Fixes #1313 

## What

Updated the design doc with a focus on the host startup hook. In practice, I found that focusing on the env vars got in the way of describing the implementation at a high level. The env vars should be covered in the configuration section as appropriate.

## Tests

n/a

## Checklist


docs/design.md
```diff
@@ -31,11 +31,13 @@ and its initialization can also be delegated to the application code.
 ### Unsupported scenarios
 
 - **Applications using Ahead-of-Time (AOT) compilation**:
-The current implementation relies on the [CLR Profiler APIs](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/)
-and doesn't support AOT.
+The current implementation relies on the [host startup hook](https://github.com/dotnet/runtime/blob/main/docs/design/features/host-startup-hook.md)
+or the [CLR Profiler APIs](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling)
+and neither is support when publishing AOT compiled applications.
```
pellared, line 36:
```suggestion
and neither is supported when publishing AOT compiled applications.
```

docs/design.md
```diff
@@ -31,11 +31,13 @@ and its initialization can also be delegated to the application code.
 ### Unsupported scenarios
 
 - **Applications using Ahead-of-Time (AOT) compilation**:
-The current implementation relies on the [CLR Profiler APIs](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/)
-and doesn't support AOT.
+The current implementation relies on the [host startup hook](https://github.com/dotnet/runtime/blob/main/docs/design/features/host-startup-hook.md)
+or the [CLR Profiler APIs](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling)
```
pellared, line 35:
```suggestion
or the [CLR Profiler API](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling)
```

docs/design.md
```diff
@@ -44,32 +46,34 @@ are logged and crash the application.
...
 
-- [**Managed Profiler**](../src/OpenTelemetry.AutoInstrumentation):
-Contains the code to set up the OpenTelemetry .NET SDK and configured instrumentations,
-as well as support code to run and implement bytecode instrumentations. Set the
-`OTEL_DOTNET_AUTO_TRACES_ENABLED` environment variable to `false` when the
-application initializes the OpenTelemetry .NET SDK Tracer on its own.
+The **.NET Framework** doesn't support the host startup hook.
+For .NET Framework applications the OpenTelemetry .NET SDK is injected using the
+[CLR Profiler APIs](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/)
+which allow modifying application code during their execution.
```
pellared, line 72:
```suggestion
[CLR Profiler API](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/)
which allows modifying application code during their execution.
```

docs/design.md
```diff
@@ -82,59 +86,115 @@ library or framework that they instrument. Some examples include:
 
 - **Bytecode Instrumentations**: Instrumentations created for libraries
 or frameworks that lack proper hooks or callbacks to allow the collection
-of observability data. These instrumentations must be implemented following
+of observability data. These instrumentations are enabled by modifying
+the application [IL code](https://en.wikipedia.org/wiki/Common_Intermediate_Language)
+during runtime using the [CLR Profiler APIs](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/).
```
pellared, line 91:
```suggestion
during runtime using the [CLR Profiler API](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/).
```

docs/design.md
```diff
@@ -44,32 +46,34 @@ are logged and crash the application.
...
-Contains the code to set up the OpenTelemetry .NET SDK and configured instrumentations,
-as well as support code to run and implement bytecode instrumentations. Set the
-`OTEL_DOTNET_AUTO_TRACES_ENABLED` environment variable to `false` when the
-application initializes the OpenTelemetry .NET SDK Tracer on its own.
+The **.NET Framework** doesn't support the host startup hook.
+For .NET Framework applications the OpenTelemetry .NET SDK is injected using the
+[CLR Profiler APIs](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/)
+which allow modifying application code during their execution.
+
+### Adding and enabling instrumentations
```
pellared, line 74:
```suggestion
### Library instrumentations
```

docs/design.md
```diff
@@ -82,59 +86,115 @@ library or framework that they instrument. Some examples include:
...
+the following actions:
+
+- If this the first user module in the current AppDomain,
+the profiler injects the IL calling the Loader `Startup` constructor.
+- If the first method observed by JITCompilationStarted is IIS startup code,
+  the profiler invokes
+  `AppDomain.CurrentDomain.SetData("OpenTelemetry_IISPreInitStart", true)`,
+  so that automatic instrumentations can correctly handle IIS startup scenarios.
+
+### Enabling and injecting instrumentations
```
pellared, line 168:
```suggestion
### Injecting instrumentations
```

pellared (approved):
👍 just few things spotted

docs/design.md
theletterf: Might be worth spelling out AOT in full the first time.

docs/design.md
```diff
@@ -82,59 +86,115 @@ library or framework that they instrument. Some examples include:
...
+
+- Adds a handler to the
+[`AssemblyResolve` event](https://docs.microsoft.com/en-us/dotnet/api/system.appdomain.assemblyresolve?view=net-5.0),
+so that it can add any assembly needed by the SDK itself or by any instrumentation.
+- Runs, through reflection, the `Initialization` method from
+  the `OpenTelemetry.AutoInstrumentation.Instrumentation` type
+  from the Managed Profiler assembly.
+  - The `Initialization` code bootstraps the OpenTelemetry .NET SDK,
+    adding configured processors, exporters, and so on,
+    and setting the mechanisms to enable any configured source instrumentations.
```
theletterf, line 145:
```suggestion
1.  Adds a handler to the [`AssemblyResolve` event](https://docs.microsoft.com/en-us/dotnet/api/system.appdomain.assemblyresolve?view=net-5.0),
so that it can add any assembly needed by the SDK itself or by any instrumentation.
2. Runs, through reflection, the `Initialization` method from
  the `OpenTelemetry.AutoInstrumentation.Instrumentation` type
  from the Managed Profiler assembly.
  a. The `Initialization` code bootstraps the OpenTelemetry .NET SDK,
    adding configured processors, exporters, and so on,
    and setting the mechanisms to enable any configured source instrumentations.
```

docs/design.md
```diff
@@ -82,59 +86,115 @@ library or framework that they instrument. Some examples include:
...
+
+#### .NET bootstrapping
+
+.NET applications rely on the [host startup hook](https://github.com/dotnet/runtime/blob/main/docs/design/features/host-startup-hook.md)
+being configured to use the `StartupHook.Initialize()` method from
+the `OpenTelemetry.AutoInstrumentation.StartupHook` assembly.
+The `Initialize` method will load the `OpenTelemetry.AutoInstrumentation.Loader`
+assembly and will create the `OpenTelemetry.AutoInstrumentation.Loader.Startup` instance.
```
theletterf, line 153:
```suggestion
.NET applications rely on the [host startup hook](https://github.com/dotnet/runtime/blob/main/docs/design/features/host-startup-hook.md)
being configured to use the `StartupHook.Initialize()` method from
the `OpenTelemetry.AutoInstrumentation.StartupHook` assembly.
The `Initialize` method loads the `OpenTelemetry.AutoInstrumentation.Loader`
assembly and creates the `OpenTelemetry.AutoInstrumentation.Loader.Startup` instance.
```

docs/design.md
```diff
@@ -82,59 +86,115 @@ library or framework that they instrument. Some examples include:
...
+.NET Framework bootstrapping is performed on the [CorProfiler::JITCompilationStarted](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/icorprofilercallback-jitcompilationstarted-method)
+callback. When that callback happens the CLR Profiler DLL takes
+the following actions:
+
+- If this the first user module in the current AppDomain,
+the profiler injects the IL calling the Loader `Startup` constructor.
+- If the first method observed by JITCompilationStarted is IIS startup code,
+  the profiler invokes
+  `AppDomain.CurrentDomain.SetData("OpenTelemetry_IISPreInitStart", true)`,
+  so that automatic instrumentations can correctly handle IIS startup scenarios.
```
theletterf, line 166:
```suggestion
1. If the module is the first user module in the current AppDomain,
the profiler injects the IL to call the Loader `Startup` constructor.
- If the first method observed by `JITCompilationStarted` is IIS startup code,
  the profiler invokes
  `AppDomain.CurrentDomain.SetData("OpenTelemetry_IISPreInitStart", true)`,
  so that automatic instrumentations can correctly handle IIS startup scenarios.
```

(I'm guessing it's a sequence.)

docs/design.md
```diff
@@ -82,59 +86,115 @@ library or framework that they instrument. Some examples include:
...
+### Enabling and injecting instrumentations
+
+#### Source instrumentation
 
-The OpenTelemetry .NET SDK and selected source instrumentations are injected
-into the target process through a series of steps started by the CLR Profiler DLL:
+Source instrumentations are injected into the target process by adding
+handlers to the [`AssemblyLoad` event](https://learn.microsoft.com/en-us/dotnet/api/system.appdomain.assemblyload?view=net-7.0)
+and when the targeted assembly is loaded triggering the source instrumentation
+initialization code.
```
theletterf, line 175:
```suggestion
Source instrumentations are injected into the target process by adding
handlers to the [`AssemblyLoad` event](https://learn.microsoft.com/en-us/dotnet/api/system.appdomain.assemblyload?view=net-7.0)
and when the targeted assembly loads and triggers the source instrumentation
initialization code.
```

docs/design.md
```diff
@@ -82,59 +86,115 @@ library or framework that they instrument. Some examples include:
...
+
+#### Bytecode instrumentation
+
+The bytecode instrumentations, called "call target" in this repo, relies on
+the JIT recompilation capability of the CLR to rewrite the IL for instrumented
+methods. This adds logic at the beginning and end of the instrumented methods
+to invoke instrumentation written in this repo, and wraps the calls with
+try-catch blocks to prevent instrumentation errors from affecting the normal operation
+of the application. This IL rewrite happens in the following steps:
```
theletterf, line 184:
```suggestion
Bytecode instrumentations, also defined as "call target" in this repo, rely on
the JIT recompilation capability of the CLR to rewrite the IL for instrumented
methods. This adds logic at the beginning and end of the instrumented methods
to invoke instrumentation included in this project, and wraps the calls with
try-catch blocks to prevent instrumentation errors from affecting the normal operation
of the application. This IL code rewrite happens in the following steps:
```

Any reason for using two names? Couldn't we just use "bytecode instrumentation" or something instead of "call target"?

theletterf (changes requested):

theletterf:
@pjanotti @pellared I suspect we're using "Zero-touch" here to avoid the term "Automatic". Where does it come from?

I'm not entirely sure I like "Zero-touch", as it's hyphenated and ambiguous. Plus, it clashes with existing Splunk product naming, such as "Zero-config". We can think of alternatives though—I'd be happy to help with that. We could then spread the usage of the new term to other instrumentations.

pellared:
> @pjanotti @pellared I suspect we're using "Zero-touch" here to avoid the term "Automatic". Where does it come from?
> 
> I'm not entirely sure I like "Zero-touch", as it's hyphenated and ambiguous. Plus, it clashes with existing Splunk product naming, such as "Zero-config". We can think of alternatives though—I'd be happy to help with that. We could then spread the usage of the new term to other instrumentations.

I think we can should replace "zero-touch" with "automatic",

pjanotti:
@pellared @theletterf "zero-touch" comes from an early version of this OTep https://github.com/open-telemetry/oteps/blob/main/text/0001-telemetry-without-manual-instrumentation.md

It seems that the OTep now uses "zero source code modification", but, I think that "automatic" will read better.

theletterf:
Thanks! Let's go with automatic then. :) I'm adding the suggestions.

docs/design.md
pjanotti: @theletterf a few lines above AOT is spelled out, do you think is still desirable to spell it out again here?

docs/design.md
pjanotti: This is more for historical reasons: there were different forms of bytecode instrumentation in the past, one was "call site" the other "call target". You are right: it is time to remove "call target" from this type of document.

docs/design.md
```diff
@@ -44,32 +46,34 @@ are logged and crash the application.
...
+The **.NET Framework** doesn't support the host startup hook.
+For .NET Framework applications the OpenTelemetry .NET SDK is injected using the
+[CLR Profiler APIs](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/)
+which allow modifying application code during their execution.
+
+### Adding and enabling instrumentations
+
+There are two broad instrumentations types injected into the applications:
 
 - **Source Instrumentations**: Instrumentations created on top of API hooks
```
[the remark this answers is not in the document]
pjanotti: 935dd0418 - searched other case between ** just found 2, except for proper names like "CLR Profiler" and ".NET Framework"

docs/design.md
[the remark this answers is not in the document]
pjanotti: ab6187388

docs/design.md
pjanotti: a1eee3830

pjanotti:
Removed "zero-touch" and used "automatic" instead in 65d2588c9

theletterf (approved):

Kielek (approved):

Issue #1683, opened 2022-11-29
Examples.AspNetCoreMvc app crashes with all profiles in launchSettings.json


rajkumar-rangaraj:
# Bug Report

## Symptom

**Describe the bug**
Running `Examples.AspNetCoreMvc` from Visual Studio 2022 crashes the process with below stack.

```
Unhandled exception. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.TypeInitializationException: The type initializer for 'OpenTelemetry.AutoInstrumentation.Loader.Startup' threw an exception.
 ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.MissingMethodException: Method not found: 'OpenTelemetry.Metrics.MeterProviderBuilder OpenTelemetry.Metrics.MeterProviderBuilderExtensions.ConfigureResource(OpenTelemetry.Metrics.MeterProviderBuilder, System.Action`1<OpenTelemetry.Resources.ResourceBuilder>)'.
   at OpenTelemetry.AutoInstrumentation.Instrumentation.Initialize()
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
   --- End of inner exception stack trace ---
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters)
   at OpenTelemetry.AutoInstrumentation.Loader.Startup.TryLoadManagedAssembly() in D:\opentelemetry-dotnet-instrumentation\src\OpenTelemetry.AutoInstrumentation.Loader\Startup.cs:line 74
   at OpenTelemetry.AutoInstrumentation.Loader.Startup..cctor() in D:\opentelemetry-dotnet-instrumentation\src\OpenTelemetry.AutoInstrumentation.Loader\Startup.cs:line 47
   --- End of inner exception stack trace ---
[...]

PR #1692, opened 2022-11-29
Validate OpenTelemetry SDK version in StartupHook

rajkumar-rangaraj:
## Why

Fixes: Follow up #1683 

If a customer has a package reference to an older version of OpenTelemetry .NET SDK than the auto-instrumentation's OpenTelemetry SDK it crashes the process. The call stack or an exception from the crash does not explain the root cause and it is difficult for users to identify such an issue is related to the OpenTelemetry package reference.

## What

This PR adds a check in StartupHook to validate if OpenTelemetry SDK referenced is the same or higher version than auto-instrumentation once. If the condition fails, startuphook throws System.NotSupportedException as below.

```
Unhandled exception. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.NotSupportedException: Application has direct or indirect reference to older version of OpenTelemetry package.
   at StartupHook.EvaluateOpenTelemetryReference(String loaderAssemblyLocation) in D:\opentelemetry-dotnet-instrumentation\src\OpenTelemetry.AutoInstrumentation.StartupHook\StartupHook.cs:line 204
   at StartupHook.Initialize() in D:\opentelemetry-dotnet-instrumentation\src\OpenTelemetry.AutoInstrumentation.StartupHook\StartupHook.cs:line 71
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
   --- End of inner exception stack trace ---
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.StartupHookProvider.CallStartupHook(StartupHookNameOrPath startupHook)
   at System.StartupHookProvider.ProcessStartupHooks()  
```

We might need to extend and have a generic solution for other OpenTelemetry Instrumentation package and System.Diagnostics.DiagnosticSource. 

## Tests

Manually tested Example project for these 3 conditions
1. No reference to OpenTelemetry package
2. Lower version of OpenTelemetry package than auto-instrumentation
3. Same version of OpenTelemetry package as auto-instrumentation

## Checklist


src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
```diff
@@ -67,7 +68,7 @@ public static void Initialize()
             // Check Instrumentation is already initialized with native profiler.
             Type profilerType = Type.GetType("OpenTelemetry.AutoInstrumentation.Instrumentation, OpenTelemetry.AutoInstrumentation");
 
-            if (profilerType == null)
+            if (profilerType == null && EvaluateOpenTelemetryReference(loaderAssemblyLocation))
```
pjanotti, line 71: `EvaluateOpenTelemetryReference` returns a boolean it gives the reader the impression that we can hit the else of the if statement due to it. However, it only returns true throwing and exception otherwise. It seem better to change its return type to `void` and move it outside the if expression. The name should reflect that it throws in case of error, my suggestion will be something like `ThrowIfReferenceIncorrectOpenTelemetryVersion`

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
```diff
@@ -164,4 +165,45 @@ private static string GetEnvironmentVariable(string variableName)
...
+
+        if (index != -1)
+        {
+            try
+            {
+                // Look up for type with an assembly name, will load the library.
+                // OpenTelemetry assembly load happens only if app has reference to the package.
+                var openTelemetryType = Type.GetType("OpenTelemetry.Sdk, OpenTelemetry");
+                var loadedOTelAssembly = Assembly.GetAssembly(openTelemetryType);
+                var loadedOTelFileVersionInfo = FileVersionInfo.GetVersionInfo(loadedOTelAssembly.Location);
```
pjanotti, line 183: Is it really necessary to trigger the load of the SDK at this point? I mean, the `AssemblyName` already has the version information. Is this to try to avoid a case in which the entry assembly references one version but it actually (somehow) ends up loading another?

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
```diff
@@ -164,4 +165,45 @@ private static string GetEnvironmentVariable(string variableName)
             return null;
         }
     }
+
+    private static bool EvaluateOpenTelemetryReference(string loaderAssemblyLocation)
+    {
+        var appReferenceAssemblies = Assembly.GetEntryAssembly().GetReferencedAssemblies();
+        var index = Array.FindIndex(appReferenceAssemblies, assembly => assembly.Name == "OpenTelemetry");
```
pjanotti, line 172: What would happen in the case of references to old versions of `OpenTelemetry.Api`? It seems a possible case: some library is instrumented with OTel but it doesn't reference the SDK.

pjanotti (commented):
One suggestion and 2 Qs. We don't need to address everything at once: having a clear message is better than limping or throwing hard-to-explain exceptions. This is already an improvement.

pjanotti:
> We might need to extend and have a generic solution for other OpenTelemetry Instrumentation package and System.Diagnostics.DiagnosticSource.

Ideally we should even automate the detection of all the dependencies. I think that we can do this at build time.

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
rajkumar-rangaraj: OpenTelemetry.dll will have an indirect reference to `OpenTelemetry.Api`. If we check for OpenTelemetry it should take care of `OpenTelemetry.Api`.

As called out PR description, later we might need to extend and have a generic solution for other OpenTelemetry Instrumentation package and System.Diagnostics.DiagnosticSource for the same issue.

pellared:
> We might need to extend and have a generic solution for other OpenTelemetry Instrumentation package and System.Diagnostics.DiagnosticSource.

That the "might" word could be removed here. Otherwise this solution only works for one concrete library. Am I correct? I imagine that the most popular problem would be using a different version of `System.Diagnostics.DiagnosticSource`.

Kielek:
> I imagine that the most popular problem would be using a different version of System.Diagnostics.DiagnosticSource.

And `System.Memory`...

rajkumar-rangaraj:
> > We might need to extend and have a generic solution for other OpenTelemetry Instrumentation package and System.Diagnostics.DiagnosticSource.
> 
> I think that the "might" word could be removed from the quoted sentence. Otherwise this would work only for one concrete library. Am I correct? I imagine that the most popular problem would be using a different version of `System.Diagnostics.DiagnosticSource`.

You're correct. Most problems would be from different version of `System.Diagnostics.DiagnosticSource`.

pjanotti (approved):

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
```diff
@@ -164,4 +167,51 @@ private static List<string> GetExcludedApplicationNames()
...
+            catch (Exception ex)
+            {
+                // Exception in evaluation should not throw or crash the process.
+                StartupHookEventSource.Log.Trace($"Couldn't evaluate reference to OpenTelemetry Sdk in an app. Exception: {ex}");
+            }
+        }
+
+        if (throwException)
+        {
+            throw new NotSupportedException("Application has direct or indirect reference to older version of OpenTelemetry package.");
```
Kielek, line 214: Is it possible to put loaded and expected versions here?

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
```diff
@@ -164,4 +167,51 @@ private static List<string> GetExcludedApplicationNames()
...
+                // Look up for type with an assembly name, will load the library.
+                // OpenTelemetry assembly load happens only if app has reference to the package.
+                var openTelemetryType = Type.GetType("OpenTelemetry.Sdk, OpenTelemetry");
+                if (openTelemetryType != null)
+                {
+                    var loadedOTelAssembly = Assembly.GetAssembly(openTelemetryType);
+                    var loadedOTelFileVersionInfo = FileVersionInfo.GetVersionInfo(loadedOTelAssembly?.Location);
+                    var loadedOTelfileVersion = new Version(loadedOTelFileVersionInfo.FileVersion);
+
+                    var profilerOTelLocation = Path.Combine(loaderAssemblyLocation, "OpenTelemetry.dll");
```
Kielek, line 195: I have some doubts about _profiler_.
Maybe _expecped_ or _autoInstrumentation_?

Kielek (approved):

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
rajkumar-rangaraj: Fixed 4b32407cc80fc3d4386c35d397d9497bb3c6e64b

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
rajkumar-rangaraj: 80f5aa68

RassK (approved):

Issue #1678, opened 2022-11-28
Enable Nullable


Kielek:
# Feature Request

Enable `<Nullable>enable</Nullable>` in whole repository.

PR #1710, opened 2022-12-05
Enable Nullable - part 1

Kielek:
## Why

Towards #1678

## What

Enable Nullable by default.
Disable other places.
Nullable for few projects.

## Tests

CI

## Checklist


src/OpenTelemetry.AutoInstrumentation.Loader/Startup.NetCore.cs
```diff
@@ -61,7 +61,7 @@ private static Assembly AssemblyResolve_ManagedProfilerDependencies(object sende
         // 2) The AssemblyVersion is lower than the version used by OpenTelemetry.AutoInstrumentation, the assembly will fail to load
         //    and invoke this resolve event. It must be loaded in a separate AssemblyLoadContext since the application will only
         //    load the originally referenced version
-        if (assemblyName.Name.StartsWith("OpenTelemetry.AutoInstrumentation", StringComparison.OrdinalIgnoreCase) && File.Exists(path))
+        if (assemblyName.Name != null && assemblyName.Name.StartsWith("OpenTelemetry.AutoInstrumentation", StringComparison.OrdinalIgnoreCase) && File.Exists(path))
```
pjanotti, line 64: why not the `?.` operator?

pjanotti (approved):

src/OpenTelemetry.AutoInstrumentation.Loader/Startup.NetCore.cs
Kielek: StartsWith returns `bool`, ?.StartWith returns lets say `bool?` which is not valid for the condition.

PR #1724, opened 2022-12-07
Fix configuration sources

RassK:
## Why

> The SDK MUST interpret an empty value of an environment variable the same way as when the variable is unset.

_src: https://opentelemetry.io/docs/reference/specification/sdk-environment-variables/_

## What

Ensures that all configuration sources follow that rule.

## Tests

Unit tests added.

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Configuration/CompositeConfigurationSource.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Configuration/CompositeConfigurationSource.cs
@@ -61,7 +61,7 @@ internal class CompositeConfigurationSource : IConfigurationSource, IEnumerable<
     public string GetString(string key)
     {
         return _sources.Select(source => source.GetString(key))
-            .FirstOrDefault(value => value != null);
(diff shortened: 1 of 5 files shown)


src/OpenTelemetry.AutoInstrumentation/Configuration/CompositeConfigurationSource.cs
```diff
@@ -61,7 +61,7 @@ public void Insert(int index, IConfigurationSource item)
     public string GetString(string key)
     {
         return _sources.Select(source => source.GetString(key))
-            .FirstOrDefault(value => value != null);
+             .FirstOrDefault(value => !string.IsNullOrEmpty(value));
```
pellared, line 64: It could be enough to implement it only here if we refactor it. I can try doing it later.

Basically the idea is to change `CompositeConfigurationSource` to e.g. `Configuration`. It will not implement `IConfigurationSource` and existing code would depend on `Configuration` instead of `IConfigurationSource`. Then we will be sure that the "sanitization" is always applied.

pellared (approved):

zacharycmontoya (approved):

PR #1729, opened 2022-12-07
[Instrumentation.GraphQL] Span kind based on the parent span context.

### LINKED ISSUE #1396: GraphQL instrumentation: Set proper span kind depending on the parent
I still think that having 2 server spans in a single request is wrong. The spec indicates that a server span should have a remote parent. I think that it would make more sense, in this embedded case (where graphql relies on aspnetcore) for aspnetcore to generate the server span, and graphql to generate an internal span. If graphql was not relying on aspnetcore, I would think that the span generated by graphql instrumentation should be a server span.

By @nrcventura 

Slack thread: https://cloud-native.slack.com/archives/C01NR1YLSE7/p1665172883242279

Kielek:
## Why

Fixes #1396

## What

Set span kind based on the parent span context.

## Tests

CI + newly added.

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Instrumentations/GraphQL/GraphQLCommon.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Instrumentations/GraphQL/GraphQLCommon.cs
@@ -58,7 +58,10 @@ internal class GraphQLCommon
                 tags.Document = query;
             }
 
-            activity = ActivitySource.StartActivityWithTags(operation, ActivityKind.Server, tags);
+            var activityCurrent = Activity.Current;
+            var activityKind = activityCurrent != null && activityCurrent.Recorded ? ActivityKind.Internal : ActivityKind.Server;
+
+            activity = ActivitySource.StartActivityWithTags(operation, activityKind, tags);
         }
         catch (Exception ex)
         {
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry.AutoInstrumentation/Instrumentations/GraphQL/GraphQLCommon.cs
```diff
@@ -58,7 +58,10 @@ internal static Activity CreateActivityFromExecuteAsync(IExecutionContext execut
                 tags.Document = query;
             }
 
-            activity = ActivitySource.StartActivityWithTags(operation, ActivityKind.Server, tags);
+            var activityCurrent = Activity.Current;
+            var activityKind = activityCurrent != null && activityCurrent.Recorded ? ActivityKind.Internal : ActivityKind.Server;
```
pellared, line 62: Shouldn't we also check if `activityCurrent.Kind` is `ActivityKind.Server` or `ActivityKind.Internal`? This covers only if the ASP.NET spans are the direct parents. What if there is some other instrumentation as parent? E.g. some in-proc HttpClient calling GraphQL with ASP.NET instrumentation disabled.

Kielek: Ref.: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spankind

I will check it. Especially scenarios, were we have a remote parent.

pjanotti (approved):

nrcventura (approved):

Issue #1824, opened 2022-12-15
Bump transitive dependencies in "netfx" to higher version


pjanotti:
As explained in this [comment](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1646#issuecomment-1349083888) with the implementation of #1646 there are good reasons to bump up the transitive dependencies of `OpenTelemetry.AutoInstrumentation` to their latest version. Unfortunately, there are no straightforward automated ways of doing that, this issue is to automate this transitive version bump as part of the project build.

Issue #1823, opened 2022-12-15
Friendly way to install "netfx" assemblies into the GAC


pjanotti:
Since we are implementing assembly redirection at runtime for the .NET Framework applications, see #1646. We need a friendly way to install the assemblies shipped under the `netfx` folder into the GAC. This is necessary in the case the application and auto instrumentation ship different versions of the same assembly.

pjanotti:
This will also help when instrumenting domain-neutral assemblies.

Issue #1727, opened 2022-12-07
AspNetCore instrumentation under .NET Framework 4.6.2


Kielek:
# Feature Request

**Are you requesting automatic instrumentation for a framework or library? Please describe.**

Support AspNetCore instrumentation under .NET Framework 4.6.2.

**Additional context**
We have lazy loading implemented, so it should works.
Manual tests while working on GraphQL passed.

Notes: Consider to separate configuration value for AspNet and AspNetCore.

pjanotti:
SIG: evaluate if we should really support this, related comment on #1646

RassK:
Such extreme cases could be left for manual instrumentation (use SDK directly), then there is much higher chance it will work.

Kielek:
Discussed with @pellared, I agree that it is to extreme. Closing.

Issue #1482, opened 2022-10-21
Refactor integration loader - native code


Kielek:
This is getting very nasty. I think it is time to refactor and introduce a `Integrations` class so that we can change the signatures to something like:

```
Integrations LoadIntegrationsFrom*(inputs...)
```

or even instead of functions create static methods in `Integrations` class so that the are named e.g. `FromJson` or `LoadJson`

We can do it in a separate PR to make it easier for reviewing.

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1475#discussion_r1001553743_

pjanotti:
As pointed in #1825 this work should revisit the skip assemblies code. We just need to keep module data for the assemblies that are going to be instrumented.

Kielek:
When #2280 will be merged there will be no loading instrumention list from json file. Closing.

PR #1825, opened 2022-12-15
Automatic netfx assembly redirection

### LINKED ISSUE #1646: Missing expected assembly when instrumenting .NET Framework apps
The automatic instrumentation injects, into the application, assemblies from multiple packages from OTel (the SDK, API, exporters, instrumentations, etc). These different packages can depend on different versions of the same assemblies. When the "runtime folders", e.g.: `tracer-home\net462`, are built the NuGet version resolution will select a single version of any conflicting assembly.

On .NET Framework this behavior becomes a problem because of the strict versioning that it implements on top of strong naming. A concrete example comes from #1647. In this PR the OTLP exporter has a dependency on `Google.Protobuf` which has a dependency on `System.Memory, Version=4.0.1.1`. However, other referenced packages depend on `System.Memory, Version=4.0.1.2` so after the build only version 4.0.1.2 ends up on `tracer-home\net462`. At runtime, when automatic instrumentation injects the assemblies, code from `Google.Protobuf` fails because it tries to load version 4.0.1.1 which is not present.

This is not an issue for manually instrumented applications because when building .NET Framework apps, by default, a binding redirect, mapping all versions to a single one, is added to the application configuration file. However, automatic instrumentation builds as a class library and there are no binding redirects for class libraries.

The way to fix this without requiring the .NET Framework applications to add a reference to the auto-instrumentation NuGet package is to do the "redirect" directly on the assemblies at the Framework "runtime folder". In order to do that it is necessary to update the assembly references in metadata tokens to the single version made available by the NuGet package resolution. This should be done during the build using assembly rewriting packages like `Mono.Cecil`. The sample listed below shows how the issue on PR #1647 can be fixed with this "redirect".

```c#
using Mono.Cecil;

const string targetModuleFile = @"C:\src\otel\dotnet\auto\bin\tracer-home\net462\Google.Protobuf.dll";

using var targetModuleDef = ModuleDefinition.ReadModule(targetModuleFile, new ReaderParameters { ReadWrite = true });
foreach (var assemblyRef in targetModuleDef.AssemblyReferences)
{
    Console.WriteLine(assemblyRef);
}

// Hard-coded the location of the assembly reference that needs to be updated.
targetModuleDef.AssemblyReferences[1].Version = new Version(4, 0, 1, 2);

targetModuleDef.Write();
Console.WriteLine("TargetModuleDef.Write()");
```

pjanotti:
## Why

To automate instrumentation of .NET Framework without requiring binding redirection and reducing the cases that adding dependencies to the projects are needed.

Fixes #1646

## What

1. During the build, build a list of the assemblies shipped in the `netfx` folder and use it to generate a map of redirections to be used by the native profiler in `_WIN32` compilation.
2. When loading a module the CLR profiler updates any assembly present in the list to use the higher version: either from the application reference of from the one available on `netfx`.

The risk is that dependencies are changed at runtime and it is possible that issues only show up when certain code paths are executed. A much safer path would be to have the application reference a NuGet package with all the dependencies required by automatic instrumentation.

To minimize the risk of downgrading packages used by the application, details [here](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1646#issuecomment-1349083888), we need the work from #1824 and #1823.

## Tests

- Removed App.Config and others that were performing binding redirects or adding packages to the test applications.
- The exception is the GraphQL test application, see #1727 and https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1646#issuecomment-1347420386
- GrpcNetClient has an explicit reference to Grpc and Protobuf packages.

## Checklist

- [X] New features are covered by tests: by removal of the previous workarounds.


CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ This beta release is built on top of [OpenTelemetry .NET](https://github.com/ope
  `ConfigureTracesOptions(StackExchangeRedisCallsInstrumentationOptions options)`.
 - Add plugin support for
  `ConfigureMetricsOptions(AspNetCoreMetricsInstrumentationOptions options)`.
+- Add automatic assembly redirection for .NET Framework applications, see
+ `OTEL_DOTNET_AUTO_NETFX_ASSEMBLY_REDIRECTION_ENABLED`.
```
Kielek, line 23: If you are using, _see_. It will be great to have link to documentation. You can change it to something like

_use_. or _It can be configured by_

src/OpenTelemetry.AutoInstrumentation.Native/netfx_assembly_redirection.h
```diff
@@ -0,0 +1,153 @@
+﻿// Auto-generated file, do not change it - generated by the AssemblyRedirectionSourceGenerator type
+
+#include "cor_profiler.h"
+
+#ifdef _WIN32
+namespace trace
+{
+void CorProfiler::InitNetFxAssemblyRedirectsMap()
+{
+    assembly_version_redirect_map_[WSTRING(L"Google.Protobuf")] = AssemblyVersionRedirection(3, 19, 4, 0);
```
Kielek, line 10: I do not see any test to check if the committed version is exactly the same as generated during build.

BTW this file is ugly.

Kielek (commented):
Briefly checked. I will revisit it on Monday.

I left only some nit comments.

build/Build.cs
```diff
@@ -159,4 +160,9 @@ string AndFilter(params string[] args)
 
         return result;
     }
+
+    string MapToFolderOutput(TargetFramework targetFramework)
```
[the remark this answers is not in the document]
pjanotti: It needs to be moved from its current location, but, keeping it on `.Steps` seems reasonable.

src/OpenTelemetry.AutoInstrumentation.Native/netfx_assembly_redirection.h
pjanotti: Changes on the sources are covered via changes in ci, see https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1825/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR85-R90

> BTW this file is ugly

I'm open to suggestions :) what really bothers you?

pjanotti: @Kielek perhaps adding a helper function could make the file easier on the eye, something like

```c++
   HelperFunctionYetToBeNamed(L"System.Memory", 4, 0, 1, 2);
```

pjanotti:
I'm having second thoughts about having a static list of versions: we could look at startup for the assemblies shipped with the app to update the version of the redirection. Anyway, I would prefer to address that in a separate PR.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -500,24 +645,6 @@ HRESULT STDMETHODCALLTYPE CorProfiler::ModuleLoadFinished(ModuleID module_id, HR
         return S_OK;
     }
 
-    for (auto&& skip_assembly : skip_assemblies)
```
Kielek, line 503: Worth to check if it was not written for the performance purposes.
I suppose it significantly reduces time for well known libraries which should be not instrumented.

Kielek (approved):
LGTM, on comment to review.

Kielek:
> I'm having second thoughts about having a static list of versions: we could look at startup for the assemblies shipped with the app to update the version of the redirection. Anyway, I would prefer to address that in a separate PR.

I suppose that it will increase cold start time. Additional measurements are needed.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
pjanotti: Let me double-check this.

pjanotti: We can't skip redirection for some assemblies in that set, so bailing early on them won't work. That said we don't need to process them in cases when redirection is disabled. We should address the set of skips as an optimization when we [refactor integrations loader](#1482) via #1482. In light of that, I opted to restore the skips for when the redirection is disabled.

Kielek (approved):

RassK (approved):

docs/config.md
```diff
@@ -237,17 +237,18 @@ Important environment variables include:
...
+| `OTEL_DOTNET_AUTO_METRICS_ENABLED`                    | Enables metrics.                                                                                                                                                                                                                                                                                                                                                                                   | `true`        |
+| `OTEL_DOTNET_AUTO_NETFX_ASSEMBLY_REDIRECTION_ENABLED` | Enables automatic redirection of the assemblies used by the automatic instrumentation on the .NET Framework.                                                                                                                                                                                                                                                                                       | `true`        |
```
theletterf, line 246: This is correct, although the var name is sooo long. :) Could it be something like:

`OTEL_DOTNET_AUTO_NETFX_ASSEMBLY_REDIRECT` ?

My only question is: Why and when would a user need to disable this? We should add this somewhere.

theletterf (changes requested):

docs/config.md
pjanotti: The pattern for boolean env vars is to have the `_ENABLED` suffix. That said I do agree that this is really long... 🤔 perhaps `OTEL_DOTNET_AUTO_NETFX_REDIRECT_ENABLED`? Still on the long side but a bit shorter...

> Why and when would a user need to disable this?

There are some cases that this feature can introduce some runtime errors due to incompatible versions of assemblies. The idea to disable it is mostly to help diagnose problems caused by the feature itself. In such cases, it is possible to suggest workarounds that require the feature to be disabled.

theletterf: If it's not immediately evident to mid-level .NET devs, I'd add it here on in the Troubleshooting docs. Or both.

pjanotti: The troubleshooting docs are in need a more careful review. Opened #1863

PR #1869, opened 2022-12-28
Add ARM64 build

dszmigielski:
## Why

Fixes #

## Checklist


.github/workflows/ci.yml
```diff
@@ -89,6 +89,39 @@ jobs:
           git diff
           [[ -z "$(git status --porcelain)" ]]
 
+  build-arm64:
+    strategy:
+      fail-fast: false
+    runs-on: ubuntu-20.04
```
RassK, line 95: There are no arm machines on the list, all available machines are x64 based. 
I think GH only supports arm on self hosted runners.

dszmigielski: I was trying to work around that with uraimo/run-on-arch-action@v2.5.0, but I cannot get it working, so I assume we have to wait for gh actions to add support.

RassK: Ah I see, it's using QEMU emulator. Might be easier to test that locally first.

.github/workflows/ci.yml
```diff
@@ -89,6 +89,39 @@ jobs:
...
+        run: |
+          touch /etc/apt/sources.list.d/backports.list
+          echo "deb http://deb.debian.org/debian stretch-backports main" >> /etc/apt/sources.list.d/backports.list
+          apt-get update
+          apt-get upgrade
+          apt-get install -y curl clang cmake make protobuf-compiler
+          apt-get install -y -t stretch-backports libgrpc++-dev libgrpc++1 libgrpc6 libgrpc-dev protobuf-compiler-grpc 
+          curl -sSL https://dot.net/v1/dotnet-install.sh --output dotnet-install.sh
+          chmod +x ./dotnet-install.sh
+          ./dotnet-install.sh -c 6.0 --install-dir /usr/share/dotnet --no-path
```
RassK, line 113:
```suggestion
          ./dotnet-install.sh -channel 6.0.xxx --install-dir /usr/share/dotnet
```

Could try this one. Seems it doesn't know about the installed dotnet versions.

dszmigielski: Will do, cheers :)

Kielek:
Closing for now. Not ARM64 env. available.

PR #1880, opened 2022-12-30
Update the usage of set-output commmand in GH workflows

vasireddy99:
This PR updates the usage of set-output command in GH actions.

Reference : https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

ChangeLog entry is not required


.github/workflows/release.yml
```diff
@@ -76,7 +76,7 @@ jobs:
       - run: cd bin-macos-11 ; zip -qq -r ../opentelemetry-dotnet-instrumentation-macos.zip . * ; cd ..
       - name: Extract Version from Tag
         id: get_version
-        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\/v}
+        run: echo "VERSION=${GITHUB_REF/refs\/tags\/v}" >> $GITHUB_OUTPUT
       - name: Create Release
         run: gh release create v${{ steps.get_version.outputs.VERSION }} --draft ./*.zip ./otel-dotnet-auto-install.sh ./OpenTelemetry.DotNet.Auto.psm1
```
pellared, line 81:
```suggestion
        run: gh release create v${{ env.VERSION }} --draft ./*.zip ./otel-dotnet-auto-install.sh ./OpenTelemetry.DotNet.Auto.psm1
```

see: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-1

Kielek (approved):

.github/workflows/release.yml
vasireddy99: oh sure, It could be implemented either way. Thanks, updated as per the suggestion.

pellared (approved):

PR #1884, opened 2023-01-03
Install netfx dlls into GAC

pellared:
## Why

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1823

## What

Install and uninstall netfx dlls into GAC via OpenTelemetry.DotNet.Auto.psm1

## Tests

Manual

> Test an Fx application referencing an older version of the System.Memory dll or something used by OTLP should cover it. In this case, if the needed dll is not on the GAC and we load a newer version from the netfx folder the runtime will error because the application actually is shipping a different version. I think I added some workaround to these issues to some specific tests when I did the netfx redirection. Let's check that.

1. Introduce a test failure

```diff
$ git diff
diff --git a/test/test-applications/integrations/TestApplication.StrongNamed/Program.cs b/test/test-applications/integrations/TestApplication.StrongNamed/Program.cs
index 3c00b352d..ddf0e49f3 100644
--- a/test/test-applications/integrations/TestApplication.StrongNamed/Program.cs
+++ b/test/test-applications/integrations/TestApplication.StrongNamed/Program.cs
@@ -14,6 +14,7 @@
 // limitations under the License.
 // </copyright>
 
+using System;
 using TestLibrary.InstrumentationTarget;
 
 namespace TestApplication.StrongNamed;
@@ -22,6 +23,10 @@ public static class Program
 {
     public static void Main(string[] args)
     {
+        var arrayMemory = new byte[100];
+        var arraySpan = new Span<byte>(arrayMemory);
+        Console.WriteLine(arraySpan.ToString());
+
         var command = new Command();
         command.Execute();
         command.InstrumentationTargetMissingBytecodeInstrumentationType();
diff --git a/test/test-applications/integrations/TestApplication.StrongNamed/TestApplication.StrongNamed.csproj b/test/test-applications/integrations/TestApplication.StrongNamed/TestApplication.StrongNamed.csproj
index f21f7478c..3ce15ced7 100644
--- a/test/test-applications/integrations/TestApplication.StrongNamed/TestApplication.StrongNamed.csproj
+++ b/test/test-applications/integrations/TestApplication.StrongNamed/TestApplication.StrongNamed.csproj
@@ -8,4 +8,8 @@
     <ProjectReference Include="..\dependency-libs\TestLibrary.InstrumentationTarget\TestLibrary.InstrumentationTarget.csproj" />
   </ItemGroup>

+  <ItemGroup>
+    <PackageReference Include="System.Memory" Version="4.5.0" />
+  </ItemGroup>
+
 </Project>
```

2. Run `nuke Workflow --test-name StrongNamedTests --test-project Integration`

3. **TEST FAILED**
4. Execute in PowerShell (as admin):
    ```pwsh
    Import-Module .\OpenTelemetry.DotNet.Auto.psm1
    Install-OpenTelemetryCore
    ```       
5. Run `nuke RunManagedIntegrationTests --test-name StrongNamedTests`
6. **TEST PASSED** 🎉 
7. Run `Uninstall-OpenTelemetryCore`
8. **TEST FAILED** 🎉 

## Checklist

  - Q: Is it worth mentioning?
  - Q: Is it worth mentioning?
  - Q: Is it worth having some automated test?


pellared:
@pjanotti Is there anything I am missing? Any recommendation on how to test it?

pellared:
> We need a friendly way to install the assemblies shipped under the netfx folder into the GAC. This is necessary in the case the application and auto instrumentation ship different versions of the same assembly.

Do we want to add strong names for all of the assemblies so that they can be added to GAC?

pjanotti:
@pellared 
> Is there anything I am missing?

This should cover it.

> Any recommendations on how to test it?

Test an Fx application referencing an older version of the `System.Memory` dll or something used by `OTLP` should cover it. In this case, if the needed dll is not on the GAC and we load a newer version from the `netfx` folder the runtime will error because the application actually is shipping a different version. I think I added some workaround to these issues to some specific tests when I did the netfx redirection. Let's check that.

pjanotti:
> Do we want to add strong names for all of the assemblies so that they can be added to GAC?

🤔 I'm unsure at this point. In principle, this is mostly for DLLs shipped by the actual Fx applications themselves but which ones would we need besides `OpenTelemetry.AutoInstrumentation`? That said it may simplify things on our side if we just strong name everything on the `netfx` folder.

OpenTelemetry.DotNet.Auto.psm1
```diff
@@ -211,6 +211,14 @@ function Install-OpenTelemetryCore() {
 
         # OpenTelemetry service locator
         [System.Environment]::SetEnvironmentVariable('OTEL_DOTNET_AUTO_INSTALL_DIR', $installDir, [System.EnvironmentVariableTarget]::Machine)
+
+        # Register .NET Framweworks dlls in GAC
+        [System.Reflection.Assembly]::Load("System.EnterpriseServices, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a")            
+        $publish = New-Object System.EnterpriseServices.Internal.Publish 
+        $dlls = Get-ChildItem -Path $installDir\netfx\ -Filter *.dll -File
+        foreach ($dll in $dlls) {
+            $publish.GacInstall($dll)
```
pjanotti, line 220: Q.: what happens if the dll is not strong-named?

OpenTelemetry.DotNet.Auto.psm1
```diff
@@ -235,6 +243,14 @@ function Uninstall-OpenTelemetryCore() {
         throw "OpenTelemetry Core is already removed."
     }
 
+    # Unregister .NET Framwework dlls from GAC
+    [System.Reflection.Assembly]::Load("System.EnterpriseServices, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a")            
+    $publish = New-Object System.EnterpriseServices.Internal.Publish 
+    $dlls = Get-ChildItem -Path $installDir\netfx\ -Filter *.dll -File
+    foreach ($dll in $dlls) {
+        $publish.GacRemove($dll)
```
pjanotti, line 251: This can be a problem with the GAC because by itself it doesn't keep some kind of reference counting, so we may end up removing something added by other applications. I think MSIs interact with some low-level Windows APIs to do some kind of reference counting.

Short of using an MSI we may want to look into whatever API MSIs use for this purpose. That said we should treat as a separate issue and address it in a separate issue/PR.

OpenTelemetry.DotNet.Auto.psm1
pellared: Right now only 3 DLLs (grpc_csharp_ext.x64.dll, grpc_csharp_ext.x86.dll, netstandard.dll) are not added to GAC and if it fails then we have a warning log in Event Viewer 

PS. Fix https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1884/commits/cf1c304388ccc3c1d19db452d0d1597b00ec72c3

OpenTelemetry.DotNet.Auto.psm1
pellared: There is a chance that `$publish.GacRemove($dll)` has some logic as `netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL` has not been removed by this line 😉

pjanotti: 🤞

pellared:
Tested manually (described in PR description). Ready for review.

pjanotti (approved):

RassK (approved):

OpenTelemetry.DotNet.Auto.psm1
pjanotti: For the record: assemblies added to the GAC by MSIs aren't removed by `$publish.GacRemove`. Assemblies added to the GAC via `$publish.GacInstall` are not protected in the same way and multiple additions from different locations for the same assembly are clobbered by a single `$publish.GacRemove`.

pjanotti: Some more info related to the GAC usage at https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1906#issuecomment-1376292814

PR #1895, opened 2023-01-04
Refactor solution setup

### LINKED ISSUE #1894: Remove OpenTelemetry.AutoInstrumentation.proj
### What

I'm proposing to remove `OpenTelemetry.AutoInstrumentation.proj` by moving `OpenTelemetry.AutoInstrumentation.Native` and `OpenTelemetry.AutoInstrumentation.Native.DLL` to `OpenTelemetry.AutoInstrumentation.Native.sln`
This should then simplify build setup, as currently `OpenTelemetry.AutoInstrumentation.proj` might need some manual tweaks if there is a conflict with main sln.

Seems also blocking https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1892

RassK:
## Why

Fixes #1894

## What

* Removes `OpenTelemetry.AutoInstrumentation.proj`
* Logic contained in `OpenTelemetry.AutoInstrumentation.proj` is transferred to nuke build.

## Tests

Existing

## Checklist


build/Build.Steps.Windows.cs
```diff
@@ -25,15 +25,17 @@ partial class Build
...
-                .CombineWith(platforms, (m, platform) => m
-                    .SetTargetPlatform(platform)));
+            foreach (var project in Solution.GetNativeSrcProjects())
+            {
+                // Can't use dotnet msbuild, as needs to use the VS version of MSBuild
+                MSBuild(s => s
+                    .SetTargetPath(project)
+                    .SetConfiguration(BuildConfiguration)
+                    .DisableRestore()
+                    .SetMaxCpuCount(null)
```
pjanotti, line 35: No from your change, but, I am curious: isn't the default good enough? It seems strange to call it and pass `null`.

RassK: > If you don't include this switch, the default value is 1. If you include this switch without specifying a value, MSBuild will use up to the number of processors in the computer.

Seems this api has quite poor readability, but null value here makes sense.

build/ProjectsHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+        return solution
+            .GetTestApplications()
+            .Where(x => x.Name.Contains("AspNet") || x.Name.EndsWith(NetFrameworkMarker));
+    }
+
+    public static IEnumerable<Project> GetCrossPlatformTestApplications(this Solution solution)
+    {
+        return solution
+            .GetTestApplications()
+            .Where(x => !x.Name.Contains("AspNet") && !x.Name.EndsWith(NetFrameworkMarker));
```
RassK, line 88: I'll refactor "AspNet" later, this project should be renamed and reusing `NetFrameworkMarker`.

build/Build.Steps.cs
```diff
@@ -73,22 +72,41 @@ partial class Build
...
+                        .SetTargetPath(project)
+                        .SetSolutionDirectory(Solution.Directory)
+                        .SetVerbosity(NuGetVerbosity.Normal)
+                        .When(!string.IsNullOrEmpty(NugetPackageDirectory), o =>
+                            o.SetPackagesDirectory(NugetPackageDirectory)));
+                }
+            }
+            else
+            {
+                foreach (var project in Solution.GetCrossPlatformManagedProjects())
```
pjanotti, line 100: What are these projects? At first, it seems redundant with the solution projects.

build/Build.Steps.cs
```diff
@@ -73,22 +72,41 @@ partial class Build
...
+                {
+                    DotNetRestore(s => s
+                        .SetProjectFile(project)
+                        .SetVerbosity(DotNetVerbosity.Normal)
+                        // .SetTargetPlatform(Platform) // necessary to ensure we restore every project
```
pjanotti, line 105: Remove commented code.

build/Projects.cs
```diff
@@ -13,6 +13,12 @@ public static class Tests
         public const string AutoInstrumentationLoaderTests = "OpenTelemetry.AutoInstrumentation.Loader.Tests";
         public const string AutoInstrumentationBootstrappingTests = "OpenTelemetry.AutoInstrumentation.Bootstrapping.Tests";
         public const string AutoInstrumentationTests = "OpenTelemetry.AutoInstrumentation.Tests";
+        public const string IntegrationTests = "IntegrationTests";
+
+        public static class Applications
```
pjanotti, line 18: Hum, in light of the other comment above a name that makes clear that this is being singled out because of the usage of `packages.config` seems better. It also should be a collection. This way we can add any other managed project that uses one here and things should just work - instead of adding the project to the concat where this one is added.

pjanotti (commented):
Some of the comments below won't make sense in light of this one. After seeing `ProjectsHelper.cs` it seems that what we want is to have something like a `ProjectsUsingPackageConfig` for the special restore using `nuget` directly. Would that simplify things in the end?

build/Build.Steps.cs
RassK: This is the sub collection that was previously done with `.proj`, now the inclusions / exclusions logic is transferred here.
Shortly, all projects that can be restored and built in Linux / MacOS.

build/Projects.cs
RassK: It's actually singled out for 2 purposes: 1 is the `packages.config` the second is for `Build.Steps.Windows`.
Why I didn't make all helpers `IEnumerable<Project>` is that, I wasn't sure if there will be additions in the near future. When it happens there could be need for some additional refactor to introduce a pattern etc. So I applied YAGNI at this moment.

pjanotti: I guess `packages.config` would only ever apply on Windows, but, YAGNI may be good for now :)

pjanotti (approved):

zacharycmontoya (approved):

PR #1983, opened 2023-01-12
Add logger to startup

### LINKED ISSUE #1864: Add Log support to StartupHook and AspNetCoreBootstrapper
Both `OpenTelemetry.AutoInstrumentation.StartupHook` and `OpenTelemetry.AutoInstrumentation.AspNetCoreBootstrapper` writes the status to `EventSource`. Additional tools like `logman`, `dotnet-trace`, `Perfview`, etc., are needed to capture these event source logs.

We have an in-built file-based logging in [`OpenTelemetry.AutoInstrumentation`](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/tree/main/src/OpenTelemetry.AutoInstrumentation/Logging) project. We should extend this feature to StartupHook and AspNetCoreBootstrapper to write the logs to file system for easier debugging.

dszmigielski:
## Why

We do not log to the file in `StartupHook` and `AspNetCoreBootstrapper`.
 
Fixes #1864

## What

Extract logging logic to separate project. Use file logging in the code mentioned above.

## Tests

CI/CD

## Checklist


src/OpenTelemetry.AutoInstrumentation.Logging/Properties/AssemblyInfo.NetFramework.cs
```diff
@@ -0,0 +1,25 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+#if NETFRAMEWORK
+using System.Reflection;
+using System.Runtime.CompilerServices;
+
+[assembly: AssemblyKeyFile("keypair.snk")]
```
pellared, line 21: Can you check if you can move move `keypair.snk` to `src` and change this like to:


```suggestion
[assembly: AssemblyKeyFile("..\\keypair.snk")]
```

The same should be then done in `src\OpenTelemetry.AutoInstrumentation.Logging\Properties\AssemblyInfo.NetFramework.cs`

Reference: https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assemblykeyfileattribute.-ctor?view=net-7.0

src/OpenTelemetry.AutoInstrumentation.Logging/OpenTelemetry.AutoInstrumentation.Logging.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\CommonExcludedAssets.props" />
+
+  <PropertyGroup>
+    <Version>0.5.1</Version>
```
pellared, line 6: Option 1: Update `releasing.md` with info to bump the version to.

Option 2: Maybe we can mve this property to `src/Directory.Build.props`?

src/OpenTelemetry.AutoInstrumentation.AspNetCoreBootstrapper/Logger.cs
```diff
@@ -14,23 +14,23 @@
...
+    private static readonly ILogger Log = OtelLogging.GetLogger();
 
-    /// <summary>
-    /// Gets the column number of the document where the error occurred
-    /// </summary>
-    public int Column;
+    internal static void LogTrace(string message)
+    {
+        Log.Information(message);
+        BootstrapperEventSource.Log.Trace(message);
```
pellared, line 28: Not directly related, but I think we should make the logging and EventSource usage consistent. `AutoInstrumentationEventSource` looks to be used in a different way. Additionally, probably we could have a single base class for all of our event sources.

I have also noticed that all event sources' `Trace` methods have `Level = EventLevel.Informational` (sic!). We should clean up these levels.

This comment should be addressed in a seperate PR. If you agree, you can create an issue out of this comment.

src/OpenTelemetry.AutoInstrumentation.AspNetCoreBootstrapper/Logger.cs
```diff
@@ -14,23 +14,23 @@
...
-    /// Gets the line number of the document where the error occurred
-    /// </summary>
-    public int Line;
+    private static readonly ILogger Log = OtelLogging.GetLogger();
 
-    /// <summary>
-    /// Gets the column number of the document where the error occurred
-    /// </summary>
-    public int Column;
+    internal static void LogTrace(string message)
```
pellared, line 25: It is not a "trace level" log, but a "Information" level log. Can we rename it to `Information` or `LogInformation` or `LogInfo`?

src/OpenTelemetry.AutoInstrumentation.Logging/CustomLogger.cs
```diff
@@ -14,7 +14,7 @@
 // limitations under the License.
 // </copyright>
 
-using System.Runtime.CompilerServices;
+using System;
```
pellared, line 17: I think we can rename `CustomLogger` to `Logger`. Then we could move the `OtelLogging` code here and rename `GetLogger()` to `Get()` or even `Value`.

This shoud be addressed in a separate PR.

src/OpenTelemetry.AutoInstrumentation.StartupHook/Logger.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+    private static readonly ILogger Log = OtelLogging.GetLogger();
+
+    internal static void LogTrace(string message)
+    {
+        StartupHookEventSource.Log.Trace(message);
+
+        try
+        {
+            Log.Information(message);
```
pellared, line 32: Shouldn't it be `Log.Trace(message);`?

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
```diff
@@ -75,20 +77,31 @@ public static void Initialize()
...
+    private static void InitLogger()
+    {
+        // Workaround to make startup hook load the logger assembly.
+        var executionPath = Assembly.GetExecutingAssembly().Location;
+        var dirname = Path.GetDirectoryName(executionPath);
+        if (!string.IsNullOrEmpty(dirname))
+        {
+            Assembly.LoadFrom(Path.Combine(dirname, "OpenTelemetry.AutoInstrumentation.Logging.dll"));
+        }
+    }
```
pellared, line 103: How about moving this code to the `Logger` static constructor?

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
```diff
@@ -75,20 +77,31 @@ public static void Initialize()
...
     }
 
+    private static void InitLogger()
+    {
+        // Workaround to make startup hook load the logger assembly.
+        var executionPath = Assembly.GetExecutingAssembly().Location;
+        var dirname = Path.GetDirectoryName(executionPath);
+        if (!string.IsNullOrEmpty(dirname))
+        {
+            Assembly.LoadFrom(Path.Combine(dirname, "OpenTelemetry.AutoInstrumentation.Logging.dll"));
```
pellared, line 101: Shouldn't we catch exception thrown by `Assembly.LoadFrom` and use event source to have any kind of indcation that something bad happened?

src/OpenTelemetry.AutoInstrumentation.StartupHook/Logger.cs
dszmigielski: Why? This is the file logger, it does not have a `Trace` method.

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
dszmigielski: Tried that and it didn't work for some reason :|

src/OpenTelemetry.AutoInstrumentation.StartupHook/Logger.cs
pellared: I see. This can be addressed together with https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1983/files/19815362edfaf339754bbd7eb8c8c15288e0e50e#r1072147421

src/OpenTelemetry.AutoInstrumentation.Logging/OpenTelemetry.AutoInstrumentation.Logging.csproj
nrcventura: It might be simpler to use option 1 for now, and perhaps try out option 2 separately.

src/OpenTelemetry.AutoInstrumentation.Logging/ILogger.cs
```diff
@@ -0,0 +1,107 @@
...
+// limitations under the License.
+// </copyright>
+
+using System;
+
+namespace OpenTelemetry.AutoInstrumentation.Logging;
+
+#pragma warning disable CS1591
+[System.Diagnostics.CodeAnalysis.SuppressMessage("StyleCop.CSharp.DocumentationRules", "SA1600:Elements should be documented", Justification = "Not")]
+public interface ILogger
```
rajkumar-rangaraj, line 23: Shall we use a different name here?
`ILogger` is a dotnet reserved name and we could avoid naming conflict.

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
```diff
@@ -75,20 +77,38 @@ public static void Initialize()
...
+    private static void InitLogger()
+    {
+        // Workaround to make startup hook load the logger assembly.
+        var executionPath = Assembly.GetExecutingAssembly().Location;
+        var dirname = Path.GetDirectoryName(executionPath);
+        if (!string.IsNullOrEmpty(dirname))
+        {
+            try
+            {
+                Assembly.LoadFrom(Path.Combine(dirname, "OpenTelemetry.AutoInstrumentation.Logging.dll"));
```
rajkumar-rangaraj, line 103: `OpenTelemetry.AutoInstrumentation.StartupHook.csproj` has project reference to `OpenTelemetry.AutoInstrumentation.Logging.csproj`, hence `OpenTelemetry.AutoInstrumentation.Logging.dll` should get loaded. Is there a reason to load this dynamically?

dszmigielski: The reason is it does not load it for some reason. My guess is that StartupHook is loaded as `DOTNET_STARTUP_HOOKS` and it just loads single assembly.

src/OpenTelemetry.AutoInstrumentation.Logging/ILogger.cs
dszmigielski: I'll rename to IOtelLogger.

src/OpenTelemetry.AutoInstrumentation.Logging/OpenTelemetry.AutoInstrumentation.Logging.csproj
dszmigielski: I have updated `releasing.md` for now.

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
pellared: https://github.com/dotnet/runtime/blob/main/docs/design/features/host-startup-hook.md#no-dependency-resolution-for-non-app-assemblies kudos to @lachmatt for finding it 😉

pjanotti:
@dszmigielski The fact that we need an explicit load from the `StartupHook` and the fact this introduces a project dependency to the managed instrumentation project makes me think if it wouldn't be better to just share the sources instead of creating a new project. I don't see a major problem with adding a new dll, but, OTOH I don't see any downside to just share the sources between the projects. The sources could go under something like `./src/shared/OpenTelemetry.AutoInstrumentation.Logging/` .

pjanotti (changes requested):
Please, see https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1983#issuecomment-1387684812

dszmigielski:
>

dszmigielski:
> @dszmigielski The fact that we need an explicit load from the `StartupHook` and the fact this introduces a project dependency to the managed instrumentation project makes me think if it wouldn't be better to just share the sources instead of creating a new project. I don't see a major problem with adding a new dll, but, OTOH I don't see any benefits versus sharing the sources between the 2 projects. The sources could go under something like `./src/shared/OpenTelemetry.AutoInstrumentation.Logging/` .

done, please review again :)

src/OpenTelemetry.AutoInstrumentation.StartupHook/Logger.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace OpenTelemetry.AutoInstrumentation.StartupHook;
+
+internal static class Logger
+{
+    private static readonly IOtelLogger Log = OtelLogging.GetLogger();
+
+    internal static void LogInformation(string message)
+    {
+        StartupHookEventSource.Log.Trace(message);
```
pjanotti, line 28: nit of nitness: just for source consistency: the other `Logger.cs`, combining EventSource with the `OtelLogging`, does it in the reverse order. Use the same order on both.

pjanotti (commented):
Thanks for moving to shared sources. A few small things to change before we can merge.

src/OpenTelemetry.AutoInstrumentation.AspNetCoreBootstrapper/Logger.cs
rajkumar-rangaraj: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2017

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
```diff
@@ -42,7 +42,7 @@ public void Add(InstrumentationInitializer loader)
 
     private class OnAssemblyLoadInitializer
     {
-        private static readonly ILogger Logger = OtelLogging.GetLogger();
+        private static readonly IOtelLogger OtelLogger = OtelLogging.GetLogger();
```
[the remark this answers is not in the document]
rajkumar-rangaraj: It was based on my feedback. I agree, its better to rename in a follow up PR.

rajkumar-rangaraj:
@dszmigielski 
Files are moved into `src/Shared/OpenTelemetry.AutoInstrumentation.Logging/` without a project file. It's best practice to create shared project and place files in it.

src/OpenTelemetry.AutoInstrumentation/Properties/AssemblyInfo.NetFramework.cs
```diff
@@ -18,7 +18,7 @@
 using System.Reflection;
 using System.Runtime.CompilerServices;
 
-[assembly: AssemblyKeyFile("keypair.snk")]
+[assembly: AssemblyKeyFile("../keypair.snk")]
```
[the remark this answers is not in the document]
dszmigielski: > @dszmigielski Files are moved into `src/Shared/OpenTelemetry.AutoInstrumentation.Logging/` without a project file. It's best practice to create shared project and place files in it.
> 
> It is not a blocker and we could create an issue to track it.

I would leave it as is for the moment. It will require build changes, as we would probably not want that to be built, and for sure we would not want the output dll to be included into tracer.

dszmigielski:
> @dszmigielski Files are moved into `src/Shared/OpenTelemetry.AutoInstrumentation.Logging/` without a project file. It's best practice to create shared project and place files in it.
> 
> It is not a blocker and we could create an issue to track it.

I would leave it as is for the moment. It will require build changes, as we would probably not want that to be built, and for sure we would not want the output dll to be included into tracer.

src/OpenTelemetry.AutoInstrumentation.Logging/CustomLogger.cs
pellared: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2022

dszmigielski:
@rajkumar-rangaraj after consulting with @pellared I have moved logger source files back to `OpenTelemetry.AutoInstrumentation` and linked them to other project. I hope that this solution works for you.

pellared:
> @rajkumar-rangaraj after consulting with @pellared I have moved logger source files back to `OpenTelemetry.AutoInstrumentation` and linked them to other project. I hope that this solution works for you.

I just want to mention that this approach is used in https://github.com/open-telemetry/opentelemetry-dotnet

src/OpenTelemetry.AutoInstrumentation/Loading/LazyInstrumentationLoader.cs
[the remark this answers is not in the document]
pellared: Maybe as part of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2022 ?

src/OpenTelemetry.AutoInstrumentation/CallTarget/Handlers/IntegrationMapper.cs
```diff
@@ -28,7 +28,7 @@ internal class IntegrationMapper
     private const string EndMethodName = "OnMethodEnd";
     private const string EndAsyncMethodName = "OnAsyncMethodEnd";
 
-    private static readonly ILogger Log = OtelLogging.GetLogger();
+    private static readonly IOtelLogger Log = OtelLogging.GetLogger();
```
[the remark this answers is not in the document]
pellared: Added as comment here: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2022

rajkumar-rangaraj:
> uld probably not want that to be built, and for sure we would not want the output dll to be included into tracer.

I was speaking here about `.shproj` and not `.csproj`. I have been following this PR since beginning and completely agree with @pjanotti feedback, we should not introduce another new library.

New proposed approach looks good to me.

rajkumar-rangaraj (approved):

pjanotti (approved):
Thanks @dszmigielski

PR #1996, opened 2023-01-18
Automate generation of integrations.json file

pjanotti:
## Why

There is no need for the generation of the integrations.json to be a manual step.

## What

- Include the run of the `IntegrationsJsonGenerator` as part of the `BuildTracer` target.
- Keep the `integrations.json` under version control so any changes to it are visible and reviewed.

## Tests

The test added when doing .NET Framework automatic direction will report any sources changed during the CI run, see https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/7d49514a0d19fb5d164073f58ce8b7661222f97c/.github/workflows/ci.yml#L85

## Checklist

### DIFF
--- a/tools/IntegrationsJsonGenerator/IntegrationsJsonGenerator.csproj
+++ b/tools/IntegrationsJsonGenerator/IntegrationsJsonGenerator.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net7.0</TargetFrameworks>
+    <TargetFramework>net7.0</TargetFramework>
   </PropertyGroup>
 
 </Project>
--- a/build/Build.Steps.cs
+++ b/build/Build.Steps.cs
@@ -231,10 +231,19 @@ partial class Build
         .DependsOn(PublishNativeProfilerLinux)
         .DependsOn(PublishNativeProfilerMacOs);
 
+    Target GenerateIntegrationsJson => _ => _
+        .After(PublishManagedProfiler)
+        .Executes(() =>
+        {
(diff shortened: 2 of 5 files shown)


nrcventura (approved):

rajkumar-rangaraj (approved):

tools/IntegrationsJsonGenerator/IntegrationsJsonGenerator.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net7.0</TargetFrameworks>
+    <TargetFramework>net7.0</TargetFramework>
```
Kielek, line 5: I would revert this file. Parents props file is using `TargetFrameworks`. Potentially it can introduce some unexpected behabiout.

pjanotti: When both `TargetFrameworks` and `TargetFramework` are present in a `.csproj` the first one is ignored. This is well-defined behavior, but, if it is confusing we can either add a comment here or add the `--framework` platform setting directly on the build source. Any preference?

Kielek: Lets keep it as it is.
Thanks for explanation.

pjanotti: It makes more sense to add the comment: it is an executable, and we want to build to a single runtime, so running with the latest available avoids forcing the code to compile to .NET Framework too. The app code can evolve with the latest runtime, BCL, without concerns.

Issue #1429, opened 2022-10-17
Improve configuration via app/web.config


pellared:
## Current behavior

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/fa6c733c6ff418f9a6d967bcd1509f65f2150717/src/OpenTelemetry.AutoInstrumentation/Configuration/TracerSettings.cs#L111

This means that only the env vars handled in OTel .NET AutoInstrumentation managed code can configured using `app/web.config`

The user cannot use it e.g. to configure an ASP.NET application deployed on IIS.

The feature is also not documented (which is probably good at this moment

## Proposal

1. Add additional code that would check if any (an allow-list) of OTel SDK configuration env vars are set via `app/web.config` if so then we should set them before bootstrapping the OTel .NET SDK.
2. Document the possibility of configuring via `app/web.config` on .NET Framework. List configuration (.NET Profiler and Runtime env vars, env vars used in our profiler code) that are not supported by `app/web.config` .

pellared:
Changing the milestone as this is important for IIS deployed ASP.NET apps

pellared:
Notice: The config in app/web.config should NOT override a setting which is already set via env var.

pellared:
Probably generalized OTEL_* whitelist should be fine.

RassK:
Native part also doesn't see app / web.config vars... but I think this will become specially important with official OTel config file.

1. Add these vars to the exception list? - will make UX more complex
2. Bridge dictionary from the managed level? - not sure if the execution location is too early for managed bridge?

tanaka-takayoshi:
Do you plan to support configuring [Resource semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#semantic-attributes-with-sdk-provided-default-value) by using Web.config?
This helps the situation that we want to identify the web application names run by a single application pool. Since these applications are run by the same process, env var can't specify different names for these applications.

pellared:
@tanaka-takayoshi Yes. That is the plan 😉 I hope that it will be finished by the end of January.

pellared:
Status after merging https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2066:

Following settings cannot be configured via `app.config`/`web.config`:

- `OTEL_DOTNET_AUTO_DEBUG` 
[...]

PR #2014, opened 2023-01-19
Add tests for configuration via app/web.config

pellared:
## Why

Towards https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1429

## What

Add integration tests for support configuration via app/web.config.
These are skipped as the implementation is not done yet and the test are checked that they fail as expected.

## Tests

This is what this PR is about 😉 

## Checklist

- [x] New features are covered by tests. (skipped for now)


test/IntegrationTests/WcfNetFrameworkTests.cs
```diff
@@ -15,15 +15,36 @@
...
 
 namespace IntegrationTests;
 
 public class WcfNetFrameworkTests : WcfTestsBase
 {
+    private const string ServiceName = "Wcf.Client.NetFramework";
+
     public WcfNetFrameworkTests(ITestOutputHelper output)
-        : base("Wcf.Client.NetFramework", output)
+        : base(ServiceName, output)
```
Kielek, line 28: ServiceName passed as testAppName is a bit misleading.

test/IntegrationTests/WcfNetFrameworkTests.cs
```diff
@@ -15,15 +15,36 @@
...
+    {
+    }
+
+    [Fact(Skip = "https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1429")]
+    [Trait("Category", "EndToEnd")]
+    public async Task TracesResource()
```
Kielek, line 34: As we discussed, I would test with independent app.
It is not related to WCF,

Can we call method TracesResourceForAppConfig?

test/IntegrationTests/WcfNetFrameworkTests.cs
pellared: Addressed by https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2014/commits/66879c70af6e8cf70d7d89c62196d7e9e6f51aee 

PTAL

test/IntegrationTests/WcfNetFrameworkTests.cs
pellared: I started to think if we should also test the `web.config` in `AspNetTests` as it is redundant, however it feels "safer" to double-check to avoid regression bugs. This is why I am leaving the `AspNetTests`.

RassK (approved):
Seems good enough to start with I think

Issue #1728, opened 2022-12-07
Refactor ConfigurationSources


pellared:
It could be enough to implement it only here if we refactor it. I can try doing it later.

Basically the idea is to change `CompositeConfigurationSource` to e.g. `Configuration`. It will not implement `IConfigurationSource` and existing code would depend on `Configuration` instead of `IConfigurationSource`. Then we will be sure that the "sanitization" is always applied.

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1724#discussion_r1042457881_

RassK:
This seems quite a big scope, since code is using composite source and tests are mainly using some specific source.
Transferring composite source to the tests makes the configuration layer quite fat for a single test + there is danger of leaking configurations (eg if env source is brought to the tests).

PR #2027, opened 2023-01-20
Refactor Configuration namespace

pellared:
## Why

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1728

## What

- Refactor `CompositeConfigurationSource` to `Configuration` which does not implement `IConfigurationSource`. This way we make sure that `Settings` classes are for sure that empty string is handled as if there was no value. It is a similar design to `Logger` and `LogSink`.
- Refactor `Settings` to have `abstract` method instead of logic inside constructor. Thanks to it we can get rid of reflection in `Settings` class.
- Other minor cleanup.

## Tests

- Existing tests
- Side note: I noticed we are missing tests for `GetInt32` , `GetDouble`, `GetBool`. I think I will add then in next PR.


pellared:
Do not merge until @RassK reviews it.

src/OpenTelemetry.AutoInstrumentation/Configuration/Settings.cs
```diff
@@ -24,56 +24,42 @@ namespace OpenTelemetry.AutoInstrumentation.Configuration;
...
-
     /// <summary>
     /// Gets the the OTLP transport protocol. Supported values: Grpc and HttpProtobuf.
     /// </summary>
-    public OtlpExportProtocol? OtlpExportProtocol { get; }
+    public OtlpExportProtocol? OtlpExportProtocol { get; private set; }
 
     public static T FromDefaultSources<T>()
-        where T : Settings
+        where T : Settings, new()
```
RassK, line 33: Could be further optimized, so there will be a single configuration instance per setting object. Should be noted, it will be still short living object and there is a conflict with lazy pattern.

Additional + would be to take account to make a possible extensibility point, so plugins can define proprietary configuration source.

Kielek: I would keep minimal extension points. As long as there is no request for that, we should avoid it.

RassK (approved):

src/OpenTelemetry.AutoInstrumentation/Configuration/Settings.cs
RassK: Sure, the ideas was just not to rewrite it like it will be a dead end, but to have some open mind when doing it.

Kielek (approved):

PR #2040, opened 2023-01-23
Move logging to separate project

dszmigielski:
## Why

Logging does not work after https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1983/
Because of how the logger is used, compiling it as a separate class into two assemblies causes file lock conflict resulting in logs not being written at all. Moving it out as a single assembly solves that.

Kudos to @lachmatt for finding this.

## What

Move logger classes to separate project.

## Tests

CI/CS

## Checklist


pellared:
Maybe we should create different file names instead?

dszmigielski:
> Maybe we should create different file names instead?

That would mean moving responsibility for creating log file name to the caller side. I also don't understand why we are so against creating additional dll if it solves the problem :(

nrcventura:
> > Maybe we should create different file names instead?
> 
> That would mean moving responsibility for creating log file name to the caller side. I also don't understand why we are so against creating additional dll if it solves the problem :(

If we use different file names, it might be easier to prevent this type of problem in the future. It might also make it easier to find the logs that we care about when trying to troubleshoot different issues. I'm not necessarily against creating additional dlls, but I have experienced cases where having too many dlls lead to slowing down application the application when starting up or lead to additional complexities when deploying things.

dszmigielski:
I have added https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2074 as an alternative solution.

Kielek:
Closing, alternative approach implemented in #2074

PR #2070, opened 2023-01-24
Document app.config limitation

pellared:
## Why

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1429

After discussion with @andrewlock and @Kielek we decided that it is safer and probably good enough NOT to support all configuration via env vars. The main problem are the configuration read by the .NET Profiler which does not know anything about app.config. We could try "refreshing" the native configuration after the managed layer has loaded the app.config. However, that would probably require PInvoke and additional synchronization. Still this might result in race conditions. 

Moreover, that env vars are per-process, but web.config can be per-appdomain; e.g.
- App1 and App2 are web apps running in IIS
- App1 sets OTEL_SOME_SETTING=1 in web.config
- App2 sets OTEL_SOME_SETTING=0 in web.config
- App1 sets the env var OTEL_SOME_SETTING=1 and Pinvokes to notify the profiler
- App2 sets the env var OTEL_SOME_SETTING=0 and Pinvokes to notify the profiler
Profiler doesn't know what to think
Essentially you'd have to check the appdomain for each callback, and some parts of the profiler are agnostic to that
So things will get complicated quickly.

At last this functionality would hopefully be not needed for the "happy path". The usual scenario would be use the app config to set the service name and other resource attributes, configure the exporter(s).

## What

Document `app.config` and `web.config` limitation.

## Tests

Manually by modyfing integration tests and test applications

## Checklist


docs/config.md
```diff
@@ -15,6 +15,16 @@ with environment variables taking precedence over `App.config` or `Web.config` f
     file  (`web.config`) or an application configuration file (`app.config`) to
     configure the `OTEL_*` settings.
 
+    ⚠️ Only settings starting with `OTEL_` can be set using `App.config` or `Web.config`.
+    Additionally, following settings are not supported:
```
theletterf, line 19:
```suggestion
    Additionally, the following settings are not supported:
```

docs/config.md
```diff
@@ -15,6 +15,16 @@ with environment variables taking precedence over `App.config` or `Web.config` f
...
+
+    - `OTEL_DOTNET_AUTO_HOME`
+    - `OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES`
+    - `OTEL_DOTNET_AUTO_INTEGRATIONS_FILE`
+    - `OTEL_DOTNET_AUTO_[TRACES|METRICS|LOGS]_[ENABLED|DISABLED]_INSTRUMENTATIONS`
+    - `OTEL_DOTNET_AUTO_LOG_DIRECTORY`
+    - `OTEL_DOTNET_AUTO_DEBUG`
```
theletterf, line 26: Might require some extra work, but I'm thinking now of re-doing the settings tables so that they have the following columns:

- Setting (env var)
- App.config / Web.config supported
- Description
- Default

theletterf:
@pellared Perhaps we can be more prescriptive as to when  a method should be used instead of another? If there's choice, that is. Of course one can use env vars for everything, but when should they use `app.config` instead?

pellared:
> @pellared Perhaps we can be more prescriptive as to when a method should be used instead of another? If there's choice, that is. Of course one can use env vars for everything, but when should they use `app.config` instead?

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2071

docs/config.md
pellared: Can we keep as it is for this PR (it makes also reviewing easier)? We can change the formatting/structure it in a separate PR.

docs/internal/config.md
```diff
@@ -14,6 +16,12 @@ This document lists advanced configuration settings rarely changed by users.
 |-------------------------------------------|--------------------------------------------------------------------------|---------------|
 | `OTEL_DOTNET_AUTO_DUMP_ILREWRITE_ENABLED` | Lets the profiler dump the IL original code and modification to the log. | `false`       |
 
+## Redirection
```
RassK, line 19: Maybe "Managing dependencies", "Managing dependency conflicts" ? ... the "Redirection" alone here is not describing much the actual issue.

RassK: btw, I think this was actually intentionally public per @pjanotti ?

zacharycmontoya: Maybe "Assembly Loading"?

docs/config.md
```diff
@@ -15,6 +15,16 @@ with environment variables taking precedence over `App.config` or `Web.config` f
     file  (`web.config`) or an application configuration file (`app.config`) to
     configure the `OTEL_*` settings.
 
+    ⚠️ Only settings starting with `OTEL_` can be set using `App.config` or `Web.config`.
+    Additionally, the following settings are not supported:
```
zacharycmontoya, line 19: Small rewording to make it more obvious that these are exclusions

```suggestion
    However, the following settings are not supported:
```

zacharycmontoya (approved):
LGTM with very small comments

docs/internal/config.md
pellared: I will bring this line back to public. I thought it is more internal 😉

pellared: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2070/commits/f1c00c816ea116a9372ff0cf39542d73720ca5d7

Issue #889, opened 2022-06-27
Refactor Nuke tasks


pellared:
The current build implementation using Nuke uses different APIs for defining and ordering dependencies, making it hard to maintain.

We should document and use some guidelines to make it more maintainable.

The proposal is to define two types of tasks:

- **target** which contains logic and use `After` to specify targets that are required to be built before this one is run,
- **workflow** which groups targets (or workflows) using `DependsOn` into a suite.

Slack thread: https://cloud-native.slack.com/archives/C01NR1YLSE7/p1656060079086989

pellared:
The other idea  is to only use `DependsOn` and one could use `--skip` to Skip all other dependencies or some specified ones.

Kielek:
Stale issue. Closing. I do not think that anybody would like to work on this.

PR #2073, opened 2023-01-24
Generate transient dependencies list

### LINKED ISSUE #1824: Bump transitive dependencies in "netfx" to higher version

RassK:
## Why

Fixes #1824

## What

1. Generates transient dependencies list to `OpenTelemetry.AutoInstrumentation`
2. Dependabot will start bumping the versions
3. A test verifies that a PR is not bringing new undefined transient dependencies

## Tests

Verifies transient deps.

## Checklist


build/Build.Steps.Windows.cs
```diff
@@ -118,6 +120,50 @@ partial class Build
             );
         });
 
+    Target GenerateNetFxTransientDependencies => _ => _
+        .Unlisted()
```
pellared, line 124: This target is unlisted and not used anywhere. Shouldn't it be invoked in GH Actions? Otherwise we will have missing entries. Personally I think we could add it to the `Workflow`. We have a `git diff` in `ci.yml - build` GH job that would make sure that everything is up to date.

build/Build.Steps.Windows.cs
```diff
@@ -13,8 +15,8 @@ partial class Build
 {
     Target CompileNativeSrcWindows => _ => _
         .Unlisted()
+        .DependsOn(GenerateNetFxAssemblyRedirectionSource)
```
pellared, line 18: Why this change? I do not remember the exact details right now, but there were some reasons that we are not using the `DependsOn` in "step targets". Reference: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/889

pjanotti:
@rajkumar-rangaraj perhaps we could use a similar approach to pull the transitive dependencies independent of the TFM that we use for the `OpenTelemetry.AutoInstrumentation`. This will ensure the presence of the needed assemblies while keeping targeting the TFM provided by the SDK. Just an idea that keeps things more under our own control.

build/Build.Steps.Windows.cs
pjanotti: I agree with this: unless it is relatively slow I would say that it should be executed before the `CompileManagedSrc` when the TFM is `net462`. If it is really slow it could be part of the `Workflow`.

rajkumar-rangaraj:
> @rajkumar-rangaraj perhaps we could use a similar approach to pull the transitive dependencies independent of the TFM that we use for the `OpenTelemetry.AutoInstrumentation`. This will ensure the presence of the needed assemblies while keeping targeting the TFM provided by the SDK. Just an idea that keeps things more under our own control.

@pjanotti I did not understand the context, could you give little background.

build/Build.Steps.Windows.cs
RassK: I currently don't see any need to auto execute it. It should be executed only within a PR that is going to break dependencies.
Secondly it will generate uncommitted changes then, but all of these changes must be committed for dependendabot to work.

build/Build.Steps.Windows.cs
RassK: This will decouple `GenerateNetFxAssemblyRedirectionSource` from `CompileNativeSrcWindows` imo?
`GenerateNetFxAssemblyRedirectionSource` is only connected to `CompileNativeSrcWindows`...

build/Build.Steps.Windows.cs
pellared: > It should be executed only within a PR that is going to break dependencies.

How we will know it if it is not automated?

> Secondly it will generate uncommitted changes then, but all of these changes must be committed for dependendabot to work.

Then we will just push commits to the dependabot PRs. Thanks to it we will make sure that `main` is always "clean".

build/Build.Steps.Windows.cs
pellared: `BuildTracer` is the "workflow" target which has the `DependsOn`. I propose to follow the current pattern otherwise we will have a mess.

build/Build.Steps.Windows.cs
pjanotti: The goal is to have the list of assemblies automatically update whenever needed. It seems to me that the simplest way to achieve that seems to have this step as part of `BuildTracer` target.

pjanotti:
@rajkumar-rangaraj I mean in the context that currently we build the tracer targeting `net6.0`, however, since some transitive dependencies don't have packages for `net6.0` we don't get the assemblies of such transitive dependencies under the `net`. I did some quick experiments changing the TFM to `netstandard2.0` and this improved the situation. However, as you mentioned, the SDK is removing the `netstandard2.0` target, which makes sense, since they are targeted to applications/exes, that will pull all the needed dependencies at build time.

For automatic instrumentation, it __seems__ that we need to pull all these references for cases when the AdditionalDeps or SharedStore can't be used. I still want to investigate this further and confirm it. Anyway, if we do need to pull these assemblies using the tool being added in this PR could pull the needed assemblies. Makes sense?

build/Build.Steps.Windows.cs
RassK: > How we will know it if it is not automated?
That's exactly why the test is there. It enforces you to run `GenerateNetFxTransientDependencies` just in the right time, so external dependencies are clearly visible and connected to the PR that caused these. 
Running it fully automated causes dependencies to be added without any clear source trail.

build/Build.Steps.Windows.cs
RassK: Which is exactly my point. I'm running these sub targets quite a lot. If you do clean and run `CompileNativeSrc` it won't compile due missing source file. Which means `CompileNativeSrc` does not do what it promises.

build/Build.Steps.Windows.cs
pjanotti: A test covering this is good too but as of right now if the csproj gets updated due to some other change CI will flag it so no accidental changes are going to go through one way or another. If you have it automated as part of the build the change will be visible locally even before the file is submitted so the author can review it.

build/Build.Steps.Windows.cs
pjanotti: In order to run `CompileNativeSrc` and get the correct bits under `tracer-home` one has to `BuildTracer` before (likely after a clean). My workflow is `BuildTracer` once and then build the components as needed, in this case, it will be nice to avoid the new target when doing `CompileNativeSrc`.

RassK: I'll revert this part, not tightly coupled to this PR.

But in general I still don't agree. `CompileNativeSrcWindows` should do exactly what it promises, otherwise it's something else. If you need to optimize `CompileNativeSrcWindows` you can always do `nuke CompileNativeSrcWindows -skip GenerateNetFxAssemblyRedirectionSource` which then skips the parts and does personalized run.

pellared: `nuke CompileNativeSrcWindows -skip` would even work

pellared: I added https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/889#issuecomment-1406277116

build/Build.Steps.Windows.cs
RassK: That actually makes sense, I originally thought that the target is somewhat expensive to be in the main pipeline. 
On my desktop it's just 2 seconds (looking at nuke's feedback), will see after a CI run as it's the weakest machine.

RassK: well `GenerateNetFxTransientDependencies Succeeded  0:13` not too terrible actually.

pellared (approved):

build/Build.Steps.Windows.cs
pjanotti: Excellent! Much better than what I expected too.

pjanotti (approved):

PR #2085, opened 2023-01-26
RHEL 8 is not supported

pellared:
## Why

- https://cloud-native.slack.com/archives/C01NR1YLSE7/p1674752373125219?thread_ts=1674679829.575859&cid=C01NR1YLSE7
- https://distrowatch.com/table.php?distribution=redhat

## What

Update compatibility docs


rajkumar-rangaraj (approved):

Kielek (approved):

theletterf:
@pellared @Kielek I think adding incompatible distributions or OS versions in a list is not sustainable. We should rather frame this as which distro versions and OS versions *are* compatible. WDYT?

Kielek:
Sure, please treat it as dirty-fix, we have a plan to extend test coverage for all dotnet-supported versions.

PR #2088, opened 2023-01-27
Env var instrumentation improvements

### LINKED ISSUE #1633: Select instrumentation libraries via `OTEL_DOTNET_AUTO_*_INSTRUMENTATION_ENABLED`
Instead of `OTEL_DOTNET_AUTO_[SIGNAL]_ENABLED_INSTRUMENTATIONS` and `OTEL_DOTNET_AUTO_[SIGNAL]_DISABLED_INSTRUMENTATIONS` we could support  `OTEL_INSTRUMENTATION_[NAME]_ENABLED` and `OTEL_INSTRUMENTATION_COMMON_DEFAULT_ENABLED` like the [OTel Java Agent](https://opentelemetry.io/docs/instrumentation/java/automatic/agent-config/#suppressing-specific-agent-instrumentation) does.

I think it may be good to be consistent with Java.

If we agree on it we should first try to bring it to the https://github.com/open-telemetry/opentelemetry-specification.

Kielek:
## Why

Fixes #1633

## What

- Change the way to manage enabled instrumentations. The following environmental
  variables:
  - `OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS`,
  - `OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS`,
  - `OTEL_DOTNET_AUTO_METRICS_ENABLED_INSTRUMENTATIONS`,
  - `OTEL_DOTNET_AUTO_METRICS_DISABLED_INSTRUMENTATIONS`,
  - `OTEL_DOTNET_AUTO_LOGS_ENABLED_INSTRUMENTATIONS`,
  - `OTEL_DOTNET_AUTO_LOGS_DISABLED_INSTRUMENTATIONS`

  are replaced by:

  - `OTEL_DOTNET_AUTO_INSTRUMENTATION_ENABLED`,
  - `OTEL_DOTNET_AUTO_TRACES_INSTRUMENTATION_ENABLED`,
  - `OTEL_DOTNET_AUTO_TRACES_{0}_INSTRUMENTATION_ENABLED`,
  - `OTEL_DOTNET_AUTO_METRICS_INSTRUMENTATION_ENABLED`,
  - `OTEL_DOTNET_AUTO_METRICS_{0}_INSTRUMENTATION_ENABLED`,
  - `OTEL_DOTNET_AUTO_LOGS_INSTRUMENTATION_ENABLED`,
  - `OTEL_DOTNET_AUTO_LOGS_{0}_INSTRUMENTATION_ENABLED`.

## Tests

CI + updated tests

## Checklist


CHANGELOG.md
```diff
@@ -29,6 +29,23 @@ This beta release is built on top of [OpenTelemetry .NET](https://github.com/ope
   to `linux-musl-x64` for Linux musl and
   `OpenTelemetry.AutoInstrumentation.Native.dylib`
   to `osx-x64` for MacOS.
+- Changed the way to managed enabled instrumentation. Following environmental variables
```
pellared, line 32:
```suggestion
- Change the way to manage enabled instrumentations. The following environmental variables:
```

CHANGELOG.md
```diff
@@ -29,6 +29,23 @@ This beta release is built on top of [OpenTelemetry .NET](https://github.com/ope
...
+  - `OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS`,
+  - `OTEL_DOTNET_AUTO_TRACES_DISABLED_INSTRUMENTATIONS`z,
+  - `OTEL_DOTNET_AUTO_METRICS_ENABLED_INSTRUMENTATIONS`,
+  - `OTEL_DOTNET_AUTO_METRICS_DISABLED_INSTRUMENTATIONS`,
+  - `OTEL_DOTNET_AUTO_LOGS_ENABLED_INSTRUMENTATIONS`,
+  - `OTEL_DOTNET_AUTO_LOGS_DISABLED_INSTRUMENTATIONS`
+
+  replaced by
```
pellared, line 40:
```suggestion
  are replaced by:
```

docs/config.md
```diff
@@ -57,15 +57,16 @@ for more details.
...
+| `OTEL_DOTNET_AUTO_INSTRUMENTATION_ENABLED`             | Enables all instrumentations.                                                                                                                                                                                                                                   | `true`                                                     |
+| `OTEL_DOTNET_AUTO_TRACES_INSTRUMENTATION_ENABLED`      | Enables all trace instrumentations. Overrides `OTEL_DOTNET_AUTO_INSTRUMENTATION_ENABLED`.                                                                                                                                                                       | Same as `OTEL_DOTNET_AUTO_INSTRUMENTATION_ENABLED`         |
+| `OTEL_DOTNET_AUTO_TRACES_{0}_INSTRUMENTATION_ENABLED`  | Configuration pattern for enabling or disabling a specific trace instrumentation, where `{0}` is the case-sensitive name of the instrumentation you want to enable. Overrides `OTEL_DOTNET_AUTO_TRACES_INSTRUMENTATION_ENABLED`.                                | Same as `OTEL_DOTNET_AUTO_TRACES_INSTRUMENTATION_ENABLED`  |
```
pellared, line 65: I would consider calling out explicitly that it also overrides `OTEL_DOTNET_AUTO_INSTRUMENTATION_ENABLED`. Same comment applies to other places.

Kielek: @theletterf?

docs/iis-instrumentation.md
```diff
@@ -66,4 +66,4 @@ to set it for all ASP.NET application running in Integrated Pipeline Mode:
 ## Enable ASP.NET instrumentation
 
 Make sure to enable the ASP.NET instrumentation by setting
-`OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS=AspNet`.
+`OTEL_DOTNET_AUTO_TRACES_AspNet_INSTRUMENTATION_ENABLED=true`.
```
pellared, line 69: This is probably conflicting and this statement is no longer true 😉  It was solved by lazy loading.

src/OpenTelemetry.AutoInstrumentation.Native/util.cpp
```diff
@@ -102,6 +103,30 @@ std::vector<WSTRING> GetEnvironmentValues(const WSTRING& name)
...
 
+std::vector<WSTRING> GetEnabledEnvironmentValues(const bool enabled_by_default, const std::unordered_map<WSTRING, WSTRING> values_map)
+{
+    std::vector<WSTRING> values;
+ 
+    for (auto value : values_map)
+    {
+        bool enabled = enabled_by_default;
+        const auto env_value = GetEnvironmentValue(value.second);
+        if (env_value == WStr("true"))
```
pellared, line 114: AFAIK everywhere else we are checking also for `1`

I think it not compliant with OTel spec, but I prefer to have consistency. Probably there is some macro for this check.

src/OpenTelemetry.AutoInstrumentation.Native/util.cpp
```diff
@@ -102,6 +103,30 @@ std::vector<WSTRING> GetEnvironmentValues(const WSTRING& name)
...
+        {
+            enabled = true;
+        }
+        else if (env_value == WStr("false"))
```
pellared, line 118: I think that everywhere else we are checking also for `0`

test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/ConfigurationTests.cs
```diff
@@ -33,106 +33,66 @@ private enum TestEnum
...
     [Fact]
-    public void ParseEnabledEnumList_Disabled()
+    public void ParseEnabledEnumList_SelectivelyEnabled()
     {
-        var source = new Configuration(new NameValueConfigurationSource(new NameValueCollection()
+        var source = new Configuration(new NameValueConfigurationSource(new NameValueCollection
         {
-            { "TEST_DISABLED_VALUES", "Test2" }
+            { "TEST_CONFIGURATION_Test1_ENABLED", "true" },
+            { "TEST_CONFIGURATION_Test3_ENABLED", "true" }
```
pellared, line 65:
```suggestion
            { "TEST_CONFIGURATION_Test3_ENABLED", "1" }
```

test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/ConfigurationTests.cs
```diff
@@ -33,106 +33,66 @@ private enum TestEnum
...
 
     [Fact]
-    public void ParseEnabledEnumList_None()
+    public void ParseEnabledEnumList_SelectivelyDisabled()
     {
-        var source = new Configuration(new NameValueConfigurationSource(new NameValueCollection()
+        var source = new Configuration(new NameValueConfigurationSource(new NameValueCollection
         {
-            { "TEST_ENABLED_VALUES", "none" }
+            { "TEST_CONFIGURATION_Test2_ENABLED", "false" },
```
pellared, line 80: can we also add test `0` for disabling?

src/OpenTelemetry.AutoInstrumentation.Native/util.cpp
Kielek: We should drop support for 0 and 1. Will be handled in scope of #2089

src/OpenTelemetry.AutoInstrumentation.Native/util.cpp
Kielek: #2089

docs/config.md
theletterf: Ok with that!

theletterf (approved):

test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/ConfigurationTests.cs
```diff
@@ -33,106 +33,66 @@ private enum TestEnum
...
 
     [Fact]
-    public void ParseEnabledEnumList_Disabled()
+    public void ParseEnabledEnumList_SelectivelyEnabled()
     {
-        var source = new Configuration(new NameValueConfigurationSource(new NameValueCollection()
+        var source = new Configuration(new NameValueConfigurationSource(new NameValueCollection
         {
-            { "TEST_DISABLED_VALUES", "Test2" }
+            { "TEST_CONFIGURATION_Test1_ENABLED", "true" },
```
RassK, line 64: Just a random example here.

Can we do keys in all caps? this format is very unfriendly in UX wise, specially in Linux where a key's case sensitivity matters.

pellared: I agree. I have forgotten to mention it in the issue. This is what Java does and also something similar was proposed here https://github.com/open-telemetry/opentelemetry-specification/issues/3025#issuecomment-1342834029

However, I think it would be better to tackle it in a separate PR.

Kielek: #2094

docs/config.md
```diff
@@ -62,16 +62,16 @@ for more details.
...
+| `OTEL_DOTNET_AUTO_INTEGRATIONS_FILE`                    | List of bytecode instrumentations JSON configuration filepaths, delimited by the platform-specific path separator (`;` on Windows, `:` on Linux and macOS). For example: `%ProfilerDirectory%/integrations.json`. It is required for bytecode instrumentations. |                                                             |
+| `OTEL_DOTNET_AUTO_INSTRUMENTATION_DISABLED`             | Disables all instrumentations.                                                                                                                                                                                                                                  | `false`                                                     |
+| `OTEL_DOTNET_AUTO_TRACES_INSTRUMENTATION_DISABLED`      | Disables all trace instrumentations. Overrides `OTEL_DOTNET_AUTO_INSTRUMENTATION_DISABLED`.                                                                                                                                                                     | Same as `OTEL_DOTNET_AUTO_INSTRUMENTATION_DISABLED`         |
```
pellared, line 69: maybe?

```suggestion
| `OTEL_DOTNET_AUTO_TRACES_INSTRUMENTATION_DISABLED`      | Disables all trace instrumentations. Overrides `OTEL_DOTNET_AUTO_INSTRUMENTATION_DISABLED`.                                                                                                                                                                     | Derived from `OTEL_DOTNET_AUTO_INSTRUMENTATION_DISABLED`         |
```

pellared (approved):
I think this PR should have ~3 reviews/approves 😋 
I do not see any way to make it smaller, but it is still not small 😉 
Still, it is a really great PR.

docs/config.md
```diff
@@ -62,15 +62,16 @@ for more details.
 
 ## Instrumentations
```
theletterf, line 64:
```suggestion
## Instrumentations

All instrumentations are enabled by default  for all signal types (traces, metrics, and logs). 

You can disable all instrumentations for a specific signal type by setting the `OTEL_DOTNET_AUTO_{SIGNAL}_ENABLED_INSTRUMENTATIONS` environment variable to `false`.

For a more granular approach, you can disable specific instrumentations for a given signal type by setting the `OTEL_DOTNET_AUTO_{SIGNAL}_{0}_INSTRUMENTATION_DISABLED` environment variable to `false`, where `{SIGNAL}` is the type of signal, for example traces, and `{0}` is the case-sensitive name of the instrumentation.

```

docs/config.md
theletterf: Suggested some changes to Piotr

docs/config.md
Kielek: Applied in d4398104102607006798baa30f67b8d22d612dba

docs/config.md
Kielek: d4398104102607006798baa30f67b8d22d612dba

pjanotti:
@Kielek why going with the `_DISABLED` instead of `_ENABLED` with reversed defaults? I ask because Java uses `_ENABLED` instead `_DISABLED`. See https://opentelemetry.io/docs/instrumentation/java/automatic/agent-config/#enable-only-specific-instrumentation

theletterf:
@pjanotti Also had the same thought. Negating an affirmation is easier than affirming a negation. @Kielek mentioned it's because of the specifications, though.

Kielek:
@pjanotti, otel-spec says that lack of value should defaults to false value. It is the only way to met this requirement.

BTW there is a commit changing ENABLED to DISABLED. I have started with more obvious version.

pjanotti:
🤔 @Kielek IIRC the conversation on the SIG with @carlosalberto was to follow Java in this regard, i.e.: use `_ENABLED` instead. These will be very similar environment variables and it will be nice to have similar languages following the same pattern.

@carlosalberto what are your thoughts in this regard?

pellared:
> @Kielek why going with the `_DISABLED` instead of `_ENABLED` with reversed defaults? I ask because Java uses `_ENABLED` instead `_DISABLED`. See https://opentelemetry.io/docs/instrumentation/java/automatic/agent-config/#enable-only-specific-instrumentation

See: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2090

pjanotti:
@pellared I'm aware of that but I'm under the impression that at the SIG meeting, we talked with Carlos about following the Java precedent in that case.

pellared:
> @pellared I'm aware of that but I'm under the impression that at the SIG meeting, we talked with Carlos about following the Java precedent in this case.

Java is still a little different. We have not discussed `_DISABLED` vs `_ENABLED` with @carlosalberto. I lean towards spec recommendation, but I will not fight for it 😄

pjanotti:
@pellared neither I will fight for it :), but, I would prefer to commit the version that has the smaller chance of having to be reverted later.

theletterf:
So, what are we going forward with in the end? 😬

src/OpenTelemetry.AutoInstrumentation.Native/bytecode_instrumentations.h
```diff
@@ -0,0 +1,13 @@
+// Auto-generated file, do not change it - generated by the IntegrationsJsonGenerator
+#ifndef BYTECODE_INSTRUMENTATIONS_H
+#define BYTECODE_INSTRUMENTATIONS_H
+
+#include "string.h"
+
+namespace trace
+{
+inline std::unordered_map<WSTRING, WSTRING> trace_integration_names = {{WStr("StrongNamedValidation"), WStr("OTEL_DOTNET_AUTO_TRACES_StrongNamedValidation_INSTRUMENTATION_DISABLED")}, {WStr("StackExchangeRedis"), WStr("OTEL_DOTNET_AUTO_TRACES_StackExchangeRedis_INSTRUMENTATION_DISABLED")}, {WStr("NServiceBus"), WStr("OTEL_DOTNET_AUTO_TRACES_NServiceBus_INSTRUMENTATION_DISABLED")}, {WStr("MySqlData"), WStr("OTEL_DOTNET_AUTO_TRACES_MySqlData_INSTRUMENTATION_DISABLED")}, {WStr("MongoDB"), WStr("OTEL_DOTNET_AUTO_TRACES_MongoDB_INSTRUMENTATION_DISABLED")}, {WStr("GraphQL"), WStr("OTEL_DOTNET_AUTO_TRACES_GraphQL_INSTRUMENTATION_DISABLED")}};
```
rajkumar-rangaraj, line 9: Should we recommend `{0}` text to be uppercase in environmental variable? 
Most generic specification recommends uppercase, numbers and underscore.

rajkumar-rangaraj:
Even, I feel it is easier for customers to use _ENABLED in the environment variables. We could hold the merge of the PR till SIG meeting for a detailed discussion.

src/OpenTelemetry.AutoInstrumentation.Native/bytecode_instrumentations.h
Kielek: Issue already created. I do not want to change to many things in one PR #2094

pellared:
Reasons why I lean towards `_DISABLED`

1. Follows [OTel spec recommendation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#boolean-value)
2. Same approach as with [`OTEL_SDK_DISABLED`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration) and I have not found any env var in the spec which defaults to `true`
3. I think that the convention with leaning towards using `false` as the default value makes sense as it: 
   - "[Users] do not need to go look up documentation elsewhere about the value, they just know because that [`false`] is the OTel default." from https://github.com/open-telemetry/opentelemetry-specification/pull/2755#discussion_r970047515
   - simplifies discussions like this 😉

pellared (approved):

src/OpenTelemetry.AutoInstrumentation.Native/bytecode_instrumentations.h
```diff
@@ -0,0 +1,13 @@
...
+#ifndef BYTECODE_INSTRUMENTATIONS_H
+#define BYTECODE_INSTRUMENTATIONS_H
+
+#include "string.h"
+
+namespace trace
+{
+inline std::unordered_map<WSTRING, WSTRING> trace_integration_names = {{WStr("StrongNamedValidation"), WStr("OTEL_DOTNET_AUTO_TRACES_StrongNamedValidation_INSTRUMENTATION_ENABLED")}, {WStr("StackExchangeRedis"), WStr("OTEL_DOTNET_AUTO_TRACES_StackExchangeRedis_INSTRUMENTATION_ENABLED")}, {WStr("NServiceBus"), WStr("OTEL_DOTNET_AUTO_TRACES_NServiceBus_INSTRUMENTATION_ENABLED")}, {WStr("MySqlData"), WStr("OTEL_DOTNET_AUTO_TRACES_MySqlData_INSTRUMENTATION_ENABLED")}, {WStr("MongoDB"), WStr("OTEL_DOTNET_AUTO_TRACES_MongoDB_INSTRUMENTATION_ENABLED")}, {WStr("GraphQL"), WStr("OTEL_DOTNET_AUTO_TRACES_GraphQL_INSTRUMENTATION_ENABLED")}};
+inline std::unordered_map<WSTRING, WSTRING> metric_integration_names = {{WStr("NServiceBus"), WStr("OTEL_DOTNET_AUTO_METRICS_NServiceBus_INSTRUMENTATION_ENABLED")}};
+inline std::unordered_map<WSTRING, WSTRING> log_integration_names = {{WStr("ILogger"), WStr("OTEL_DOTNET_AUTO_LOGS_ILogger_INSTRUMENTATION_ENABLED")}};
```
rajkumar-rangaraj, line 11: We should separate the logs in a different namespace. This could be handled in a different PR.

Kielek: Probably metrics also. :)

Kielek: #2138

src/OpenTelemetry.AutoInstrumentation/Configurations/ConfigurationKeys.cs
```diff
@@ -189,14 +194,14 @@ public static class Logs
         public const string IncludeFormattedMessage = "OTEL_DOTNET_AUTO_LOGS_INCLUDE_FORMATTED_MESSAGE";
 
         /// <summary>
-        /// Configuration key for comma separated list of enabled logs instrumentations.
+        /// Configuration key for enabling all log instrumentations.
```
rajkumar-rangaraj, line 197: enabling or disabling? - traces and metrics has disabling and her enabling.

Kielek: Both are correct, but it is good to have same versions for all of them. Fixed aea539c6d22f5eec8be2d3e7b92b332828f4d128

PR #2128, opened 2023-01-30
Make dotnet CLI issue more visible in the docs

pjanotti:
## Why

Minimize the chances of people hitting known issues with self-contained and `dotnet` CLI in particular.

## What

Added explicit text to README.md in the `Get started` section.

## Tests

N/A.

## Checklist


dev/envvars.sh
```diff
@@ -1,16 +1,42 @@
 #!/bin/bash
 
-case "$(uname -s | tr '[:upper:]' '[:lower:]')" in
-  linux*)
-    if [ "$(ldd /bin/ls | grep -m1 'musl')" ]; then
-      DOTNET_RUNTIME_ID="linux-musl"
-    else
-      DOTNET_RUNTIME_ID="linux-x64"
-    fi
+# guess OS_TYPE if not provided
```
pjanotti, line 3: This code is the same from instrument.sh plus the initialization for OS_TYPE

pjanotti:
CI failure on mac seems infrastructure flakiness, see https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/4048439221/jobs/6963680274#step:7:10

Kielek (approved):

instrument.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 # guess OS_TYPE if not provided
+OS_TYPE=${OS_TYPE:-}
```
pellared, line 4: Why this change?

pellared (approved):

docs/README.md
```diff
@@ -76,6 +76,24 @@ See [config.md#instrumented-libraries-and-frameworks](config.md#instrumented-lib
...
+the automatic instrumentation. Current recommendations to build and launch the
+instrumented application:
+
+- Do not set the automatic instrumentation environment variables in the same session
+used to run the `dotnet` tool.
+- Do not launch the application to be instrumented using `dotnet run` or
+`dotnet <dll>`. Build the application in an isolated shell, without the
+automatic instrumentation environment variables set and use a separate
+session, with the automatic instrumentation variables, to directly launch
+the executable.
```
theletterf, line 95:
```suggestion

### Considerations on scope

Currently, instrumenting [`self-contained`](https://learn.microsoft.com/en-us/dotnet/core/deploying/#publish-self-contained)
applications is not supported. Note that a `self-contained` applications is
automatically generated in .NET 7.0 whenever the `dotnet publish` or `dotnet build`
command is used with a Runtime Identifier (RID) parameter, for example when `-r` or
`--runtime` is used when running the command. The `dotnet` CLI is a `self-contained` application and it's incompatible with automatic instrumentation.

To build and launch an instrumented application, take the following into account:

- Don't set the automatic instrumentation environment variables in the same session
used to run the `dotnet` tool.
- Don't launch the application to be instrumented using `dotnet run` or
`dotnet <dll>`. Build the application in an isolated shell, without the
automatic instrumentation environment variables set, and use a separate
session with the automatic instrumentation variables to directly launch
the executable.
```

instrument.sh
pjanotti: Ops, this was to be part of #2129 to keep the code consistent between the two scripts. Without that the script fails on `if [ -z "$OS_TYPE" ]; then`: it doesn't like that test without defining `OS_TYPE` first.

Issue #1984, opened 2023-01-12
Avoid benign GAC warnings logged by install/uninstall


pjanotti:
Running the PowerShell install/uninstall ends up logging warning events to the Windows Event Logger that are benign and not actionable by the user:

![image](https://user-images.githubusercontent.com/5600755/212187243-3603c5f7-6d8e-4acd-a040-7ed674587f6d.png)

This issue is a follow-up to #1906

pellared:
Do you have any preference on how to implement it?
I see two possibilities:

1. Use https://learn.microsoft.com/en-us/dotnet/standard/assembly/identify#using-the-pereader-class.

    Pros:
    - should not require a lot of maintenance in future

    Cons:
    - would probably not cover `netstandard.dll`. I guess this one would need special handling.
    - Would require using https://www.nuget.org/packages/System.Reflection.Metadata/ inside the powershell module

2. Have a "exclude list" which would currently be the 3 assemblies.

    Pros:
    - no dependencies and possibly best performance
    
    Cons:
    - would require more maintaince (updating the exclude list)

We could consider also writing a PowerShell unit test to check if after using the module there are no new errors or warnings in the Windows Event Log.

rajkumar-rangaraj:
It's worth exploring option 1, at a different it could help in different places too.

pellared:
I spend a few hours trying the option 1, but there are too many obstacles to make it worth in my opinion. Example problems:

- we would have to download all needed packages (together with dependencies) and then remove them at the end. See: https://github.com/OneGet/oneget/issues/475
- We would have to `Register-PackageSource` and later remove it.
- For some reasons I have failed to have a working `Install-Package` which worked fine and downloaded exactly what I wanted.

Notable article: https://endjin.com/blog/2020/12/how-to-consume-a-nuget-package-in-powershell Maybe someone else would have more luck...

Issue #1906, opened 2023-01-09
GAC - install and uninstall clarification 


Kielek:
Hi @Kielek @pellared , with [v0.5.1-beta.3](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/releases/tag/v0.5.1-beta.3), I can not reproduce this issue, even without registering netfx assemblies in GAC. 

I have 2 quick questions about registering / unregistering assemblies in GAC:
- Some assemblies below did not get added to GAC. is it because they are not strong named? any impact if they are not showing up in GAC?
  - netstandard.dll
  - grpc_csharp_ext.x64.dll
  - grpc_csharp_ext.x86.dll
- if I have an assembly already in GAC before installing dotnet auto-instrumentation, and this assembly is duplicated with an assembly under tracer-home/netfx/ folder. what will happen to it after installing dotnet auto-instrumentation? and what will happen to it after `uninstalling` auto-instrumentation? will it be removed from GAC by auto-instrumentation uninstall?

for info, I used this command to register assemblies in GAC:
```
        # Register .NET Framweworks dlls in GAC
        [System.Reflection.Assembly]::Load("System.EnterpriseServices, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a") | Out-Null
        $publish = New-Object System.EnterpriseServices.Internal.Publish 
        $dlls = Get-ChildItem -Path .\netfx\ -Filter *.dll -File
        foreach ($dll in $dlls) {
            $publish.GacInstall($dll.FullName)
        }
```
and I used command below to unregister assemblies from GAC:
```
    [System.Reflection.Assembly]::Load("System.EnterpriseServices, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a") | Out-Null
    $publish = New-Object System.EnterpriseServices.Internal.Publish 
    $dlls = Get-ChildItem -Path .\netfx\ -Filter *.dll -File
    foreach ($dll in $dlls) {
        $publish.GacRemove($dll.FullName)
    }
```
before auto-instrument install, my GAC: 
[gac.1.txt](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/files/10355048/gac.1.txt)
after auto-instrument install, my GAC:
[gac.2.txt](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/files/10355052/gac.2.txt)

[...]

PR #2164, opened 2023-02-03
Install netfx dlls to GAC for tests

### LINKED ISSUE #2155: Sync production and integration tests environment for .NET Framework
### What

Currently .NET Framework integration tests are not using GAC but the PowerShell module is installing netfx dlls to GAC. 
This fixes some assembly versioning conflicts.

example dependabot PR: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2122

RassK:
## Why

Fixes #2155

## What

Installs `tracer-home/netfx` dlls to GAC before running integration tests.

## Tests

Existing.

## Checklist


tools/GacInstallTool/Program.cs
```diff
@@ -0,0 +1,39 @@
...
+        var dir = args[0];
+        var publisher = new System.EnterpriseServices.Internal.Publish();
+        var files = Directory.GetFiles(dir, "*.dll");
+
+        foreach (var file in files)
+        {
+            publisher.GacInstall(file);
+        }
+
+        Console.WriteLine($"Success: Installed {files.Length} libraries.");
```
pjanotti, line 37: nit: 
```suggestion
        Console.WriteLine($"Success: Installed {files.Length} libraries in the GAC.");
```

tools/GacInstallTool/Program.cs
```diff
@@ -0,0 +1,39 @@
...
+            throw new ArgumentNullException(nameof(args));
+        }
+
+        var dir = args[0];
+        var publisher = new System.EnterpriseServices.Internal.Publish();
+        var files = Directory.GetFiles(dir, "*.dll");
+
+        foreach (var file in files)
+        {
+            publisher.GacInstall(file);
```
pjanotti, line 34: This API can fail silently, it is worth commenting about it.

pjanotti (commented):
@Rassk What do you think about providing an uninstall target? It seems useful during development. While we don't need it in CI it helps keep the dev box clean.

RassK:
> @RassK What do you think about providing an uninstall target? It seems useful during development. While we don't need it in CI it helps keep the dev box clean.

I really wanted to implement it but I think it's not that straightforward. I'm a bit afraid to remove something that overlaps and breaks other applications. To fix that I would need to save a state that tracks what's new and could be removed... although even this is valid only short time.

build/Build.Steps.Windows.cs
```diff
@@ -177,4 +179,20 @@ partial class Build
 
             AssemblyRedirectionSourceGenerator.Generate(netFxAssembliesFolder, generatedSourceFile);
         });
+
+    Target InstallNetFxAssembliesGAC => _ => _
+        .Unlisted()
+        .Before(RunManagedTests)
```
Kielek, line 185: Could you please convert this condition to `.After(InstallNetFxAssembliesGAC )` on `RunManagedTests`?

tools/GacInstallTool/GacInstallTool.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net462</TargetFramework>
```
Kielek, line 5: As it is fully internal tool, we could target it to `net48` version?

build/Build.Steps.Windows.cs
```diff
@@ -177,4 +179,20 @@ partial class Build
 
             AssemblyRedirectionSourceGenerator.Generate(netFxAssembliesFolder, generatedSourceFile);
         });
+
+    Target InstallNetFxAssembliesGAC => _ => _
```
Kielek, line 183: If I execute `nuke` or `nuke Workflow` with elevated privileges it will overwrite GAC,
It will be great to have reverse step included into it, or at least provide easy way to remove all this files.

I think `nuke Workflow CleanupGAC` execution form is fine.

Kielek (commented):
I think that we should update documentation that instrumenting .NET Fx stuff on Windows requires GAC update (done by psm1 file).
I do not think that it is obvious https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/README.md?plain=1#L79-L211

build/Build.Steps.Windows.cs
pellared: I would prefer to have it as part of `nuke` or `nuke Workflow`

pellared:
> I really wanted to implement it but I think it's not that straightforward. I'm a bit afraid to remove something that overlaps and breaks other applications. To fix that I would need to save a state that tracks what's new and could be removed... although even this is valid only short time.

@pjanotti was testing it and it will not remove the assemblies from GAC that were added by MSIs. See https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1906#issuecomment-1376292814

tools/GacInstallTool/Program.cs
RassK: Any way to catch some warnings etc? since it's a .net framework app, it probably can integration into another windows built in service that reports those errors?

build/Build.Steps.Windows.cs
RassK: There are 2 ways which might work according to your own risk.
1. Add CleanupGAC to the workflow and a warning to use it on your own risk or skip the step.
2. Add CleanupGAC as a non referenced step. You need to run it yourself manually.

tools/GacInstallTool/Program.cs
pellared: I think you would have to check for Windows Event logs 😬 
See: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1984

build/Build.Steps.Windows.cs
pellared: Probably point 2 is better/safer after all you said 😉 
Make it could somehow part of `nuke Cleanup`?

tools/GacInstallTool/Program.cs
RassK: Seems doable > 
https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.eventlog.entrywritten?redirectedfrom=MSDN&view=netframework-4.6.1

build/Build.Steps.Windows.cs
RassK: Found better place for the `After` method.

tools/GacInstallTool/Program.cs
pellared: I think @pjanotti just suggested to give a comment in code. We can improve this later.

build/Build.Steps.Windows.cs
```diff
@@ -177,4 +179,31 @@ partial class Build
...
+    /// <remarks>
+    /// Warning: This target could cause potential harm to your system by removing a required library from GAC.
+    /// </remarks>
+    Target UninstallNetFxAssembliesGAC => _ => _
+        .Unlisted()
+        .After(BuildTracer)
+        .OnlyWhenStatic(() => IsWin)
+        .Executes(() => RunNetFxGacOperation(install: false));
+
+    private void RunNetFxGacOperation(bool install)
```
pellared, line 198: nit: I would rather pass `string operation` that would be `"-i"` or `"-u"`. I do not like bool parameters😉 No need to address it.

build/Build.Steps.Windows.cs
```diff
@@ -177,4 +179,31 @@ partial class Build
...
+        .Unlisted()
+        .After(BuildTracer)
+        .OnlyWhenStatic(() => IsWin)
+        .Executes(() => RunNetFxGacOperation(install: true));
+
+    /// <remarks>
+    /// Warning: This target could cause potential harm to your system by removing a required library from GAC.
+    /// </remarks>
+    Target UninstallNetFxAssembliesGAC => _ => _
+        .Unlisted()
```
pellared, line 193: It must have description and be listed😉

Kielek (approved):

pjanotti (approved):

build/Build.Steps.Windows.cs
pjanotti: @pellared should we list everything always or there is some criteria to unlist some targets?

PR #2171, opened 2023-02-06
Improve GAC registration in PowerShell module

pellared:
## Why

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/1984

## What

- Do not install/uninstall DLLs which cannot be registered in GAC
- Fix the condition in the loop used for enumarating dlls
- Add progress bar for uninstalllation
- Remove the progress bar once installation/uninstallation is completed

## Tests

Manual. I was considering to automate it, but I resisted as I think it is not worth the cost (as we only get warnings in Event Logs if we miss something). I created an issue anyway https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2172

## Checklist

- [x] ~~`CHANGELOG.md` is updated.~~ I do not think it is worth mentioning.
- [x] ~~New features are covered by tests.~~ _See above._

### DIFF
--- a/OpenTelemetry.DotNet.Auto.psm1
+++ b/OpenTelemetry.DotNet.Auto.psm1
@@ -199,6 +199,15 @@ function Get-OpenTelemetry-Archive([string] $Version, [string] $LocalPath) {
     return Download-OpenTelemetry $Version $tempDir
 }
 
+function Test-AssemblyNotForGAC([string] $Name) {
+    switch ($Name) {
+        "netstandard.dll" { return $true }
+        "grpc_csharp_ext.x64.dll" { return $true }
+        "grpc_csharp_ext.x86.dll" { return $true }
+    }
+    return $false 
+}
+
 <#
     .SYNOPSIS
     Installs OpenTelemetry .NET Automatic Instrumentation.
@@ -246,13 +255,7 @@ function Install-OpenTelemetryCore() {
                 -Status "Module $($i+1) out of $($dlls.Count). Installing $($dlls[$i].Name):" `
                 -PercentComplete $percentageComplete
 
-            if ($dlls[$i].Name -eq "netstandard.dll") {
-                continue
-            }
-            if ($dlls[$i].Name -eq "grpc_csharp_ext.x86.dll") {
-                continue
-            }
-            if ($dlls[$i].Name -eq "grpc_csharp_ext.x64.dll") {
+            if (Test-AssemblyNotForGAC $dlls[$i].Name) {
                 continue
             }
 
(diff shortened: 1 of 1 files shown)


OpenTelemetry.DotNet.Auto.psm1
```diff
@@ -240,13 +240,25 @@ function Install-OpenTelemetryCore() {
...
         $publish = New-Object System.EnterpriseServices.Internal.Publish 
         $dlls = Get-ChildItem -Path $installDir\netfx\ -Filter *.dll -File
-        for ($i = 0; $i -le $dlls.Count; $i++) {
+        for ($i = 0; $i -lt $dlls.Count; $i++) {
             $percentageComplete = $i / $dlls.Count * 100
             Write-Progress -Activity "Registering .NET Framweworks dlls in GAC" `
                 -Status "Module $($i+1) out of $($dlls.Count). Installing $($dlls[$i].Name):" `
                 -PercentComplete $percentageComplete
+
+            if ($dlls[$i].Name -eq "netstandard.dll") {
```
RassK, line 249: maybe just to define ignore list and check dll name not in the list? prob scales better in the future.

Kielek: and we can have it in one place, instead of two

pjanotti: Whenever a new assembly gets added we need to review the list.

xiang17: It looks like the explorations in #1984 pointed a good direction in using C# to determine whether a DLL is an assembly, and the blockers revolves around having to download/unload dependencies in PowerShell and having to do C# things in PowerShell.

I have an idea to automate it which combines the option 1 and option 2: if we generate the "ignore list" or "allow list" in test projects like in #1156 or build projects like in #2073, there is no need to worry about having to load/unload C# packages in PowerShell module.

The list can be generated in a text file or directly as a few lines of HashMap initialization in the `OpenTelemetry.DotNet.Auto.psm1` file itself to avoid having to read the text file from PowerShell, kind of like the auto-generated block "Transient dependencies auto-generated by GenerateNetFxTransientDependencies" in the `OpenTelemetry.AutoInstrumentation.csproj` file.

pellared: I will address it in an "easy" way first. https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2171#discussion_r1097978479 is a good idea for a separate PR. @xiang17 do you want to try it?

pellared: Addressed https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2171/commits/692d9cb1b73035575ae23c43ee524760589c0aba

Kielek (approved):

xiang17 (approved):
LGTM

OpenTelemetry.DotNet.Auto.psm1
xiang17: I got a few higher priority tasks recently. I might not have time for it until later.

Issue #2098, opened 2023-01-28
Remove Microsoft.Bcl.AsyncInterfaces library from auto-instrumentation package


xiang17:
# Bug Report

## Symptom

**Describe the bug**

`Microsoft.Bcl.AsyncInterfaces.dll` library is present in the `tracer-home` folder.

**Expected behavior**

The DLL file should NOT be present. Customer would bring in this file if they depend on it.

This library is brought in by `StackExchange.Redis/2.1.58`, which is brought in by `OpenTelemetry.Instrumentation.StackExchangeRedis/1.0.0-rc9.7`, etc.

**Screenshots**

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/2b73fc435ad8706120c56a67637288fac7d3b0ac/test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt#L10

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/2b73fc435ad8706120c56a67637288fac7d3b0ac/test/IntegrationTests/BuildTests.DistributionStructure_linux.verified.txt#L11

And a few more.

<!--
NOTE: If possible, also include the log files found (by default) in:
Windows: %ProgramData%\OpenTelemetry .NET AutoInstrumentation\logs\
Linux and macOS: /var/log/opentelemetry/dotnet/
--!>

xiang17:
I will work on this.

PR #2190, opened 2023-02-10
Dependabot bumps

Kielek:
## Why

Cleanup PRs.

## What

Update all transitive dependencies.

## Tests

CI + local nuke Workflow on Windows with Linux Docker

## Checklist


src/Directory.Packages.props
```diff
@@ -28,7 +28,7 @@
 
   <!-- Versions from CommonExcludedAssets.props -->
   <ItemGroup>
-    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.1" />
+    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="7.0.0" />
```
Kielek, line 31: `System.Text.Json` requires 7.0.0+ version.

Should we really avoid shipping it? Ref #2135 and #2098 

@xiang17, @rajkumar-rangaraj?

xiang17: We can have dependency on it, upgrade the version as needed. The `CommonExcludedAssets`'s purpose is to exclude them from being included in the binaries we release.

To my understanding, we exclude them because the customer who has dependency will bring in the assemblies themselves.

Kielek: I understand the reason of `CommonExcludedAssets` - and it should be valid only for dependencies which are only referenced by MySQL.Data and StackExchange.Redis. If it is transitive dependency also by another library we should ship it.
I think that we should not treat `Microsoft.Bcl.AsyncInterfaces` as this kind of assets. The main reason - we ship the latest version of `System.Text.Json`. If the user will reference older version `Microsoft.Bcl.AsyncInterfaces` it can leads to unexpected scenarios.

pellared (approved):

src/Directory.Packages.props
RassK: I think all of the netfx libs that are upgraded to the latest must be shipped. There is a very high chance that customer does not bring the required dependency.

Kielek: Nvm. This lib is referenced only for .NET 4.6.2.
Now I have only some doubts where to keep it. I am adding some comment related to this state.

pjanotti: @Kielek good point: the exclusion of some libraries depends on the targeted framework. Ideally, we should do the exclusion according to that. Anyway, related but separate problem.

pjanotti (approved):

pjanotti:
@RassK could you clarify what is the issue with Google.Protobu?

RassK:
> @RassK could you clarify what is the issue with Google.Protobu?

No issues, just it's currently perfect to test bumping issues and https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2192

Kielek:
You can test anything you want on your fork. Including dependabot. Just to revert for some older commit

RassK:
> You can test anything you want on your fork. Including dependabot. Just to revert for some older commit

Nvm, it was already decided to go with manual option. No further blocking.

RassK (approved):

PR #2192, opened 2023-02-13
Remove netfx_assembly_redirection.h from Git

RassK:
## Why

Follow-up to https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2164

## What

Removes `netfx_assembly_redirection.h` from Git as it's generated during build.

## Tests

Existing.


Kielek:
It is generated in build pipeline as few other files.
My preference is to keep it as it is:
- you know exactly what you are using while compiling,
- clear history with redirection,
- potentially with this we can avoid some security concerns related to compilation of not-committed code.

The only issue I see with this file - dependabot could work a but smoother.

RassK:
> It is generated in build pipeline as few other files. My preference is to keep it as it is:
> 
> * you know exactly what you are using while compiling,
> * clear history with redirection,
> * potentially with this we can avoid some security concerns related to compilation of not-committed code.
> 
> The only issue I see with this file - dependabot could work a but smoother.

This header is only for `netfx`. All `netfx` dependencies are already tracked, there is no need for secondary verification.
This header is rendering dependabot completely unuseful, that means a lot of manual work, since we track quite many deps.

Kielek:
Please check for example `System.ComponentModel.Annotations` in #2192. Version is different (and it is expected) in props and .h file. And the difference is not 5.1.2.3 in props and 5.0.0.0 in .h (which could be common).

We tracking different things in these files.

RassK:
> Please check for example `System.ComponentModel.Annotations` in #2190. Version is different (and it is expected) in props and .h file. And the difference is not 5.1.2.3 in props and 5.0.0.0 in .h (which could be common).
> 
> We tracking different things in these files.

Ok, bit overlapping opinions how to go forward. I leave others to decide. 

1. Leave as is - all conflicting netfx dependabot PRs must be manually checked out and updated, but the history remains in git.
2. Make the change, automated dependabot PRs work in case of non-connected dependencies. Lose `netfx_assembly_redirection.h` tracking but build is still ensured by the tests.

pjanotti:
@RassK, @Kielek point is something that I missed when we talked about this: the h file tracks assembly versions. While package and assembly versions typically match there is no guarantee that they will. We can't replace the h file with the package versions.

Closing this PR - my mistake for missing that @RassK.

Issue #1213, opened 2022-09-09
Testing - Support for different versions of libraries 


Kielek:
# Feature Request

**Are you requesting automatic instrumentation for a framework or library? Please describe.**

We need a way to execute our pipeline with different versions of libraries.
For example, we should execute `StackExchange.Redis` tests with at least 4 different versions. To fully cover all cases. See https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/4ac185f24b92e8c6515b7a0ae8546441df43b353/src/OpenTelemetry.AutoInstrumentation/Instrumentations/StackExchangeRedis/StackExchangeRedisIntegration.cs#L27-L54

**Describe the solution you'd like**
TBD

**Describe alternatives you've considered**
TBD

**Additional Context**
Discussion started at https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1171

pellared:
I think it would be easier if the integration test would build and run the instrumented application. What I mean is that I would move the code that builds the test application from Nuke target to the integration tests.

Kielek:
Leftovers from https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2241#pullrequestreview-1310937584

Asp.NET MVS - separate issue created #2245
HttpClient - only the latest release for .NET Fx is not deprecated. Nothing to extend for now.

Closing.

PR #2230, opened 2023-02-20
[IntegrationTests] Support for testing multiple nuget package versions

Kielek:
## Why

Towards #1213 

## What

Allow to execute tests against multiple versions.
Tests for StackExchange.Redis and MassTransit.

Set of versions - hardcoded list + `*` taken from props file.

If executed from VS it uses version from props file.

Follow ups
 - cover other test applications.
 - Potentially add nuget.org check if there is no new minor for the old version release which should be included into set.

## Tests

CI

## Checklist


.github/workflows/ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           dotnet publish -f net7.0 -c Release ./test/test-applications/integrations/TestApplication.Smoke
           export OTEL_DOTNET_AUTO_HOME="${PWD}/bin/tracer-home"
           . ./instrument.sh
-          ./test/test-applications/integrations/TestApplication.Smoke/bin/Release/net7.0/publish/TestApplication.Smoke
+          ./test/test-applications/integrations/bin/TestApplication.Smoke/Release/net7.0/publish/TestApplication.Smoke
```
RassK, line 56: Isn't it going to overwrite the same dir contents always? 

upstream was using `./test/test-applications/integrations/{Application}/bin/{Config}/{Framework}/{PackageVersion}/` pattern, this seems also more natural to me.

test/test-applications/integrations/TestApplication.MongoDB/TestApplication.MongoDB.csproj
```diff
@@ -5,23 +5,19 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <ApiVersion Condition="'$(ApiVersion)' == ''">2.13.3</ApiVersion>
-    <DefineConstants Condition="'$(ApiVersion)'>='2.15.0'">$(DefineConstants);MONGODB_2_15</DefineConstants>
+    <TestedPackageVersion Condition="'$(TestedPackageVersion)' == ''">2.13.3</TestedPackageVersion>
```
RassK, line 8: just `PackageVersion` instead of `TestedPackageVersion` ?

Kielek: `PackageVersion` is used by https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/3d3f5c395ac6eb0a2d774f25c87c9f74b2a8f266/test/Directory.Packages.props#L4
Even if it is tenically possible, I would like to avoid colision name.

I am open for the better naming, but I do not like `ApiVersion`.

.github/workflows/ci.yml
Kielek: I have to change default `obj` and `bin` folder location to be able to compile test application with multiple librariyversions.
The reason - dotnet creates `cs` files under the bin folder. When I have tried to use
`./test/test-applications/integrations/{Application}/bin/{Config}/{Framework}/{PackageVersion}/` for version 1 and version 2. 1st complilation was passing, but the second one was failing due to duplicates attribiutes in this `cs`
 files.

See https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/0182a430d13cd129a429daa4655319e3585b4ff6/test/test-applications/integrations/Directory.Build.props#L6-L9

build/Build.Steps.cs
```diff
@@ -134,11 +150,26 @@ partial class Build
 
             foreach (var app in testApps)
             {
-                DotNetBuild(x => x
-                    .SetProjectFile(app)
-                    .SetConfiguration(BuildConfiguration)
-                    .SetPlatform(Platform)
-                    .SetNoRestore(true));
+                if (TestPackageVersions.Versions.TryGetValue(app.Name, out var testedPackageVersions))
```
pjanotti, line 153: If feels like this should be generated during the build of our `_build.csproj` since we are using the compiled version not the one generated during the workflow itself. It may make sense overall, but, at first it seems strange.

build/Build.Steps.cs
```diff
@@ -66,15 +66,43 @@ partial class Build
...
+            {
+                if (TestPackageVersions.Versions.TryGetValue(project.Name, out var testedPackageVersions))
+                {
+                    foreach (var testedPackageVersion in testedPackageVersions)
+                    {
+                        DotNetRestore(s => s
+                            .SetProjectFile(project)
+                            .SetVerbosity(DotNetVerbosity.Normal)
+                            .SetProperty("configuration", BuildConfiguration.ToString())
+                            .SetProperty("TestedPackageVersion", testedPackageVersion)
```
pjanotti, line 86: Just calling attention that this means that only a single package can "rotate" versions on the test app. It is reasonable, if there is a need to test different packages in a single app it is not hard to adapt the code to do that.

build/Build.Steps.cs
```diff
@@ -66,15 +66,43 @@ partial class Build
...
+                    }
+                }
+                else
+                {
+                    DotNetRestore(s => s
+                            .SetProjectFile(project)
+                            .SetVerbosity(DotNetVerbosity.Normal)
+                            .SetProperty("configuration", BuildConfiguration.ToString())
+                            .SetPlatform(Platform)
+                            .When(!string.IsNullOrEmpty(NugetPackageDirectory), o =>
```
pjanotti, line 99: I _think_ that this could be done with a `.CombineWith` to avoid the `if` statement and the almost identical branches.

pjanotti: I mean not this specific line but the `DotNetRestore` itself.

build/Build.Steps.cs
```diff
@@ -134,11 +150,26 @@ partial class Build
...
-                    .SetPlatform(Platform)
-                    .SetNoRestore(true));
+                if (TestPackageVersions.Versions.TryGetValue(app.Name, out var testedPackageVersions))
+                {
+                    foreach (var testedPackageVersion in testedPackageVersions)
+                    {
+                        DotNetBuild(x => x
+                            .SetProjectFile(app)
+                            .SetConfiguration(BuildConfiguration)
+                            .SetProperty("TestedPackageVersion", testedPackageVersion)
```
pjanotti, line 160: Same suggestion of trying to use `.CombineWith` to avoid the branching.

build/Build.Steps.cs
```diff
@@ -66,15 +66,43 @@ partial class Build
...
-                    .SetVerbosity(DotNetVerbosity.Normal)
-                    .SetProperty("configuration", BuildConfiguration.ToString())
-                    .When(!string.IsNullOrEmpty(NugetPackageDirectory), o =>
-                        o.SetPackageDirectory(NugetPackageDirectory)));
+                projectsToRestore = projectsToRestore.Concat(Solution.GetWindowsOnlyTestApplications());
+            }
 
+            foreach (var project in projectsToRestore)
+            {
+                if (TestPackageVersions.Versions.TryGetValue(project.Name, out var testedPackageVersions))
```
pjanotti, line 78: As we add this to more projects we may want it disabled by default and run the test against a single version. Anyway, I'm unsure of the current impact.

test/test-applications/integrations/dependency-libs/TestApplication.Shared/TestApplication.Shared.csproj
```diff
@@ -6,8 +6,4 @@
     <StartupObject />
   </PropertyGroup>
```
pjanotti, line 8: Which projects still need this `TestApplication.Shared`?

tools/TestedPackageVersionsGenerator/CSharpFileBuilder.cs
```diff
@@ -0,0 +1,72 @@
...
+    internal StringBuilder Builder { get; } = new();
+
+    public CSharpFileBuilder AddAutoGeneratedHeader()
+    {
+        Builder.AppendLine(@"//------------------------------------------------------------------------------
+// <auto-generated>
+//     This code was generated by the TestedPackageVersionsGenerator tool. To safely
+//     modify this file, edit PackageVersionDefinitions.cs and
+//     re-run the TestedPackageVersionsGenerator project in Visual Studio.
+
```
pjanotti, line 32: nit: prefix all header lines with `//`

pjanotti (commented):
Looks good to me, small suggestions and a few Qs.

build/Build.Steps.cs
```diff
@@ -66,15 +66,43 @@ partial class Build
...
+
             if (IsWin)
             {
-                DotNetRestore(s => s
-                    .SetProjectFile(Solution)
-                    .SetVerbosity(DotNetVerbosity.Normal)
-                    .SetProperty("configuration", BuildConfiguration.ToString())
-                    .When(!string.IsNullOrEmpty(NugetPackageDirectory), o =>
-                        o.SetPackageDirectory(NugetPackageDirectory)));
+                projectsToRestore = projectsToRestore.Concat(Solution.GetWindowsOnlyTestApplications());
```
xiang17, line 73: What about projects that are not in `GetWindowsOnlyTestApplications` nor `GetCrossPlatformManagedProjects`, for example native source projects?

test/test-applications/integrations/dependency-libs/TestApplication.Shared/TestApplication.Shared.csproj
xiang17: Same question. The methods in this project are still getting called in other projects, e.g. `ConsoleHelper.WriteSplashScreen` is called in `TestApplication.CustomSdk`. I'm surprised that the CI passes despite of this.

test/test-applications/integrations/TestApplication.StackExchangeRedis/Program.cs
```diff
@@ -57,8 +56,7 @@ public static async Task Main(string[] args)
 
             db.Ping();
         }
-
-        // SentinelConnect and SentinelConnectAsync introduced in 2.1.50
+#if !REDIS_2_1_50_OR_LOWER
```
zacharycmontoya, line 59: nit: Following the convention for .NET compiler directives, I prefer renaming this to `REDIS_2_1_50_OR_GREATER` so the directives seem like they're adding code paths for newer versions

test/test-applications/integrations/TestApplication.MongoDB/TestApplication.MongoDB.csproj
zacharycmontoya: What about `LibraryVersion`?

test/test-applications/integrations/TestApplication.StackExchangeRedis/Program.cs
zacharycmontoya: Also in this case the code wouldn't run if the library version was 2.1.50, even though I think it should

test/test-applications/integrations/Integrations.props
```diff
@@ -9,16 +9,14 @@
     <PlatformTarget>$(Platform)</PlatformTarget>
     <LangVersion>10.0</LangVersion>
 
-    <BaseIntermediateOutputPath Condition="'$(ApiVersion)'!=''">obj\$(ApiVersion)\</BaseIntermediateOutputPath>
-    <BaseOutputPath Condition="'$(ApiVersion)'!=''">bin\$(ApiVersion)\</BaseOutputPath>
+    <BaseIntermediateOutputPath Condition="'$(TestedPackageVersion)'!=''">..\obj\$(MSBuildProjectName)\$(TestedPackageVersion)\</BaseIntermediateOutputPath>
+    <BaseIntermediateOutputPath Condition="'$(TestedPackageVersion)'==''">..\obj\\$(MSBuildProjectName)\</BaseIntermediateOutputPath>
```
zacharycmontoya, line 13:
```suggestion
    <BaseIntermediateOutputPath Condition="'$(TestedPackageVersion)'==''">..\obj\$(MSBuildProjectName)\</BaseIntermediateOutputPath>
```

tools/TestedPackageVersionsGenerator/PackageVersionDefinitions.cs
```diff
@@ -0,0 +1,49 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+namespace TestedPackageVersionsGenerator;
+
+internal static class PackageVersionDefinitions
+{
+    public static IReadOnlyCollection<PackageVersionDefinition> Definitions => new List<PackageVersionDefinition>
```
zacharycmontoya, line 21: In this same PR, can you add a README somewhere that directs people to this file, if a library needs to be tested against multiple versions?

tools/TestedPackageVersionsGenerator/PackageVersionDefinition.cs
```diff
@@ -0,0 +1,28 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+namespace TestedPackageVersionsGenerator;
+
+internal record PackageVersionDefinition
```
zacharycmontoya, line 19: It's might be confusing that this file is named `PackageVersionDefinition` and the other one is `PackageVersionDefinitions`. I don't have a rename in mind though, but I thought I would share my first impression

build/Build.Steps.cs
Kielek: It is good candidate for improvement.
I have a branch with additional changes related to cover all other our library-depended test application by this tool.
The results 
image

And the results for the latest main execution:
image

The problem is probably ASP.NET core build/.NET Framework, not the few more test cases.

test/test-applications/integrations/dependency-libs/TestApplication.Shared/TestApplication.Shared.csproj
Kielek: There is not traditional reference to this project, but still all `TestApplication`s are referencing it.
See  6a1fc5b04366ca789e5e9452315fc7de5e7337d8 for details.

The reason - I was not able to execute restore/build process for multi-lib scenarios with this dependency.
Shared lib was build a lot of times.
It can be improved in the future if needed.

test/test-applications/integrations/TestApplication.MongoDB/TestApplication.MongoDB.csproj
Kielek: Fixed in b85b07a9c43c743d1c442b679d963f7f57da6b1b
Technically LibVersion can be different than PackageVersion but it is probably better naming.

tools/TestedPackageVersionsGenerator/PackageVersionDefinition.cs
Kielek: Fixed in 698e9eb974bcfa6002fd2eb2e39e985886f379f1

Now we have one file with internal class only.

test/test-applications/integrations/Integrations.props
Kielek: c8c8a20215ac405d3341e2658c1b03b3372056c6

tools/TestedPackageVersionsGenerator/CSharpFileBuilder.cs
Kielek: 87219384f3b6553df01eecc5e00e9dc3b2c3a211

build/Build.Steps.cs
Kielek: Nothing change here. See lines 104-120.

build/Build.Steps.cs
Kielek: I am not sure if I understand this comment correctly but I have added generation tool to the build pipeline to ensure that there is no missing generated files in the repository.

test/test-applications/integrations/TestApplication.StackExchangeRedis/Program.cs
Kielek: Thanks!
Fixed f5c61a756afe85e9f76bed6c2e8d085d64103502

build/Build.Steps.cs
Kielek: ba965160f94c1aad4e45de4f84e3987069e0d399

build/Build.Steps.cs
Kielek: I do not find the way to fully remove branching (When is not working well with CobineWith - due to possible return type differences), but I some improvements was possible. ba965160f94c1aad4e45de4f84e3987069e0d399

tools/TestedPackageVersionsGenerator/PackageVersionDefinitions.cs
Kielek: I have follow up PR in progress with other library versions included to pipeline. I will put documentation there.

test/test-applications/integrations/dependency-libs/TestApplication.Shared/TestApplication.Shared.csproj
RassK: Could change the csproj project type.

Did not find better docs reference for this feature. Basically a project type for shared code and not compiled.
https://learn.microsoft.com/en-us/xamarin/cross-platform/app-fundamentals/shared-projects?tabs=windows

Kielek: Shared projects based on old style files. The current approach is used by .NET SDK/Contrib repository

```xml
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <PropertyGroup Label="Globals">
    <ProjectGuid>30648dda-7487-4754-b6aa-a9a0c87a57f7</ProjectGuid>
    <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
  </PropertyGroup>
  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.Common.Default.props" />
  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.Common.props" />
  <PropertyGroup />
  <Import Project="SharedProject1.projitems" Label="Shared" />
  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.CSharp.targets" />
</Project>
```

.github/workflows/ci.yml
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2240

RassK (approved):
Seems excellent base, some cleanups extracted and not blocking this stage.

build/Build.Steps.cs
pjanotti: ci will detect if the file changed so we are covered, but, `dotnet nuke Workflow` will run with the source file that was checked in, i.e.: before the targeting generating/updating the source was run. It should work, it is just somehow unexpected (at least to me).

build/Build.Steps.cs
pjanotti: Windows is already pretty bad independent of the current change.

build/Build.Steps.cs
Kielek: Well we can rearrange pipeline a bit, but it looks like a bigger rework.
Restore, build and execute generator before restore/build tests.
It is doable, if you are fine, I would create separate issue.

pjanotti (approved):

PR #2231, opened 2023-02-21
[CI] verify native headers

### LINKED ISSUE #416: Add license headers for C++ source files
Shouldn't we do the same for native C++ code? 

Another issue and PR if so. For reference, I guess there may be some automation here: https://github.com/open-telemetry/opentelemetry-cpp. Personally, I am aware of https://github.com/open-telemetry/opentelemetry-go/blob/b1c1e781e6df6ac8a280da21766a00d16a45a9ed/Makefile#L171-L179

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/415#pullrequestreview-913150243_

lachmatt:
## Why

Fixes #416 

## What

Add job to workflow to verify all native source files have license headers.

## Checklist

### DIFF
--- a/.github/workflows/format-native.yml
+++ b/.github/workflows/format-native.yml
@@ -30,7 +30,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3.3.0
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v3.5.0
         with:
           go-version: '1.20.1'
       - name: Verify all native files have license headers
--- a/.github/workflows/format-native.yml
+++ b/.github/workflows/format-native.yml
@@ -24,3 +24,20 @@ jobs:
     - name: Format native code
       shell: bash
       run: ./scripts/format-native.sh
+
+  check-native-headers:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v3.3.0
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '1.20.1'
+      - name: Verify all native files have license headers
+        shell: bash
+        run: |
+          go install github.com/google/addlicense@v1.1.1
+          addlicense -v -y "" -l "apache" -c "The OpenTelemetry Authors" -s=only -ignore **/lib/** src/OpenTelemetry.AutoInstrumentation.Native/
+          git status
+          git diff
+          test -z "$(git status --porcelain)"


Kielek (approved):

.github/workflows/format-native.yml
```diff
@@ -24,3 +24,20 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v3.3.0
+      - uses: actions/setup-go@v3.5.0
+        with:
+          go-version: '1.20.1'
+      - name: Verify all native files have license headers
+        shell: bash
+        run: |
+          go install github.com/google/addlicense@v1.1.1
```
pellared, line 39: Later (as a seperate PR) I can improve it in a way to have dependabot bumps for `addlicense`.

Question. Do you prefer to use a container image or use Go to build the binary?

.github/workflows/format-native.yml
```diff
@@ -24,3 +24,20 @@ jobs:
...
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v3.3.0
+      - uses: actions/setup-go@v3.5.0
+        with:
+          go-version: '1.20.1'
+      - name: Verify all native files have license headers
+        shell: bash
+        run: |
```
pellared, line 38: Can you check if `run` is not retruning only the status of the last command?
I mean I think that if `go install github.com/google/addlicense@v1.1.1` fails and `test -z "$(git status --porcelain)"` passed then probably the job will be marked as passed.

How about moving this code to a `.sh` script like I have done here: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/.github/workflows/format-native.yml

It would also make it easier for local usage.

.github/workflows/format-native.yml
```diff
@@ -24,3 +24,20 @@ jobs:
...
       run: ./scripts/format-native.sh
+
+  check-native-headers:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v3.3.0
+      - uses: actions/setup-go@v3.5.0
+        with:
+          go-version: '1.20.1'
```
pellared, line 35: Do you prefer to use a container image or use Go to build the binary? I think that having a script that uses a container image would be easier for local usage.

.github/workflows/format-native.yml
lachmatt: > Can you check if `run` is not retruning only the status of the last command?
> I mean I think that if `go install github.com/google/addlicense@v1.1.1` fails and `test -z "$(git status --porcelain)"` passed then probably the job will be marked as passed.

My testing on fork indicates that failure for any command will fail the step and the job.

> How about moving this code to a `.sh` script like I have done here: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/.github/workflows/format-native.yml
> 
> It would also make it easier for local usage.

Sounds good, let me know if you'd like to do that before merging.

pellared: > Can you check if run is not returning only the status of the last command?

If fails the step at the first failing command.

I would still prefer to have it as a script, but maybe it would make more sense to refactor to a Nuke target later... For now, I am OK to keep it as it is.

PR #2283, opened 2023-03-02
docs: Compatibility update

pellared:
Update the compatibility docs.

- .NET Core 3.1 is already out of support for some time. If we want to keep the info which was the last supported version then probably we should do the same for .NET 5. Anyway all this info is in changelog.
- Document that we test against CentOS 
- Rephrase what .NET and .NET Framework versions we support.


Kielek (approved):

pjanotti (commented):
👍

Issue #1382, opened 2022-10-07
Remove indirect dependencies from additional store


Kielek:
The change itself looks good. 

I do have a question about the dependencies that are being added to the additional deps. Aren't they dependencies of the library being instrumented? It seems that they would be already present on the application it self. I'm assuming that instrumentation helper/middleman indirectly depends on those. Is that correct?

Anyway, the change looks good. This question of indirect dependencies, if real, can and should be addressed separately.

_Originally posted by @pjanotti in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1320#pullrequestreview-1134854978_

The same question under: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1320#discussion_r990367416

----

We should probably remove indirect dependencies like in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/src/CommonExcludedAssets.props

Kielek:
Done in scope of #1829

RassK:
Reopening. Discovered that we don't need to ship MongoDB transient packages. 
Some more scans must be made to understand the missing scope (mostly issue with packages that have multi layer transient packages).

RassK:
@rajkumar-rangaraj do you know if there are any straightforward ways to remove transient deps for specific package? ... or do you think this is entirely manual job? 

The MongoDB case is the following:
* `MongoDB.Driver.Core.Extensions.DiagnosticSources` is the main package, **keep**
* `MongoDB.Driver.Core` is transient package of `MongoDB.Driver.Core.Extensions.DiagnosticSources`, **delete**
* `DnsClient` is transient package of `MongoDB.Driver.Core`, **delete**
* `MongoDB.Bson` is transient package of `MongoDB.Driver.Core`, **delete**
* `MongoDB.Libmongocrypt`  is transient package of `MongoDB.Driver.Core`, **delete**
* `SharpCompress` is transient package of `MongoDB.Driver.Core`, **delete**

rajkumar-rangaraj:
There are no easy ways to remove it. We need to customize in build steps.

Kielek:
Moving to post 1.0.0 release. Most important parts was removed while working on MognoDB task.
Further optimization can be handled after GA.

RassK:
irrelevant after https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4783

PR #2285, opened 2023-03-02
Use System.Text.Json.Nodes instead of Regex

RassK:
## Why

Towards #1382 

## What

Use writable json tools (System.Text.Json.Nodes) instead of Regex.
This makes future work with json content simpler and less error prone. 

## Tests

Existing.

## Next

Need to analyse packages we instrument. If dependencies contents is the same as instrumented library is already shipping, then we can remove common part from our package.

### DIFF
--- a/build/Build.Steps.cs
+++ b/build/Build.Steps.cs
@@ -475,8 +475,7 @@ partial class Build
                         var rawJson = File.ReadAllText(file);
                         var depsJson = JsonNode.Parse(rawJson).AsObject();
 
-                        var folderRuntimeName = GetFolderRuntimeName(depsJson);
-
+                        var folderRuntimeName = depsJson.GetFolderRuntimeName();
                         var architectureStores = new List<string>
                         {
                             Path.Combine(StoreDirectory, "x64", folderRuntimeName),
@@ -505,19 +504,6 @@ partial class Build
                 AdditionalDepsDirectory.GlobFiles("**/*.dll", "**/*.pdb", "**/*.xml", "**/*.dylib", "**/*.so").ForEach(DeleteFile);
                 AdditionalDepsDirectory.GlobDirectories("**/runtimes").ForEach(DeleteDirectory);
             });
-
-        string GetFolderRuntimeName(JsonObject jsonDocument)
-        {
-            var runtimeName = jsonDocument["runtimeTarget"]["name"].GetValue<string>();
-            var folderRuntimeName = runtimeName switch
-            {
-                ".NETCoreApp,Version=v6.0" => "net6.0",
-                ".NETCoreApp,Version=v7.0" => "net7.0",
-                _ => throw new ArgumentOutOfRangeException(nameof(runtimeName), runtimeName,
-                    "This value is not supported. You have probably introduced new .NET version to AutoInstrumentation")
-            };
-            return folderRuntimeName;
-        }
     };
 
     Target InstallDocumentationTools => _ => _
--- a/build/Build.Steps.cs
+++ b/build/Build.Steps.cs
@@ -1,5 +1,4 @@
-using System.Text.Json;
-using System.Text.RegularExpressions;
+using System.Text.Json.Nodes;
 using Extensions;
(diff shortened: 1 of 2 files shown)


build/Build.Steps.cs
```diff
@@ -473,8 +472,8 @@ partial class Build
                 AdditionalDepsDirectory.GlobFiles("**/*deps.json")
                     .ForEach(file =>
                     {
-                        var depsJsonContent = File.ReadAllText(file);
-                        using var jsonDocument = JsonDocument.Parse(depsJsonContent);
+                        var rawJson = File.ReadAllText(file);
+                        var jsonDocument = JsonNode.Parse(rawJson).AsObject();
```
pjanotti, line 476: suggestion:
```suggestion
                        var depsJson = JsonNode.Parse(rawJson).AsObject();
```

build/Build.Steps.cs
```diff
@@ -486,43 +485,46 @@ partial class Build
 
                         CopyNativeDependenciesToStore(file, jsonDocument, architectureStores);
 
-                        depsJsonContent = RemoveOpenTelemetryAutoInstrumentationAdditionalDepsFromDepsContent(depsJsonContent);
+                        RemoveOpenTelemetryAutoInstrumentationAdditionalDepsFromDepsContent(jsonDocument);
```
pjanotti, line 488: nit: since the `jsonDocument` is mutable, it may work better as an extension method (possibly paired with a rename for the method)

build/Build.Steps.cs
```diff
@@ -486,43 +485,46 @@ partial class Build
...
 
                         if (folderRuntimeName == TargetFramework.NET6_0)
                         {
                             // To allow roll forward for applications, like Roslyn, that target one tfm
                             // but have a later runtime move the libraries under the original tfm folder
                             // to the latest one.
-                            depsJsonContent = RollFrameworkForward(TargetFramework.NET6_0, TargetFramework.NET7_0, architectureStores, depsJsonContent);
+                            RollFrameworkForward(TargetFramework.NET6_0, TargetFramework.NET7_0, architectureStores, jsonDocument);
```
pjanotti, line 495: similar nit: it may read better as an extension method.

build/Build.Steps.cs
```diff
@@ -486,43 +485,46 @@ partial class Build
...
-                        File.WriteAllText(file, depsJsonContent);
+                        File.WriteAllText(file, jsonDocument.ToJsonString(new()
+                        {
+                            WriteIndented = true
+                        }));
                     });
                 RemoveFilesFromAdditionalDepsDirectory();
 
-                void CopyNativeDependenciesToStore(AbsolutePath file, JsonDocument jsonDocument, IReadOnlyList<string> architectureStores)
+                void CopyNativeDependenciesToStore(AbsolutePath file, JsonObject jsonDocument, IReadOnlyList<string> architectureStores)
```
pjanotti, line 506:
```suggestion
                void CopyNativeDependenciesToStore(AbsolutePath file, depsJson jsonDocument, IReadOnlyList<string> architectureStores)
```

build/Build.Steps.cs
```diff
@@ -551,12 +556,31 @@ void RemoveFilesFromAdditionalDepsDirectory()
                     AdditionalDepsDirectory.GlobDirectories("**/runtimes").ForEach(DeleteDirectory);
                 }
 
-                string RollFrameworkForward(string runtime, string rollForwardRuntime, IReadOnlyList<string> architectureStores, string depsJsonContent)
+                void RollFrameworkForward(string runtime, string rollForwardRuntime, IReadOnlyList<string> architectureStores, JsonObject jsonDocument)
```
pjanotti, line 559: nit: static method if possible, if not possible I think it is worth adding a type wrapping the `jsonDocument` to implement these methods.

pjanotti (commented):
Nice! I made a bunch of nit suggestions, but, much better than the regular expressions.

rajkumar-rangaraj (approved):

Issue #2037, opened 2023-01-21
Review instructions for macos


pjanotti:
We need to ensure that instructions for macos give the proper visibility to the need of `brew install coreutils` and the usage of popular shells like `zsh`. See conversation in Slack thread: https://cloud-native.slack.com/archives/C01NR1YLSE7/p1674162631314489

Related to #2036

PR #2360, opened 2023-03-22
Make shell scripts work on sh (Bourne shell)

pjanotti:
## Why

`instrument.sh` fails on macOS with 
```terminal
$ . $HOME/.otel-dotnet-auto/instrument.sh
/Users/pjanotti/.otel-dotnet-auto/instrument.sh:62: = not found
```

Related to #2037 

## What

Using the way recommended for POSIX instead of a way that only works on bash. A good explanation can be found [here]( https://stackoverflow.com/a/20449556).

## Tests

Manually checked on macOS, still pending Linux validation.

## Checklist

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -41,7 +41,7 @@ else
     # If global.json exists, load expected version
     if [[ -f "$DOTNET_GLOBAL_FILE" ]]; then
         DOTNET_VERSION=$(FirstJsonValue "version" "$(cat "$DOTNET_GLOBAL_FILE")")
-        if [[ "$DOTNET_VERSION" == ""  ]]; then
+        if [[ "$DOTNET_VERSION" = ""  ]]; then
             unset DOTNET_VERSION
         fi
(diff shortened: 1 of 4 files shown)


Kielek (approved):
😱

pjanotti:
🤔 one alternative that I didn't try was to switch the script to `#!/bin/bash` instead. I'm not sure if there is a reason to prefer one over the other, `#!/bin/sh` is older and probably available everywhere, but, nowadays is there any *nix OS that requires us to use `sh`?

pjanotti:
> 🤔 one alternative that I didn't try was to switch the script to #!/bin/bash instead. I'm not sure if there is a reason to prefer one over the other, #!/bin/sh is older and probably available everywhere, but, nowadays is there any *nix OS that requires us to use sh?

nvm, it doesn't work on macOS.

pellared (approved):

PR #2385, opened 2023-03-27
Fix MongoDB Server bumping

RassK:
## Why

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2341

## What

Bumps only MongoDB Server patch version.

## Tests

Existing.


RassK:
does not work that way... it will always check against `latest` tag.

PR #2390, opened 2023-03-28
MongoDB - instrumenation for .NET Framework

### LINKED ISSUE #2328: NuGet package missing assemblies for MongoDB instrumentation
The package required for MongoDB bytecode instrumentation, [`MongoDB.Driver.Core.Extensions.DiagnosticSources`](https://www.nuget.org/packages/MongoDB.Driver.Core.Extensions.DiagnosticSources/), is only added via AdditionalDeps so it is not shipped with the current NuGet package.

cc @RassK

Kielek:
## Why

Fixes #2328 

## What

* MongoDB - instrumentation for .NET Framework
* Move instrumentation library dependency from Addition Dependencies to main project.

## Tests

Executed locally on Docker for Windows with Linux.
Testocontainers for now does not support Docker for Windows with Windows. No possibility to enable it in pipeline: https://github.com/testcontainers/moby-ryuk/pull/40#issuecomment-1467732109

## Checklist


theletterf (approved):

Kielek:
Alternative approach discussed with @RassK - manually copy MongoDB instrumentation library instead of removing non-needed libraries from the output folder.

As the goal, for now, is to make nuget scenario working (+ .NET Framework support) I would keep current scenario. Proposal from Rasmus can be handled later as future optimisations.

RassK:
> @RassK, I think that #2294 can be adapted also to direct references approach. Now, ExcludedAssets are generated manually.

No, it's only addressed for additional deps / shared store purpose.

pjanotti:
@Kielek @RassK this PR is an interesting possible path forward for the "usual" distribution (ie.: non-Nuget). That said I think we have to step back and design the overall solution.

Regarding:
> Testocontainers for now does not support Docker for Windows with Windows. No possibility to enable it in pipeline: https://github.com/testcontainers/moby-ryuk/pull/40#issuecomment-1467732109

I thought the issue was that GH Windows runners can only run Windows containers. The ASP.NET test launches a Windows container without the `ryuk` stuff. That said it is true that `ryuk` launches a Linux container. If we want to test MongoDB instrumentation on .NET Framework it will be necessary to run `MongoDB` in a Windows container.

RassK:
> @Kielek @RassK this PR is an interesting possible path forward for the "usual" distribution (ie.: non-Nuget). That said I think we have to step back and design the overall solution.

For me it seems to be in any way step backwards. 
* It's a complex package (many dependencies + native dependencies)
* It has dependencies not connected to MongoDB (DnsClient and SharpCompress)
* I think we should be more dynamic and adaptive to the application, than more static and ship a relatively large package.
* Being static narrows down the supported version range. Currently it's theoretically possible to extend the support by having multiple versions of the adapter module and then dynamically loading the correct version.

theletterf (approved):
@Kielek LMK when this goes downstream—thx!

src/OpenTelemetry.AutoInstrumentation/Instrumentations/MongoDb/MongoClientIntegration.cs
```diff
@@ -65,12 +60,10 @@ public static class MongoClientIntegration
...
         clusterConfiguratorProperty?.SetValue(settings, Delegate.Combine(existingDelegate, setListenerDelegate));
-#endif
 
         return CallTargetState.GetDefault();
     }
 
-#if NET6_0_OR_GREATER
     private static object GetInstrumentationOptions()
     {
         Type optionsType = Type.GetType("MongoDB.Driver.Core.Extensions.DiagnosticSources.InstrumentationOptions, MongoDB.Driver.Core.Extensions.DiagnosticSources")!;
```
RassK, line 69: Can take direct reference now

src/CommonExcludedAssets.props
```diff
@@ -21,4 +21,29 @@
     <!-- Microsoft.Bcl.AsyncInterfaces is required by StackExchange.Redis. ExcludeAssets="all" prevents copying it to the output -->
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" ExcludeAssets="all" />
   </ItemGroup>
+
+  <ItemGroup>
+    <!-- DnsClient is required by MongoDB.Driver.Core.Extensions.DiagnosticSources. ExcludeAssets="all" prevents copying it to the output -->
```
pjanotti, line 26: To avoid the verbosity and the copy/paste mistake it seems better to have a single comment before the `ItemGroup`. Something like "Transient dependencies from MongoDB.Driver.Core.Extensions.DiagnosticSources that are shipped with the applications. ExcludeAssets="all" prevents copying them to the output"

pjanotti (approved):
LGTM w/ a minor suggestion.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/MongoDb/MongoClientIntegration.cs
pjanotti: It seems better to change to direct reference in a separate PR.

Kielek: It is great comment. This type was removed https://github.com/jbogard/MongoDB.Driver.Core.Extensions.OpenTelemetry/commit/7652f6286d7be1fbd1c50bb37b57966de004d6f1

I will check how it impact us and if we really need to ship this package at all.

Kielek: #2454

src/CommonExcludedAssets.props
Kielek: Extracted common comment. b7ca218073c744f0b8d980346c47e89224c65dcb
I would keep explanation per library - I expect more to come. Part of them probably can be linked by more than one library.

PR #2405, opened 2023-04-04
Add Rule Engine to the StartupHook

### LINKED ISSUE #2392: Proposal: Add Rule Engine to StartupHook
## Overview
Add a rule engine to the StartupHook. This rule engine will validate all OpenTelemetry Auto-Instrumentation assemblies for conflicts and ensure that it backs off in unsupported scenarios instead of crashing. The feature should provide a way for vendors to use this rule engine to validate their custom logic. It should validate all the rules and log to an event source/logger. Each rule will return a Boolean value, and even a single rule failure should back off the auto-instrumentation.  This feature could be extended to solve other issues apart from assembly load as well. The .NET event handler or delegate could be used to implement this feature.

## Feature Details
- The rule engine will be implemented in the StartupHook to validate all OpenTelemetry Auto-Instrumentation assemblies for conflicts and ensure that it backs off in unsupported scenarios instead of crashing.
- The feature should provide an extension for vendors to use this rule engine to validate their custom logic.
- The rule engine will be designed to validate all the rules and log to an event source or logger.
- Each rule will return a Boolean value, and even a single rule failure should back off the auto-instrumentation.
- The .NET event handler or delegate could be used to implement this feature.

## Motivation
Currently, the OpenTelemetry .NET Auto-Instrumentation lacks a mechanism to validate for conflicts and unsupported scenarios, which can lead to crashes during runtime. This feature will ensure that the Auto-Instrumentation is more robust and stable.

rajkumar-rangaraj:
## Why

Fixes #2392

## What

This PR adds a rule engine to the StartupHook. The rule engine is designed to validate all OpenTelemetry assemblies and ensures that it backs off in unsupported scenarios instead of crashing. It validates all the rules and logs error to logger.

Implemented one of the rule, want to follow up other rules in a follow up PR.  Verified with manual test, it works as expected. CI results should confirm that it does not change behavior. 

**Follow-up items**
* Add implementation of other rules
* Add tests to validate rules
* Update changelog

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/OpenTelemetrySdkRule.cs
+++ b/src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/OpenTelemetrySdkRule.cs
@@ -1,0 +1,64 @@
+// <copyright file="OpenTelemetrySdkRule.cs" company="OpenTelemetry Authors">
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System.Diagnostics;
+using System.Reflection;
+using OpenTelemetry.AutoInstrumentation.Logging;
+
+namespace OpenTelemetry.AutoInstrumentation.RulesEngine;
+
+internal class OpenTelemetrySdkRule : Rule
+{
+    private static readonly IOtelLogger Logger = OtelLogging.GetLogger("StartupHook");
+
+    internal override bool Evaluate()
+    {
+        string? oTelPackageVersion = null;
+
+        try
+        {
+            var openTelemetryType = Type.GetType("OpenTelemetry.Sdk, OpenTelemetry");
+            if (openTelemetryType != null)
+            {
+                var loadedOTelAssembly = Assembly.GetAssembly(openTelemetryType);
+                var loadedOTelFileVersionInfo = FileVersionInfo.GetVersionInfo(loadedOTelAssembly?.Location);
+                var loadedOTelFileVersion = new Version(loadedOTelFileVersionInfo.FileVersion);
+
+                var autoInstrumentationOTelLocation = Path.Combine(StartupHook.LoaderAssemblyLocation ?? string.Empty, "OpenTelemetry.dll");
+                var autoInstrumentationOTelFileVersionInfo = FileVersionInfo.GetVersionInfo(autoInstrumentationOTelLocation);
+                var autoInstrumentationOTelFileVersion = new Version(autoInstrumentationOTelFileVersionInfo.FileVersion);
+
+                if (loadedOTelFileVersion < autoInstrumentationOTelFileVersion)
+                {
+                    oTelPackageVersion = loadedOTelFileVersionInfo.FileVersion;
+                }
+            }
+        }
+        catch (Exception ex)
+        {
+            // Exception in evaluation should not throw or crash the process.
+            Logger.Information($"Couldn't evaluate reference to OpenTelemetry Sdk in an app. Exception: {ex}");
+        }
+
+        if (oTelPackageVersion != null)
+        {
+            Logger.Error($"Application has direct or indirect reference to older version of OpenTelemetry package {oTelPackageVersion}.");
+            return false;
+        }
+
+        return true;
+    }
+}
--- a/src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/Rule.cs
+++ b/src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/Rule.cs
@@ -1,0 +1,26 @@
+// <copyright file="Rule.cs" company="OpenTelemetry Authors">
+// Copyright The OpenTelemetry Authors
+//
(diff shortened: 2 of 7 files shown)


src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/Rule.cs
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+// </copyright>
+
+namespace OpenTelemetry.AutoInstrumentation.RulesEngine;
+
+internal abstract class Rule
+{
+    internal string? Name { get; set; }
+
+    internal string? Description { get; set; }
```
Kielek, line 23:
```suggestion
    internal string Name { get; }

    internal string Description { get; }
```

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngine.cs
```diff
@@ -0,0 +1,67 @@
...
+internal class RuleEngine
+{
+    private static readonly IOtelLogger Logger = OtelLogging.GetLogger("StartupHook");
+
+    private readonly List<Rule> _rules = new()
+    {
+        new OpenTelemetrySdkRule
+        {
+            Name = "OpenTelemetry SDK Validator",
+            Description = "Ensure that the OpenTelemetry SDK version is not older than the version used by the Auto-Instrumentation"
```
Kielek, line 30: I think that it could be declares in Rule classes (together with abstract class contract change).

Kielek (approved):
In general looks good.
It will be great to have at least scaffolding/smoke tests in this PR but I will not block this.

RassK (approved):
Seems good. +1 to the thing that @Kielek mentioned. 
Not applying to this PR but DS rule and Instrumentation assembly rule should have integration tests to be sure that these execute properly and application is not crashing before.

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
```diff
@@ -60,15 +63,20 @@ public static void Initialize()
...
-        string loaderAssemblyLocation = GetLoaderAssemblyLocation();
+        LoaderAssemblyLocation = GetLoaderAssemblyLocation();
 
         try
         {
-            ThrowIfReferenceIncorrectOpenTelemetryVersion(loaderAssemblyLocation);
+            var ruleEngine = new RuleEngine();
+            if (!ruleEngine.Validate())
+            {
+                Logger.Error("Rule Engine Failure: One or more rules failed validation");
```
pjanotti, line 73: suggestion: make explicit in the message that automatic instrumentation won't be loaded in this case.

rajkumar-rangaraj:
> Seems good. +1 to the thing that @Kielek mentioned. Not applying to this PR but DS rule and Instrumentation assembly rule should have integration tests to be sure that these execute properly and application is not crashing before.
> 
> Turn off flag is also needed (I think in a new PR), so who has staging environment can optimize loading speed in production.

* Very good call out, in the follow up PRs add integration tests to ensure that we are not crashing with lower version of libraries.
* Along with turn-off flag, plan is to make the `RuleEngine` smart to execute once for an app. Yet to design this part.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/Rule.cs
pjanotti: Given that these properties are being initialized via object initializers it is good to add the `required` keyword (with the `init` too).

src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs
```diff
@@ -27,6 +28,8 @@ internal class StartupHook
 {
     private static readonly IOtelLogger Logger = OtelLogging.GetLogger("StartupHook");
 
+    internal static string? LoaderAssemblyLocation { get; set; }
```
pjanotti, line 31: Add a comment that this property must be initialized before any rule is evaluated since they may use it. In light of that it seems better to initialize its value here.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/Rule.cs
pjanotti: Hum, I'm having second thoughts here: the current way makes the name and description be declared on the creation of the instance while they make sense to be specific to for each rule type. It is possible to have some rule class that is used with different parameters but that case should be handled by the constructor of such class.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/OpenTelemetrySdkRule.cs
```diff
@@ -0,0 +1,64 @@
...
+                if (loadedOTelFileVersion < autoInstrumentationOTelFileVersion)
+                {
+                    oTelPackageVersion = loadedOTelFileVersionInfo.FileVersion;
+                }
+            }
+        }
+        catch (Exception ex)
+        {
+            // Exception in evaluation should not throw or crash the process.
+            Logger.Information($"Couldn't evaluate reference to OpenTelemetry Sdk in an app. Exception: {ex}");
```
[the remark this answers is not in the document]
rajkumar-rangaraj: This won't fail the rule, it could not validate the OpenTelemetry version and the execution will continue.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/Rule.cs
rajkumar-rangaraj: @pjanotti Modified the code , please take a look.

rajkumar-rangaraj:
I want to keep this PR simple. Adding the tests makes it complex. My next PRs will be to address the test gaps for the RuleEngine.

pjanotti (approved):

pjanotti:
Merging it @rajkumar-rangaraj ...

PR #2407, opened 2023-04-05
RuleEngineTracker to control rule engine execution

rajkumar-rangaraj:
## Why

Follow up of #2405 
Covers part of #2392 

## What

This pull request adds a new implementation to control whether the rule engine should be executed again or not by using the `OTEL_DOTNET_AUTO_RULE_ENGINE_ENABLED` environment variable.

## Tests

This PR includes a set of unit tests to ensure results are not impacted with environment variable values. 

## Checklist


src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngineTracker.cs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            DeleteTrackerFile();
+        }
+
+        return _shouldTrack == true || File.Exists(filePath);
+    }
+
+    private string GetFilePath()
+    {
+        var processId = Process.GetCurrentProcess().ProcessName;
```
Kielek, line 91: processId? I think it will leads to new sha256 for each process execution.
OpenTelemetry Version should be part of the hash/file name. Without this, rules will be not checked after upgrade.

test/OpenTelemetry.AutoInstrumentation.StartupHook.Tests/OpenTelemetry.AutoInstrumentation.StartupHook.Tests.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
```
Kielek, line 6: Both should be enabled by inherited props files.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngineTracker.cs
rajkumar-rangaraj: Fixed. I was testing around this value and without modifying send the change.

rajkumar-rangaraj: f464738a06505f8b1a45adbfabb0bb1a6ef85131

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngineTracker.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        return _shouldTrack == true || File.Exists(filePath);
+    }
+
+    internal string GetFilePath()
+    {
+        var processName = Process.GetCurrentProcess().ProcessName;
+        var path = AppContext.BaseDirectory;
+        var username = Environment.UserName;
+        var dataToHash = $"{processName}-{path}-{username}";
```
Kielek, line 94: Consider adding auto-instrumentation version to execute rules while updating it.
What happens if the application will be upgraded in place from v1 to v2? Should we execute rules or no?

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngineTracker.cs
```diff
@@ -0,0 +1,106 @@
...
+
+internal class RuleEngineTracker
+{
+    internal const string FileNamePrefix = "rule_engine_tracker_";
+    private static readonly IOtelLogger Logger = OtelLogging.GetLogger("StartupHook");
+    private readonly bool? _shouldTrack;
+
+    public RuleEngineTracker()
+    {
+        var ruleEngineEnabled = Environment.GetEnvironmentVariable("OTEL_RULE_ENGINE_ENABLED");
```
RassK, line 33: Shouldn't the prefix be `OTEL_DOTNET_AUTO` in our case?

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngineTracker.cs
```diff
@@ -0,0 +1,106 @@
...
+            File.Delete(filePath);
+        }
+        catch (Exception ex)
+        {
+            // Not a critical exception, when file exists rule engine does not validate.
+            Logger.Warning($"Failed to delete RuleEngine file: {ex.Message}");
+        }
+    }
+
+    public bool IsFileExists()
```
RassK, line 77: nit

```suggestion
    public bool FileExists()
```

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngineTracker.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Text;
+using OpenTelemetry.AutoInstrumentation.Logging;
+
+namespace OpenTelemetry.AutoInstrumentation.RulesEngine;
+
+internal class RuleEngineTracker
+{
+    internal const string FileNamePrefix = "rule_engine_tracker_";
+    private static readonly IOtelLogger Logger = OtelLogging.GetLogger("StartupHook");
+    private readonly bool? _shouldTrack;
```
RassK, line 29: 3 state can be simplified to 2 state 

```suggestion
    private readonly bool _shouldTrack = true;
```

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngineTracker.cs
```diff
@@ -0,0 +1,106 @@
...
+
+    public RuleEngineTracker()
+    {
+        var ruleEngineEnabled = Environment.GetEnvironmentVariable("OTEL_RULE_ENGINE_ENABLED");
+        if (ruleEngineEnabled == null)
```
RassK, line 34: after 3 state bool simplification

```csharp
if (bool.TryParse(ruleEngineEnabled, out var shouldTrack))
{
    _shouldTrack = shouldTrack;
}
```

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngineTracker.cs
RassK: One more additional thing, not related to this PR.

Currently we have a lot of places using `Environment.GetEnvironmentVariable("SOME_VAR")` but should be using configuration instead. I know that in .NET Core there is only environment configuration reader, but as soon as we have otel file, this will break anyway.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngineTracker.cs
RassK: Unfortunately in our case I think even the patch version could break this, if we do version bumps in patch versions.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngineTracker.cs
RassK: Might need logging also if the parsing fails, but configuration reader should actually have it also 🤔

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngineTracker.cs
rajkumar-rangaraj: We move all our environment variables to enabled pattern. I will use `OTEL_DOTNET_AUTO_RULE_ENGINE_ENABLED`.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngine.cs
```diff
@@ -29,10 +29,24 @@ internal class RuleEngine
         new InstrumentationAssemblyRule()
     };
 
+    internal RuleEngine(List<Rule>? rules = null)
```
Kielek, line 32: I suppose it is only for testing purposes.

I would prepare 2 separate constructors and mark one with proper comment:

```
internal RuleEngine()
internal RuleEngine(List<Rule> rules)
```

test/OpenTelemetry.AutoInstrumentation.StartupHook.Tests/RuleEngineTests.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        // Assert
+        Assert.True(result);
+        Assert.True(testRule.IsEvaluated);
+    }
+
+    public void Dispose()
+    {
+        Environment.SetEnvironmentVariable("OTEL_DOTNET_AUTO_RULE_ENGINE_ENABLED", null);
+    }
```
Kielek, line 106: Can it be moved just after class name? Is is easier to find what happen before/after tests by default.

docs/config.md
```diff
@@ -325,6 +325,22 @@ Important environment variables include:
...
+back off during unsupported scenarios instead of crashing.
+
+It is recommended to enable the RuleEngine only during the first run of the
+application, or when the deployment changes or the Auto-Instrumentation library
+is upgraded, as once validated, there is no need to re-validate the rules when
+the application restarts.
+
+| Environment variable                   | Description         | Default value  | Status                                                                                                                            |
+|--------------------------------------- |---------------------|----------------|-----------------------------------------------------------------------------------------------------------------------------------|
+| `OTEL_DOTNET_AUTO_RULE_ENGINE_ENABLED` | Enables RuleEngine. | `true`         | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
```
Kielek, line 342: Some comment needed that it is supported only on .NET (Core)

docs/config.md
```diff
@@ -325,6 +325,22 @@ Important environment variables include:
...
 
+## RuleEngine
+
+The RuleEngine is a feature that validates OpenTelemetry API, SDK,
+Instrumentation, and Exporter assemblies for unsupported scenarios. It can
+ensure that the OpenTelemetry Auto-Instrumentation is more robust and stable by
+back off during unsupported scenarios instead of crashing.
+
+It is recommended to enable the RuleEngine only during the first run of the
+application, or when the deployment changes or the Auto-Instrumentation library
```
Kielek, line 336: As it is official doc we should avoid inconsitent naming: Auto-Instrumentation -> Automatic Instrumentation.

docs/config.md
```diff
@@ -328,14 +328,15 @@ Important environment variables include:
 ## RuleEngine
 
 The RuleEngine is a feature that validates OpenTelemetry API, SDK,
-Instrumentation, and Exporter assemblies for unsupported scenarios. It can
-ensure that the OpenTelemetry Auto-Instrumentation is more robust and stable by
-back off during unsupported scenarios instead of crashing.
+Instrumentation, and Exporter assemblies for unsupported scenarios.
+It is supported only on .NET and .NET Core platforms. It can ensure
```
Kielek, line 332: .NET Core is in general not-supported. .NET 6+ only.

docs/config.md
```diff
@@ -325,6 +325,23 @@ Important environment variables include:
...
 | `OTEL_DOTNET_AUTO_PLUGINS`                          | Colon-separated list of OTel SDK instrumentation plugin types, specified with the [assembly-qualified name](https://docs.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname?view=net-6.0#system-type-assemblyqualifiedname). _Note: This list must be colon-separated because the type names may include commas._ See more info on how to write plugins at [plugins.md](plugins.md). |               | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
 
+## RuleEngine
+
+The RuleEngine is a feature that validates OpenTelemetry API, SDK,
+Instrumentation, and Exporter assemblies for unsupported scenarios.
+It is supported only on .NET and .NET Core platforms. It can ensure
+that the OpenTelemetry Automatic Instrumentation is more robust and
+stable by back off during unsupported scenarios instead of crashing.
```
theletterf, line 334:
```suggestion
RuleEngine is a feature that validates OpenTelemetry API, SDK,
Instrumentation, and Exporter assemblies for unsupported scenarios,
ensuring that OpenTelemetry automatic instrumentation is more
stable by backing of instead of crashing. It works on .NET 6 and higher.
```

docs/config.md
```diff
@@ -325,6 +325,23 @@ Important environment variables include:
...
+
+It is recommended to enable the RuleEngine only during the first run of the
+application, or when the deployment changes or the Automatic Instrumentation
+library is upgraded, as once validated, there is no need to re-validate
+the rules when the application restarts.
```
theletterf, line 339:
```suggestion
Enable RuleEngine only during the first run of the application,
or when the deployment changes or the automatic instrumentation
library is upgraded. Once validated, there's no need to revalidate
the rules when the application restarts.
```

Kielek (approved):

rajkumar-rangaraj:
Thanks @Kielek for the doc update.

PR #2411, opened 2023-04-06
Add DiagnosticSource version check rule

xiang17:
## Why

Follow up of #2405 for DS rule
For issue #2392

## What

This PR adds a version check rule for System.Diagnostics.DiagnosticSource. It will back off the auto instrumentation in case a `System.Diagnostics.DiagnosticSource` is included that can cause the auto instrumentation to crash.

Automated tests are to be added as follow up.

## Tests

Unit tests

## Checklist


src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
```diff
@@ -14,13 +14,59 @@
...
     internal override bool Evaluate()
     {
-        // TODO: implement checking logic
+        string? diagnosticSourcePackageVersion = null;
+
+        try
+        {
+            // Look up for type with an assembly name, will load the library.
+            // The loaded version depends on the app's reference to the package.
+            var diagnosticSourceType = Type.GetType("System.Diagnostics.DiagnosticSource, System.Diagnostics.DiagnosticSource");
```
rajkumar-rangaraj, line 41: This looks incorrect, namespace for `System.Diagnostics.DiagnosticSource` is `System.Diagnostics`. Could you please verify?

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
```diff
@@ -14,13 +14,59 @@
...
+                }
+            }
+        }
+        catch (Exception ex)
+        {
+            // Exception in evaluation should not throw or crash the process.
+            Logger.Information($"Couldn't evaluate reference to System.Diagnostics.DiagnosticSource in an app. Exception: {ex}");
+        }
+
+        if (diagnosticSourcePackageVersion != null)
```
Kielek, line 64: The name for a variable is misleading. It suggests that if any DiagnosticSource version was loaded then it should return false.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
```diff
@@ -14,13 +14,59 @@
...
+
+        try
+        {
+            // Look up for type with an assembly name, will load the library.
+            // The loaded version depends on the app's reference to the package.
+            var diagnosticSourceType = Type.GetType("System.Diagnostics.DiagnosticSource, System.Diagnostics.DiagnosticSource");
+            if (diagnosticSourceType != null)
+            {
+                var loadedDiagnosticSourceAssembly = Assembly.GetAssembly(diagnosticSourceType);
+                var loadedDiagnosticSourceFileVersionInfo = FileVersionInfo.GetVersionInfo(loadedDiagnosticSourceAssembly?.Location);
```
rajkumar-rangaraj, line 45: Consider using the below, it is faster than making a I/O call. We might need to fix this in the other rule. That could be in a different PR.
AssemblyFileVersionAttribute fileVersionAttribute = Assembly.GetAssembly(diagnosticSourceType).GetCustomAttributes(typeof(AssemblyFileVersionAttribute), false)
    .FirstOrDefault() as AssemblyFileVersionAttribute;

xiang17: I updated it to a similar one: `Assembly.GetAssembly(diagnosticSourceType)?.GetCustomAttribute<AssemblyFileVersionAttribute>()?.Version`

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
xiang17: Renamed to `olderDiagnosticSourcePackageVersion`.

xiang17:
Removed unit tests. It's probably better to test in integration tests, because most of the logic are checking the environment (the Type.GetType() call and version check of DiagnosticSource that's in `tracer-home\net` folder) outside the class anyways, except the less than check.

For example, checking the DiagnosticSource version of the auto instrumentation library depends on checking `StartupHook.LoaderAssemblyLocation`, which is the `tracer-home\net` folder. The unit test project `OpenTelemetry.AutoInstrumentation.StartupHook.Tests.csproj` doesn't have it.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
```diff
@@ -14,13 +14,62 @@
...
+            {
+                var loadedDiagnosticSourceAssembly = Assembly.GetAssembly(diagnosticSourceType);
+                var loadedDiagnosticSourceAssemblyFileVersionAttribute = loadedDiagnosticSourceAssembly?.GetCustomAttribute<AssemblyFileVersionAttribute>();
+                if (loadedDiagnosticSourceAssemblyFileVersionAttribute != null)
+                {
+                    var loadedDiagnosticSourceFileVersion = new Version(loadedDiagnosticSourceAssemblyFileVersionAttribute.Version);
+
+                    var autoInstrumentationDiagnosticSourceLocation = Path.Combine(StartupHook.LoaderAssemblyLocation ?? string.Empty, "System.Diagnostics.DiagnosticSource.dll");
+                    var autoInstrumentationDiagnosticSourceFileVersionInfo = FileVersionInfo.GetVersionInfo(autoInstrumentationDiagnosticSourceLocation);
+                    var autoInstrumentationDiagnosticSourceFileVersion = new Version(autoInstrumentationDiagnosticSourceFileVersionInfo.FileVersion);
```
Kielek, line 52: I seen Raj comment which suggest that it could be expensive this is expensive. If so, it potentially can be avoided by typing here manually the version used by automatic-instrumentation which is known on compilation status.

I see two possibilities how to achieve this:
 - source code generator (newer tested in fact)
 - `class DiagnosticSourceVersionProvider` with unit tests testing exactly what you are doing here

```csharp
var autoInstrumentationDiagnosticSourceLocation = Path.Combine(StartupHook.LoaderAssemblyLocation ?? string.Empty, "System.Diagnostics.DiagnosticSource.dll");
var autoInstrumentationDiagnosticSourceFileVersionInfo = FileVersionInfo.GetVersionInfo(autoInstrumentationDiagnosticSourceLocation);
var autoInstrumentationDiagnosticSourceFileVersion = new Version(autoInstrumentationDiagnosticSourceFileVersionInfo.FileVersion);

Assert.Equal(autoInstrumentationDiagnosticSourceFileVersion, DiagnosticSourceVersionProvider.Version);
```

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
```diff
@@ -14,13 +14,62 @@
...
+                    {
+                        olderDiagnosticSourcePackageVersion = loadedDiagnosticSourceAssemblyFileVersionAttribute.Version;
+                    }
+                }
+            }
+        }
+        catch (Exception ex)
+        {
+            // Exception in evaluation should not throw or crash the process.
+            Logger.Information($"Couldn't evaluate reference to System.Diagnostics.DiagnosticSource in an app. Exception: {ex}");
```
Kielek, line 64: Totally agree that we should not crash, should we allow execute auto-instrumentation?

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
rajkumar-rangaraj: > If so, it potentially can be avoided by typing here manually the version used by automatic-instrumentation which is known on compilation status.

I'm working on `InstrumentationAssemblyRule` which addresses this problem. At the build time, creating a file with list of package name and file version. We read that specific file to determine the versions from auto-instrumentation.

We could accept the PR with the current change. Based on feedback from `InstrumentationAssemblyRule`, we could change in a later PR. Another factor to consider is we have a rule engine switch to control the behavior.

xiang17: I tried to add a `DiagnosticSourceVersionProvider` to hold the version, but testing it also needs the `StartupHook.LoaderAssemblyLocation` property which requires the code to be invoked as auto instrumentation (unit tests won't have access to tracer-home folder).

I've added the new unit tests that focuses on the logic within the `DiagnosticSourceRule` class by filling in both the versions of customer's app and auto instrumentation with an override method.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
xiang17: A warning makes more sense I think. Updated: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2411/commits/893c25cea598b1987ac7fac0aa690386528b31ea

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
```diff
@@ -14,13 +14,77 @@
...
+
+    protected virtual Version? GetVersionFromApp()
+    {
+        // Look up for type with an assembly name, will load the library.
+        // The loaded version depends on the app's reference to the package.
+        var diagnosticSourceType = Type.GetType("System.Diagnostics.DiagnosticSource, System.Diagnostics.DiagnosticSource");
+        if (diagnosticSourceType != null)
+        {
+            var loadedDiagnosticSourceAssembly = Assembly.GetAssembly(diagnosticSourceType);
+            var loadedDiagnosticSourceAssemblyFileVersionAttribute = loadedDiagnosticSourceAssembly?.GetCustomAttribute<AssemblyFileVersionAttribute>();
```
pjanotti, line 72: Q: why the file version instead of the assembly version?

rajkumar-rangaraj: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1692#discussion_r1035419689 - This is applicable for all OpenTelemetry assembly.

For diagnosticsource it is not applicable, but to keep a standard implementation we could use file version.

xiang17: The assembly version is not always upgraded for minor/path version upgrades: Both for 7.0.0 and 7.0.1 the assembly version is `[assembly: AssemblyVersion("7.0.0.0")]`.

xiang17: In comparison, the File Versions are: `[assembly: AssemblyFileVersion("7.0.22.51805")]`, `[assembly: AssemblyFileVersion("7.0.323.6910")]`.

pjanotti: So, in practice, the file version is a proxy for the NuGet package version which is not easily accessible at runtime. It seems that it deserves a comment why the file version is being used.

xiang17: I've added a comment on why file version is being used.

xiang17: > Just as example for two assemblies used by the project:
> 
> ```powershell
> PS C:\src\otel\dotnet\auto> dir "C:\Program Files\dotnet\shared\System.Diagnostics.DiagnosticSource.dll" -r  | % { (Get-Command $_.FullName) | Select-Object -ExpandProperty FileVersionInfo -Property Version  | Select-Object Version, FileVersion, FileName }
> 
> Version  FileVersion    FileName
> -------  -----------    --------
> 6.0.15.0 6.0.1523.11507 C:\Program Files\dotnet\shared\Microsoft.NETCore.App\6.0.15\System.Diagnostics.DiagnosticSource.dll
> 7.0.1.0  7.0.122.56804  C:\Program Files\dotnet\shared\Microsoft.NETCore.App\7.0.1\System.Diagnostics.DiagnosticSource.dll
> 7.0.4.0  7.0.423.11508  C:\Program Files\dotnet\shared\Microsoft.NETCore.App\7.0.4\System.Diagnostics.DiagnosticSource.dll
> 
> 
> PS C:\src\otel\dotnet\auto> dir ".\bin\tracer-home\store\x64\System.Diagnostics.DiagnosticSource.dll" -r  | % { (Get-Command $_.FullName) | Select-Object -ExpandProperty FileVersionInfo -Property Version  | Select-Object Version, FileVersion, FileName }
> 
> Version FileVersion   FileName
> ------- -----------   --------
> 7.0.4.0 7.0.423.11508 C:\src\otel\dotnet\auto\bin\tracer-home\store\x64\net6.0\system.diagnostics.diagnosticsource\7.0.2\lib\net7.0\System.Diagnostics.DiagnosticSource.dll
> ```
> 
> This leads to a question unrelated to this PR: why S.D.DS version for `OpenTelemetry.AutoInstrumentation` .NET Framework and AdditionalDeps diverged? Anyway, not related to this PR.

Created an issue for this question: #2431.

nrcventura (approved):

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
```diff
@@ -14,13 +14,88 @@
...
+        {
+            logger.Error($"Rule Engine: Application has direct or indirect reference to older version of System.Diagnostics.DiagnosticSource.dll {olderDiagnosticSourcePackageVersion}.");
+            return false;
+        }
+
+        logger.Information("Rule Engine: DiagnosticSourceRule evaluation success.");
         return true;
     }
+
+    protected virtual Version? GetVersionFromApp()
```
pjanotti, line 75: nit: `GetResolvedVersion()` or `GetActualVersion()` it is not necessary that S.D.DS is coming from the app itself.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
```diff
@@ -14,13 +14,88 @@
...
+    {
+        // Look up for type with an assembly name, will load the library.
+        // The loaded version depends on the app's reference to the package.
```
pjanotti, line 78: nit: sorry, I'm being very pedantic here :) this is true if the application has a (direct or transitive) reference to it, otherwise, it should be the version from AdditionalDeps (or shared runtime if no or incorrect AdditionalDeps/SharedStore).

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
```diff
@@ -14,13 +14,88 @@
...
+                if (loadedDiagnosticSourceFileVersion < autoInstrumentationDiagnosticSourceFileVersion)
+                {
+                    olderDiagnosticSourcePackageVersion = loadedDiagnosticSourceFileVersion.ToString();
+                }
+            }
+        }
+        catch (Exception ex)
+        {
+            // Exception in evaluation should not throw or crash the process.
+            logger.Warning($"Rule Engine: Couldn't evaluate reference to System.Diagnostics.DiagnosticSource in an app. Exception: {ex}");
```
pjanotti, line 61:
```suggestion
            logger.Warning($"Rule Engine: Couldn't evaluate System.Diagnostics.DiagnosticSource version. Exception: {ex}");
```

pjanotti (approved):
Minor suggestions, but, otherwise LGTM. There should be a follow-up to get the version wanted by auto-instrumentation at build time or from the json created in a separate PR by Raj.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
xiang17: Removed. I meant the version could be different if the app has reference to the package or the lack thereof. Anyways the code is more self-explanatory now with the new method name `GetResolvedVersion`.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
xiang17: Updated to `GetResolvedVersion()`.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
```diff
@@ -14,13 +14,88 @@
...
+            if (versionString != null)
+            {
+                return new Version(versionString);
+            }
+        }
+
+        return null;
+    }
+
+    protected virtual Version? GetVersionFromAutoInstrumentation()
```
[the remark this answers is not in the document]
xiang17: Updated.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
xiang17: Updated.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/DiagnosticSourceRule.cs
```diff
@@ -14,13 +14,88 @@
...
+    {
+        var autoInstrumentationDiagnosticSourceLocation = Path.Combine(StartupHook.LoaderAssemblyLocation ?? string.Empty, "System.Diagnostics.DiagnosticSource.dll");
+        var autoInstrumentationDiagnosticSourceFileVersionInfo = FileVersionInfo.GetVersionInfo(autoInstrumentationDiagnosticSourceLocation);
+        var autoInstrumentationDiagnosticSourceFileVersion = new Version(autoInstrumentationDiagnosticSourceFileVersionInfo.FileVersion);
```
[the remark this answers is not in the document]
xiang17: Updated.

PR #2415, opened 2023-04-06
Container resource detectors + manage resource detectors

### LINKED ISSUE #2414: Docker resource detector
Needs to be configurable similar to what is done for instrumentations (fine and coarse level control)
Short run: add "low dependencies" detector: docker firs

part of #2395

Kielek:
## Why

Fixes #2414

## What

Container resource detectors + manage resource detectors
by

- `OTEL_DOTNET_AUTO_RESOURCE_DETECTOR_ENABLED`
- `OTEL_DOTNET_AUTO_{0}_RESOURCE_DETECTOR_ENABLED`.

Available resource detectors:

- ~~`ENVIRONMENTALVARIABLES`~~
- ~~`TELEMETRYSDK`~~
- `CONTAINER`

## Tests

CI

## Checklist

- [ ] New features are covered by tests. - unit test created. Lack of integration test for Container

## WIP

For now, it is referencing obsolete Docker package, the new one will be released after merge:  
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1132
`DistributionStructure` and `ModuleTests` tests will be updated with new package


RassK (approved):
Seems excellent!

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceDetector.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    /// <summary>
+    /// Environmental variables resource detector.
+    /// </summary>
+    EnvironmentalVariables = 0,
+
+    /// <summary>
+    /// TelemetrySdk resource detector.
+    /// </summary>
+    TelemetrySdk = 1,
```
pellared, line 32: I think that these two should be always enabled.

pellared: > That would require custom plugin then to override this behavior.

Exactly. Is there any reason why these should not be always enabled? Let me share my reasoning.

I think the env var detector is critical for auto-instrumentation as this is the recommended way to add additional resources. Otherwise, we would need to document that [this](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/config.md#resources) section relies on this detector and has to be enabled. I feel that these env vars should always be supported.

Regarding TelemetrySDK  [the specification](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#telemetry-sdk) it is stable and I am not sure if there is a good use case to disable it.

I believe a common setup would be to set `OTEL_DOTNET_AUTO_RESOURCE_DETECTOR_ENABLED=false` to rely on [Collector's Resource Detection Processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor)

lachmatt: > Regarding TelemetrySDK [the specification](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#telemetry-sdk) it is stable and I am not sure if there is a good use case to disable it.

TelemetrySDK attributes seem to be required.

Kielek: Related to Env.Vars. and TelemetrySdk. Both of them are enabled by default.

Now I am introducing new layer of configuration. It is not different than call SetReourceBuilde(ResourceBuilder.CreateEmpty()). As I remember there is also possibility to manage resource on container side. So the users can drop them there.

For sure, I can remove this two options from the configuration but I do not see any good reason to do it.

pellared: > I believe a common setup would be to set OTEL_DOTNET_AUTO_RESOURCE_DETECTOR_ENABLED=false to rely on [Collector's Resource Detection Processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor)

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
   ensuring back off instead of crashing, improving overall stability.
 - The environment variable `OTEL_DOTNET_AUTO_RULE_ENGINE_ENABLED` could be
   used to enable or disable the rule engine.
+- Support for Container resource detector.
+- Support for enabling well known resource detectors such by
```
nrcventura, line 24: Something like the following might be clearer.
```suggestion
- Support for enabling well known resource detectors by using the environment variables
```

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceDetector.cs
nrcventura: It would probably be confusing if setting the environment variable to disable resource detectors disables something that is really "required", and it would also be confusing if it did not disable all resource detectors. I wonder if we need to call out that these settings really only apply to "optional" resource detectors?

src/OpenTelemetry.AutoInstrumentation/Configurations/ConfigurationKeys.cs
```diff
@@ -60,6 +60,16 @@ internal static class ConfigurationKeys
     /// </summary>
     public const string InstrumentationEnabled = "OTEL_DOTNET_AUTO_INSTRUMENTATION_ENABLED";
 
+    /// <summary>
+    /// Configuration key for disabling all trace instrumentations.
```
rajkumar-rangaraj, line 64: nit: change comment

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceDetector.cs
rajkumar-rangaraj: I complete agree with the inputs shared by @pellared. `TelemetrySdk` is marked as stable in specification and this should be enabled by default. We could provide an option to turn off.

rajkumar-rangaraj:
Changes in this PR looks good to me. Once this conversation https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2415#discussion_r1159687803 is sorted we could merge.

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceDetector.cs
RassK: @rajkumar-rangaraj that's what done currently. I think the question here is to enable default vs subset enabled constantly (so you cannot turn off the required ones).

I think required is satisfied already by enabling default + that leaves some room for extreme cases, where you can disable the required ones but you still know what you are doing.

theletterf (approved):

rajkumar-rangaraj (approved):

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceDetector.cs
Kielek: This will be handled by #2442

docs/config.md
```diff
@@ -62,7 +62,23 @@ However, if given setting supports it, then:
...
+|--------------------------|--------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------|
+| `ENVIRONMENTALVARIABLES` | Environment variables detector | See /[1/]                                                                                                                                                                           | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
+| `TELEMETRYSDK`           | Telemetry SDK detector         | [Telemetry SDK resources documentation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md?plain=1#L117) | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
```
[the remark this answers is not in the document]
pellared: Addressed https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2415/commits/34abca9d812f55e0a80b2f055d5db65a6f61efb3

docs/config.md
```diff
@@ -64,11 +64,26 @@ A resource is the immutable representation of the entity producing the telemetry
 See [Resource semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions)
 for more details.
 
+Resources can be configured by following environmental variables
```
pellared, line 67:
```suggestion
### Resource attributes
```

docs/config.md
```diff
@@ -64,11 +64,26 @@ A resource is the immutable representation of the entity producing the telemetry
...
 | `OTEL_SERVICE_NAME`        | Sets the value of the [`service.name`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service) resource attribute. If `service.name` is provided in `OTEL_RESOURCE_ATTRIBUTES`, the value of `OTEL_SERVICE_NAME` takes precedence. | See [Service name automatic detection](#configuration-methods) under Configuration method section.                                                                                                                                  | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
 
+and detected by resource detectors.
```
pellared, line 74:
```suggestion
### Resource detectors
```

docs/config.md
```diff
@@ -64,11 +64,26 @@ A resource is the immutable representation of the entity producing the telemetry
...
+
+Supported resource detectors:
+
+| ID          | Description        | Documentation                                                                                                                                                                 | Status                                                                                                                            |
+|-------------|--------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------|
+| `CONTAINER` | Container detector | [Container resource detector documentation](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.ResourceDetectors.Container/README.md) | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
```
pellared, line 85:
```suggestion
| `CONTAINER` | Container detector | [Container resource detector documentation](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.ResourceDetectors.Container/README.md) | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/ResourceDetectors.Container-1.0.0-beta.3/specification/versioning-and-stability.md) |
```

pellared (approved):

docs/config.md
Kielek: 1ac46da0ab8a6e36bb02bf73c9d6470d7d09169e

docs/config.md
Kielek: 1ac46da0ab8a6e36bb02bf73c9d6470d7d09169e

docs/config.md
Kielek: 1ac46da0ab8a6e36bb02bf73c9d6470d7d09169e

PR #2419, opened 2023-04-07
Add InstrumentationAssemblyRule to validate OpenTelemetry Instrumentation libraries

rajkumar-rangaraj:
## Why

Follow up of #2405 
Covers part of #2392 

## What

This PR adds a new target to the build process that creates a `ruleengine.json` file which contains the assembly names and versions of the OpenTelemetry Instrumentation libraries that are used by the Auto-Instrumentation. This file is used during the rule validation to ensure that the version of the OpenTelemetry Instrumentation libraries is not older than the version used by the Auto-Instrumentation.

Creating the `ruleengine.json` file during the build process has the benefit of reducing the performance impact during the rule validation. The file is generated only once during the build and then used by the rule engine.

Modified `InstrumentationAssemblyRule` class to read the `ruleengine.json` file and compare the versions of the referenced assemblies with the ones specified in the file. If a referenced assembly has an older version than the one specified in the file, the rule validation fails.

**Change Summary**

* Added new target `PublishRuleEngineJson` to generate `ruleengine.json` file during build process
* Updated `InstrumentationAssemblyRule` class to read `ruleengine.json` file and compare versions of referenced assemblies

## Tests

* Manual testing performed by generating a `ruleengine.json` file and validating that the `InstrumentationAssemblyRule` correctly compares the assembly versions
* **TODO:** Once this gets merged, we will add integration tests which validate all of the rules, including the `InstrumentationAssemblyRule`, to ensure that the Auto-Instrumentation works as expected.

## Checklist


nrcventura (commented):
Looks good to me other than needing to add the new json file to the list of distributed files.

rajkumar-rangaraj:
> Looks good to me other than needing to add the new json file to the list of distributed files.

Reading the instrumentation libraries from the disk one after the other will increase the app's start up time. Generating the information at the build time reduce the I/O call and decreases the auto-instrumentation load time.

nrcventura (approved):

nrcventura:
> > Looks good to me other than needing to add the new json file to the list of distributed files.
> 
> Reading the instrumentation libraries from the disk one after the other will increase the app's start up time. Generating the information at the build time reduce the I/O call and decreases the auto-instrumentation load time.

Yeah, I was just referring to the tests that failed, that you fixed by adding that file to the distribution lists.

RassK:
> > Looks good to me other than needing to add the new json file to the list of distributed files.
> 
> Reading the instrumentation libraries from the disk one after the other will increase the app's start up time. Generating the information at the build time reduce the I/O call and decreases the auto-instrumentation load time.

I'm fine with current changes for the first step but preferred next step would be to generate cs file with these constants, so no extra IO at all.

rajkumar-rangaraj:
> I'm fine with current changes for the first step but preferred next step would be to generate cs file with these constants, so no extra IO at all.

We are making a IO call during our build process and it has no impact on user application. I don't think we need to fine-tune this approach. Any manual intervention is a lot of maintenance work. If we design it as a constant, an update to package reference needs an update to constant file which will break dependabot kind of automation.

RassK:
> > I'm fine with current changes for the first step but preferred next step would be to generate cs file with these constants, so no extra IO at all.
> 
> We are making a IO call during our build process and it has no impact on user application. I don't think we need to fine-tune this approach. Any manual intervention is a lot of maintenance work. If we design it as a constant, an update to package reference needs an update to constant file which will break dependabot kind of automation.

We already do source generations for some things, I guess it won't be any different. If you track the generated file in git then dependabot will break, otherwise it's good. 
I did mean the IO call to read and parse `ruleengine.json` in `InstrumentationAssemblyRule`, which is not in the build time but application's runtime / startup time. The same kind of optimisation we do to get rid of `integrations.json`.

rajkumar-rangaraj:
> I did mean the IO call to read and parse `ruleengine.json` in `InstrumentationAssemblyRule`, which is not in the build time but application's runtime / startup time. The same kind of optimisation we do to get rid of `integrations.json`.

Reading a single file is not a costlier operation. Going back and forth only causes an issue. We have a disable switch for rule engine, so it should not be an issue. 

If you something on top of your mind which could further improve it, we should definitely give it a try.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/InstrumentationAssemblyRule.cs
```diff
@@ -14,13 +14,77 @@
...
     {
-        // TODO: implement checking logic
+        var result = true;
+
+        try
+        {
+            var ruleEngineFileLocation = Path.Combine(StartupHook.LoaderAssemblyLocation ?? string.Empty, "ruleEngine.json");
+            var ruleEngineContent = File.ReadAllText(ruleEngineFileLocation);
+            var ruleFileInfoList = JsonSerializer.Deserialize<List<RuleFileInfo>>(ruleEngineContent);
+            var entryAssembly = Assembly.GetEntryAssembly();
```
pjanotti, line 43: It seems worth a comment on why checking the entry assembly is reasonable even when that doesn't guarantee that another assembly wouldn't reference the OpenTelemetry SDK (which implies that an old one was shipped with the application). The intention is that after this point auto-instrumentation will inject the OpenTelemetry SDK assemblies (the loading of instrumentation assemblies is delayed) and will fail with "could not load X.Y.Z version n.n.n" if there is a reference to earlier versions by some other assembly. Lots of to just say that the assumption here is that if there is any reference to the OpenTelemetry SDK it is very likely to be on the entry assembly.

🤔 I wonder if instead, we shouldn't catch the exceptions when trying to load the SDK and do some further probing to investigate the error. It becomes a kind of pay-for-play: if it works, it works, if it doesn't we try to get better messages on the way down.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/InstrumentationAssemblyRule.cs
```diff
@@ -14,13 +14,77 @@
...
+            var referencedAssemblies = entryAssembly?.GetReferencedAssemblies();
+
+            if (referencedAssemblies == null)
+            {
+                Logger.Information($"Rule Engine: Could not get referenced assembly (GetReferencedAssemblies()) from an application.");
```
pjanotti, line 48:
```suggestion
                Logger.Warning($"Rule Engine: Could not get referenced assembly (GetReferencedAssemblies()) from an application. Skipping rule evaluation.");
```

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/InstrumentationAssemblyRule.cs
pjanotti: Sorry, forgot that the goal here is to minimize the chance of crashing the application.

RassK:
> > I did mean the IO call to read and parse `ruleengine.json` in `InstrumentationAssemblyRule`, which is not in the build time but application's runtime / startup time. The same kind of optimisation we do to get rid of `integrations.json`.
> 
> Reading a single file which is around a KB is not a costlier operation. Going back and forth only causes an issue. We have a disable switch for rule engine, so it should not be an issue.
> 
> If you something on top of your mind which could further improve it, we should definitely give it a try.

Just currently I think there isn't even a single benefit generating json over cs. 

Both -
* generated
* non tracked

Json[^1] - 
* More logic / increased complexity
* IO read
* Json parse
* Shipping extra file
* Leaves file editable / could cause potential issues if changed

I can definitely pick up this small optimisation after. If it's discovered that there are any flaws or issues afterwards and it will be discarded, then I really don't mind.

[^1]: Currently the file size is so small that the performance loss is extremely minimal. Plus as mentioned there is no loss in case of feature turned off.

Kielek (approved):
I agree with @RassK. It will be great to remove json file and generate cs files instead of this file.

I won't block this, as we can fix it in follow up and it does not require any new configuration variable.

rajkumar-rangaraj:
@RassK / @Kielek Created an issue to improve the performance. Ref: #2430

PR #2432, opened 2023-04-12
StartupHook: Move all checks to mandatory rules

rajkumar-rangaraj:
## Why

Fixes #

## What

Moves the implementation of `IsApplicationInExcludeList` from `StartupHook.cs` to `ApplicationInExcludeListRule`. It's a refactor, no major code change. 

## Tests

Current CI tests.

## Checklist


Kielek:
This rule should works always, not only when `OTEL_DOTNET_AUTO_RULE_ENGINE_ENABLED` is set to `true`.
After this change, it will not work in this way.

rajkumar-rangaraj:
> This rule should works always, not only when `OTEL_DOTNET_AUTO_RULE_ENGINE_ENABLED` is set to `true`. After this change, it will not work in this way.

Good catch, I sent a part of the changes in this PR. Will add the entire changes to introduce mandatory and other rules.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/ApplicationInExcludeListRule.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        Logger.Information("Rule Engine: ApplicationInExcludeListRule evaluation success.");
+        return true;
+    }
+
+    private static string GetApplicationName()
+    {
+        try
+        {
+            return AppDomain.CurrentDomain.FriendlyName;
```
pjanotti, line 50: fyi: I opened #2441 to address the inconsistency between how native and managed code are handling OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/ApplicationInExcludeListRule.cs
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    private static bool IsApplicationInExcludeList(string applicationName)
+    {
+        return GetExcludedApplicationNames().Contains(applicationName);
+    }
+
+    private static List<string> GetExcludedApplicationNames()
+    {
+        var excludedProcesses = new List<string>();
```
pjanotti, line 66: Not from your change, but, a suggestions (perhaps for later): instead of creating the list here, it could delay the allocation until after it was confirmed that environment variable is set. At that point it is also can allocate the list with its final capacity (given the array returned by Split).

pjanotti (commented):
@rajkumar-rangaraj are you planning to update this PR or submit a separate one making this rule mandatory?

Anyway, this PR LGTM.

rajkumar-rangaraj:
> are you planning to update this PR or submit a separate one making this rule mandatory?

Thanks @pjanotti for the review. I updated the PR.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/ApplicationInExcludeListRule.cs
```diff
@@ -0,0 +1,98 @@
...
+    public ApplicationInExcludeListRule()
+    {
+        Name = "Application is in exclude list validator";
+        Description = "This rule checks if the application is included in the exclusion list specified by the OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES environment variable. If the application is in the exclusion list, the rule skips initialization.";
+    }
+
+    internal override bool Evaluate()
+    {
+        var applicationName = GetApplicationName();
+        Logger.Information($"Rule Engine: StartupHook loaded for application with name {applicationName}.");
```
pjanotti, line 34: When the code was moved this log line became redundant. My preference would be to remove this log line and add `applicationName` to the two below.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/ApplicationInExcludeListRule.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        if (IsApplicationInExcludeList(applicationName))
+        {
+            Logger.Information("Rule Engine: Application is in the exclusion list. Skipping initialization.");
+            return false;
+        }
+
+        Logger.Information("Rule Engine: ApplicationInExcludeListRule evaluation success.");
```
pjanotti, line 42:
```suggestion
        Logger.Debug("Rule Engine: {applicationName} is not in the exclusion list. ApplicationInExcludeListRule evaluation success.");
```

pjanotti (commented):
Just some minor suggestions.

Kielek (approved):

PR #2280, opened 2023-03-01
Update native code

zacharycmontoya:
## Why

Updates the native code to the version of the Datadog upstream components at git tag `v2.3.0`. The last update was over 1.5 years ago, so there are many improvements that have not yet been adopted.

## What

List of major commits:
- Remove CallTarget Integrations from json file: https://github.com/DataDog/dd-trace-dotnet/commit/54f0ad690afff4df0c0b2cd447f150ee23d47d4f
- Fix duplicating integrations due to multiple Initialize calls from different AppDomains: https://github.com/DataDog/dd-trace-dotnet/commit/9e2cbf81cd32beb1bab6e823f02daae3f3c2a308
- Ensure ConfigureAwait is added: https://github.com/DataDog/dd-trace-dotnet/commit/777c7da31f5a0526b4cd905871b927c640c8e1ed
- Detect current module at runtime and use it for PInvoke rewriting: https://github.com/DataDog/dd-trace-dotnet/commit/9f8c9e4b267aaeff7abde0726497dbad4486bf92
- Remove CallSite instrumentation: https://github.com/DataDog/dd-trace-dotnet/commit/62bf9f8d9fdc682158a9e14e49291b3b8e8e41b1
- Native stats fixes: https://github.com/DataDog/dd-trace-dotnet/commit/6a217a7a629da9296afaac22006ef6a64f5c373f
- Refactor Integration struct, and remove unused code: https://github.com/DataDog/dd-trace-dotnet/commit/71416610fdaa631981c5de6de68f346f5963f3b9
- Enable CallTarget instrumentation to pass in parameters by reference: https://github.com/DataDog/dd-trace-dotnet/commit/f2cbb4ba933e8a5be3a4af0fdf6e5b3d74ce49d9
- Change the CallTargetReturn struct to a ref struct: https://github.com/DataDog/dd-trace-dotnet/commit/3419d355b1d10b8c246c544d1dd6473ce30a19b9
- Add a big fat warning in NativeCallTargetDefinition file: https://github.com/DataDog/dd-trace-dotnet/commit/d4dc6224518fa92d0b805f19bb5cb4f9074422c2
- Minor improvements: https://github.com/DataDog/dd-trace-dotnet/commit/d939f81bdfbf85272b349ca1899590d2b76b2707
- CallTarget abstract methods support: https://github.com/DataDog/dd-trace-dotnet/commit/cee205b58877fd337a6dd2648860b0ce4d48fee7
[...]

PR #2472, opened 2023-04-17
Make the ReJIT call from the ModuleLoadFinished callback a blocking call

### LINKED ISSUE #2452: After native code update, remove race condition for OnModuleLoadFinished ReJIT request
From this native code update, this seems like we are once again susceptible to a race condition with instrumentation when a new assembly is loaded. Let me explain where and how we issue ReJIT requests:
1. At startup we invoke `CorProfiler::InternalAddInstrumentation`, which will initialize the complete list of enabled bytecode instrumentations. We iterate through all loaded modules and do a _blocking_ ReJIT request for instrumentations whose assemblies have been loaded by calling `EnqueueRequestRejitForLoadedModules` with a [promise](https://github.com/zacharycmontoya/opentelemetry-dotnet-instrumentation/blob/c232661aba13dd5ef85468a9504adcff5ff1c673/src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp#L1155-L1156).
2. When a new assembly is loaded we invoke `CorProfiler::ModuleLoadFinished`, which will see if the assembly contains instrumentation(s) and if so it will issue _nonblocking_ ReJIT requests by calling `RequestRejitForLoadedModules` with parameter `enqueueInSameThread=false` (this is required since we can't call ReJIT from the CorProfiler thread).

So number 1 is fine, but number 2 needs to be fixed. All we have to do is call the same API as number 1 and we can block with a certain timeout. I'll create an issue to track this and we can resolve this after this PR is merged.

_Originally posted by @zacharycmontoya in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2280#discussion_r1166155533_

zacharycmontoya:
## Why

In the recent native code update (https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2280), there was a large amount of restructuring of code that organizes the `CorProfiler` and the ReJIT requests. One change was that the ReJIT call from the ModuleLoadFinished callback was no longer blocking. This PR ensures that it blocks (with a reasonable timeout) so that target methods will receive bytecode instrumentation on their first (and later) invocations.

Fixes #2452

## What

Updates the ModuleLoadFinished callback to invoke the `EnqueueRequestRejitForLoadedModules` API, which accepts a promise that will be completed when the ReJIT is finished.

## Tests

Uses the existing integration tests to validate the instrumentation

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
@@ -755,9 +755,13 @@ HRESULT STDMETHODCALLTYPE CorProfiler::ModuleLoadFinished(ModuleID module_id, HR
         // We call the function to analyze the module and request the ReJIT of integrations defined in this module.
         if (tracer_integration_preprocessor != nullptr && !integration_definitions_.empty())
         {
-            const auto numReJITs =
-                tracer_integration_preprocessor->RequestRejitForLoadedModules(std::vector<ModuleID>{module_id},
-                                                                              integration_definitions_);
+            std::promise<ULONG> promise;
+            std::future<ULONG>  future = promise.get_future();
+            tracer_integration_preprocessor->EnqueueRequestRejitForLoadedModules(std::vector<ModuleID>{module_id},
+                                                                                 integration_definitions_, &promise);
+
+            // wait and get the value from the future<int>
+            const auto& numReJITs = future.get();
             Logger::Debug("Total number of ReJIT Requested: ", numReJITs);
         }
     }
--- a/src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
@@ -32,6 +32,8 @@
 #include <mach-o/getsect.h>
 #endif
 
+using namespace std::chrono_literals;
+
 #ifdef _WIN32
 #include "netfx_assembly_redirection.h"
 #endif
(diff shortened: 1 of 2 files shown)


nrcventura (approved):

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -755,9 +755,13 @@ HRESULT STDMETHODCALLTYPE CorProfiler::ModuleLoadFinished(ModuleID module_id, HR
...
-            const auto numReJITs =
-                tracer_integration_preprocessor->RequestRejitForLoadedModules(std::vector<ModuleID>{module_id},
-                                                                              integration_definitions_);
+            std::promise<ULONG> promise;
+            std::future<ULONG>  future = promise.get_future();
+            tracer_integration_preprocessor->EnqueueRequestRejitForLoadedModules(std::vector<ModuleID>{module_id},
+                                                                                 integration_definitions_, &promise);
+
+            // wait and get the value from the future<int>
+            const auto& numReJITs = future.get();
```
lachmatt, line 764: looks good, but what about a timeout that was mentioned in the PR description?

pjanotti (commented):
The promise needs to be set before the return at https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/e9ed0550ee49be066de6b4a8129813459eae7036/src/OpenTelemetry.AutoInstrumentation.Native/rejit_preprocessor.cpp#L413-L416

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
pjanotti: Related Q.: while the description mentions the timeout I don't understand when it can really timeout. Is it about "good form" and preventing a deadlock or is there a real case that we can expect it to timeout?

zacharycmontoya: Mostly it's about good form and how much delay we would want on the first request. It should never hit the timeout (I've just added it at 100ms), but I think any timeout is good to have versus leaving it unbounded, especially when we don't necessarily control when it will get invoked

pjanotti (approved):
Nice, LGTM.

zacharycmontoya:
> The promise needs to be set before the return at
> 
> https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/e9ed0550ee49be066de6b4a8129813459eae7036/src/OpenTelemetry.AutoInstrumentation.Native/rejit_preprocessor.cpp#L413-L416

Fixed in 233137c4e766ac861270afd42c423c81867bb325

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
zacharycmontoya: I've added a timeout of 100ms in 0d0ce9158dd53160d546d25b7cdd4861d112bfad

PR #2479, opened 2023-04-18
Restore assembly redirection for .NET Framework

### LINKED ISSUE #2466: .NET Fx redirection was removed
Redirection is not happening anymore for .NET Framework: calls to `RedirectAssemblyReferences` were removed.

pjanotti:
## Why

To restore assembly redirection for .NET Framework.

Fixes #2466 

## What

* Add a dedicated test to ensure that .NET Framework redirection is working as intended.
* Put the call to redirection back in the native code.

## Pending

* Find a way for `dependabot` to ignore the test project: it needs to use an old version that we shouldn't upgrade. The initial search doesn't look promising: https://github.com/dependabot/dependabot-core/issues/4364

## Tests

Smoke and new test validated locally.

## Checklist


nrcventura:
> * Find a way for `dependabot` to ignore the test project: it needs to use an old version that we shouldn't upgrade.

Would it be possible to simply tell dependabot to ignore the one outdated dependency in that project? I do not know if adding that ignore configuration for that subdirectory will have higher precedence than the configuration for the parent directory.

nrcventura:
> > Would it be possible to simply tell dependabot to ignore the one outdated dependency in that project?
> 
> I don't know the best alternative here. One thought that crossed my mind was to break the build somehow if the version is changed. However, the ideal, seems to me, for `dependabot` to never open PRs to that specific dependency on that specific project.

Maybe adding a comment to that dependency will help us for now. If we at least do that, there is a chance that we will read that comment if dependabot submits a PR to update that dependency.

nrcventura (approved):

rajkumar-rangaraj (approved):

Kielek (approved):

Issue #2480, opened 2023-04-19
Configuration to control fail fast behavior during application startup


nrcventura:
# Feature Request

**Are you requesting automatic instrumentation for a framework or library? Please describe.**
No

**Is your feature request related to a problem? Please describe.**
The Rules Engine is useful for detecting issue that we are aware of. However, there may be some edge cases that we are not aware of that could cause a crash during application startup. 

**Describe the solution you'd like**
We should provide a configuration that allows the application to handle errors more gracefully during application startup and prevent the application from crashing. For example, it may be useful to fail quickly and crash in a dev or test environment, but in production we do not want to cause the application to crash.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

pellared:
Related: 
- https://github.com/open-telemetry/opentelemetry-specification/issues/2817
- https://github.com/open-telemetry/opentelemetry-specification/issues/2005

Kielek:
@nrcventura,
I start looking into possible implementations of fail fast scenarios.
1. Support only things related to Rules Engine - just fail process instead of silently continue.
2. Unknown issues like https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/a4116522f43115bea6692205502541c98ffbc46e/src/OpenTelemetry.AutoInstrumentation.StartupHook/StartupHook.cs#L65-L69
3. Issues related to paring not supported configuration  - like passing to `OTEL_LOG_LEVEL` `SomeNotSupportedValue`.

I think we could introduce `OTEL_DOTNET_AUTO_FAIL_FAST_ENABLED` to cover all 3 cases, but before implementing this I would like to know your opinion.

nrcventura:
Those are the things that I am aware of, and I think that having a single setting like you described to cover all of those scenarios makes sense to me. I do not think that it will be helpful to have separate settings for each of those scenarios.

When someone is testing things, I think they will want to enable fail fast to make it easier to diagnose problems without having to read through log files. However, in production environments they will want fail fast to be disabled. To be safe, we may want the default setting for fail fast to be "false/off".

Kielek:
One more thing. Part of the parsing is done on the SDK/libraries side. I do not think that we will be able to introduce such feature in our RC timeframe/1.5.0 of OTel SDK/API release. I will implement and document that not all cases are handled by this.

nrcventura:
[...]

PR #2524, opened 2023-05-09
Support OTEL_DOTNET_AUTO_FAIL_FAST_ENABLED for StartupHook

Kielek:
## Why

Towards #2480

## What

`OTEL_DOTNET_AUTO_FAIL_FAST_ENABLED` for StartupHook

## Tests

CI + Manual testing + added smoke test

## Checklist

## Notes

Further work:
 - parsing values (our code only)
 - fail fast for native code


test/IntegrationTests/SmokeTests.cs
```diff
@@ -89,6 +89,28 @@ public void ApplicationIsExcluded()
         VerifyTestApplicationNotInstrumented();
     }
 
+#if NET6_0_OR_GREATER
+    [Fact]
+    [Trait("Category", "EndToEnd")]
+    public void ApplicationFailFast()
```
nrcventura, line 95: Should we also add a test that confirms that we do not crash the application?

test/IntegrationTests/SmokeTests.cs
```diff
@@ -89,6 +89,28 @@ public void ApplicationIsExcluded()
...
+    {
+        SetEnvironmentVariable("OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES", $"dotnet,dotnet.exe,{EnvironmentHelper.FullTestApplicationName},{EnvironmentHelper.FullTestApplicationName}.exe");
+        SetEnvironmentVariable("OTEL_DOTNET_AUTO_FAIL_FAST_ENABLED", "true");
```
pjanotti, line 98: The test is showing a problem: if some processes are excluded one would expect them to keep running without instrumentations, but, with the fail fast these processes can't be started. At first it seems that the process exclusion shouldn't be a rule like the others.

test/IntegrationTests/SmokeTests.cs
Kielek: Few lines above there is a test ApplicationIsExcluded which is doing exactly what you are asking for.

Should I add here some comment refering this test?

test/IntegrationTests/SmokeTests.cs
pjanotti: I agree this probably won't be a frequent/common case. However, it is a (small) bug. I'm fine with fixing it separately, but, likely something worth fixing before GA.

test/IntegrationTests/SmokeTests.cs
nrcventura: I think that it is not intuitive that the `ApplicationIsExluded` test covers that scenario. I think that a comment is necessary (at a minimum). Ideally we would have a separate test so that the 2 tests can evolve independently.

Kielek: New test added. For now, the content is exactly the same as for old test, but I like the idea to be open for evolving.

test/IntegrationTests/SmokeTests.cs
Kielek: I think it is feature (FAIL_FAST takes precedence over another settings) not a bug.
I documented behavior.
Please let me know if you are fine with this 83c684f7818bd480241c24fd8954005a4b4a6e6a

rajkumar-rangaraj (approved):
LGTM.

docs/config.md
```diff
@@ -52,11 +53,16 @@ However, if given setting supports it, then:
...
+| `OTEL_DOTNET_AUTO_FAIL_FAST_ENABLED` | Enables possibility to fail process when automatic instrumentation cannot be executed. It is designed for debugging purposes. It should not be used in production environment. \[1\]                                               | `false`       | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
+| `OTEL_LOG_LEVEL`                     | SDK log level. (supported values: `none`,`error`,`warn`,`info`,`debug`)                                                                                                                                                            | `info`        | [Stable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md)       |
+
+\[1\] If `OTEL_DOTNET_AUTO_FAIL_FAST_ENABLED` is set to `true` then processes
```
pjanotti, line 63: 👍

pjanotti (approved):

nrcventura (approved):

PR #2527, opened 2023-05-11
Support OTEL_DOTNET_AUTO_FAIL_FAST_ENABLED in native code

Kielek:
## Why

Towards #2480

## What

`OTEL_DOTNET_AUTO_FAIL_FAST_ENABLED` for Native code

## Tests

CI + Manual testing + extended smoke test

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
@@ -38,15 +38,15 @@ using namespace std::chrono_literals;
 #include "netfx_assembly_redirection.h"
 #endif
 
-#define FailProfiler(LEVEL, MESSAGE)    \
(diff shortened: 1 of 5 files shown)


pjanotti (approved):
I guess that a macro is being used because of the Logger::Level. Is it possible to use `Logger::Log` and pass the level instead? This way we could use a static inline. Other than that LGTM.

nrcventura (approved):

rajkumar-rangaraj (approved):

zacharycmontoya (approved):

Kielek:
> I guess that a macro is being used because of the Logger::Level. Is it possible to use `Logger::Log` and pass the level instead? This way we could use a static inline. Other than that LGTM.

I do no found any good way to pass the log level, without introducing switch-case method calling then appropriate methods.
For now, I will keep it as it is.

PR #2532, opened 2023-05-16
FailFast - configuration

### LINKED ISSUE #2480: Configuration to control fail fast behavior during application startup

Kielek:
## Why

Fixes #2480

## What

Support for Fail Fast while parsing data.

## Tests

CI + newly created tests.

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Configurations/FailFastSettings.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Configurations/FailFastSettings.cs
@@ -1,0 +1,27 @@
+// <copyright file="FailFastSettings.cs" company="OpenTelemetry Authors">
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+namespace OpenTelemetry.AutoInstrumentation.Configurations;
+
+internal class FailFastSettings : Settings
+{
+    public bool FailFast { get; private set; }
+
+    protected override void OnLoad(Configuration configuration)
+    {
+        FailFast = configuration.GetBool(ConfigurationKeys.FailFast) ?? false;
+    }
+}
--- a/src/OpenTelemetry.AutoInstrumentation/Configurations/Settings.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Configurations/Settings.cs
@@ -28,10 +28,10 @@ internal abstract class Settings
     /// </summary>
     public OtlpExportProtocol? OtlpExportProtocol { get; private set; }
 
-    public static T FromDefaultSources<T>()
+    public static T FromDefaultSources<T>(bool failFast)
         where T : Settings, new()
     {
-        var configuration = new Configuration(new EnvironmentConfigurationSource());
+        var configuration = new Configuration(failFast, new EnvironmentConfigurationSource(failFast));
         var settings = new T();
         settings.Load(configuration);
         return settings;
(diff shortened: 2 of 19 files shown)


RassK (approved):

src/OpenTelemetry.AutoInstrumentation/Configurations/Settings.cs
```diff
@@ -28,10 +28,10 @@ internal abstract class Settings
     /// </summary>
     public OtlpExportProtocol? OtlpExportProtocol { get; private set; }
 
-    public static T FromDefaultSources<T>()
+    public static T FromDefaultSources<T>(bool failFast)
         where T : Settings, new()
     {
-        var configuration = new Configuration(new EnvironmentConfigurationSource());
+        var configuration = new Configuration(failFast, new EnvironmentConfigurationSource(failFast));
```
zacharycmontoya, line 34: This seems like a chicken and egg problem, trying to embed a `failFast` value into the configuration source even though its value can be provided from a configuration source. I'm not requesting any changes, but this looks like an interesting circular dependency

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -242,6 +244,11 @@ private static void AddLazilyLoadedMetricInstrumentations(LazyInstrumentationLoa
                     break;
                 default:
                     Logger.Warning($"Configured metric instrumentation type is not supported: {instrumentation}");
+                    if (FailFastSettings.Value.FailFast)
```
zacharycmontoya, line 247: Will this be overridden if the FAIL FAST environment variable is set? Is there already a smoke test that covers this scenario?

zacharycmontoya (approved):

nrcventura (approved):

src/OpenTelemetry.AutoInstrumentation/Configurations/TracerSettings.cs
```diff
@@ -63,7 +63,7 @@ internal class TracerSettings : Settings
     /// <summary>
     /// Gets the instrumentation options.
     /// </summary>
-    public InstrumentationOptions InstrumentationOptions { get; private set; } = new(new Configuration());
+    public InstrumentationOptions InstrumentationOptions { get; private set; } = new(new Configuration(false));
```
pjanotti, line 66: Is it possible to use a named parameter consistently when creating types that take it as a parameter?

src/OpenTelemetry.AutoInstrumentation/Configurations/FailFastSettings.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace OpenTelemetry.AutoInstrumentation.Configurations;
+
+internal class FailFastSettings : Settings
+{
+    public bool FailFast { get; private set; }
+
+    protected override void OnLoad(Configuration configuration)
+    {
+        FailFast = configuration.GetBool(ConfigurationKeys.FailFast) ?? false;
```
pjanotti, line 25: 😄 can we make fail fast fail fast if fail fast fail?

test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/ConfigurationTests.cs
```diff
@@ -35,7 +35,7 @@ private enum TestEnum
     [Fact]
     public void ParseEnabledEnumList_Default_Enabled()
     {
-        var source = new Configuration(new NameValueConfigurationSource(new NameValueCollection()));
+        var source = new Configuration(false, new NameValueConfigurationSource(false, new NameValueCollection()));
```
pjanotti, line 38: For the tests, I don't think that naming the parameter is worth it.

pjanotti (approved):

src/OpenTelemetry.AutoInstrumentation/Configurations/TracerSettings.cs
Kielek: It was the only place under src with non-named parameter. Fixed.

src/OpenTelemetry.AutoInstrumentation/Configurations/FailFastSettings.cs
Kielek: :D no!

Issue #1865, opened 2022-12-21
Add arm64 build for native components


pjanotti:
We removed this quite some time ago since the CI used at the time didn't support arm64. As time goes by arm64 is getting more popular and we should build our native components to support arm64.

pjanotti:
https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1869

DerrickMartinez:
Is there any guidance on how to build it with arm64 support?

baterja:
https://github.blog/changelog/2023-10-30-accelerate-your-ci-cd-with-arm-based-hosted-runners-in-github-actions/

arm64 runners are there. What about reviving this thread? :)

RassK:
@baterja we just got some from actuated, see https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3277

PR #2536, opened 2023-05-16
Add build fixes for arm64

### LINKED ISSUE #2181: Add arm64 architecture to the shared store
The list at [dotnet/runtime]( https://github.com/dotnet/runtime/blob/42ef1970047ace8972e22b13e5b05a7e2be656d2/src/native/corehost/hostmisc/utils.cpp#LL201-LL209) has other values but because of Apple M1 boxes, we may want to provide `arm64` by default.

Adding all `arch` to the shared store doesn't seem scalable, ideally, we would have a way to automatically select the proper `|arch|/|tfm|` pair upon installing on the box. However, as mentioned, `arm64` may be popular enough to justify its addition even without the general mechanism.

**Temporary Workaround**: rename `$DOTNET_SHARED_STORE/x64/` to `$DOTNET_SHARED_STORE/arm64/`. The assemblies under the folder are IL only so [source instrumentations](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/design.md#library-instrumentations) are expected to work without issues. The native component is not supported on arm64 yet and given that the default rule engine will fail startup if it is enabled it needs to be disabled removing or setting the environment variable `CORECLR_ENABLE_PROFILING` to `0`.

[EDIT 01: Update workaround since it is not going to work with the default rule engine checks.]

RassK:
## Why

Fixes #2181
Towards #1865

## What

Fixes to be able to build and run arm64.

## Tests

* Existing
* Manual: Cortex A72 - arm64, running Raspbian (Debian) 64bit 

## Next

* Fix arm64 distribution folder (currently distributes to linux-x64 but should be linux-arm64).[^1]
* Build arm64 distribution in GHA.

[^1]: https://learn.microsoft.com/en-us/dotnet/core/rid-catalog#linux-rids


pjanotti:
Related to:
> Fix arm64 distribution folder (currently distributes to linux-x64 but should be linux-arm64)

Notice that it also needed to be added to the Native.Runtime NuGet package: just require the publication of the arm64 artifacts.

pjanotti (approved):
Looks good so far. Perhaps let's merge as it is and do the other steps incrementally?

pellared (approved):

Kielek:
I would like to merge it after 1.0 release. We have agreement that we can do this release without additional focus on arm64

pellared:
> I would like to merge it after 1.0 release. We have agreement that we can do this release without additional focus on arm64

Can't we simply say that the support for arm64 is experimental?

Kielek:
We do not have CI for this, also native part is missing. We have agreed that we should have whole support or nothing.

Kielek (changes requested):
After discussion with Rasmus, I am blocking merge till 1.0.0 release. Then we can focus on next architecture including native code part.

RassK:
Will recreate it after 1.0.0 release, way too old and missing bits.

Issue #2463, opened 2023-04-17
Revisit WCF instrumentation


Kielek:
# Feature Request

**Are you requesting automatic instrumentation for a framework or library? Please describe.**

Make WCF instrumentation more user-friendly.

**Is your feature request related to a problem? Please describe.**

For now, usage requires a lot of manual modification either - xml configuration (.NET Framework) or even compilation for .NET.
For details see https://github.com/Kielek/opentelemetry-dotnet-instrumentation/blob/75b3c1f768cdbbcd1b9f97a29ced0199259cf2a4/docs/wcf-config.md#L3.

**Describe the solution you'd like**

Fully bytecode instrumentation for WCF communication both on the .NET Framework and .NET side (server and client).
It should fully remove configuration steps.

**Describe alternatives you've considered**

Current instrumentation. Which kills most of the benefits from Automatic Instrumentation usage.

**Additional context**

Server side on .NET can be handled as a separate step.
Should we have separate configs to enable particular instrumentations, such us: `WCFCLIENT`, `WCFSERVIER`, `WCFCORECLIENT` and `WCFCORESERVER`?

pellared:
> Should we have separate configs to enable particular instrumentations, 

I think this is nice to have.

We could also consider having having byte code an source code WCF instrumentations if a user would not be able to set the .NET Profiler.

Kielek:
Tasks needed for 1.0.0-rc.1/GA are done.

Some follow ups post GA:
* #2673 
* #2672 
* #2671 
* #2665 
* #2664 
* #2663
* #2662

PR #2541, opened 2023-05-17
[WCF] NetFx client bytecode instrumentation

lachmatt:
## Why

Towards https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2463

## What

WCF client bytecode instrumentation for .Net framework.

Removes the manual step (configuration change) required to enable instrumentation.

Targets the same type/method as NewRelic's dotnet [agent](https://github.com/newrelic/newrelic-dotnet-agent/blob/181a628ff1cb7a0f0b7a347378644782f085f3ab/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Wcf3/Instrumentation.xml#L28).

## Tests

Existing test.

## Checklist


nrcventura (approved):
I like this approach. It simplifies the setup process, and still allows us to leverage the existing WCF instrumentation package. 

To further improve things (in a separate effort) we can support more binding types if we update the instrumentation library to support the different "header" types that the different bindings rely on. Some example code can be found in https://github.com/newrelic/newrelic-dotnet-agent/blob/181a628ff1cb7a0f0b7a347378644782f085f3ab/src/Agent/NewRelic/Agent/Extensions/Providers/Wrapper/Wcf3/ChannelFactoryWrapper.cs#L140-L158. In particular, the current code only looks at the "Headers" collection on the message, but some bindings do not use that collection, and instead rely on the HttpRequestRequestHeaders object that is defined as a "Property" on the "Properties" collection of the message. Similarly the service code we need to be updated to look for the trace context and baggage from the two possible locations (with some appropriate safe guards to prevent triggering exceptions).

Kielek (approved):
Please remember to create changelog entry, when you finish job with WCF improvements.

PR #2554, opened 2023-05-21
NuGet packages: do not bring instrumentation packages by default

pjanotti:
## Why

Avoid bringing unnecessary instrumentation and the instrumented packages when adding a reference to the `OpenTelemetry.AutoInstrumentation` package. 

Fixes #2286 [NuGet packages: separate instrumentation packages for fine tuning deployment](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2286)

## What

Before this change, installing the `OpenTelemetry.AutoInstrumentation` package meant that the files needed to instrument libraries like `MongoDB.Driver.Core` and `StackExchange.Redis` were always copied to the application's build. After the change, the instrumentation and target libraries are not added by default, however, after adding a reference to `OpenTelemetry.AutoInstrumentation` the application build will, if necessary, request the user to either ignore the request or add the required instrumentation package (aka an adapter package).

Initially, only the packages `MongoDB.Driver.Core` and `StackExchange.Redis` are being removed from the `OpenTelemetry.AutoInstrumentation.Runtime.Managed` package, note that the distribution folder is not affected by this change. The removed packages are being added to a `.targets` file shipped with the `OpenTelemetry.AutoInstrumentation.BuildTasks` package. This package is a build-time-only package that checks for ranges of specified NuGet packages and fails the build if an instrumentation package should be added to the build.

## Tests

- Unit tests for the new build task.
- Add a NuGet package test that validates the expected developer's workflow when adding the `OpenTelemetry.AutoInstrumentation`. 
- 
## Checklist

 
~~- [ ] `CHANGELOG.md` is updated.~~ CHANGELOG.md should be updated when the NuGet packages are published via nuget.org


.cspell/dot-net.txt
```diff
@@ -1,38 +1,39 @@
...
-idls
-midl
-stdlibc
-corelib
-ASPNETCORE
-HOSTINGSTARTUPASSEMBLIES
-Bootstrapper
-NETFX
-iisreset
+Xunit
```
pjanotti, line 39: This was the added word but VS put all the words in order.

src/OpenTelemetry.AutoInstrumentation.BuildTasks/OpenTelemetry.AutoInstrumentation.BuildTasks.targets
```diff
@@ -0,0 +1,36 @@
+<Project>
+
+    <ItemGroup Label="InstrumentationTargetItems">
```
pjanotti, line 3: Two follow-up issues:

1. the generation of entries here can be auto-generated from `src/OpenTelemetry.AutoInstrumentation/InstrumentationDefinitions.cs` #2555 
2. move the remaining instrumentation packages from the `Runtime,Managed` package to `InstrumentationTarget` handled by the `BuildTasks` package #2556

test/NuGetPackagesTests/InstrumentationTargetTests.cs
```diff
@@ -119,21 +119,20 @@ public void InstrumentApplication()
...
-            public static class Program
-            {
-                public static void Main()
-                {
-                    System.Console.WriteLine("Hello World!");
-                }
-            }
-            
-            """;
+        const string ProgramContent = @"
```
pjanotti, line 122: I tried to suppress dotnet-format around this via the same `#pragma warning disable format` used else where in the code but couldn't: opted to use old style in the end.

nuget/OpenTelemetry.AutoInstrumentation/contentFiles/any/any/instrument.cmd
```diff
@@ -8,6 +8,9 @@ set BASE_PATH=%~dp0
 set COR_ENABLE_PROFILING=1
 set COR_PROFILER={918728DD-259F-4A6A-AC2B-B85E1B658318}
 set COR_PROFILER_PATH=%BASE_PATH%OpenTelemetry.AutoInstrumentation.Native.dll
+:: On .NET Framework automatic assembly redirection MUST be disabled. This setting
```
nrcventura, line 11: We might want to briefly explain why it's disabled by default.

pjanotti:
It seems that doc validation is having intermittent issues reaching nuget.org, see: 

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/5179592348/jobs/9332987985#step:3:446

pjanotti:
Each time failing on a different package: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/5179592348/jobs/9333312523#step:3:448

I will give it a break and retry a bit later.

src/OpenTelemetry.AutoInstrumentation.BuildTasks/CheckForInstrumentationPackages.cs
```diff
@@ -0,0 +1,161 @@
...
+    public override bool Execute()
+    {
+        _logMessageImportance = UseVerboseLog ? MessageImportance.High : MessageImportance.Low;
+        try
+        {
+            return ExecuteImplementation();
+        }
+        catch (Exception ex)
+        {
+            Log.LogError($"{LogPrefix}{ex}");
```
lachmatt, line 70: nit: `Log.LogErrorFromException` could be used instead

src/OpenTelemetry.AutoInstrumentation.BuildTasks/CheckForInstrumentationPackages.cs
```diff
@@ -0,0 +1,161 @@
...
+        // Process each instrumentation target item.
+        foreach (var instrumentationTarget in InstrumentationTargetItems ?? Array.Empty<ITaskItem>())
+        {
+            missingAnyInstrumentationAdapterPackage |=
+                MissingInstrumentationAdapterPackage(instrumentationTarget, runtimeCopyLocalItemsDictionary);
+        }
+
+        // If missing any instrumentation adapter package then fail the build by returning false, the method
+        // Log.LogError was already called.
+        return !missingAnyInstrumentationAdapterPackage;
```
lachmatt, line 108: I think instead of tracking it manually you could simply return `Log.HasLoggedErrors`

docs/internal/using-the-nuget-packages.md
```diff
@@ -41,6 +41,15 @@ the `OpenTelemetry.AutoInstrumentation` package to your project:
 dotnet add [<PROJECT>] package OpenTelemetry.AutoInstrumentation --source <PATH_TO_AUTO_INSTRUMENTATION_PACKAGES> --prerelease
 ```
 
+If the application references packages that can be instrumented, but, require
+other packages for the instrumentation to work the build will fail and prompt
+you to either add the missing instrumentation package or to skip the
+instrumentation of the corresponding package:
```
lachmatt, line 47: I'd add an example how to skip the instrumentation (especially for scenario when instrumentation of multiple packages is to be skipped, where trying to separate package names with comma or semicolon would cause errors)

lachmatt (approved):

lachmatt:
👍

docs/internal/using-the-nuget-packages.md
pjanotti: 4396cfd1f808110481c627e8733e9f49935078d7

src/OpenTelemetry.AutoInstrumentation.BuildTasks/CheckForInstrumentationPackages.cs
pjanotti: d1b9876d7df02e869fa748fa3e19275c71d27865

src/OpenTelemetry.AutoInstrumentation.BuildTasks/CheckForInstrumentationPackages.cs
pjanotti: 10727e6e08dbb33055b637bf0e29148697d67eec

nrcventura (approved):

rajkumar-rangaraj (approved):
LGTM.

Validated the Nuget from build with dotnet CLI and VS 2022.

PR #2584, opened 2023-05-29
Fix double licensing in DependencyListGenerator

RassK:
## Why

https://github.com/cncf/foundation/blob/main/copyright-notices.md#what-about-third-party-code

## What

There is already a separate license file in `DotNetOutdated` folder. 

> If a file only contains code that originates from a third party source who didn't contribute it themselves, then you would not want to add the notices above.


tools/DependencyListGenerator/GlobalSuppressions.cs
```diff
@@ -18,3 +18,4 @@
 
 [assembly: SuppressMessage("StyleCop.CSharp.DocumentationRules", "SA1600:Elements should be documented", Justification = "Reviewed.")]
 [assembly: SuppressMessage("StyleCop.CSharp.DocumentationRules", "SA1604:Element documentation should have summary", Justification = "Reviewed.")]
+[assembly: SuppressMessage("StyleCop.CSharp.DocumentationRules", "SA1633:File should have header", Justification = "Preserving original license.")]
```
RassK, line 21: I really would like to scope it only for a namespace or a folder, but somehow stylecop does not address it.

pellared (approved):

rajkumar-rangaraj (approved):

nrcventura (approved):

PR #2614, opened 2023-06-07
Instrumentation generator

### LINKED ISSUE #2428: Generate InstrumentationDefinitions
Consider creating separate GitHub issue.

_Originally posted by @Kielek in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2280#discussion_r1156968030_

Kielek:
## Why

Fixes #2428

## What

Generate instrumentation definitions for production code.

Based on  https://andrewlock.net/creating-a-source-generator-part-1-creating-an-incremental-source-generator/ , https://andrewlock.net/creating-a-source-generator-part-6-saving-source-generator-output-in-source-control/ and #1466

## Tests

CI

Based on https://andrewlock.net/creating-a-source-generator-part-2-testing-an-incremental-generator-with-snapshot-testing/ - snapshot testing should be considered as best option. I think that saving generated code to the git is good enough test.

## Checklist


nrcventura (approved):

andrewlock:
Nice! FYI, I have updates to make to that series, as there's more perf you're leaving on the table! 😄 

Here are the highlights: 

- Use `ForAttributeWithMetadataName` - it gives a big perf boost over `CreateSyntaxProvider`. You can check out [this generator for an example](https://github.com/andrewlock/NetEscapades.EnumGenerators/blob/main/src/NetEscapades.EnumGenerators/EnumGenerator.cs#L22).
- Never return ISymbol or similar things from the chained values (Doesn't look like you are) 👍.
-  Always return immutable record types that are compared by value (explicitly if necessary). This is essential for caching. Everything in these records should be value types, or records (i.e. compared by value). Your `IntegrationToGenerate` class will be compared by _reference_, so will never be cached
- Avoid LINQ
- Never use the `CompilationProvider` in later stages, as this changes with every keystroke, so will completely destroy any caching you do achieve!

I _really_ need to get on with it and write another blog post about this stuff 😂

src/SourceGenerators/InstrumentationDefinitionsGenerator.cs
```diff
@@ -0,0 +1,280 @@
...
+
+    private static ClassDeclarationSyntax? GetClassesMarkedByInstrumentMethodAttribute(GeneratorSyntaxContext context)
+    {
+        var classDeclarationSyntax = (ClassDeclarationSyntax)context.Node;
+
+        foreach (var attributeListSyntax in classDeclarationSyntax.AttributeLists)
+        {
+            foreach (var attributeSyntax in attributeListSyntax.Attributes)
+            {
+                if (attributeSyntax.Name.ToString() is "InstrumentMethod" or "InstrumentMethodAttribute" or "GraphQLExecuteAsync" or "AdoNetTargetSignature")
```
pjanotti, line 271: You can remove ` or "GraphQLExecuteAsync" or "AdoNetTargetSignature"` from the condition.

You need only one of `"InstrumentMethod" or "InstrumentMethodAttribute"`, no?

src/SourceGenerators/InstrumentationDefinitionsGenerator.cs
```diff
@@ -0,0 +1,280 @@
...
+using Microsoft.CodeAnalysis.Text;
+
+namespace SourceGenerators;
+
+[Generator]
+public class InstrumentationDefinitionsGenerator : IIncrementalGenerator
+{
+    public void Initialize(IncrementalGeneratorInitializationContext context)
+    {
+        var classesWithAttributes = context.SyntaxProvider.CreateSyntaxProvider(
```
pjanotti, line 30: Any reason to not use [SyntaxValueProvider.ForAttributeWithMetadataName<T> Method](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.syntaxvalueprovider.forattributewithmetadataname?view=roslyn-dotnet-4.6.0)?

pjanotti:
```suggestion
        var instrumentationClasses = context.SyntaxProvider.CreateSyntaxProvider(
```

src/SourceGenerators/InstrumentationDefinitionsGenerator.cs
```diff
@@ -0,0 +1,280 @@
...
+                static (node, _) => IsAttributedClass(node),
+                static (node, _) => GetClassesMarkedByInstrumentMethodAttribute(node))
+            .Where(static m => m != null);
+
+        IncrementalValueProvider<(Compilation, ImmutableArray<ClassDeclarationSyntax?>)> compilationAndClasses
```
pjanotti, line 35: What does this step does?

src/SourceGenerators/InstrumentationDefinitionsGenerator.cs
```diff
@@ -0,0 +1,280 @@
...
+            = context.CompilationProvider.Combine(classesWithAttributes.Collect());
+
+        context.RegisterSourceOutput(
+            compilationAndClasses,
+            static (context, source) => Generate(source.Item1, source.Item2, context));
+    }
+
+    private static void Generate(Compilation compilation, ImmutableArray<ClassDeclarationSyntax?> classesWithAttributes, SourceProductionContext context)
```
pjanotti, line 43: If my understanding is correct:
```suggestion
    private static void Generate(Compilation compilation, ImmutableArray<ClassDeclarationSyntax?> instrumentationClasses, SourceProductionContext context)
```

src/SourceGenerators/InstrumentationDefinitionsGenerator.cs
```diff
@@ -0,0 +1,280 @@
...
+    {
+        if (classesWithAttributes.IsDefaultOrEmpty)
+        {
+            return;
+        }
+
+        var distinctClasses = classesWithAttributes.Distinct();
```
pjanotti, line 50: Why distinct is necessary here?

src/SourceGenerators/InstrumentationDefinitionsGenerator.cs
```diff
@@ -0,0 +1,280 @@
...
+        context.RegisterSourceOutput(
+            compilationAndClasses,
+            static (context, source) => Generate(source.Item1, source.Item2, context));
+    }
+
+    private static void Generate(Compilation compilation, ImmutableArray<ClassDeclarationSyntax?> classesWithAttributes, SourceProductionContext context)
+    {
+        if (classesWithAttributes.IsDefaultOrEmpty)
+        {
+            return;
```
pjanotti, line 47: hum ... should a warning be issued in this case? It seems valid to build everything without bytecode instrumentations, but, it seems atypical...

src/SourceGenerators/InstrumentationDefinitionsGenerator.cs
```diff
@@ -0,0 +1,280 @@
...
+        context.AddSource("InstrumentationDefinitions.g.cs", SourceText.From(result, Encoding.UTF8));
+    }
+
+    private static (IReadOnlyCollection<IntegrationToGenerate> TraceIntegrations, IReadOnlyCollection<IntegrationToGenerate> MetricsIntegrations, IReadOnlyCollection<IntegrationToGenerate> LogsIntegrations) GetIntegrationsToGenerate(Compilation compilation, IEnumerable<ClassDeclarationSyntax?> classes, CancellationToken contextCancellationToken)
+    {
+        var instrumentAttribute = compilation.GetTypeByMetadataName("OpenTelemetry.AutoInstrumentation.Instrumentations.InstrumentMethodAttribute");
+
+        if (instrumentAttribute == null)
+        {
+            return (Array.Empty<IntegrationToGenerate>(), Array.Empty<IntegrationToGenerate>(), Array.Empty<IntegrationToGenerate>());
```
pjanotti, line 64: How this can happen?

pjanotti (commented):
Minor stuff and some Qs, LG in general.

src/Directory.Packages.props
```diff
@@ -28,6 +28,12 @@
     <PackageVersion Include="OpenTelemetry.ResourceDetectors.Container" Version="1.0.0-beta.3" />
   </ItemGroup>
 
+  <!-- Versions from OpenTelemetry.AutoInstrumentation.SourceGenerators.csproj -->
```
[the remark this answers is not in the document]
Kielek: Long project name leads to too long path exception on windows.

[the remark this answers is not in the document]
Kielek: 7b0f29e0c03ed4f0521fa45cf6ba602d33582f3b

src/SourceGenerators/SourceGenerators.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <EnablePublicApi>false</EnablePublicApi>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
+    <IsRoslynComponent>true</IsRoslynComponent>
+    <NoWarn>$(NoWarn);SA1600;CS1591</NoWarn>
```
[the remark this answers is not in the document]
Kielek: 80974e2705f34789dd189c692856107d45f860a4

src/SourceGenerators/InstrumentationDefinitionsGenerator.cs
Kielek: f02feab0ac6f361fbf3b6babe83aa2e3aa99abbf

src/SourceGenerators/InstrumentationDefinitionsGenerator.cs
Kielek: No longer used.

src/SourceGenerators/InstrumentationDefinitionsGenerator.cs
Kielek: Copy paste from linked instruction. Removed locally, everything is working fine.

src/SourceGenerators/InstrumentationDefinitionsGenerator.cs
Kielek: It is just fast handle scenario taken from linked example. I do not think it will harm us, also adding warnings is not what I would like to do becasue adding bytecode instrumentation is not typical task for us. If new instrumentation will not occur locally we should detect it on

src/SourceGenerators/InstrumentationDefinitionsGenerator.cs
Kielek: It shouldn't if referenced only by OpenTelemetry.AutoInstrumentation project. 

Now it is handled differently.

Kielek:
@andrewlock, @pjanotti thank you for review. I hope that I do not missed any recommendation. Please review once more time.

test/IntegrationTests/Helpers/TestHelper.cs
```diff
@@ -42,9 +42,6 @@ protected TestHelper(string testApplicationName, ITestOutputHelper output, strin
 
     protected ITestOutputHelper Output { get; }
 
-    /// <summary>
```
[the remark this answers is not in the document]
pjanotti: I'm fine with all or none approach. That said, the content of this comment seems worth to keep, could you please make it a standard comment inside the method.

pjanotti (approved):

src/SourceGenerators/IntegrationToGenerate.cs
```diff
@@ -0,0 +1,25 @@
...
+// </copyright>
+
+using System.Collections.Immutable;
+
+namespace SourceGenerators;
+
+internal readonly record struct IntegrationToGenerate(string IntegrationType, ImmutableArray<TargetToGenerate> Targets)
+{
+    public readonly string IntegrationType = IntegrationType;
+    public readonly ImmutableArray<TargetToGenerate> Targets = Targets;
```
andrewlock, line 24: Unfortunately `ImmutableArray<>` doesn't implement value equality - I suggest you use a wrapper type that tests for value equality, something like this:

https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace.SourceGenerators/Helpers/EquatableArray.cs

PR #2657, opened 2023-06-15
InstrumentationGenerator - use EquatableArray for better performance

Kielek:
## Why

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2614#discussion_r1229208407

## What

InstrumentationGenerator - use EquatableArray for better performance

## Tests

CI

## Checklist


pjanotti:
> InstrumentationGenerator - use EquatableArray for better performance

Could you clarify how this improves performance?

Kielek:
Two messages from @adnrewlock

> Always return immutable record types that are compared by value (explicitly if necessary). This is essential for caching. Everything in these records should be value types, or records (i.e. compared by value). Your IntegrationToGenerate class will be compared by reference, so will never be cached

> Unfortunately ImmutableArray<> doesn't implement value equality - I suggest you use a wrapper type that tests for value equality, something like this: https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace.SourceGenerators/Helpers/EquatableArray.cs


Probably followin description is oversimplified, but it should describe general idea.:
1. Record struct is implicitly implementing `IEquatable<>`
1. Generator can be executed even on every key stroke while working with IDE
1. If possible, eah step of Initialize method should returns/accept data which met condition decribed by Andrew.
1. If data are the same, there is no need to execute further steps => cache is used.

pjanotti:
Thanks for giving the context @Kielek.

What is not clear to me yet is who/where the check for "same data as before" is happening. I don't recall that happening in the generator code - maybe I misremember it. Anyway, I can review it as it is and we can learn more about it later.

src/SourceGenerators/EquatableArray.cs
```diff
@@ -0,0 +1,70 @@
...
+    public EquatableArray(T[] array)
+    {
+        _array = array;
+    }
+
+    public int Count => _array.Length;
+
+    public bool Equals(EquatableArray<T> array)
+    {
+        return _array.AsSpan().SequenceEqual(array.AsSpan());
```
pjanotti, line 36: I don't think this matter for the specific case at hand, but, seems a good thing for the type in general.

```suggestion
        return _array == array || _array.AsSpan().SequenceEqual(array.AsSpan());
```

pjanotti (approved):
LGTM, very minor suggestion.

nrcventura (approved):

Issue #2464, opened 2023-04-17
GraphQL - instrumentation for 7.5.0+ versions


Kielek:
# Feature Request

**Are you requesting automatic instrumentation for a framework or library? Please describe.**

GraphQL added native support for DiagnosticSource with OTel AutoInstrumentation friendly way in 7.5.0+ versions.
https://github.com/graphql-dotnet/graphql-dotnet/pull/3467

Support for this version should be added to this distribution.

**Additional context**

Still we Will have huge gap 3.0.*-7.3.*. It can be covered separately.

RassK:
Detected multiple paths to resolve this, @open-telemetry/dotnet-maintainers any preferences?

```csharp
public void ConfigureServices(IServiceCollection services)
{
    services.AddGraphQL(b => b
        ... // General setup here
        .UseTelemetry(options => // <-- not very auto instrumentation friendly
        {
            options.RecordDocument = false; // must come from OTEL_DOTNET_AUTO_GRAPHQL_SET_DOCUMENT
        }));
}
```

1. Let user call `.UseTelemetry()`
P: safe, no double service registrations
N: cannot be called as "auto instrumentation", requires manual code modifications
N:  not reusing GraphQL env vars

2. Use bytecode instrumentation to tap into `.AddGraphQL()` and register `.UseTelemetry()` automatically.
N: Makes source code instrumentation to byte code instrumentation dependent
N: Could cause double `IServiceCollection` registration (and crash?)

3. Reuse `OpenTelemetry.AutoInstrumentation.AspNetCoreBootstrapper` and call internals of `.UseTelemetry()` with reflection.
N: Complicated reflection
N: Could cause double `IServiceCollection` registration (and crash?)
N: `.UseTelemetry()` can change

4. ???

I definitely see, it's much better than MongoDB instrumentation, no additional dependencies needed. In the same time I think we need to workout auto instrumentation friendly way, how to set service to publish activities with simple code and consume these. (At least consuming part is quite good if it works like this only -> `services.AddOpenTelemetry().WithTracing(b => b           .AddSource("GraphQL"))`

RassK:
Detected multiple paths to resolve this, @open-telemetry/dotnet-instrumentation-maintainers any preferences?

```csharp
public void ConfigureServices(IServiceCollection services)
{
    services.AddGraphQL(b => b
        ... // General setup here
        .UseTelemetry(options => // <-- not very auto instrumentation friendly
        {
[...]

PR #2681, opened 2023-06-20
GraphQL v7.5.0 instrumentation

RassK:
## Why

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2464

## What

* Adds native support for GraphQL version 7.5.0+
* Removes existing versions [2.3-2.4] support.

## Tests

* Added new set of integration tests

## Follow-ups

* https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2699

## Checklist


test/IntegrationTests/LibraryVersions.g.cs
```diff
@@ -47,6 +47,14 @@ public static class LibraryVersion
 #else
         new object[] { "2.3.0" },
         new object[] { "2.4.0" },
+#endif
+    };
+    public static readonly IReadOnlyCollection<object[]> GraphQLNativeSupport = new List<object[]>
```
pjanotti, line 52:
```suggestion
    public static readonly IReadOnlyCollection<object[]> GraphQLNativeInstrumentation = new List<object[]>
```

pjanotti: or perhaps
```suggestion
    public static readonly IReadOnlyCollection<object[]> GraphQLNativelyInstrumented = new List<object[]>
```

test/IntegrationTests/GraphQLByteCodeTests.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+    }
+
+    public override bool SupportsExceptions => true;
+
+    public override string InstrumentationScope => "OpenTelemetry.AutoInstrumentation.GraphQL";
+
+    [Theory]
+    [Trait("Category", "EndToEnd")]
+    [MemberData(nameof(GetData), new object[] { false })]
```
pjanotti, line 34: The `MemberData` here is pretty opaque, a comment or a constant to convey the meaning of the boolean would be helpful.

pjanotti: Same for the "*.NativeTests.cs"

test/IntegrationTests/GraphQLNativeTests.cs
```diff
@@ -0,0 +1,43 @@
...
+// limitations under the License.
+// </copyright>
+
+#if NET6_0_OR_GREATER
+
+using Xunit.Abstractions;
+
+namespace IntegrationTests;
+
+public class GraphQLNativeTests : GraphQLTests
```
pjanotti, line 23: Similar to other suggested renames:
```suggestion
public class GraphQLNativelyInstrumented : GraphQLTests
```

pjanotti (commented):
LGTM - a few naming suggestions, and an important Q about `netstandard2.0`.

test/test-applications/integrations/TestApplication.GraphQL/TestApplication.GraphQL.csproj
```diff
@@ -1,10 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
```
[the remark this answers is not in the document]
RassK: @pjanotti thinking now to remove all that "natively" and instead rename the old project `TestApplication.GraphQL.Legacy` so the current (newest) way will have "normal" name.

test/IntegrationTests/GraphQLByteCodeTests.cs
RassK: @pjanotti I was also concerned about this. Either a new enum or 

```csharp
new object[] { /* isNativelyInstrumented */ false }
```

any preferences?

test/test-applications/integrations/TestApplication.GraphQL/TestApplication.GraphQL.csproj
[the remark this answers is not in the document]
pjanotti: Sounds good.

Kielek:
What's about fully dropping legacy instrumentation?
We will have huge gap between [3.0, 7.5.0) with non-instrumented versions. It is unlikely that we would like to support these versions.

I do not see any important differences with supported .NET and .NET Frameworl versions for 
https://www.nuget.org/packages/GraphQL/7.5.0#supportedframeworks-body-tab
and https://www.nuget.org/packages/GraphQL/2.4.0#supportedframeworks-body-tab

docs/internal/instrumentation-libraries.md
Kielek, line 1: You have forgot also about https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/54804fef53e9522113bd77392fbbd5d80fec015d/docs/config.md#traces-instrumentations

RassK:
> What's about fully dropping legacy instrumentation? We will have huge gap between [3.0, 7.5.0) with non-instrumented versions. It is unlikely that we would like to support these versions.
> 
> I do not see any important differences with supported .NET and .NET Frameworl versions for https://www.nuget.org/packages/GraphQL/7.5.0#supportedframeworks-body-tab and https://www.nuget.org/packages/GraphQL/2.4.0#supportedframeworks-body-tab

@Kielek I think I would be fine dropping [2.0.0, 2.4.0]. Just wondering that 2.4.0 is by far the most popular version and it's around 4.5 years old. Kind of in the border range to consider as legacy. 

@Shane32 you may have better insights whether version range [2.0.0, 2.4.0] is still topical?

image

Shane32:
2.4.0 was (is?) used by some large legacy systems. Occasionally we get asked questions about that version but it is getting rare now. Those users seem to be hesitant to upgrade - again probably because it’s a legacy system, but also because 3.0 changed some fundamentals.

Also, there was a very large time gap (and feature gap) between 2.4 and 3.0 which is why there are so many more downloads of 2.4 than newer versions. Since 3.0 there has been a fairly regular release cadence, so no single version amassed a great quantity of downloads. For instance, 4.x versions total nearly the same number of downloads as 2.x. 

We also do not provide “support” for 2.4 any more, although we usually try to answer questions for newer versions when possible. Generally all fixes go into the latest version although it is fairly easy for us to hotfix 4.x and newer if a critical flaw were identified.  Not so for 2.x

Considering that OpenTelemetry is a fairly new platform (?), I wouldn’t expect users of legacy systems containing 2.4 to be adding OpenTelemetry to it. They likely have legacy telemetry systems. Or, I’d expect them to be trying to keep all their dependencies up to date. That’s just a guess.

Kielek:
@Shane32, this project is able to support only officially supported .NET Framework and .NET versiosn. It means .NET Framewrok 4.6.2+ and .NET 6+.

Older versions are not supported by OpenTelemetry SDK/API which is our base.

Kielek:
SIG meeting decisions: lets drop 2.x support for now. We are just before GA. If needed we can bring it back later.

CHANGELOG.md
```diff
@@ -33,12 +33,16 @@ This release is built on top of [OpenTelemetry .NET](https://github.com/open-tel
   Both supported only on .NET Framework.
 - `WCFCLIENT` and `WCFSERVICE` no longer requires manual modification
   of config files to include `TelemetryEndpointBehaviorExtensionElement`.
+- [GraphQL](https://www.nuget.org/packages/GraphQL/) instrumentation is now
+  supported from version 7.5.0
```
Kielek, line 37:
```suggestion
- [GraphQL](https://www.nuget.org/packages/GraphQL/) instrumentation is now
  supported from version 7.5.0.
```

CHANGELOG.md
```diff
@@ -33,12 +33,16 @@ This release is built on top of [OpenTelemetry .NET](https://github.com/open-tel
...
 
 ### Deprecated
 
 ### Removed
 
 - Removed `WCF` instrumentation for Core WCF Client working on .NET.
+- Removed [GraphQL](https://www.nuget.org/packages/GraphQL/) instrumentation
+  for versions 2.0.0-2.4.*
```
Kielek, line 45: 2.0.0-2.2.0 was never supported.

```suggestion
- Removed [GraphQL](https://www.nuget.org/packages/GraphQL/) instrumentation
  for versions 2.3.0-2.4.*.
```

docs/config.md
```diff
@@ -127,7 +127,7 @@ due to lack of stable semantic convention.
...
 | `ENTITYFRAMEWORKCORE` | [Microsoft.EntityFrameworkCore](https://www.nuget.org/packages/) **Not supported on .NET Framework**                                                                                            | ≥6.0.12            | source                  | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
-| `GRAPHQL`             | [GraphQL](https://www.nuget.org/packages/GraphQL)                                                                                                                                               | ≥2.3.0 & < 3.0.0   | bytecode                | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
+| `GRAPHQL`             | [GraphQL](https://www.nuget.org/packages/GraphQL)                                                                                                                                               | ≥7.5.0             | source                  | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
```
Kielek, line 130:
```suggestion
| `GRAPHQL`             | [GraphQL](https://www.nuget.org/packages/GraphQL) **Not supported on .NET Framework**                                                                                                           | ≥7.5.0             | source                  | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
```

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -270,9 +270,9 @@ private static void AddLazilyLoadedMetricInstrumentations(LazyInstrumentationLoa
         }
     }
 
-    private static void AddLazilyLoadedTraceInstrumentations(LazyInstrumentationLoader lazyInstrumentationLoader, PluginManager pluginManager, IReadOnlyList<TracerInstrumentation> enabledInstrumentations)
+    private static void AddLazilyLoadedTraceInstrumentations(LazyInstrumentationLoader lazyInstrumentationLoader, PluginManager pluginManager, TracerSettings tracerSettings)
```
Kielek, line 273: Why this change?

test/IntegrationTests/GraphQLTests.cs
Kielek, line 1: As we dropped bytecode instrumentation, there is no need to these cases.
Consider to simplify test to the level of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/1f77dc1eb43c4fa350203dea9b56e8ee26404b32/test/IntegrationTests/MySqlDataTests.cs

test/test-applications/integrations/TestApplication.GraphQL/Properties/launchSettings.json
```diff
@@ -12,20 +12,9 @@
       "commandName": "Project",
       "launchBrowser": false,
       "environmentVariables": {
-        "ASPNETCORE_ENVIRONMENT": "Development",
```
Kielek, line 15: What is the reason to remove easy way to attach OTel Auto?

test/Directory.Packages.props
```diff
@@ -5,9 +5,10 @@
...
     <PackageVersion Include="FluentAssertions" Version="6.11.0" />
     <PackageVersion Include="Google.Protobuf" Version="3.23.3" />
-    <PackageVersion Include="GraphQL" Version="2.4.0" />
-    <PackageVersion Include="GraphQL.Server.Transports.AspNetCore" Version="3.4.0" />
-    <PackageVersion Include="GraphQL.Server.Ui.Playground" Version="3.4.0" />
+    <PackageVersion Include="GraphQL" Version="7.5.0" />
+    <PackageVersion Include="GraphQL.MicrosoftDI" Version="7.5.0" />
+    <PackageVersion Include="GraphQL.Server.Transports.AspNetCore" Version="7.4.0" />
+    <PackageVersion Include="GraphQL.Server.Ui.Playground" Version="7.4.0" />
     <PackageVersion Include="GraphQL.StarWars" Version="1.0.0" />
```
Kielek, line 12: Remove this line. It is no loner used.

Kielek:
Everything under https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/tree/1f77dc1eb43c4fa350203dea9b56e8ee26404b32/src/OpenTelemetry.AutoInstrumentation/Instrumentations/GraphQL should be removed.

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
RassK: same method & `case TracerInstrumentation.GraphQL:`

docs/config.md
RassK: Is supported but we don't have tests.

test/IntegrationTests/GraphQLTests.cs
RassK: I started to think here that we did have those tests in .NET SDK side in case of official or community adapter library.
But if library itself is ejecting activities, then if we should also verify OTel compatibility - or just leave it for the library author?

test/test-applications/integrations/TestApplication.GraphQL/Properties/launchSettings.json
RassK: part of cleanup, outdated values - actually most of our launchsettings are clean, so I just followed the pattern there.

test/IntegrationTests/GraphQLTests.cs
Kielek: For this kind of instrumentation, we have agreed to only check if instrumentation is working for the simplest case.
Any tags verification should be on library author.

If we decide to verify on our side we should expect different spans for 2.0 and probably 2.3/2.4 when the specification will be stabilized.

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
Kielek: I have missed that. Thanks.

docs/config.md
Kielek: Fixed b663f953e38f070356c9350ef0fa0977f3fe00ee

src/OpenTelemetry.AutoInstrumentation/Configurations/EnvironmentConfigurationTracerHelper.cs
```diff
@@ -52,6 +52,7 @@ internal static class EnvironmentConfigurationTracerHelper
                 TracerInstrumentation.MySqlData => Wrappers.AddMySqlClientInstrumentation(builder, pluginManager, lazyInstrumentationLoader),
                 TracerInstrumentation.StackExchangeRedis => builder.AddSource("OpenTelemetry.Instrumentation.StackExchangeRedis"),
                 TracerInstrumentation.EntityFrameworkCore => Wrappers.AddEntityFrameworkCoreInstrumentation(builder, pluginManager, lazyInstrumentationLoader),
+                TracerInstrumentation.GraphQL => Wrappers.AddGraphQLInstrumentation(builder, pluginManager, lazyInstrumentationLoader, settings),
```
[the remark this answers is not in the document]
RassK: I think we need to make proper .NET Framework test application. Microsoft.AspnetCore is also now deprecated, so prob good time separate ASP.NET Core from running on netfx.

CHANGELOG.md
```diff
@@ -33,12 +33,16 @@ This release is built on top of [OpenTelemetry .NET](https://github.com/open-tel
   Both supported only on .NET Framework.
 - `WCFCLIENT` and `WCFSERVICE` no longer requires manual modification
   of config files to include `TelemetryEndpointBehaviorExtensionElement`.
+- [GraphQL](https://www.nuget.org/packages/GraphQL/) instrumentation is now
+  supported from version 7.5.0.
```
Shane32, line 37: GraphQL.NET 7.5 only supports OpenTelemetry on .NET 5+ because the necessary APIs do not exist within .NET Framework or .NET Standard.

docs/internal/instrumentation-libraries.md
```diff
@@ -59,7 +59,7 @@ submit an issue to request its inclusion.
...
 | [gRPC-dotnet](https://github.com/grpc/grpc-dotnet)                                |                          Partial[^grpc_client_only]                          |                                                                               | Client and service support should be added.                                                                                                            |
-| [GraphQL](https://www.nuget.org/packages/GraphQL/)                                |                     Partial (versions ≥2.3.0 & < 3.0.0)                      |                                                                               | The current instrumentation needs updates to match the semantic conventions.                                                                           |
+| [GraphQL](https://www.nuget.org/packages/GraphQL/)                                |                                     Yes                                      |                                                                               |                                                                                                                                                        |
```
Shane32, line 62: Specify version?  Specify .NET 5+?

src/OpenTelemetry.AutoInstrumentation/Loading/Initializers/GraphQLInitializer.cs
```diff
@@ -0,0 +1,54 @@
...
+    public GraphQLInitializer(PluginManager pluginManager, TracerSettings tracerSettings)
+    : base("GraphQL")
+    {
+        _pluginManager = pluginManager;
+        _tracerSettings = tracerSettings;
+    }
+
+    public override void Initialize(ILifespanManager lifespanManager)
+    {
+        var initializerType = Type.GetType("OpenTelemetry.AutoInstrumentation.Initializer, GraphQL");
```
Shane32, line 37: I'm a bit disappointed that this code is specific to GraphQL.  I would have expected this to be generic to any library -- just add an internal class with the appropriate name to the library to support auto instrumentation (when not automatically emitted by the library).  In other words, it seems that if I write another library and add a similar class named OpenTelemetry.AutoInstrumentation.Initializer, it's not going to work unless there's specific support added to OpenTelemetry.

src/OpenTelemetry.AutoInstrumentation/Loading/Initializers/GraphQLInitializer.cs
```diff
@@ -0,0 +1,54 @@
...
+        if (initializerType == null)
+        {
+            return;
+        }
+
+        var optionsType = Type.GetType("GraphQL.Telemetry.GraphQLTelemetryOptions, GraphQL")!;
+        var optionsInstance = Activator.CreateInstance(optionsType)!;
```
Shane32, line 44: The the same vein, I would have expected this to by a dynamic mapping of options -- or, to not support options at all, just as if the library emitted tracing events by default.

docs/internal/instrumentation-libraries.md
RassK: There's different doc for that https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2681/files#diff-77b3519370ff48f6caf7c316f0ece966f2f0871c40e9381d451d2c5b1c76b40fR130

But according to https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2681#discussion_r1238544573, the information has to be adjusted.

src/OpenTelemetry.AutoInstrumentation/Loading/Initializers/GraphQLInitializer.cs
RassK: Unfortunately we cannot scan every loaded library in the startup time, that's going to add up. 

The automated way is to use `OTEL_DOTNET_AUTO_TRACES_ADDITIONAL_SOURCES` https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/config.md#additional-settings. It's expected that every library in the future will use activities by default without custom bootstrapping.

Initializer is already one of many ways to bootstrap activity broadcasting. Definitely the newest way, still not sure how many are going to adapt this pattern.

src/OpenTelemetry.AutoInstrumentation/Loading/Initializers/GraphQLInitializer.cs
RassK: This should become better when OTel config file is finally speced. Currently introducing new envs is also on hold.
But in the end environment controller (end user) and OTel vendor (via plugins) must have control over options. Some of the options can control sending sensitive data over the wire, which is extremely important to control.

test/IntegrationTests/GraphQLTests.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2699

Kielek (approved):

src/OpenTelemetry.AutoInstrumentation/Configurations/EnvironmentConfigurationTracerHelper.cs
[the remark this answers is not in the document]
Shane32: Just FYI, Microsoft.AspNetCore 2.1 is supported for .NET Framework. 

https://dotnet.microsoft.com/en-us/platform/support/policy/aspnet/2.1-packages

PR #2686, opened 2023-06-21
Adds MySql.Data bytecode instrumentation

### LINKED ISSUE #2542: Rewrite MySql.Data instrumentation to bytecode instrumenation
MySql.Data v 8.0.33 introduced breaking changes.
It is not feasible to longer use sourcecode instrumentation without some other magic (like Harmony https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1157).

Issue detected by #2478

The goal is to rewrite it to bytecode instrumentation.

zacharycmontoya:
## Why

Starting with version 8.0.33 of the MySql.Data nuget package, source instrumentation does not work because of how `Activity.Current` (and AsyncLocals in general) are propagated through the async method.

Fixes #2542 

## What

This PR adds automatic instrumentation all versions of MySql, including 8.0.33+. Changes include:
- v8.0.33+: Automatic instrumentation of `MySql.Data.MySqlClient.MySqlCommand.ExecuteReaderAsync`
- Lower versions: Automatic instrumentation of `MySql.Data.MySqlClient.MySqlCommand.ExecuteReader`
- Automatic instrumentation of `OpenTelemetry.Instrumentation.MySqlData.MySqlDataInstrumentation.ctor` so we can consume options if the user configures instrumentation themselves

The instrumentation generates an Activity and decorates it in the same way that the `OpenTelemetry.Instrumentation.MySqlData` does.

It's important to note that this does not generate duplicate Activity objects. Only the Activity generated from the automatic instrumentation is finished and exported. Here's the order in which Activity events are generated:
- When `MySqlCommand.ExecuteReader`/`MySqlCommand.ExecuteReaderAsync` begins, the automatic instrumentation _starts_ a new Activity
  - Note: Activity.Previous is either null or was created by another ActivitySource
- Further within the callstack, `OpenTelemetry.Instrumentation.MySqlData` _starts_ a new Activity
  - Note: Activity.Previous is the automatic instrumentation Activity
- When `ExecuteReader`/`ExecuteReaderAsync` exits, the automatic instrumentation _stops_ the Activity it generates.
  - Note: This changes Activity.Current to the value of Activity.Previous
- When `MySqlCommand.Close` is called, `OpenTelemetry.Instrumentation.MySqlData` checks Activity.Current and if it was generated by its ActivitySource, it closes the Activity. With our automatic instrumentation, Activity.Current is either null or its ActivitySource is different, so the original Activity generated by the instrumentation library never closes.

## Tests

The MySqlDataTests has been updated to use snapshot verification to make sure the data on the exported Activities is consistent from library version to library version. This has been used to validate that the Activity generated with the new bytecode instrumentation is identical to the Activity generated by previous instrumentation

## Checklist


Kielek:
@zacharycmontoya, what's about the second option discussed one week ago? Maybe we could fully drop MySqlData instrumentation library and fully apply bytecode instrumentation?

src/OpenTelemetry.AutoInstrumentation/Instrumentations/MySqlData/MySqlDataCommon.cs
```diff
@@ -0,0 +1,115 @@
...
+    internal const string DbUser = "db.user";
+
+    internal const string NetPeerIp = "net.peer.ip";
+    internal const string NetPeerPort = "net.peer.port";
+    internal const string NetPeerName = "net.peer.name";
+
+    internal const string PeerService = "peer.service";
+
+    internal const string MysqlDatabaseSystemName = "mysql";
+    internal static readonly ActivitySource ActivitySource = new ActivitySource(
```
zacharycmontoya, line 39: We can update the name of the ActivitySource to `OpenTelemetry.AutoInstrumentation.MySqlData` if that would be preferred. As long as the outcome is the same (tests pass) that should be fine

zacharycmontoya: Updated in d4bcef1a1e1bfc68d3b5fba34557f423b3a88681

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -290,9 +290,6 @@ private static void AddLazilyLoadedTraceInstrumentations(LazyInstrumentationLoad
                 case TracerInstrumentation.AspNetCore:
                     DelayedInitialization.Traces.AddAspNetCore(lazyInstrumentationLoader, pluginManager);
                     break;
-                case TracerInstrumentation.MySqlData:
-                    DelayedInitialization.Traces.AddMySqlClient(LazyInstrumentationLoader, pluginManager);
-                    break;
                 case TracerInstrumentation.EntityFrameworkCore:
```
Kielek, line 293:
```suggestion
                    break;
                case TracerInstrumentation.MySqlData:
                    break;
                case TracerInstrumentation.EntityFrameworkCore:
```

Otherwise it will leads to
```
Logger.Warning($"Configured trace instrumentation type is not supported: {instrumentation}");
                    if (FailFastSettings.Value.FailFast)
                    {
                        throw new NotSupportedException($"Configured trace instrumentation type is not supported: {instrumentation}");
                    }
```

src/OpenTelemetry.AutoInstrumentation/Instrumentations/MySqlData/ExecuteReaderIntegration.cs
```diff
@@ -0,0 +1,86 @@
...
+/// MySql.Data.MySqlClient.MySqlCommand.ExecuteReaderAsyncIntegration calltarget instrumentation
+/// </summary>
+[InstrumentMethod(
+    assemblyName: "MySql.Data",
+    typeName: "MySql.Data.MySqlClient.MySqlCommand",
+    methodName: "ExecuteReader",
+    returnTypeName: "System.Threading.Tasks.Task`1<MySql.Data.MySqlClient.MySqlDataReader>",
+    parameterTypeNames: new[] { "System.Data.CommandBehavior" },
+    minimumVersion: "6.10.7",
+    maximumVersion: "8.0.32",
```
Kielek, line 36: Just to double check, what's about 8.0.32.1?

src/OpenTelemetry.AutoInstrumentation/Generated/net6.0/SourceGenerators/SourceGenerators.InstrumentationDefinitionsGenerator/InstrumentationDefinitions.g.cs
Kielek, line 1: Whats about other instrumentations defined in upsteam?
For now it seems that here only ExecuteReader is instrumented: https://github.com/DataDog/dd-trace-dotnet/pull/4109/files#diff-b7a5ab305895159ec230fc6ecb7bec90b944fb30ddee8aa7558d141daa42376aR222-R255

Kielek (commented):
First round of review. I need to revisit how the instrumentation is implemented.

docs\config.md needs to be update - all versions are covered by bytecode instrumentation
CHANGELOG.md needs a record that we have changed the way how mysql is instrumented + info about fixes known issue from rc.1

nrcventura:
> @zacharycmontoya, what's about the second option discussed one week ago? Maybe we could fully drop MySqlData instrumentation library and fully apply bytecode instrumentation?

This might be necessary. The `Activity` class is disposable, and when disposing will stop the activity. The `Dispose` method also has a call to suppress finalization, which makes me think there are versions with a finalizer or there are plans to include a finalizer. And if that is the case, we could see both activities collected.

pjanotti:
Closing this one given that #2836 was merged. Please, re-open if appropriate. /cc @zacharycmontoya @Kielek

PR #2754, opened 2023-07-17
Make OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES consistent

### LINKED ISSUE #2441: Inconsistency between native and managed code handling of OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES
Most of the time the inconsistency shouldn't be a problem, but, for `dotnet` CLI it can be noticeable:

- The native code uses the process module name
- The managed code uses the `AppDomain.CurrentDomain.FriendlyName`

RassK:
## Why

Fixes #2441

## What

According to https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2441#issuecomment-1635030911

Instead of `Process.GetCurrentProcess().ProcessName` uses `Process.GetCurrentProcess().MainModule.ModuleName` to get module name instead of process name (example: `dotnet.exe` vs `dotnet`).

New behavior (Windows):
dotnet MyApp.dll -> dotnet.exe
MyApp.exe -> MyApp.exe

New behavior (Linux / Mac):
dotnet MyApp.dll -> dotnet
MyApp.exe -> MyApp

## Tests

Existing.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
 - In plugins `ConfigureTracerProvider` and `ConfigureMeterProvider` are changed now
   to `AfterConfigureTracerProvider` and `AfterConfigureMeterProvider`.
   See [plugins documentation](/docs/plugins.md) for details.
+- Changed `OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES` behavior in case of using
+StartupHook only. Now consistent with native logic by excluding only by
+application module name.
 
 ### Deprecated
 
--- a/src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/ApplicationInExcludeListRule.cs
+++ b/src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/ApplicationInExcludeListRule.cs
@@ -14,6 +14,7 @@
 // limitations under the License.
 // </copyright>
 
+using System.Diagnostics;
 using OpenTelemetry.AutoInstrumentation.Logging;
 
 namespace OpenTelemetry.AutoInstrumentation.RulesEngine;
@@ -30,23 +31,23 @@ internal class ApplicationInExcludeListRule : Rule
 
     internal override bool Evaluate()
     {
-        var applicationName = GetApplicationName();
+        var processModuleName = GetProcessModuleName();
 
-        if (IsApplicationInExcludeList(applicationName))
+        if (IsProcessInExcludeList(processModuleName))
         {
-            Logger.Information($"Rule Engine: {applicationName} is in the exclusion list. Skipping initialization.");
+            Logger.Information($"Rule Engine: {processModuleName} is in the exclusion list. Skipping initialization.");
             return false;
         }
 
-        Logger.Debug($"Rule Engine: {applicationName} is not in the exclusion list. ApplicationInExcludeListRule evaluation success.");
+        Logger.Debug($"Rule Engine: {processModuleName} is not in the exclusion list. ApplicationInExcludeListRule evaluation success.");
         return true;
     }
 
-    private static string GetApplicationName()
+    private static string GetProcessModuleName()
     {
         try
         {
-            return AppDomain.CurrentDomain.FriendlyName;
+            return Process.GetCurrentProcess().MainModule.ModuleName;
         }
         catch (Exception ex)
         {
@@ -55,12 +56,12 @@ internal class ApplicationInExcludeListRule : Rule
         }
     }
 
-    private static bool IsApplicationInExcludeList(string applicationName)
+    private static bool IsProcessInExcludeList(string processName)
     {
-        return GetExcludedApplicationNames().Contains(applicationName);
+        return GetExcludedProcessNames().Contains(processName);
     }
 
-    private static List<string> GetExcludedApplicationNames()
+    private static List<string> GetExcludedProcessNames()
     {
         var excludedProcesses = new List<string>();
 
--- a/test/IntegrationTests/Helpers/EnvironmentHelper.cs
+++ b/test/IntegrationTests/Helpers/EnvironmentHelper.cs
@@ -159,7 +159,10 @@ public class EnvironmentHelper
             extension = "dll";
         }
 
-        var appFileName = $"{FullTestApplicationName}.{extension}";
+        var appFileName = EnvironmentTools.IsWindows()
+            ? $"{FullTestApplicationName}.{extension}"
+            : FullTestApplicationName;
+
         var testApplicationPath = Path.Combine(GetTestApplicationApplicationOutputDirectory(packageVersion: packageVersion, framework: framework), appFileName);
         return testApplicationPath;
     }
--- a/test/IntegrationTests/Helpers/EnvironmentHelper.cs
+++ b/test/IntegrationTests/Helpers/EnvironmentHelper.cs
@@ -154,7 +154,7 @@ public class EnvironmentHelper
     {
         string extension = "exe";
 
-        if (IsCoreClr() || _testApplicationDirectory.Contains("aspnet"))
+        if (_testApplicationDirectory.Contains("aspnet"))
         {
             extension = "dll";
         }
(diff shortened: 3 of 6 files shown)


pellared (commented):
How about adding additional test cases  in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/e64f697f8445cb611cbc1c4eeebc96ab4d2921eb/test/IntegrationTests/SmokeTests.cs#L74-L90 that would have

```csharp
        SetEnvironmentVariable("COR_ENABLE_PROFILING", "0");
        SetEnvironmentVariable("CORECLR_ENABLE_PROFILING", "0");
```
to make sure that the behavior is consistent?

rajkumar-rangaraj (approved):
LGTM

pjanotti (commented):
@RassK can you take a look if the process name is consistent across Linux/Mac/Windows? It seems more "portable" than the module name. Unless there is some way to change the process name, it seems a better option than the file name.

RassK:
The biggest issue why tests are not passing currently:

* Both (native profiler and StartupHook) are configured to bootstrap at the same time
* .NET Core smoke tests are executed as "dotnet.exe TestApplication.Smoke.dll"
* While native profiler fails bootstrapping, StartupHook still bootstraps TestApplication.Smoke due `AppDomain.CurrentDomain.FriendlyName` != `dotnet,dotnet.exe`. This gives false positive test run for `ApplicationIsNotExcluded`.

There are 2 ways to fix this thing:
1. `ApplicationIsNotExcluded` and `ApplicationIsExcluded` should remove `dotnet,dotnet.exe`
2. Run .NET Core apps directly from exe

nrcventura:
If we do not handle the `dotnet myapp.dll` use case well, I am concerned that it will lead to some confusion for people trying to instrument an application. Running `dotnet myapp.dll` is common enough that people WILL run into problems. For example the Visual Studio template that creates a docker container for an aspnetcore application by default will use `ENTRYPOINT ["dotnet", "ContainerizedAspNetCoreApp.dll"]`.

RassK:
> If we do not handle the `dotnet myapp.dll` use case well, I am concerned that it will lead to some confusion for people trying to instrument an application. Running `dotnet myapp.dll` is common enough that people WILL run into problems. For example the Visual Studio template that creates a docker container for an aspnetcore application by default will use `ENTRYPOINT ["dotnet", "ContainerizedAspNetCoreApp.dll"]`.

Currently all our test apps are using `dotnet myapp.dll`. I remember there were  some issues with this, but do not remember what happened with this. But currently according to tests you definitely can run  `dotnet myapp.dll` - just need to be sure not to combine `dotnet myapp.dll` with `OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES=dotnet,dotnet.exe`

RassK:
> We need to be careful with `dotnet run` too. If I am remembering correctly, when you execute `dotnet run` there are 2 processes that get started:
> 
> 1. A `dotnet` tool process that we do not want to instrument
> 2. The application we want to instrument
> 
> It can be confusing to our customers/users to have to know when they should use the exclude process setting for `dotnet,dotnet.exe`. This is especially relevant when `dotnet myapp.dll` and `myapp.exe` has no problems, but `dotnet run` has problems.

Is there any other use than development stage? AFAIK the `dotnet run` command requires project with source code. Do we really want to focus to development stage?

nrcventura:
I do not think usage of `dotnet run` is limited to the development stage. I saw instances of production apps reporting "double" memory usage because both processes were instrumented. In the OTel use case, especially now that we have a nuget package, I think the developer experience is also important in order to build confidence in our solution to those developers. The previously solution that we had in place with the exclusion logic based on the app domain, handled both use cases for us simply (in addition to the crash/error that used to be triggered when the startuphook tried to bootstrap otel into dotnet.exe). The downside to that approach was that it behaved differently than the native code filtering logic. If we need to do something different to handle the `dotnet run` scenario, while also ensuring that the process exclusion logic works as expected consistently we need to get an issue submitted to better handle `dotnet run`.

pjanotti:
@RassK In the past, we excluded `dotnet` by default. However, since we have fixed it there is no need to exclude `dotnet` by default for the tests.

@nrcventura yes, `dotnet run` end up creating another process to run the actual application. If dotnet CLI telemetry is enabled we may end up capturing a trace for the parent process (the build step). If the installation follows our current instructions this shouldn't be a case that the users hit. That said, I wonder if we could simply document the issue for machine-wide installations. By disabling the `dotnet` CLI telemetry we disable the most visible side effect, but, as you pointed out, it may came back in the future if we collect some other signals for the processes.

pjanotti:
@RassK @nrcventura I didn't have a chance to fully think about the implications, but, let me throw out one idea. What if, by default, we bail out on the StartupHook if the domain-friendly name is `dotnet`? It doesn't prevent the CLR profiler from attaching and instrumenting, but, since our exporters are on the managed side it prevents exporting any data from `dotnet` CLI build, run, etc it works with `dotnet <DLL>` out of the box. This makes adding `dotnet,dotnet.exe` to the process exclusion a kind of optimization, a fine-tuning.

nrcventura:
> @RassK @nrcventura I didn't have a chance to fully think about the implications, but, let me throw out one idea. What if, by default, we bail out on the StartupHook if the domain-friendly name is `dotnet`? It doesn't prevent the CLR profiler from attaching and instrumenting, but, since our exporters are on the managed side it prevents exporting any data from `dotnet` CLI build, run, etc it works with `dotnet <DLL>` out of the box. This makes adding `dotnet,dotnet.exe` to the process exclusion a kind of optimization, a fine-tuning.

I think that would solve both problems. I do not think that we need any configuration to change that "hard-coded" behavior.

RassK:
I think it will be still inconsistent for .NET Fx, It doesn't have StartupHook to fallback on `AppDomain.FriendlyName`.
+ all of the TestApplications will be named dotnet.

nrcventura:
> I think it will be still inconsistent for .NET Fx, It doesn't have StartupHook to fallback on `AppDomain.FriendlyName`.
> Plus all of the TestApplications will be named dotnet.

.net framework will still have the process include/exclude lists. Now the process include/exclude will only apply to the profiler code. For .net (and the startuphook), we are now always excluding the dotnet tool itself. Our test applications, when using this specific appdomain filter (not the process check) should not be excluded because their appdomains are not `dotnet`.

RassK:
Quick overview here.

Possible combinations:
* Native only (.NET Fx and when missing StartupHook env)
* Native + StartupHook (default)
* StartupHook (disable profiler)


| Command Line                            | native     | StartupHook (FriendlyName) | StartupHook (MainModule) |
|-----------------------------------------|------------|----------------------------|--------------------------|
| dotnet MyApp.dll [Windows] (NO .NET FX) | dotnet.exe | MyApp                      | dotnet                   |
| MyApp.exe [Windows]                     | MyApp.exe  | MyApp                      | MyApp.exe                |
| dotnet MyApp.dll [Linux]                | dotnet     | MyApp                      | dotnet                   |
| MyApp [Linux]                           | MyApp      | MyApp                      | MyApp                    |


OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES=MyApp,MyApp.exe

| Command Line                           | native     | StartupHook (FriendlyName) | StartupHook (MainModule) |
|----------------------------------------|------------|----------------------------|--------------------------|
| dotnet MyApp.dll [Windows] (Core only) | no exclude | exclude (exception)        | no exclude               |
| MyApp.exe [Windows] (Core only)        | exclude    | exclude                    | exclude (exception)      |
| MyApp.exe [Windows] (NET FX only)      | exclude    | -                          | -                        |
| dotnet MyApp.dll [Linux]               | no exclude | exclude (exception)        | no exclude               |
| MyApp [Linux]                          | exclude    | exclude (exception)        | exclude (exception)      |

RassK:
> @RassK what in the tables above is meant by `exclude (exception)`? Is it crashing?

Yes, crash shutdown after `RuleEngine` evaluated all the rules.

```csharp
if (!ruleEngine.ValidateRules())
{
    throw new Exception("Rule Engine Failure: One or more rules failed validation. Automatic Instrumentation won't be loaded.");
}
```

nrcventura:
I can see the inconsistency that is being talked about between using the `OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES` setting in both the profiler and the startuphook in that example. But from what I understand in Paulo's suggestion, is that `OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES` will only be checked in the profiler and the startup hook will only ever exclude a process if the appdomain was either `dotnet` or `dotnet.exe` and it would not check the `OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES` setting (in the startup hook).

RassK:
I think that would just move inconsistency from one place to another, not making it overall consistent. 
That brings me back to the original idea of a separate var: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2441#issuecomment-1633804006

if Native and StartupHook (MainModule) are equal, that makes any combination working. Separately if you use dotnet to bootstrap the application (which is .NET Core only, with an exception of `dotnet run`) you can specify `OTEL_DOTNET_AUTO_EXCLUDE_APPDOMAIN` to fix this case.

PR #2772, opened 2023-07-25
Versioning based on release tag

### LINKED ISSUE #2601: CI builds produces artifacts with last release version
> AutoInstrumentation with the downloaded binaries (also seem to be labeled 0.7.0, but whatever).

_Originally posted by @danielodievich in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2588#issuecomment-1574336701_

We should ensure that CI produces version with higher/different version than last release.

Kielek:
## Why

Fixes #2601

## What

MinVer - natively for C# code
Native code versioned by Nuke tasks (version taken from MinVer).

No need to modify code, except installation script, before release.

## Tests

CI.
Tested with tag `v5.2.6` on my fork: https://github.com/Kielek/opentelemetry-dotnet-instrumentation/actions/runs/5656282012

## Checklist

## Notes

Possible follow up - find a way to avoid manual changes in installation scripts. Possibly we could do it by nuke and then copy it to artifacts.

Related issues:
* #2778
* #2777


build/VersionHelper.cs
```diff
@@ -0,0 +1,24 @@
+using System.Reflection;
+
+public static class VersionHelper
```
nrcventura, line 3: Should we use something like https://learn.microsoft.com/en-us/dotnet/api/system.version.tryparse?view=net-7.0 to help us instead of the manual string parsing that we are doing here?

Kielek: `Version.Parse` is not working correctly with format we need, eg `4.2.1-rc.1.54` or even `4.2.1-rc.1.54+commithashhere`.

Technically I could rewrite GetVersionsParts method, but I do not see big benefits to parse it to ints.

rajkumar-rangaraj (commented):
LGTM

PR #2810, opened 2023-08-07
Add smoke test to cover exclude process when CLR profiler is disabled

### LINKED ISSUE #2771: Smoke tests for `OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES` when profiler disabled
Am I correct that there are no tests for the changed code? All existing tests involving `OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES` work with .NET Profiler therefore the applications would be excluded on native code layer.

See https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2754#pullrequestreview-1532601427

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2765#discussion_r1270575214_

pjanotti:
## Why

Missing test the covers process exclusion when CLR profiler is disabled.
Fix #2771 

## What

Added a smoke test that validates process exclusion when the CLR profiler is not enabled. Notice that this scenario doesn't apply to .NET Framework since there is no StartupHook to instrument applications without the CLR profiler.

## Checklist

### DIFF
--- a/test/IntegrationTests/SmokeTests.cs
+++ b/test/IntegrationTests/SmokeTests.cs
@@ -71,6 +71,31 @@ public class SmokeTests : TestHelper
 #endif
     }
 
+#if !NETFRAMEWORK
+    [Theory]
+    [Trait("Category", "EndToEnd")]
+    [InlineData(TestAppStartupMode.DotnetCLI)]
+    [InlineData(TestAppStartupMode.Exe)]
+    public void WhenClrProfilerIsNotEnabledStartupHookIsEnabledApplicationIsExcluded(TestAppStartupMode testAppStartupMode)
+    {
+        switch (testAppStartupMode)
+        {
+            case TestAppStartupMode.DotnetCLI:
+                SetEnvironmentVariable("OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES", "dotnet,dotnet.exe");
+                break;
+            case TestAppStartupMode.Exe:
+                SetEnvironmentVariable("OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES", $"{EnvironmentHelper.FullTestApplicationName},{EnvironmentHelper.FullTestApplicationName}.exe");
+                break;
+            default:
+                throw new ArgumentException($"{testAppStartupMode} is not supported by this test", nameof(testAppStartupMode));
+        }
+
+        SetEnvironmentVariable("CORECLR_ENABLE_PROFILING", "0");
+
+        VerifyTestApplicationNotInstrumented(testAppStartupMode);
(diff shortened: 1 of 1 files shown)


test/IntegrationTests/SmokeTests.cs
```diff
@@ -71,6 +71,31 @@ public void WhenClrProfilerIsNotEnabled()
...
+                SetEnvironmentVariable("OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES", "dotnet,dotnet.exe");
+                break;
+            case TestAppStartupMode.Exe:
+                SetEnvironmentVariable("OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES", $"{EnvironmentHelper.FullTestApplicationName},{EnvironmentHelper.FullTestApplicationName}.exe");
+                break;
+            default:
+                throw new ArgumentException($"{testAppStartupMode} is not supported by this test", nameof(testAppStartupMode));
+        }
+
+        SetEnvironmentVariable("CORECLR_ENABLE_PROFILING", "0");
```
Kielek, line 93: As I remember, by default profiling is disabled for .NET.

Kielek (approved):

test/IntegrationTests/SmokeTests.cs
RassK: I think it might be best to combine it into a single test with 2 params: `TestAppStartupMode` and `IsProfilerEnabled`.

Currently it also double reflects `ApplicationIsExcluded`

nrcventura: I think that it is fine to either set that environment variable to "0" or to either "unset" that environment variable. Having that code in place can serve 2 purposes:

1. Make it clear that the native profiler is disabled.
2. "Defensively" protect the test code when the test runner accidentally inherits that environment variable from the environment.

nrcventura (approved):

test/IntegrationTests/SmokeTests.cs
pjanotti: > by default profiling is disabled for .NET.

Yes, that is correct. However, as Chris noticed, the presence of the statement makes that clear that the test doesn't have CLR profiling.

@RassK I didn't want to combine the tests because the StartupHook doesn't apply to .NET Framework while the `ApplicationIsExcluded` applies to all runtimes.

pjanotti:
Flaky test? https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/5801757827/job/15726930172?pr=2810#step:7:876

Re-running failed jobs...

PR #2825, opened 2023-08-10
Fix return value in InstrumentationAssemblyRule

rajkumar-rangaraj:
## Why

`InstrumentationAssemblyRule` always returns true and does not respect result value.

## What

Modify the function to return `result` instead of `true`.

## Tests

CI

## Checklist


nrcventura (commented):
Good catch! Is there a way to add a test for this?

rajkumar-rangaraj:
> Good catch! Is there a way to add a test for this?

There are currently no tests for any of the rules. We definitely need to add them. Thanks for pointing it out!

pjanotti (commented):
👍

Issue #2833, opened 2023-08-12
ASP.NET intrumentation failed on .NET framework


muhaook:
**Describe the bug**
ASP.NET intrumentation failed sometimes on .NET framework. 
I have 2 webapi ASP.NET applications. One is `WebApiMySql80Netfx462`; the other is `WebApiMySql81Netfx462`. I accessed the first application and saw the ASP.NET instrumentation succeeded. after that, when I tried to access the second application, I did not see any ASP.NET trace/span got collected. Looks like the second ASP.NET instrumentation failed. If I access the second web app in the first place, its traces/spans can be collected.

Attached please find logs:
[logs_aspnet_instrumentation_failure.zip](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/files/12326848/logs_aspnet_instrumentation_failure.zip)

**Runtime environment (please complete the following information):**

- OpenTelemetry Automatic Instrumentation version: 1.0.0-rc.2
- OS: Windows Server 2012 R2
- .NET version: .NET Framework 4.6.2

muhaook:
[web.configs.zip](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/files/12337670/web.configs.zip)
Attached please find web.config files

pjanotti:
Some suspicions, but, not clear to me yet exactly what is the problem. The current top hypothesis is that the binding redirect for `System.Runtime.CompilerServices.Unsafe` of  `WebApiMySql80Netfx462` ends up loading a version lower than `6.0.0.0` that is needed for `System.Diagnostics.DiagnosticSource` to work.

@muhaook could add/update the package reference to https://www.nuget.org/packages/System.Diagnostics.DiagnosticSource/7.0.2 and give it another try? Please, share the resulting web.config files and traces. Thanks!

muhaook:
Thanks @pjanotti for taking care of this issue. I added the suggested package reference to project `WebApiMySql80Netfx462 `and it works - both web applications got instrumented as expected. Here is what I did to the sample source code:
![image](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/assets/52417764/8db718cc-8210-4f94-8936-19647864b708)

Attached please find logs and web.config files:
[log_and_web.configs.zip](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/files/12341483/log_and_web.configs.zip)

pjanotti:
[...]

PR #2841, opened 2023-08-15
Skip more assemblies by their prefix

pjanotti:
## Why

Reduce the amount of work done on the `CorProfilerBase::GetAssemblyReferences` callback.

Collateral of #2833

## What

Assemblies that are not going to be instrumented don't need to be internally tracked. This generates a lot of noise on debug logs and waste work on normal operations. Besides `OpenTelemetry` prefix I'm also adding some also from the current upstream: https://github.com/DataDog/dd-trace-dotnet/blob/5292fc6b09b29e924e62eb6c1cb82107b43d4789/tracer/src/Datadog.Tracer.Native/debugger_constants.h#L16

There are more differences with upstream, but, since we are close to GA let's be careful with these changes, so, at least for now not adding `Microsoft` or `System` to skip prefixes.

## Tests

N/A

## Checklist


pjanotti:
2 CI failures on mac 1 on Alpine: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/5863049402/job/15903137724?pr=2841

🤔

pjanotti:
The failures are consistent. I need to investigate them.

PR #2876, opened 2023-08-25
[WCF] use wcf client instrumentation from contrib

lachmatt:
## Why

Recently released version of `OpenTelemetry.Instrumentation.Wcf` fixes handling of communication exceptions.
There is no longer a need for us to have custom bytecode instrumentation for `WCF` client.

## What

- Drop custom bytecode instrumentation
- Simplify our `WCF` tests as there is no longer custom instrumentation
- Bring back support for WCF client on .NET

## Tests

Existing tests adjusted.

## Checklist

~~- [] New features are covered by tests.~~


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
 ### Added
 
 - Added support for Azure SDK traces instrumentation on .NET Framework.
+- Added support for `WCFCLIENT` instrumentation on .NET.
```
rajkumar-rangaraj, line 13: Should we also update what is removed as a part of this PR?

nrcventura: I think that we should call out the switch from custom byte code instrumentation to the contrib library.

nrcventura:
@lachmatt Do you know if the contrib instrumentation library has test coverage for the different async testing scenarios that we are removing with this PR?

docs/wcf-config.md
```diff
@@ -8,16 +8,31 @@ Example project available in
...
-[TAP-style](https://learn.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/task-based-asynchronous-pattern-tap)
-calls.
-
 ## WCF Client Configuration (.NET)
 
-Instrumentation for WCF Client on .NET is not supported.
+Add `OpenTelemetry.Instrumentation.Wcf` and `System.Diagnostics.DiagnosticSource`
+package to the project. The version of `OpenTelemetry.Instrumentation.Wcf`
+should match the one used by AutoInstrumentation library to avoid compatibility
+issues.
```
Kielek, line 16: Is there any way to avoid manual steps?
If so, we can wait with introducing it post GA. This version is not autoinstrumentation friendly.

CHANGELOG.md
nrcventura: You are right, when I look back at the changelog updates for WCF I think that we do not need to call out the switch. We are already calling out what should be the only noticeable change. Even with this PR the current solution relies on both byte-code instrumentation and the contrib instrumentation library, so we are not eliminating the byte code instrumentation.

nrcventura (approved):

pjanotti (approved):

PR #2958, opened 2023-09-14
Disable optional rules by default on the NuGet package scripts

pjanotti:
## Why

What these rules validate can't happen when using the NuGet.

## What

Set `OTEL_DOTNET_AUTO_RULE_ENGINE_ENABLED` to `false` on the script shipped by the NuGet package.

## Tests

Locally run StartupHook.Tests to validate cea5e15389c3735f77a0c5412eeee5ec58b25472

## Checklist


src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngine.cs
```diff
@@ -28,23 +28,8 @@ internal class RuleEngine
...
-
-    internal RuleEngine()
-    {
-    }
-
-    // This constructor is used for test purpose.
-    internal RuleEngine(List<Rule> rules)
-    {
-        _otherRules = rules;
-    }
```
nrcventura, line 47: Did you need to change the way the rules are injected? You probably could have kept the existing injection pattern but added a new private field for the `_optionalTests` and made `_otherRules` static. Then you could update the .ctors to do something like the following.
```
internal RuleEngine() : this(_otherRules) {}

//This constructor is used for test purpose
internal RuleEngine(List<Rule> optionalRules)
{
    _optionalRules = optionalRules;
}
```

pjanotti: It is not required to change the way that the injection happens as long as the list of optional rules is only created when it is going to be used. Let me play with it.

pjanotti: Returned to prior injection pattern.

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngine.cs
```diff
@@ -65,8 +59,16 @@ internal bool ValidateRules()
             return result;
         }
 
+        var optionalRules = _optionalRules ?? new()
+        {
+            new OpenTelemetrySdkMinimumVersionRule(),
+            new DiagnosticSourceRule(),
+            new InstrumentationAssemblyRule(),
+            new NativeProfilerDiagnosticsRule()
```
nrcventura, line 67: Why do we need to defer the creation of these rules?

pjanotti: If we create them before knowing if they will be used or not we can load assemblies, make allocations, etc... that end up being not used. Delaying this creation until we are sure that they will be used avoid that, it becomes pay-for-play.

rajkumar-rangaraj (approved):

src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/RuleEngine.cs
pjanotti: Check lines 56-60 above.

pjanotti: > So just the act of creating these types causes other dependencies to load?

It depends on what the constructors of each type do. The current implementations don't do that, but, there is always the possibility of some later change creating or modifying a rule type that does that.

I like the idea of a static method to make clear that we are creating the default rules. Updating the code...

pjanotti: 5efcf1e04ddf61b405022472826f386113d35495

nrcventura (approved):

PR #2975, opened 2023-09-22
Fix for Missing Logs in ASP.NET Core 6.0 Web and Razor Apps

rajkumar-rangaraj:
## Why

ASP.NET Core Razor and several types of web applications do not produce logs when auto-instrumentation is enabled. This issue occurs exclusively on .NET 6.0.

**Reproduction Steps:**
To reproduce this issue, create a default ASP.NET Core 6.0 Razor web application. After auto-instrumenting with the 1.0.1 release, no logs will be collected.

## What

The problem originated from a specific line in the `LogBuilderExtensions` code. For the affected application types, the return type is `Microsoft.AspNetCore.WebApplicationServiceCollection`. Consequently, the following logic fails:

```
if (!(builder.Services is ServiceCollection services))
```

The fix involves check builder.Services is null in the aforementioned line. During the debugging process, I noticed that the logs did not provide enough clarity regarding this issue. To improve this, I've added a verbose log for better traceability.

## Recommendation:

Considering this issue affects all ASP.NET Core 6.0 applications and prevents any log emissions, it's crucial to **release** version **1.0.2** after merging this PR.

## Tests

Existing CI.

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Logger/LogBuilderExtensions.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Logger/LogBuilderExtensions.cs
@@ -31,7 +31,7 @@ internal static class LogBuilderExtensions
     {
         try
         {
-            if (!(builder.Services is IServiceCollection services))
+            if (builder.Services == null)
             {
                 AutoInstrumentationEventSource.Log.Verbose("Logs: The builder.Services property is not of the IServiceCollection type, so we're skipping the integration of logs with ServiceCollection.");
                 return builder;
@@ -40,7 +40,7 @@ internal static class LogBuilderExtensions
             // Integrate AddOpenTelemetry only once for ServiceCollection.
 
             _loggingProviderSdkType ??= Type.GetType("OpenTelemetry.Logs.LoggerProviderBuilderSdk, OpenTelemetry");
-            var openTelemetryLoggerProviderDescriptor = services.FirstOrDefault(descriptor => descriptor.ImplementationType == _loggingProviderSdkType);
+            var openTelemetryLoggerProviderDescriptor = builder.Services.FirstOrDefault(descriptor => descriptor.ImplementationType == _loggingProviderSdkType);
             if (openTelemetryLoggerProviderDescriptor != null)
             {
                 AutoInstrumentationEventSource.Log.Verbose("Logs: AddOpenTelemetry already called on logging builder instance.");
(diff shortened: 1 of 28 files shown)


pjanotti (approved):

src/OpenTelemetry.AutoInstrumentation/Logger/LogBuilderExtensions.cs
```diff
@@ -31,8 +31,9 @@ public static ILoggingBuilder AddOpenTelemetryLogs(this ILoggingBuilder builder)
     {
         try
         {
-            if (!(builder.Services is ServiceCollection services))
+            if (!(builder.Services is IServiceCollection services))
```
Kielek, line 34: In general, you can change the condition to builder.Services == null.

ILoggingBuilder Services is always `IServiceCollection`.

Shouldn't it check some extended supported set of types?

Kielek (commented):
Please include also test to the codebase to cover such scenario. 

I am fine with 1.0.2 just after this release.

src/OpenTelemetry.AutoInstrumentation/Logger/LogBuilderExtensions.cs
rajkumar-rangaraj: Good thought, we could check this for null. I don't recall the reason for checking ServicesCollection here.

rajkumar-rangaraj:
> Please include also test to the codebase to cover such scenario. If possible without bootstrap and jQuery files from default razor template.
> 
> I am fine with 1.0.2 just after this is merged.

We already have tests for logs. I could add another one for the Razor app. However, I'm uncertain about its value since this test would never reproduce the issue. If there's a problem, other tests should detect it.

src/OpenTelemetry.AutoInstrumentation/Logger/LogBuilderExtensions.cs
rajkumar-rangaraj: Fixed.

rajkumar-rangaraj:
@Kielek Added tests. We could remove it later if we find this as duplicate.

test/test-applications/integrations/TestApplication.Azure/Program.cs
Kielek, line 1: I think that this files was added by mistake.
When reverted we are fine to merge.

rajkumar-rangaraj: Thanks for spotting out. Fixed.

Kielek (approved):

PR #3016, opened 2023-10-19
[docs] List supported processor architecture in Compatibility section

pellared:
## Why

Relatively frequently, people do not notice that ARM is not supported.

## What

I think that the lack of support for ARM should be noted in "Compatibility" section. I also decided to explicitly call out proc archs that we support.


pellared:
CC @theletterf

Kielek (approved):

pjanotti (approved):

Issue #3074, opened 2023-11-07
CPU/Allocation profiler


Kielek:
# Feature Request

Implement CPU and memory profiler based on description in Splunk repository:

* CPU profiler : https://github.com/signalfx/signalfx-dotnet-tracing/blob/main/docs/always-on-profiling.md
* Memory profiler: https://github.com/signalfx/signalfx-dotnet-tracing/blob/main/docs/internal/memory-profiling.md

Most of the native code is stored under: https://github.com/signalfx/signalfx-dotnet-tracing/blob/main/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp (and other prefixed by always_on_*).
Most of the managed code part: https://github.com/signalfx/signalfx-dotnet-tracing/tree/main/tracer/src/Datadog.Trace/AlwaysOnProfiler
Integration tests: https://github.com/signalfx/signalfx-dotnet-tracing/blob/main/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AlwaysOnProfilerPprofTests.cs
Tests application (covers all .NET languages): https://github.com/signalfx/signalfx-dotnet-tracing/tree/main/tracer/test/test-applications/integrations/Samples.AlwaysOnProfiler

pellared:
I propose to add a pprof exporter as it could be useful for both the users as well as OTLP Profiler SIG for reference.

PR #3196, opened 2023-12-21
Continuous profiler

Kielek:
## Why

Towards #3074

## What

Implementation for extendible continuous profiler for thread and allocations. It is donation of Splunk code from [SignalFx repository](https://github.com/signalfx/signalfx-dotnet-tracing/tree/main/tracer/src/Datadog.Trace/AlwaysOnProfiler) with some dedicated adjustments.

Separate comments describe:

- [How thread sampling is working](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3196#issuecomment-1870310409)
- [How allocation sampling is working](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3196#issuecomment-1873712139)
- [How to implement plugin](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3196#issuecomment-1873853132)

As there is no way to export these data in any OTel common way, I would like to keep this documentation only in PR.
Also Changelog is omitted. There is a plan to address it in the future, when  https://github.com/open-telemetry/oteps/pull/239 or https://github.com/open-telemetry/oteps/pull/237 will be ready and merged.

## Tests

CI + testing with Splunk exporter implemented by plugin: https://github.com/signalfx/splunk-otel-dotnet/pull/393

## Checklist

- [x] ~~[Documentation is updated.~~ Documentation is included only in this PR

## How to review

As the PR is huge, my recommendation is to start with plugin implementation and the exporter itself (TestApplication folder). Then check which code is calling it. Last part - verification of the native code which do the most of the tricks.

If you need, I will be happy to make real time peer review. I have did it already with @pellared and we have found couple of things to improve. 

## Notes

While merging, please add following comment:

```
Co-authored-by: John Bley <jbley@splunk.com>
Co-authored-by: Paulo Janotti <pjanotti@splunk.com>
Co-authored-by: Robert Pająk <rpajak@splunk.com>
Co-authored-by: Mateusz Łach <mateusza@splunk.com>
Co-authored-by: Dawid Szmigielski <dszmigielski@splunk.com>
```


Kielek:
# About the Continuous Profiling - Thread sampling

Thread sampling can be enabled by the custom plugin.
Plugin is responsible for parsing dense, data format and exporting it in the appropriate format.

## How does the thread sampler work?

The profiler leverages [.NET profiling](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/) to perform periodic call stack sampling. For every sampling period, the runtime is suspended and the samples for all managed thread are saved into the buffer, then the runtime resumes.

The separate managed-thread is processing data from the buffer and export it the way defined by the plugin.

To make the process more efficient, the sampler uses two independent buffers to store samples alternatively.

# Requirements

* .NET 6.0 or higher (`ICorProfilerInfo12` available in runtime). Technically It could be ICorProiler10, and .NET Core 3.1/.NET 5.0, but these versions are not supported by OpenTelemetry .NET AutoInstrumentation.
* .NET Framework is not supported. ICorProiler10 nor ICorProiler12 are not available in .NET Fx.

# Enable the profiler

Implement custom plugin.

# Configuration settings by plugin

`threadSamplingEnabled = true;`
`var threadSamplingInterval = 10000u;` // in ms. Splunk is using 10000. There is no possibility to set value lower than 1000.
`var exportInterval = TimeSpan.FromMilliseconds(500);` // Interval to read data from buffers and call exporter, common for Thread and Allocation sampling
`object continuousProfilerExporter = new ConsoleExporter();` // Exporter, common for Thread and Allocation sampling

# Escape hatch

The profiler limits its own behavior when both buffers used to store sampled data are full.

This scenario might happen when the data processing thread is not able
to export data the given period of time.

Thread sampler will resume when any of the buffers are empty.

# Troubleshooting the .NET profiler

## How do I know if it's working?

At the startup, the SignalFx Instrumentation for .NET logs the string `ContinuousProfiler::StartThreadSampling` at `info` log level.

You can grep for this in the native logs for the instrumentation to see something like this:

```text
10/12/22 12:10:31.962 PM [12096|22036] [info] ContinuousProfiler::StartThreadSampling
```

## How can I see Continuous Profiling configuration?

The OpenTelemetry .NET AutomaticInstrumentation logs the profiling configuration
at `Debug` log level during the startup. You can grep for the string `Continuous profiling configuration: `
to see the configuration.

## What does the escape hatch do?

The escape hatch automatically discards profiling data
if the ingest limit has been reached.

If the escape hatch activates, it logs the following message:

```text
Skipping a thread sample period, buffers are full.
```

You can also look for:

```text
** THIS WILL RESULT IN LOSS OF PROFILING DATA **.
```

If you see these log messages, check the exporter implementation.

## What if I'm on an unsupported .NET version?

None of the .NET Framework versions is supported. You have to switch to supported .NET version.

## Can I tell the sampler to ignore some threads?

There is no such functionality. All managed threads are captured by the profiler.

Kielek:
# Plugin

For now, the plugins is responsible for
 - defining configuration for continuous profiling
 - providing exporter for the allocation and profiling data
 - *parsing data* prepared by the native code.

## Plugin contract

> [!IMPORTANT]  
> It will be subject to change, when https://github.com/open-telemetry/oteps/pull/239 or https://github.com/open-telemetry/oteps/pull/237 will be ready and merged.


As other methods, `GetContinuousProfilerConfiguration` is called by reflection and convention.

```csharp
    /// <summary>
    /// Configure Continuous Profiler.
    /// </summary>
    /// <returns>(threadSamplingEnabled, threadSamplingInterval, allocationSamplingEnabled, maxMemorySamplesPerMinute, exportInterval, continuousProfilerExporter)</returns>
    public Tuple<bool, uint, bool, uint, TimeSpan, object> GetContinuousProfilerConfiguration()
    {
        var threadSamplingEnabled = true; // enables thread sampling
        var threadSamplingInterval = 10000u; // interval to stop CLR runtime and fetch stacks. 10 000ms is Splunk default. 1000ms is the lowest supported value by Splunk. The code does not contains any limitations for this. Plugins is responsible for checks.
        var allocationSamplingEnabled = true; // enables allocation sampling
        var maxMemorySamplesPerMinute = 200u; // max number of samples in minutes. 200 is tested default value by Splunk.
        var exportInterval = TimeSpan.FromMilliseconds(500); // Pause time before next execution of exporting/reading buffer  process
        object continuousProfilerExporter = new ConsoleExporter();

        return Tuple.Create(threadSamplingEnabled, threadSamplingInterval, allocationSamplingEnabled, maxMemorySamplesPerMinute, exportInterval, continuousProfilerExporter);
    }
```

if more than one plugin implement `GetContinuousProfilerConfiguration` only the firtst one will be used. Other will be ignored.

## Exporter contract

Works in the similar way as for Splunker.

Two methods has to be implemented by Exporter

```csharp
public void ExportThreadSamples(byte[] buffer, int read);
public void ExportAllocationSamples(byte[] buffer, int read);
```

Both accept buffer produced by the native code together with the length of filled data.
The Exporter is responsible both for parsing this buffer and exporting it.

Example: ConsoleExporter in this PR.


## Native parser:

As there is no default OpenTelemetry Protocol format there is not easy way to create good contract between OpenTelemetry Automatic Instrumentation and the plugin. The plugin has to implement (copy) our version of the parser.
It should be changed when the OTEP will be merged, and we can start implementing real OTLP exporter.

Implementation can be found in `SampleNativeFormatParser` class in this PR.

src/OpenTelemetry.AutoInstrumentation.Native/clr_helpers.cpp
```diff
@@ -610,6 +617,10 @@ WSTRING GetSigTypeTokName(PCCOR_SIGNATURE& pbCur, const ComPtr<IMetaDataImport2>
     {
         tokenName += WStr("&");
     }
+    if (pointer_flag)
```
RassK, line 620: nit: formatting error or intentional? 
missing blank lines before `if` and before `return`.

src/OpenTelemetry.AutoInstrumentation/Plugins/PluginManager.ContrinuousProfiler.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright The OpenTelemetry Authors
```
RassK, line 1: File naming error `PluginManager.ContrinuousProfiler.cs` => `PluginManager.ContinuousProfiler.cs`

src/OpenTelemetry.AutoInstrumentation.Native/clr_helpers.cpp
Kielek: Added lines in 0ca789cd5957f17546dbfed2fca0b2732dee6490

Technically it was passing native code check. There is not strict rule for this.

src/OpenTelemetry.AutoInstrumentation/Plugins/PluginManager.ContrinuousProfiler.cs
Kielek: 7d1fe8237992e60b0ec368efaecb80625424bc28

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.h
```diff
@@ -0,0 +1,246 @@
...
+
+constexpr auto unknown_managed_thread_id = -1;
+
+#ifdef _WIN32
+#define EXPORTTHIS __declspec(dllexport)
+#else
+#define EXPORTTHIS __attribute__((visibility("default")))
+#endif
+
+
```
RassK, line 27: nit: extra blank line

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.h
```diff
@@ -0,0 +1,246 @@
...
+    trace::WSTRING thread_name_;
+    ThreadState()
+    {
+    }
+    ThreadState(ThreadState const& other) : thread_name_(other.thread_name_)
+    {
+    }
+};
+
+
```
RassK, line 90: nit: extra blank line

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler_clr_helpers.cpp
```diff
@@ -0,0 +1,366 @@
...
+            const auto generic_info = GetFunctionInfo(metadata_import, parent_token);
+            std::memcpy(function_name, generic_info.name.c_str(), sizeof(WCHAR) * (generic_info.name.length() + 1));
+            function_name_len = DWORD(generic_info.name.length() + 1);
+        }
+        break;
+        default:
+            trace::Logger::Warn("[trace::GetFunctionInfo] unknown token type: {}", token_type);
+            return {};
+    }
+    if (FAILED(hr) || function_name_len == 0)
```
RassK, line 60: nit: missing blank before if

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler_clr_helpers.cpp
```diff
@@ -0,0 +1,366 @@
...
+        }
+        break;
+        case mdtModuleRef:
+            metadata_import->GetModuleRefProps(token, type_name, trace::kNameMaxSize, &type_name_len);
+            break;
+        case mdtMemberRef:
+        case mdtMethodDef:
+            return GetFunctionInfo(metadata_import, token).type;
+    }
+    if (FAILED(hr) || type_name_len == 0)
```
RassK, line 124: nit: missing blank before if

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.h
Kielek: 28480b3a26a7044e53f5976f9e67400ed7a1ad9c

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler_clr_helpers.cpp
Kielek: 28480b3a26a7044e53f5976f9e67400ed7a1ad9c

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.h
Kielek: 28480b3a26a7044e53f5976f9e67400ed7a1ad9c

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler_clr_helpers.cpp
Kielek: 28480b3a26a7044e53f5976f9e67400ed7a1ad9c

test/test-applications/integrations/TestApplication.ContinuousProfiler/Exporter/ConsoleExporter.cs
```diff
@@ -0,0 +1,33 @@
...
+    public void ExportAllocationSamples(byte[] buffer, int read)
+    {
+        var allocationSamples = SampleNativeFormatParser.ParseAllocationSamples(buffer, read);
+        var allocationSample = allocationSamples.FirstOrDefault();
+        if (allocationSample == null)
+        {
+            return;
+        }
+
+        Console.WriteLine("ExportAllocationSamples[" + allocationSample.AllocationSizeBytes + "]" + string.Join("\n", allocationSample.ThreadSample.Frames));
```
RassK, line 31:
```suggestion
        Console.WriteLine($"ExportAllocationSamples[{allocationSample.AllocationSizeBytes}]" + string.Join("\n", allocationSample.ThreadSample.Frames));
```

test/test-applications/integrations/TestApplication.ContinuousProfiler/Exporter/SampleNativeFormatParser.cs
```diff
@@ -0,0 +1,308 @@
...
+                {
+                    var version = ReadInt(buffer, ref position);
+                    if (version != 1)
+                    {
+                        return null; // not able to parse
+                    }
+
+                    sampleStartMillis = ReadInt64(buffer, ref position);
+
+                    /* if (IsLogLevelDebugEnabled)
```
RassK, line 47: commented code (here and below), if worth adding TODO comment seems appropriate. Possibly a separate issue also for the future to track.

test/test-applications/integrations/TestApplication.ContinuousProfiler/Exporter/SampleNativeFormatParser.cs
```diff
@@ -0,0 +1,308 @@
...
+                        TraceIdLow = traceIdLow,
+                        SpanId = spanId,
+                        ManagedId = managedId,
+                        ThreadName = threadName,
+                        ThreadIndex = threadIndex
+                    };
+
+                    ReadStackFrames(code, threadSample, codeDictionary, buffer, ref position);
+
+                    if (threadName == "OpenTelemetry Continuous Profiler Thread")
```
RassK, line 87: magic string, can it take a reference to the existing constant declaration?

test/test-applications/integrations/TestApplication.ContinuousProfiler/Exporter/SampleNativeFormatParser.cs
Kielek: #3205 and 77f6a6048795efa5c50e2c4eb962c40d4e895b32

test/test-applications/integrations/TestApplication.ContinuousProfiler/Exporter/ConsoleExporter.cs
Kielek: f9daeff00d72d2174a3d411bbe5e325059540efc

test/test-applications/integrations/TestApplication.ContinuousProfiler/Exporter/SampleNativeFormatParser.cs
Kielek: a2066c9a955f70521c5084e0a546e96c2d7df6aa, please let me know if you are fine with this solution

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -0,0 +1,1042 @@
...
+// FIXME make configurable (hidden)?
+// These numbers were chosen to keep total overhead under 1 MB of RAM in typical cases (name lengths being the biggest
+// variable)
+constexpr auto kMaxFunctionNameCacheSize         = 5000;
+constexpr auto kMaxVolatileFunctionNameCacheSize = 2000;
+
+// If you squint you can make out that the original bones of this came from sample code provided by the dotnet project:
+// https://github.com/dotnet/samples/blob/2cf486af936261b04a438ea44779cdc26c613f98/core/profiling/stacksampling/src/sampler.cpp
+// That stack sampling project is worth reading for a simpler (though higher overhead) take on thread sampling.
+
```
johnbley, line 41: I think this would be a good place to document the threading and locking behavior, and in particular the weird cases where things are different than the simple "this method acquires a lock to modify a shared data structure".  I've written up a bit below:
```
  Locking/threading design:
  We have the following shared data structures:
  - A buffer for captured thread samples, used by a single writing thread and a single (managed) reading one
  - A buffer for captured allocation samples, used by any application thread at any time
  - a name cache (data structure for building humnan-readable stack traces), used during both thread and
    allocation sampling                    
  - a cache of thread id->thread name (set by each thread itself, used during either sample type)
  - a cache of thread span context state (set by any application thread at any time, used during either sample type)
  
  In general we want to keep locks "adjacent" to just one data structure and usage of them local to one 
  modifying/reading method to simplify analysis.  However, there are some special cases.  
  Here are the locks in use:
  - cpu_buffer_lock guarding access to the thread samples buffer
  - allocation_buffer_lock guarding access to the buffer for allocation samples
  - name_cache_lock, guarding the
  - thread_state_lock_ guarding the thread name map
  - thread_span_context_lock guarding that data structure
  - (special) a profiling_lock so only one type of profiling (thread stacks or allocation sample) runs at a time 
     
  The special cases worth calling out about locking behavior are:  
  - Because the stack sampler pauses the CLR, to avoid deadlock it needs to know that no application thread
    is holding a lock it needs (e.g., thread_span_context_lock) while it is paused.  So, it acquires
    all the necessary locks at once before pausing the CLR.
  - Because the thread sampler pauses the whole CLR and then proceeds to walk every
    thread's stack (using the name cache for each stack entry), it obviously needs the   
    name_cache_lock but we don't want to pay the overhead of locking/unlocking it in that tight loop.
    So, it is acquired once before the iteration of threads start, and unlike other locks,
    the methods for this are specifically coded assuming you own
  - AllocationTick(s - possibly on multiple threads) and the thread sampler can happen concurrently,
    so the profiling_lock is used with unique_lock and shared_lock to ensure that any allocation samples
    are fully processed before pausing the CLR for thread samples.
```

Kielek: Thanks! 0502696d7b12b7fb15adad25cb2b2715bc5753de

pellared:
> Separate comments describe:

This should live in documentation e.g. in `docs/design.md`. I propose to make a separate issue for it.

lachmatt (approved):
LGTM

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -79,6 +85,26 @@ public static void Initialize()
             _pluginManager = new PluginManager(GeneralSettings.Value);
             _pluginManager.Initializing();
 
+#if NET6_0_OR_GREATER
+            var profilerEnabled = GeneralSettings.Value.ProfilerEnabled;
+            var (threadSamplingEnabled, threadSamplingInterval, allocationSamplingEnabled, maxMemorySamplesPerMinute, exportInterval, continuousProfilerExporter) = _pluginManager.GetFirstContinuousConfiguration();
```
rajkumar-rangaraj, line 90: Perform these operation and below condition check only when profiling is enabled.

src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
```diff
@@ -0,0 +1,80 @@
...
+        thread.Start();
+
+        Logger.Information("Continuous Profiler export thread initialized.");
+    }
+
+    private static void SampleReadingThread(BufferProcessor sampleExporter, TimeSpan exportInterval)
+    {
+        while (true)
+        {
+            Thread.Sleep(exportInterval);
```
rajkumar-rangaraj, line 75: Consider using Task.Delay with async pattern and if possible send cancellation Token as parameter.

src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        var exportAllocationSamplesMethod = continuousProfilerExporterType.GetMethod("ExportAllocationSamples");
+        if (exportAllocationSamplesMethod == null)
+        {
+            Logger.Warning("Exporter does not have ExportAllocationSamples method. Continuous Profiler initialization failed.");
+            return;
+        }
+
+        var thread = new Thread(() =>
```
rajkumar-rangaraj, line 58: Is there a way to dispose this thread?

src/OpenTelemetry.AutoInstrumentation/Plugins/PluginManager.ContinuousProfiler.cs.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright The OpenTelemetry Authors
```
rajkumar-rangaraj, line 1: Is there an issue with this file name, it has two `.cs'

rajkumar-rangaraj:
nit: I recommend placing a launchSettings.json file with environment variables in the `test/test-applications/integrations/TestApplication.ContinuousProfiler/` app. This will make it easier for people to test this project.

src/OpenTelemetry.AutoInstrumentation/Plugins/PluginManager.ContinuousProfiler.cs.cs
Kielek: Fixed 37c7c5f50ba0593b6cdd63cf8a74549992ab0f50

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
Kielek: Fixes, please check: d00db398cc3fd31fd9d3721fce96187840c058c2

src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
Kielek: 1. `Thread` is not implementing `IDisposable`. 
2. This thread is marked as background. [It will be killed by the runtime when all foreground tasks will be finished](https://learn.microsoft.com/en-us/dotnet/standard/threading/foreground-and-background-threads).
3. There is not need to care if all data are correctly _exported_. By the nature, we are sampling. In the worst case the process of handling one buffer will be not handled correctly. IMO it is fine.

Technically, we could pass cancellation token and call it from 'Instrumentation.OnExit'. But it might extend killing process time. I would keep it as is, but if you thing otherwise, we can improve in follow up issue. What do you thint?

Kielek:
> nit: I recommend placing a launchSettings.json file with environment variables in the `test/test-applications/integrations/TestApplication.ContinuousProfiler/` app. This will make it easier for people to test this project.

Done in 2b957d15b79f5d18ac35b66d0163c23044d22a3d

src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
pellared: Then `sampleExporter.Process();` should also accept a cancellation token so that the export can be interrupted during application shutdown

Kielek: Is it a blocker for you? If now, I would keep it as it is for this PR and consider a follow up.

BTW I am not sure if adding Task.Delay here bill brings a lot of benefits. We are in the dedicated thread which handles calls synchrnously.

pellared: I think it will give benefits only if we decide to address https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3196#discussion_r1440919726

rajkumar-rangaraj: Not a blocker from my side. Leave a TODO comment to revisit this part later.

src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
rajkumar-rangaraj: Same here, leave a TODO we could investigate the need for it later.

RassK (approved):
LGTM

nit: may need to recheck native formatting rules, loads of missing lines, different styles etc. Too much to deal with in this branch.

src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
Kielek: Done: 93a5e853fb88cd4d3f053057281b3ea25eee0531 #3216

src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
Kielek: Done: 93a5e853fb88cd4d3f053057281b3ea25eee0531 #3216

src/OpenTelemetry.AutoInstrumentation/Plugins/PluginManager.ContinuousProfiler.cs
```diff
@@ -0,0 +1,24 @@
...
+        foreach (var plugin in _plugins)
+        {
+            var mi = plugin.Type.GetMethod("GetContinuousProfilerConfiguration");
+            if (mi?.Invoke(plugin.Instance, null) is Tuple<bool, uint, bool, uint, TimeSpan, object> continuousProfilerConfiguration)
+            {
+                return (continuousProfilerConfiguration.Item1, continuousProfilerConfiguration.Item2, continuousProfilerConfiguration.Item3, continuousProfilerConfiguration.Item4, continuousProfilerConfiguration.Item5, continuousProfilerConfiguration.Item6);
+            }
+        }
+
+        return (false, 0u, false, 0u, TimeSpan.Zero, new object());
```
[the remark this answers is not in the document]
Kielek: Lets treat  new object() here as creating instance of new NoopExporter(). It will work in the same way.

I will keep it as is, as you marked it as nit. I am open for follow up PRs to change accept it. :)

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -1126,6 +1130,54 @@ void CorProfiler::InternalAddInstrumentation(WCHAR* id, CallTargetDefinition* it
...
+    DWORD pdvEventsLow;
+    DWORD pdvEventsHigh;
+    auto  hr = this->info12_->GetEventMask2(&pdvEventsLow, &pdvEventsHigh);
+    if (FAILED(hr))
+    {
+        Logger::Warn("ConfigureContinuousProfiler: Failed to take event masks for continuous profiler.");
+        return;
+    }
+
+    pdvEventsLow |= COR_PRF_MONITOR_THREADS | COR_PRF_ENABLE_STACK_SNAPSHOT;
```
zacharycmontoya, line 1157: Are these flags needed if only allocation sampling is enabled?

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -137,6 +167,12 @@ public static void Initialize()
                     Logger.Information("Initialized lazily-loaded metric instrumentations without initializing sdk.");
                 }
             }
+#if NET6_0_OR_GREATER
+            if (profilerEnabled && (threadSamplingEnabled || allocationSamplingEnabled))
```
zacharycmontoya, line 171: Is there a reason this is done last? Could this be moved to the earlier conditional blocks?

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -0,0 +1,1076 @@
...
+    allocation sampling
+  - a cache of thread id->thread name (set by each thread itself, used during either sample type)
+  - a cache of thread span context state (set by any application thread at any time, used during either sample type)
+
+  In general we want to keep locks "adjacent" to just one data structure and usage of them local to one
+  modifying/reading method to simplify analysis.  However, there are some special cases.
+  Here are the locks in use:
+  - cpu_buffer_lock guarding access to the thread samples buffer
+  - allocation_buffer_lock guarding access to the buffer for allocation samples
+  - name_cache_lock, guarding the
```
zacharycmontoya, line 57: This looks to be missing the rest of the line

zacharycmontoya (approved):
Overall LGTM

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
Kielek: Ref: https://learn.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/cor-prf-monitor-enumeration

* `COR_PRF_MONITOR_THREADS` - needed to have the info about thread events. Needed to corelate stack traces with Traces/Spans. Needed for both allocation and thread sampling
* `COR_PRF_ENABLE_STACK_SNAPSHOT` - Enables calls to the [DoStackSnapshot](https://learn.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/icorprofilerinfo2-dostacksnapshot-method) method. `DoStackSnapshot` method is also executed for thread and allocation sampling.

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
Kielek: Technically yes, but for traces and metrics we would like to take all possible data (at least for AlwaysOn Sampler).
For the profiler, we are heavily dropping data/sampling once per second(s). So dropping data is acceptable, both at the beggining and end of the process.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
pellared: Wouldn't it be worth adding a comment in code?

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
johnbley: 😞 Oops!  `guarding the data structures used for function/class name lookup`

Kielek:
I am merging this PR as we agreed yesterday. Please let me know if you have any other comments.

PR #3213, opened 2024-01-04
Bump OTel to 1.7.0

### LINKED ISSUE #3204: Bump OTel to 1.7.0.
Bump OTel to 1.7.0.

Kielek:
## Why

Fixes #3204

## Checklist


test/IntegrationTests/BuildTests.DistributionStructure_linux.verified.txt
```diff
@@ -9,6 +9,17 @@
   /net/Grpc.Core.Api.dll,
   /net/Grpc.Net.Client.dll,
   /net/Grpc.Net.Common.dll,
+  /net/Microsoft.Extensions.Configuration.Abstractions.dll,
```
rajkumar-rangaraj, line 12: These libraries are present in the runtime store. Do we need to keep a copy of it in the net directory?

CHANGELOG.md
```diff
@@ -7,20 +7,64 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
...
+- .NET Framework only, following packages updated
+  - `Grpc.Core.Api` from `2.59.0` to `2.60.0`
+  - `OpenTelemetry.Instrumentation.AspNet` from `1.6.0-beta.2` to `1.7.0-beta.1`,
 
 ### Deprecated
 
 ### Removed
 
+Removed support for `Microsoft.Extensions.Logging`
+for `ILOGGER` for versions older than `8.0.0`.
```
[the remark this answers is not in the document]
rajkumar-rangaraj: In our documentation, `ILogger` is marked as experimental. A minor version bump would be appropriate in this case.

test/IntegrationTests/BuildTests.DistributionStructure_linux.verified.txt
Kielek: Previously referenced versions were lower than the versions in runtime store (mostly 3.1.0 vs 6.0.0). The compiler and runtime in such cases is loading never versions.

Now all of the referenced packages are in 8.0.0 which are newer than the runtime store. I think we should keep it.
It will be removed when we drop support for .NET7 and .NET6. Folder for .NET8 will not need it.

CHANGELOG.md
```diff
@@ -7,20 +7,67 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
...
+  for plugins.
 
 ### Changed
 
+- Changed minimal supported version of `Microsoft.Extensions.Logging`
+  for `ILOGGER` instrumentation from `6.0.0` to `8.0.0`.
+- Changed supported method parameters for plugins
+  - from `OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreInstrumentationOptions`
+    to `OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreTraceInstrumentationOptions`,
+  - from `OpenTelemetry.Instrumentation.Http.HttpClientTraceInstrumentationOptions`
```
lachmatt, line 27:
```suggestion
  - from `OpenTelemetry.Instrumentation.Http.HttpClientInstrumentationOptions`
```

src/OpenTelemetry.AutoInstrumentation/Loading/Initializers/AspNetMetricsInitializer.cs
```diff
@@ -24,7 +28,11 @@ private void InitializeOnFirstCall(ILifespanManager lifespanManager)
         }
 
         var instrumentationType = Type.GetType("OpenTelemetry.Instrumentation.AspNet.AspNetMetrics, OpenTelemetry.Instrumentation.AspNet");
-        var instrumentation = Activator.CreateInstance(instrumentationType);
+
+        var options = new OpenTelemetry.Instrumentation.AspNet.AspNetMetricsInstrumentationOptions();
+        _pluginManager.ConfigureTracesOptions(options);
```
Kielek, line 33:
```suggestion
        _pluginManager.ConfigureMetricsOptions(options);
```

test/IntegrationTests/BuildTests.DistributionStructure_linux.verified.txt
pjanotti: In principle they will always be loaded from the shared store, so this seems wasteful. That said it is an optimization regarding the size of instrumentation. How many bytes are they altogether?

rajkumar-rangaraj: If the runtime store contains a lower version of the Microsoft.Extensions.* package than the one found in the `net` folder, it is considered a bug. We rely on AddDeps and the Runtime Store to upgrade these libraries. If they fail to upgrade, and if we force load it, this will result in an Assembly Load Exception.

test/test-applications/integrations/TestApplication.EntityFrameworkCore.Pomelo.MySql/TestApplication.EntityFrameworkCore.Pomelo.MySql.csproj
```diff
@@ -8,9 +8,14 @@
   <ItemGroup>
     <PackageReference Include="Pomelo.EntityFrameworkCore.MySql" VersionOverride="$(LibraryVersion)" />
 
-    <!-- Microsoft.Extensions.Configuration.Abstractions is needed to bump reference to 7.0.0
+    <!-- Microsoft.Extensions.Configuration.Abstractions, Microsoft.Extensions.Logging.Abstractions,
+    Microsoft.Extensions.Options, and Microsoft.Extensions.DependencyInjection
+    are needed to bump reference to 8.0.0
     some versions of Pomelo.EntityFrameworkCore.MySql brings to old references -->
     <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" VersionOverride="8.0.0" />
```
pjanotti, line 16: I assume that the comment still applies to the packages added here.

test/test-applications/integrations/TestApplication.EntityFrameworkCore/TestApplication.EntityFrameworkCore.csproj
```diff
@@ -8,5 +8,6 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" VersionOverride="$(LibraryVersion)" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" />
+    <PackageReference Include="Microsoft.Extensions.Logging" />
```
pjanotti, line 11: I assume we need these because some of the referenced packages bring old version, is that correct?

pjanotti (commented):
A general concern are the explicit reference to some of the Microsoft.Extensions.* packages that were moved to version 8.0 - it seems that if we needed to those for tests many users will have to do the same. That said it is a known issue if there are conflicting versions brought by the application.

Could you just verify that these are really being brought by the test applications? Just to ensure that we are not missing anything on AdditionalDeps and SharedStore.

test/IntegrationTests/BuildTests.DistributionStructure_linux.verified.txt
pjanotti: The AdditionalDeps and SharedStore are built from the deps.json so they should always be referencing the same versions.  I checked the artifacts of this PR and they match that expectation. Looking at the artifacts for `v1.2.0` they also reference `8.0.0`, but, they were not on the `net` folder. I'm not sure why they were in the deps.json but not in the `net` folder (it could be that the 3.1.0 was considered part of the referenced framework), but, they look correct in this PR.

@Kielek I think that we should add duplicated libraries to `RemoveDuplicatedLibraries` in the build. Anyway, I think it is better to do it in a separate PR.

Kielek: > In principle they will always be loaded from the shared store, so this seems wasteful. That said it is an optimization regarding the size of instrumentation. How many bytes are they altogether?

548 KB (561,520 bytes)

Kielek: @rajkumar-rangaraj, I think that we have some mess in communication. I will try to summarize based on `Microsoft.Extensions.Logging.Abstractions.dll`. I hope I do not make any mistakes in the naming convention.

1. Runtime store - it is the store created by the .NET installer. For Windows, typically under the `C:\Program Files\dotnet` folder.
2. Additional deps - list of additional dependencies that are not in application location directly nor in the folder `%otel-installation-folder\net` . It references our shared store. For us, only json file, for each of supported .NET versions. -  `%otel-installation-folder%/store`
3. Shared store - store created an managed by us containing libraries defined in Additional deps. - `%otel-installation-folder%/AdditionalDeps`
4. net folder - `%otel-installation-folder%/net`

`Microsoft.Extensions.Logging.Abstractions.dll` versions

| .NET Version | Runtime version | Shared store | net folder |
| - | -|  -| -|
| .NET6 | 6.0.25 | 8.0.0 | 8.0.0 |
| .NET7 | 7.0.14 | 8.0.0 | 8.0.0 |
| .NET8 | 8.0.0 | 8.0.0 | 8.0.0 |

There is no conflict between net folder and shared store/additional dependencies.

Previously `Microsoft.Extensions.Logging.Abstractions.dll` was not in this folder because OTel SDK was referencing 3.1.0 version. It was changed in https://github.com/open-telemetry/opentelemetry-dotnet/pull/5015 and  https://github.com/open-telemetry/opentelemetry-dotnet/pull/5051.
What is more, it should be removed if we retarget `OpenTelemetry.AutoInstrumentation.csproj` from .NET6 to .NET8. Runtime store will contain expected/never version, so that the compiler will not include it the output.

If I understand it correctly, exactly the same case we have for `System.Diagnostics.DiagnosticSource.dll`.

| OTel AutoVersion | Shared store | net folder |
| - | - | -| 
| 1.2.0 | 8.0.0 | 7.0.4 |
| This PR/1.3.0 | 8.0.0 | 8.0.0 |.

Maybe I missed something, but I do not see reason to change anything in the distribution structure.

@pjanotti, `RemoveDuplicatedLibraries` is used when you have two versions of the same library in the additional store. It is not this case.

Kielek:
> A general concern are the explicit reference to some of the Microsoft.Extensions.* packages that were moved to version 8.0 - it seems that if we needed to those for tests many users will have to do the same. That said it is a known issue if there are conflicting versions brought by the application.
> 
> Could you just verify that these are really being brought by the test applications? Just to ensure that we are not missing anything on AdditionalDeps and SharedStore.

Double checked, all libraries are transitive dependencies referenced by instrumented libraries.

test/IntegrationTests/BuildTests.DistributionStructure_linux.verified.txt
pellared: 1. IMO this comment should not be addressed in this PR and it should create a separate issue for it.
2. I think that before addressing the issue above the [`design.md`](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/design.md) should be updated. We should describe all the directories (and NuGet package), how they affect assembly resolution, what directories are optional etc. I feel that we are discussing the same things around these folders back and forth.

pjanotti (approved):

test/IntegrationTests/BuildTests.DistributionStructure_linux.verified.txt
rajkumar-rangaraj: I'm fine with addressing this in a new PR, but let's agree to hold the release until this issue is addressed.

@Kielek We will discuss this in tomorrow's SIG. If either of us misses the SIG, I will try to chat with you separately.

pjanotti: SIG: removal of added assemblies will not be done in this PR but prior to 1.3.0 release.

pjanotti: Follow-up via https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3223

rajkumar-rangaraj (approved):

PR #3273, opened 2024-02-14
Tests - support for library versions only for subset of target frameworks

Kielek:
## Why

Handles case from #3097

## What

Tests - support for library versions only for subset of target frameworks

## Tests

CI + manual build in VS.

## Checklist


tools/LibraryVersionsGenerator/Models/PackageVersion.cs
```diff
@@ -5,10 +5,13 @@ namespace LibraryVersionsGenerator.Models;
 
 internal class PackageVersion
 {
-    public PackageVersion(string version)
+    public PackageVersion(string version, params string[] supportedFrameworks)
```
RassK, line 8: I think using a second constructor removes all of that `Array.Empty<string>()` bloat.

build/LibraryVersions.g.cs
```diff
@@ -19,149 +19,151 @@ public static class LibraryVersion
...
         },
         {
             "TestApplication.EntityFrameworkCore.Pomelo.MySql",
             new List<PackageBuildInfo>
             {
-                new("6.0.2"),
-                new("7.0.0"),
+                new("6.0.2", Array.Empty<string>()),
+                new("7.0.0", Array.Empty<string>()),
+                new("8.0.0", new string[] {"net8.0"}),
```
RassK, line 50: Wondering if it is supported already. 

```suggestion
                new("8.0.0", ["net8.0"]),
```

Kielek: .NET 7 on centos, it can be hard, I would keep as is for now

tools/LibraryVersionsGenerator/Models/PackageVersion.cs
Kielek: 38319538bc8a57a994866d45b4f5f1bf0cd2c663

build/LibraryVersions.g.cs
RassK: I mean the new shorter array literals with latest C# version 😄

Kielek: We do not have the latest version on Centos builds :] There is only .NET7

RassK (approved):

nrcventura (approved):

rajkumar-rangaraj (approved):

Issue #2453, opened 2023-04-14
Update native code based on upstream to bring ARM64 support


Kielek:
I haven't tried on Linux ARM64 yet, but I just tried it on OSX ARM64 and it _does not_ work. There are upstream changes made at a later point in time that should add support.

_Originally posted by @zacharycmontoya in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2280#discussion_r1166134898_

Issue #2314, opened 2023-03-14
ARM64 - support for opentelemetry-operator


Kielek:
# Feature Request

Based on ##2296. First we need #1865 and #2181.

rkferreira:
Hi, it seems still not working for arm64 on Otel Operator for Auto Instrumentation.

pjanotti:
Hi @rkferreira - could you provide details about the error that you are getting? In principle this work is done and we need some info to identify if this is an issue for dotnet-autoinstrumentation or the operator itself.

navaneeth-dev:
With Jellyfin for example it loads x64 versions by default?

```bash
I have no name!@jellyfin-fbcbb88ff-985fq:/$ cat /tmp/otel-dotnet-auto-jellyfin-1-StartupHook-20260207.log
[2026-02-07T06:12:27.0805696Z] [Debug] Rule Engine: jellyfin is not in the exclusion list. ApplicationInExcludeListRule evaluation success.
[2026-02-07T06:12:27.1343173Z] [Information] Rule Engine: MinSupportedFrameworkRule evaluation success.
[2026-02-07T06:12:27.1379877Z] [Information] Rule Engine: OpenTelemetrySdkMinimumVersionRule evaluation success.
[2026-02-07T06:12:27.2082886Z] [Debug] Error checking if native profiler is attached.
Exception: Unable to load shared library '/otel-auto-instrumentation-dotnet/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so' or one of its dependencies. In order to help diagnose loading problems, consider using a tool like strace. If you're using glibc, consider setting the LD_DEBUG environment variable:
/otel-auto-instrumentation-dotnet/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so: cannot open shared object file: No such file or directory

System.DllNotFoundException: Unable to load shared library '/otel-auto-instrumentation-dotnet/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so' or one of its dependencies. In order to help diagnose loading problems, consider using a tool like strace. If you're using glibc, consider setting the LD_DEBUG environment variable:
/otel-auto-instrumentation-dotnet/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so: cannot open shared object file: No such file or directory

   at System.Runtime.InteropServices.NativeLibrary.Load(String libraryPath)
[...]

PR #3277, opened 2024-02-16
Add ARM64 Support

### LINKED ISSUE #1865: Add arm64 build for native components

### LINKED ISSUE #2181: Add arm64 architecture to the shared store

RassK:
## Why

Fixes #1865
Fixes #2181

Towards #2314
Towards #2453

## What

Adds ARM64 support for ubuntu, debian and alpine.
CentOS based images are **not supported** with .NET 6 and .NET 7

## Tests

Tests .NET 6, .NET 7 and .NET 8.

* arm64 ubuntu 22.04 (native platform)
* arm64 debian (container)
* arm64 alpine (container)
* arm64 nuget package tests

## Follow up

`Confluent.Kafka 1.4.0`  is not supported by arm64, is disabled temporarily due missing platform filtration capabilities.
`Confluent.Kafka 1.6.2` does not work on x64 and arm64. This seems to require separate focus.
Follow up task here > https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3306

## Checklist


RassK:
Note: Docs issue, it's looking for a dockerfile from a main branch, that does not exist before merge.

.github/workflows/build.yml
```diff
@@ -19,52 +19,85 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v4.1.1
         with:
           fetch-depth: 0 # fetching all, needed to correctly calculate version
-      - uses: actions/setup-dotnet@v4.0.0
+
+      - name: Setup ARM64 Environment Variables
+        if: ${{ matrix.machine == 'actuated-arm64' }}
+        run: |
+          echo "DOTNET_INSTALL_DIR=~/.dotnet" >> $GITHUB_ENV
```
pjanotti, line 39: Why this is not needed by the other architectures?

test/IntegrationTests/Helpers/EnvironmentTools.cs
```diff
@@ -92,6 +92,17 @@ public static string GetPlatform()
         return RuntimeInformation.ProcessArchitecture.ToString();
     }
 
+    public static string GetPlatformDir()
+    {
+        return RuntimeInformation.ProcessArchitecture switch
+        {
+            Architecture.X86 => "x86",
+            Architecture.X64 => "x64",
+            Architecture.Arm64 => "ARM64",
```
pjanotti, line 101: why the different casing here?

test/IntegrationTests/IntegrationTests.csproj
```diff
@@ -4,12 +4,21 @@
     <AddTestStrongNameAssemblyKeyOnNetFramework>true</AddTestStrongNameAssemblyKeyOnNetFramework>
     <!-- No warn is needed for autogenerated code by protobuf: CS8981 - The type name only contains lower-cased ascii characters. -->
     <NoWarn Condition="'$(TargetFramework)' == 'net8.0' OR '$(TargetFramework)' == 'net7.0'">CS8981</NoWarn>
+    <Platforms>x64;ARM64</Platforms>
```
pjanotti, line 7: Is this really needed?

.github/workflows/build.yml
RassK: `actions/setup-dotnet@v4.0.0` had somekind of issues running without root privileges. I guess other runners are configured differently.

test/IntegrationTests/Helpers/EnvironmentTools.cs
RassK: Not to cause (weird) issues with Visual Studio. I have no idea why they defined ARM64 all uppercase in $(AvailablePlatforms).

test/IntegrationTests/IntegrationTests.csproj
RassK: Yes, this is connected to _X64 and _ARM64 constant. Visual Studio is using a `$(Platform)` variable to define these constants. It also makes Visual Studio interacting with the code that is marked as `#if _X64` so as soon as you select x64 the code has intellisense.

.github/workflows/build.yml
RassK:
```
Run actions/setup-dotnet@v4.0.0
/home/runner/actions-runner/_work/_actions/actions/setup-dotnet/v4.0.0/externals/install-dotnet.sh --skip-non-versioned-files --runtime dotnet --channel LTS
mkdir: cannot create directory ‘/usr/share/dotnet’: Permission denied
```

.github/workflows/release-publish.yml
Kielek, line 1: Did you have chance to test it, e.g. on your fork? Maybe you can make something like 1.4.0-alpha.1?

docs/README.md
```diff
@@ -3,6 +3,12 @@
 [![Slack](https://img.shields.io/badge/slack-@cncf/otel--dotnet--auto--instr-brightgreen.svg?logo=slack)](https://cloud-native.slack.com/archives/C01NR1YLSE7)
 [![NuGet](https://img.shields.io/nuget/v/OpenTelemetry.AutoInstrumentation.svg)](https://www.nuget.org/packages/OpenTelemetry.AutoInstrumentation)
 [![NuGet](https://img.shields.io/nuget/dt/OpenTelemetry.AutoInstrumentation.svg)](https://www.nuget.org/packages/OpenTelemetry.AutoInstrumentation)
+<a href="https://actuated.dev/">
+  <img
+    alt="Arm CI sponsored by Actuated"
+    src="https://docs.actuated.dev/images/actuated-badge.png" width="120px">
+  </img>
+</a>
```
Kielek, line 11: Tracked this syntax on: https://github.com/self-actuated/docs.actuated.dev/issues/18

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
 
 - .NET only, warning in logs about End of Support date
   and upcoming End of Support date for .NET version.
+- ARM64 support for Ubuntu, Alpine and Debian based images.
+- ARM64 support for the `OpenTelemetry.AutoInstrumentation` NuGet package.
```
Kielek, line 15: I would consider putting something like - experimental support for (here and in the docs). With this, we can release ARM64 in stable packages. Without this, it should be supported only on prereleases. It applies also to REDME.md

Kielek (commented):
Good job. Some minor comments.

.github/workflows/build-container.yml
```diff
@@ -13,20 +13,30 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - base-image: alpine
+          - machine: ubuntu-20.04
+            base-image: alpine
             os-type: linux-musl
-    runs-on: ubuntu-20.04
+            architecture: x64
+          - machine: actuated-arm64
```
pellared, line 20: Is there any list of what Actuated machines are available? Any docs that would help us in troubleshooting?

RassK: I think maintainers should see it? Actions -> (left side label: management) runners

pellared: I see only regular GitHub runners.

FYI related discussion: https://github.com/open-telemetry/opentelemetry-go/pull/4994#discussion_r1508641578

RassK: Yes that would have been next issue after there was more information, which runners are even available.
I was told also that they are going to remove the default label somewhere in the near? future.

tools/LibraryVersionsGenerator/PackageVersionDefinitions.cs
```diff
@@ -216,7 +216,8 @@ internal static class PackageVersionDefinitions
             TestApplicationName = "TestApplication.Kafka",
             Versions = new List<PackageVersion>
             {
-                new("1.4.0"),
+                // TODO: Temporarily disable version 1.4.0, Doesn't support ARM64
+                // new("1.4.0"),
```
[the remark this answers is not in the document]
RassK: Interestingly 1.6.2 causes issues also on x64 while 1.4.0 was ok 🤔 This is the first release that ships arm64 native binaries also. Might need bigger research, we might have holes in the supported range.

.github/workflows/build-container.yml
pellared: Look like it would be removed rather sooner than later. I suggest changing this line to 

```suggestion
          - machine: actuated-arm64-4cpu-8gb
```

https://github.com/open-telemetry/opentelemetry-go/pull/4994#discussion_r1509178904

RassK: will update within this scope then 👍

test/IntegrationTests/IntegrationTests.csproj
zacharycmontoya: Instead of assigning a Platform here and in the Nuke code, can we 1) compile these as AnyCPU, 2) remove the ifdefs, and 3) check the architecture dynamically when the test is run and skip the test then?

The reason I ask is that the solution file is building this project as AnyCPU for release builds but platform-specific for debug builds, and I'd like to build managed code as AnyCPU as much as possible.

RassK: The only issue with runtime check was that it has to be done in multiple places. 
1. In the test itself, need to change Theory to SkippableTheory and then skip the test if arm64
2. In the `SqlServerFixture` (or any fixture that does not work) cause this executes way before the test and fails there.

Is a bit more fragmented approach but still not that complicated to do. Personally don't have strong feelings about either of the approach.

.github/workflows/build.yml
RassK: https://github.com/actions/setup-dotnet/issues/486

build/Build.cs
```diff
@@ -11,8 +11,8 @@ partial class Build : NukeBuild
     [Parameter("Configuration to build. Default is 'Release'")]
     readonly Configuration BuildConfiguration = Configuration.Release;
 
-    [Parameter("Platform to build - x86 or x64. Default is 'x64'")]
-    readonly MSBuildTargetPlatform Platform = MSBuildTargetPlatform.x64;
+    [Parameter("Platform to build - x86, x64, ARM64. Default is current architecture.")]
```
rajkumar-rangaraj, line 14: nit: Provide a better clarity her - "Default is current architecture."

RassK: any suggestions?

.github/workflows/build-container.yml
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3277/commits/2cbc49253876099d927e59f27c1ba9ef194c6f53

CHANGELOG.md
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3277/commits/33c5bf20fe54381f78566fe8b0b74f13be8b8915

Kielek (approved):

test/IntegrationTests/IntegrationTests.csproj
pjanotti: I like @zacharycmontoya suggestion, but, it can come in a follow up.

pjanotti (approved):

test/IntegrationTests/IntegrationTests.csproj
RassK: I think @Kielek was currently the only one who preferred more compile time filter than runtime... which is definitely cleaner solution. But up to you... I'd prefer to do it in the current branch (while there is history to check just in case)

docs/README.md
RassK: Badge is merged 👍 I'll update the docs also in the current branch tomorrow.

test/IntegrationTests/IntegrationTests.csproj
Kielek: From my perspective, it is just a preference. Any working options should be fine :)

docs/README.md
Kielek:
```suggestion
[![Arm CI sponsored by Actuated](https://img.shields.io/badge/SA_actuated.dev-004BDD)](https://actuated.dev/)
```

Great!

.markdownlint.yaml
```diff
@@ -12,3 +12,8 @@ MD024:
   allow_different_nesting: true
   # Only check sibling headings
   siblings_only: true
+
+# MD033/no-inline-html : Inline HTML : https://github.com/DavidAnson/markdownlint/blob/v0.33.0/doc/md033.md
+MD033:
+  # Allowed elements
+  allowed_elements: ['a', 'img']
```
Kielek, line 19:
```suggestion
```

This should be also not needed.

pellared (approved):

Kielek:
@rajkumar-rangaraj, @zacharycmontoya, do you see any objections to merge it?

If no, I will adjust merge build requierments.

pellared:
> @rajkumar-rangaraj, @zacharycmontoya, do you see any objections to merge it?
> 
> If no, I will adjust merge build requierments.

See https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3307

zacharycmontoya (approved):

test/IntegrationTests/IntegrationTests.csproj
zacharycmontoya: Sounds good, it's good to merge and we can perform this as follow-up 👍🏼

PR #3285, opened 2024-02-21
Remove Windows x86 shared store files

### LINKED ISSUE #3279: Remove x86 output for AdditionalDeps
### What

Currently we output both x86 and x64 for AdditionalDeps Windows project.

Here is the general overview of what we have:
o - we have output but no compatible binary in `net` folder.
x - we have properly supported output.
a - we use the old AnyCpu flag that makes compatible binary for x86 and x64 for .NET Framework.

(arm64) - means the planned route and not yet available.

|                                              | x86 | x64 | (arm64) | 
| -------------------------------| ----| ---- | --------- |
| Windows (.NET Framework) |  a  |   a   |        -      |
| Windows (.NET Core)           |  o   |   x   |        -      |
| Linux                                    |  -   |   x   |        x      |
| MacOS                                 |  -   |   x   |        -      |

### Propose

Cleanup AdditionalDeps by removing x86 output.

CC: @rajkumar-rangaraj

RassK:
## Why

Fixes #3279

## What

Removes x86 binaries from Windows Shared Store, which aren't used anyway.

## Tests

Updated distribution tests.

## Checklist


Kielek (changes requested):
Based on @zacharycmontoya comment: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3279#issuecomment-1958775093

Issue #3216, opened 2024-01-05
Continuous profiler - exporter thread - graceful shutdown


Kielek:
### Dispose thread

> Is there a way to dispose this thread?

_Originally posted by @rajkumar-rangaraj in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3196#discussion_r1440919726_

> 1. `Thread` is not implementing `IDisposable`. 
> 2. This thread is marked as background. [It will be killed by the runtime when all foreground tasks will be finished](https://learn.microsoft.com/en-us/dotnet/standard/threading/foreground-and-background-threads).
> 3. There is not need to care if all data are correctly _exported_. By the nature, we are sampling. In the worst case the process of handling one buffer will be not handled correctly. IMO it is fine.
>
> Technically, we could pass cancellation token and call it from 'Instrumentation.OnExit'. But it might extend killing process time. I would keep it as is, but if you thing otherwise, we can improve in follow up issue. What do you thint?

_Originally posted by @Kielek in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3196#discussion_r1441381638_

### Task.Delay/Cancellation Token

> Consider using Task.Delay with async pattern and if possible send cancellation Token as parameter.

_Originally posted by @rajkumar-rangaraj in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3196#discussion_r1440919353_
            
> Then `sampleExporter.Process()` should also accept a cancellation token so that the export can be interrupted during application shutdown

_Originally posted by @pellared in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3196#discussion_r1441395986_

> BTW I am not sure if adding Task.Delay here bill brings a lot of benefits. We are in the dedicated thread which handles calls synchrnously.

_Originally posted by @Kielek in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3196#discussion_r1441474296_

pellared:
SIG meeting:

The issue is not that minor as .NET does NOT always get rid of Background threads which are still running.

PR #3294, opened 2024-02-28
[continuous-profiler] graceful export thread shutdown

lachmatt:
## Why

Towards #3216

## What

Graceful export thread shutdown.

## Tests

Existing tests.

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
@@ -105,15 +105,13 @@ internal class ContinuousProfilerProcessor : IDisposable
         while (true)
         {
             // TODO Task.Delay https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3216
-            var shutdownRequested = _shutdownTrigger.Wait(_exportInterval);
-
-            sampleExporter.Process();
-
-            if (shutdownRequested)
+            if (_shutdownTrigger.Wait(_exportInterval))
             {
                 Logger.Debug("Shutdown requested, exiting continuous profiler's exporter thread.");
                 return;
             }
+
+            sampleExporter.Process();
         }
     }
 }
--- a/src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
@@ -21,25 +21,22 @@ internal class ContinuousProfilerProcessor : IDisposable
     private readonly Thread _thread;
     private readonly ManualResetEventSlim _shutdownTrigger = new(false);
     private readonly TimeSpan _exportInterval;
+    private readonly BufferProcessor _bufferProcessor;
 
-    public ContinuousProfilerProcessor(bool threadSamplingEnabled, bool allocationSamplingEnabled, Action<byte[], int> threadSamplesMethod, Action<byte[], int> allocationSamplesMethod, TimeSpan exportInterval)
+    public ContinuousProfilerProcessor(BufferProcessor bufferProcessor, TimeSpan exportInterval)
     {
         Logger.Debug("Initializing Continuous Profiler export thread.");
 
         _exportInterval = exportInterval;
+        _bufferProcessor = bufferProcessor;
         _supportingActivityAsyncLocal = new AsyncLocal<Activity?>(ActivityChanged);
 
-        _thread = new Thread(() =>
-        {
-            SampleReadingThread(new BufferProcessor(threadSamplingEnabled, allocationSamplingEnabled, threadSamplesMethod, allocationSamplesMethod));
-        })
+        _thread = new Thread(SampleReadingThread)
         {
             Name = BackgroundThreadName,
             IsBackground = true
         };
         _thread.Start();
-
-        Logger.Information("Continuous Profiler export thread initialized.");
     }
 
     public void Activity_CurrentChanged(object? sender, ActivityChangedEventArgs e)
@@ -78,8 +75,10 @@ internal class ContinuousProfilerProcessor : IDisposable
         NativeMethods.ContinuousProfilerSetNativeContext(0, 0, 0);
     }
 
-    private void SampleReadingThread(BufferProcessor sampleExporter)
+    private void SampleReadingThread()
     {
+        Logger.Information("Continuous Profiler export thread initialized.");
+
         while (true)
         {
             // TODO Task.Delay https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3216
@@ -89,7 +88,7 @@ internal class ContinuousProfilerProcessor : IDisposable
                 return;
             }
 
-            sampleExporter.Process();
+            _bufferProcessor.Process();
         }
     }
 }
--- a/src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
@@ -5,29 +5,23 @@
 
 using System.Diagnostics;
 using System.Globalization;
-using System.Reflection;
 using OpenTelemetry.AutoInstrumentation.Logging;
 
 namespace OpenTelemetry.AutoInstrumentation.ContinuousProfiler;
 
-internal static class ContinuousProfilerProcessor
+internal class ContinuousProfilerProcessor : IDisposable
 {
-    public const string BackgroundThreadName = "OpenTelemetry Continuous Profiler Thread";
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -87,20 +91,18 @@ public static void Initialize()
...
                 Logger.Debug($"Continuous profiling configuration: Thread sampling enabled: {threadSamplingEnabled}, thread sampling interval: {threadSamplingInterval}, allocation sampling enabled: {allocationSamplingEnabled}, max memory samples per minute: {maxMemorySamplesPerMinute}, export interval: {exportInterval}, continuous profiler exporter: {continuousProfilerExporter.GetType()}");
 
                 if (threadSamplingEnabled || allocationSamplingEnabled)
                 {
                     NativeMethods.ConfigureNativeContinuousProfiler(threadSamplingEnabled, threadSamplingInterval, allocationSamplingEnabled, maxMemorySamplesPerMinute);
-                    Activity.CurrentChanged += ContinuousProfilerProcessor.Activity_CurrentChanged;
+
+                    _profilerProcessor = new ContinuousProfilerProcessor(threadSamplingEnabled, allocationSamplingEnabled, exportInterval, continuousProfilerExporter);
+                    Activity.CurrentChanged += _profilerProcessor.Activity_CurrentChanged;
```
RassK, line 105: it may create more activities if `_profilerProcessor` is disposed, so could trigger nullreference exception?

src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
```diff
@@ -86,13 +94,20 @@ private static void ActivityChanged(AsyncLocalValueChangedArgs<Activity?> sender
...
+    private void SampleReadingThread(BufferProcessor sampleExporter, TimeSpan exportInterval)
     {
         while (true)
         {
-            // TODO Graceful shutdown and Task.Delay https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3216
-            Thread.Sleep(exportInterval);
+            // TODO Task.Delay https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3216
+            var shutdownRequested = _shutdownTrigger.Wait(exportInterval);
+
             sampleExporter.Process();
```
pellared, line 104: This call may never exit. `sampleExporter.Process()` can starve the shutdown as there is no timeout nor cancelation handling.

Ideas how this can be handled:
1. We can pass a "export timeout" or CancelationToken to the `threadSamplesMethod` and `allocationSamplesMethod` methods. The responsibility of cancelling is moved to the called method (plugin). The good think is that we will get e.g. logs what happened wrong.
2. We can also `_thread?.Abort()` the thread. I think this is a "last resort" if the plugin does handles the cancelation properly. We can execute it e.g. if the plugin does not cancel in `2 * exportTimeout`.

lachmatt: That's a very good point.
Option 2 won't work as `Thread.Abort` is [not supported](https://learn.microsoft.com/en-us/dotnet/api/system.threading.thread.abort?view=net-8.0) for .NET5+.
That leaves us with option 1, will update the PR accordingly.

pellared: We discussed that instead of option 2 we can simply replace `_thread?.Join()` with `_thread?.Join(exportTimeout)`

pellared: We can start with option 2 and apply option 1 in a separate PR.

lachmatt: fixed in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3294/commits/0b6fd27fbeee942c36031a6987919cc79b918eb8

src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
```diff
@@ -86,13 +100,20 @@ private static void ActivityChanged(AsyncLocalValueChangedArgs<Activity?> sender
...
+            // TODO Task.Delay https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3216
+            var shutdownRequested = _shutdownTrigger.Wait(_exportInterval);
+
             sampleExporter.Process();
+
+            if (shutdownRequested)
+            {
+                Logger.Debug("Shutdown requested, exiting continuous profiler's exporter thread.");
+                return;
+            }
```
Kielek, line 116: I would put it before process. loosing one set of sampled data should not be a problem.
The bigger problem should be waiting for finishing the process.

src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
```diff
@@ -86,13 +100,20 @@ private static void ActivityChanged(AsyncLocalValueChangedArgs<Activity?> sender
...
 
-    private static void SampleReadingThread(BufferProcessor sampleExporter, TimeSpan exportInterval)
+    private void SampleReadingThread(BufferProcessor sampleExporter)
     {
         while (true)
         {
-            // TODO Graceful shutdown and Task.Delay https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3216
-            Thread.Sleep(exportInterval);
+            // TODO Task.Delay https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3216
+            var shutdownRequested = _shutdownTrigger.Wait(_exportInterval);
```
Kielek, line 108: What is the typical timeout, where OTLPexporters are waiting?

src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
```diff
@@ -47,25 +42,44 @@ public static void Initialize(bool threadSamplingEnabled, bool allocationSamplin
...
+        if (_supportingActivityAsyncLocal != null)
+        {
+            _supportingActivityAsyncLocal.Value = e.Current;
+        }
+    }
+
+    public void Dispose()
+    {
+        _shutdownTrigger.Set();
+        if (_thread != null && !_thread.Join(_exportInterval))
```
pellared, line 75: Could `_thread` be ever null?

src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
```diff
@@ -5,29 +5,24 @@
...
-    public const string BackgroundThreadName = "OpenTelemetry Continuous Profiler Thread";
+    private const string BackgroundThreadName = "OpenTelemetry Continuous Profiler Thread";
 
     private static readonly IOtelLogger Logger = OtelLogging.GetLogger();
+
     // Additional async local required to get full set of notifications,
     // see https://github.com/dotnet/runtime/issues/67276#issuecomment-1089877762
-    private static AsyncLocal<Activity?>? _supportingActivityAsyncLocal;
+    private readonly AsyncLocal<Activity?>? _supportingActivityAsyncLocal;
+    private readonly Thread? _thread;
```
pellared, line 21:
```suggestion
    private readonly Thread _thread;
```

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
lachmatt: changed in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3294/commits/d6db263c07d68c664dfa91fa46118c3deb375827

src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/ContinuousProfilerProcessor.cs
lachmatt: Default timeout is [5s](https://github.com/open-telemetry/opentelemetry-dotnet/blob/77ef12327f720ca3defd0c9590c0197cceb5952e/src/OpenTelemetry/BatchExportProcessor.cs#L18).

lachmatt: Adjusted the code in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3294/commits/0a3b315354cf12b2a50421747a62ee4cb390a0ee to wait for the same amount (5s) as when tracer/meter provider are being disposed.

Kielek (approved):

pellared (approved):

nrcventura (approved):

PR #3316, opened 2024-03-12
Define supported Kafka versions

### LINKED ISSUE #3306: Redefine supported Kafka ranges
### What

* Define Kafka supported ranges for arm64
* Find holes in the x64 supported range

context: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3277#discussion_r1509106239

RassK:
## Why

Fixes #3306

## What

* Makes `LibraryVersion` partial (logic + generated definitions)
* Can generate different versions for different platforms (x64 vs ARM64).
* Fixes Kafka tests to support other versions
* Adds Kafka supported range for ARM64 (docs)

## Tests

Existing.

Tests do not work with versions 1.8.0-1.8.1 on ARM64. I'm not sure if it's because of the OS version or the range itself is broken.

## Checklist


docs/config.md
```diff
@@ -127,38 +127,42 @@ for example `TRACES`, and `{0}` is the case-sensitive name of the instrumentatio
...
+\[2\]: `ASP.NET (.NET Framework) MVC / WebApi` is not supported on ARM64.
 
-\[3\]: `Elastic.Clients.Elasticsearch` version ≥8.0.0 and <8.10.0.
+\[3\]: `Azure.` prefixed packages, released after October 1, 2021.
+
+\[4\]: `Elastic.Clients.Elasticsearch` version ≥8.0.0 and <8.10.0.
         Version ≥8.10.0 is supported by `Elastic.Transport` instrumentation.
 
-\[4\]: `Microsoft.Data.SqlClient` v3.* is not supported on .NET Framework,
+\[5\]: `Confluent.Kafka` is supported from version ≥1.6.2 on ARM64.
```
nrcventura, line 163: Should we just say that it is supported from version >= 1.8.2 on ARM64 because of the problems that you noticed with the tests?

RassK: 1.6.3 and 1.7.0 seem to be one of the top downloaded versions. Unfortunately I don't know how important are these versions for arm64. In the same time 1.8.0 and 1.8.1 are one of the least downloaded versions plus there is 1.8.2 also extremely popular. So we may just add that 1.8.0-1.8.1 are not guaranteed to work.

nrcventura: I am fine with either calling out that we support 1.8.2+, or saying that we support >= 1.6.2 with the caveat of not supporting 1.8.0 and 1.81 (supporting 1.8.2+ is easier to understand). I just think that it is misleading to say that we support 1.6.2+ when we were not able to validate 1.8.0 and 1.8.1 (we do not understand why it did not work yet).

RassK: Agree, let's start with 1.8.2+. If anyone feels it should be lower, then they can always open an issue.

nrcventura (approved):

Kielek (approved):

Issue #3168, opened 2023-11-30
Version mismatch


Kielek:
>Same issue here. Where the project, or dependant projects, have direct dependencies on Microsoft.Extensions.* 6.0, we are having `Could not load file or assembly` issues. Adding OpenTelemetry.AutoInstrumentation nuget package does not solve the issue.
> 
> I am not familiar on how the injection mechanism work, but why a net6 project should hard reference Microsoft.Extensions 7.* packages?
Shouldn't be that depending on the target framework, the dependancy is >= [targetframework].*

_Originally posted by @federicobarera in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3075#issuecomment-1832168095_

Kielek:
@federicobarera, could you please create [Minimal, Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example)? It should include all steps you are using to build the package.

federicobarera:
Hi, you can find a repro with instructions here: https://github.com/federicobarera/otel-auto-repro

rajkumar-rangaraj:
To resolve an issue, you have to either remove this package reference or update it to version 7.0.0: `<PackageReference Include="Microsoft.Extensions.Configuration" Version="6.0.1" />` in the https://github.com/federicobarera/otel-auto-repro/blob/master/Dependencies/Dependencies.csproj. We are aware of the root cause of the issue and will address the application crash issue in the next version. Instead of crashing the app, it should safely back off and provide more detailed information in the log files to aid in fixing the issue.

federicobarera:
@rajkumar-rangaraj, @Kielek 

I have a couple more questions regarding this issue.

I did read the [Troubleshooting](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/troubleshooting.md#assembly-version-conflicts) and we eventually managed to align the dependencies in the test project, however it is stated on the docs that adding OpenTelemetry.AutoInstrumentation should have automatically sorted the dependencies. Given we are planning a fairly large rollout, across tens of teams and projects, we hoped for the autoinstrumention not to require code changes. Is there any plan for auto the auto instrumentation libraries to work with available libraries instead of forcing the requirement on 7.x

net8. We also had problem with net8 and the 0.43 operator (referencing 1.11 tag - Microsoft.Extensions 7.x).
[...]

PR #3187, opened 2023-12-14
Add Diagnostic Logging for Runtime Store Assembly Version Validation

rajkumar-rangaraj:
## Why

Addresses part of #3168, #3075 by adding logging about an issue.

Adding the below package reference to `Examples.AspNetCoreMvc` project for net6.0 target it will recreate crash issue.
 
```
<PackageReference Include="Microsoft.Extensions.Configuration" Version="6.0.1" />
```

## What

This rule logs diagnostic messages about the assembly versions in the runtime store compared to the application's assembly versions. It is designed to inform rather than enforce, ensuring that any potential versioning issues are noted without affecting the runtime behavior of the application.

**Example:** 
```
[2024-01-05T20:39:28.2247519Z] [Warning] Rule Engine: Application has direct or indirect reference to lower version of runtime store assembly C:\repo\opentelemetry-dotnet-instrumentation\bin\tracer-home\store\x64\net6.0\microsoft.extensions.configuration\8.0.0\lib\net8.0\Microsoft.Extensions.Configuration.dll - 6.0.322.12309.  
[2024-01-05T20:39:28.3405484Z] [Debug] Rule Engine: Runtime store assembly C:\repo\opentelemetry-dotnet-instrumentation\bin\tracer-home\store\x64\net6.0\microsoft.extensions.configuration.abstractions\8.0.0\lib\net8.0\Microsoft.Extensions.Configuration.Abstractions.dll is validated successfully. 
[2024-01-05T20:39:28.4660596Z] [Debug] Rule Engine: Runtime store assembly C:\repo\opentelemetry-dotnet-instrumentation\bin\tracer-home\store\x64\net6.0\microsoft.extensions.configuration.binder\8.0.0\lib\net8.0\Microsoft.Extensions.Configuration.Binder.dll is validated successfully. 

```

## Tests

* Tested manually
* Existing CI
* All Rule Engine test needs to be covered through - #2826. We need to design integration tests for these scenarios.

## Checklist

- ~~[ ] `CHANGELOG.md` is updated.~~
- ~~[ ] Documentation is updated.~~
- ~~[ ] New features are covered by tests.~~

Issue #3075, opened 2023-11-07
Version mismatch: Could not load file or assembly 'Microsoft.Extensions.DependencyInjection.Abstractions'


lorenzoiuri:
# Bug Report

## Symptom

**Describe the bug**
Instrumented pod crashes with error (note `Version=6.0.0.0` and `Version=3.1.0.0`)
```
Unhandled exception. System.IO.FileLoadException: Could not load file or assembly 'Microsoft.Extensions.DependencyInjection.Abstractions, Version=6.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. Uncaught exception during type initialization. (0x80131534)
File name: 'Microsoft.Extensions.DependencyInjection.Abstractions, Version=6.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'
 ---> System.TypeInitializationException: The type initializer for 'OpenTelemetry.AutoInstrumentation.Loader.Loader' threw an exception.
 ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.Extensions.DependencyInjection.Abstractions, Version=3.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.

File name: 'Microsoft.Extensions.DependencyInjection.Abstractions, Version=3.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'
   at OpenTelemetry.Sdk.CreateTracerProviderBuilder()
   at OpenTelemetry.AutoInstrumentation.Instrumentation.Initialize() in /project/src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs:line 67
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Span`1& arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters)
   at OpenTelemetry.AutoInstrumentation.Loader.Loader.TryLoadManagedAssembly() in /project/src/OpenTelemetry.AutoInstrumentation.Loader/Loader.cs:line 72
[...]

PR #3343, opened 2024-03-29
Resolve file loading issues originating from Runtime store libraries

rajkumar-rangaraj:
## Why

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3168, https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3075 
Follow up https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3187

**Steps to recreate an issue:**: Adding the below package reference to `Examples.AspNetCoreMvc` project for net6.0 target it will reproduce crash issue.
 
```
<PackageReference Include="Microsoft.Extensions.Configuration" Version="6.0.1" />
```

## What

The .NET loader takes control when the assembly load fails. If an assembly is part of the runtime store, we load it in a different context. This approach prevents crashes and allows ILogger data collection.

Rule engine needs to be turned off to test this approach. Diagnostics rule has a backoff logic.

## Tests

Existing CI

## Checklist


rajkumar-rangaraj:
@open-telemetry/dotnet-instrumentation-approvers Please review and test the behavior with Examples app in our repo. This approach resolves the crash issues that occur due to runtime store conflicts. Do set `OTEL_DOTNET_AUTO_RULE_ENGINE_ENABLED` to `false` to test it.

Kielek:
@rajkumar-rangaraj, could you please add CHANGELOG entry for this fix?

I suppose it could convince some users to upgrade.

src/OpenTelemetry.AutoInstrumentation.Loader/Loader.Net.cs
```diff
@@ -21,6 +23,29 @@ private static string ResolveManagedProfilerDirectory()
...
+            if (storeDirectory == null || !Directory.Exists(storeDirectory))
+            {
+                return null;
+            }
+
+            var architecture = Environment.Is64BitProcess ? "x64" : "x86";
+            var targetFramework = $"net{Environment.Version.Major}.{Environment.Version.Minor}";
+            var finalPath = Path.Combine(storeDirectory, architecture, targetFramework);
+
+            var storeFiles = Directory.GetFiles(finalPath, "Microsoft.Extensions*.dll", SearchOption.AllDirectories);
```
pjanotti, line 40: Shouldn't the same be done for other assemblies on the shared store?

pjanotti (approved):
Very nice @rajkumar-rangaraj!

I will give a spin later today to validate some cases, but, at first LGTM. Per @Kielek comment a note on changelog will be good.

src/OpenTelemetry.AutoInstrumentation.Loader/Loader.Net.cs
rajkumar-rangaraj: Here is the list of assemblies in the shared store. It's riskier to handle DiagnosticSource because we always want it to load in the application context. I just got this filter `Microsoft.Extensions*.dll` to keep it future-proof and handle other libraries that might be added here on a need basis.

image

rajkumar-rangaraj:
> @rajkumar-rangaraj, could you please add CHANGELOG entry for this fix?
> 
> I suppose it could convince some users to upgrade.

@Kielek Added changelog, please take a look.

src/OpenTelemetry.AutoInstrumentation.Loader/Loader.Net.cs
```diff
@@ -21,6 +23,29 @@ private static string ResolveManagedProfilerDirectory()
...
+    {
+        try
+        {
+            var storeDirectory = Environment.GetEnvironmentVariable("DOTNET_SHARED_STORE");
+            if (storeDirectory == null || !Directory.Exists(storeDirectory))
+            {
+                return null;
+            }
+
+            var architecture = Environment.Is64BitProcess ? "x64" : "x86";
```
nrcventura, line 36: Do we need to handle arm64 here too?

rajkumar-rangaraj: arm64 experimental support, and I'm not sure if we need it. This is a good point and needs to be explored further as a follow-up.

rajkumar-rangaraj: Changes for arm64 was very small and I could cover in this PR.
Diagnostic rule has a similar code and arm64 was not updated in it. This should be followed up in a different PR as it is not related to this issue.

rajkumar-rangaraj: Fixed f2410d6d882fa05040e0319ba66e4d3e801b5ed5

nrcventura (approved):

PR #3375, opened 2024-04-17
[kafka] update messaging operation attribute name

lachmatt:
## Why

Update attribute name to match current semantic conventions.

Fixes #

## What

Attribute rename.

## Tests

Existing tests.

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Instrumentations/Kafka/KafkaInstrumentation.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Instrumentations/Kafka/KafkaInstrumentation.cs
@@ -124,7 +124,7 @@ internal static class KafkaInstrumentation
         object? key,
         INamedClient? client)
     {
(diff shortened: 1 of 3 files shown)


pyohannes:
The change itself looks good, however, we're currently actively working on stabilizing semantic conventions for messaging, and other breaking changes might come up. That's why we have this [disclaimer](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/messaging-spans.md) on the messaging semantic convention document:

> **Warning** Existing messaging instrumentations that are using [v1.24.0 of this document](https://github.com/open-telemetry/semantic-conventions/blob/v1.24.0/docs/messaging/messaging-spans.md) (or prior) SHOULD NOT change the version of the messaging conventions that they emit until a transition plan to the (future) stable semantic conventions has been published. Conventions include, but are not limited to, attributes, metric and span names, and unit of measure.

So to avoid churn, we recommend to avoid releasing any changes to users until semantic conventions for messaging are stable. The result would be just one (bigger) breaking change instead of multiple (smaller) ones.

PR #3444, opened 2024-06-04
Capture HTTP Headers/gRPC Metadata as span attributes

Kielek:
## Why

Towards https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1696
We have agreement that it is to early to implement this feature on contrib side, as part of the functionalities will be moved directly to .NET code.
For sure, env. variables will not be handled on contrib side.

## What

Possibility to capture HTTP Headers/gRPC Metadata as span attributes.
Configurable by the pair of environmental variables. Separate for each instrumentation package.

## Tests

Added new tests.

## Checklist

## ~~TODO~~

*  ~~Fix MacOs builds. I do not have such machine locally. Scheduled to work with @pellared for tomorrow.~~
* ~~Fix ASP.NET tests/implemnetation.~~

~~Please review other parts without waiting for the green builds. It looks like specific for mac only.~~


theletterf (approved):

test/OpenTelemetry.AutoInstrumentation.Tests/Diagnostics/SdkSelfDiagnosticsEventListenerTests.cs
```diff
@@ -1,7 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-// Source originated from https://github.com/open-telemetry/opentelemetry-dotnet/blob/23609730ddd73c860553de847e67c9b2226cff94/test/OpenTelemetry.Tests/Internal/SelfDiagnosticsEventListenerTest.cs
+// Key originated from https://github.com/open-telemetry/opentelemetry-dotnet/blob/23609730ddd73c860553de847e67c9b2226cff94/test/OpenTelemetry.Tests/Internal/SelfDiagnosticsEventListenerTest.cs
```
Kielek, line 4:
```suggestion
// Source originated from https://github.com/open-telemetry/opentelemetry-dotnet/blob/23609730ddd73c860553de847e67c9b2226cff94/test/OpenTelemetry.Tests/Internal/SelfDiagnosticsEventListenerTest.cs
```

.github/workflows/ci.yml
```diff
@@ -57,6 +57,9 @@ jobs:
         with:
           fetch-depth: 0 # fetching all, needed to correctly calculate version
 
+      - name: Set GITHUB_RUNNER_SYSTEM environemental variable
```
pjanotti, line 60:
```suggestion
      - name: Set the GITHUB_RUNNER_SYSTEM environment variable
```

.github/workflows/ci.yml
```diff
@@ -57,6 +57,9 @@ jobs:
...
+        run: echo "GITHUB_RUNNER_SYSTEM=${{matrix.machine}}" >> $GITHUB_ENV
```
pjanotti, line 61:
```suggestion
        run: echo "GITHUB_RUNNER_SYSTEM=${{  matrix.machine  }}" >> $GITHUB_ENV
```

pjanotti (commented):
Partial... will finish later.

pjanotti (approved):
LGTM

src/OpenTelemetry.AutoInstrumentation/Configurations/ConfigurationKeys.cs
```diff
@@ -108,27 +108,69 @@ public static class Traces
         /// </summary>
         public static class InstrumentationOptions
         {
+#if NETFRAMEWORK
             /// <summary>
-            /// Configuration key for GraphQL instrumentation to enable passing query as a document attribute.
+            /// Configuration key for ASP.NET instrumentation to enable capturing HTTP request headers as span tags.
             /// </summary>
-            public const string GraphQLSetDocument = "OTEL_DOTNET_AUTO_GRAPHQL_SET_DOCUMENT";
+            public const string AspNetInstrumentationCaptureRequestHeaders = "OTEL_DOTNET_AUTO_TRACES_ASPNET_INSTRUMENTATION_CAPTURE_REQUEST_HEADERS";
```
RassK, line 115: Do we really need to separate aspnet and aspnet core?

Kielek: It was the main question I have raised on SIG meeting Other parts of the PR are more or less boring.

For this version - we have separate settings for SqlDatabaseStatements. And it was the reason I started with this approach.

Unification, less things to maintain/configure.

src/OpenTelemetry.AutoInstrumentation/Loading/Initializers/AspNetCoreInitializer.cs
```diff
@@ -22,12 +28,33 @@ public override void Initialize(ILifespanManager lifespanManager)
         var httpInListenerType = Type.GetType("OpenTelemetry.Instrumentation.AspNetCore.Implementation.HttpInListener, OpenTelemetry.Instrumentation.AspNetCore")!;
 
         var options = new OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreTraceInstrumentationOptions();
+
+        if (_tracerSettings.InstrumentationOptions.AspNetCoreInstrumentationCaptureRequestHeaders.Count != 0)
+        {
+            options.EnrichWithHttpRequest = EnrichWithHttpRequest;
```
RassK, line 34: for plugins we need to create dynamic call that wraps both, the current value and our function call.

RassK: maybe it's that simple > https://stackoverflow.com/questions/40401610/possible-to-combine-two-actions-in-c

src/OpenTelemetry.AutoInstrumentation/HeadersCapture/HeaderEnrichment.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Collections.Specialized;
+using System.Diagnostics;
+using System.Net.Http.Headers;
+#if NET6_0_OR_GREATER
+using Microsoft.AspNetCore.Http;
+#endif
+
+namespace OpenTelemetry.AutoInstrumentation.HeadersCapture;
+
+internal static class HeaderEnrichment
```
RassK, line 13: Might use different name, to me it sounds more like it's enriching headers instead of activities with headers.

Maybe something like: "ActivityExtensions.EnrichWithHeaders" ?

src/OpenTelemetry.AutoInstrumentation/Configurations/ConfigurationKeys.cs
RassK: For sure this approach is safer as if only one integration is requested to extract headers. If it's the same header name for eg asp.net and grpc and ones value is important not to be captured.
But from the other hand adds a lot of configuration overhead.

src/OpenTelemetry.AutoInstrumentation/HeadersCapture/HeaderNormalizer.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace OpenTelemetry.AutoInstrumentation.HeadersCapture;
+
+internal static class HeaderNormalizer
+{
+    public static string NormalizeHeader(string httpHeaderName)
```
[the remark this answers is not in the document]
Kielek: fbb0488cad65d40354134d9c672b0ba3f3b82cc9

src/OpenTelemetry.AutoInstrumentation/HeadersCapture/HeaderEnrichment.cs
Kielek: 4037c6ded4e7fcb8c375bbf601b5eaa006bda361

src/OpenTelemetry.AutoInstrumentation/Loading/Initializers/AspNetCoreInitializer.cs
Kielek: > For plugins we need to create dynamic call that wraps both, the current value and our function call. Also need to make sure that plugins assign the value first.

Current implementation - we are setting what we need, then sending it to the plugin. I think that it should stay as is. Plugin potentially would like to remove the feature. I agree that we should document this change (in changelog and in plugin.md).

RassK: On the other hand, plugin can handle this also when documented. We definitely use it already, so it's important for us also.

src/OpenTelemetry.AutoInstrumentation/Configurations/ConfigurationKeys.cs
Kielek: Please check a15789f722179866b935d6d1dcb6acb225f34d10

src/OpenTelemetry.AutoInstrumentation/Loading/Initializers/AspNetCoreInitializer.cs
Kielek: Please check a15789f722179866b935d6d1dcb6acb225f34d10

RassK (approved):

src/OpenTelemetry.AutoInstrumentation/Configurations/ConfigurationKeys.cs
Kielek: @rajkumar-rangaraj, @zacharycmontoya, @nrcventura, could you please check (whole PR, but env. vars, are most important IMO)?

lachmatt (approved):

src/OpenTelemetry.AutoInstrumentation/Configurations/ConfigurationKeys.cs
nrcventura: It looks like we have a precedent with some of our other environment variables for separating aspnet and aspnetcore, so I think that keeping the env vars separate here also make sense. I think separating the aspnet* and *client variables make sense because I can see different use cases for those. The client libraries are also different enough that I think it makes sense to keep their settings separated.

nrcventura (approved):

rajkumar-rangaraj (approved):

PR #3456, opened 2024-06-17
[kafka] fix return type name for `ProduceAsync` integration

lachmatt:
Shared by @zacharycmontoya 
- It seems that return type name is [not being checked](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/927d65b0c78be46825dbc88eea43aaac7b54e2db/src/OpenTelemetry.AutoInstrumentation.Native/rejit_preprocessor.cpp#L86-L97) when deciding whether to ReJIT.
- This is not a problem (it is not valid C# to have the same arguments but different return type)

Instrumentation works as expected even with previous return type name.


Kielek (approved):

lachmatt:
CI fails due to:
> This is a scheduled macOS-11 brownout. The macOS-11 environment is deprecated and will be removed on June 28th, 2024.

RassK (approved):

rajkumar-rangaraj (approved):

nrcventura (approved):

pjanotti (approved):

PR #3489, opened 2024-07-04
[docs] Add warning about the breaking change in .NET 8 regarding the DOTNET_EnableDiagnostics variable

zacharycmontoya:
## Why

If a user has set the environment variable `DOTNET_EnableDiagnostics=0`, they will see instrumentation when running on .NET 7 or earlier but no instrumentation starting with .NET 8. This calls out the breaking change in our documentation.

## What

Updates documentation in our main README

## Tests

N/A

## Checklist

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -157,6 +157,9 @@ When running your application, make sure to:
 > [!NOTE]
 > Some settings can be omitted on .NET. For more information, see [config.md](config.md#net-clr-profiler).
 
+> [!IMPORTANT]
+> Starting in .NET 8, the environment variable `DOTNET_EnableDiagnostics=0` disables all diagnostics, including the CLR Profiler facility which is needed to launch the instrumentation, if not using .NET Startup hooks. Ensure that `DOTNET_EnableDiagnostics=1`, or if you'd like to limit diagnostics only to the CLR Profiler, you may set both `DOTNET_EnableDiagnostics=1` and `DOTNET_EnableDiagnostics_Profiler=1` while setting other diagnostics features to 0. See this [issue](https://github.com/dotnet/runtime/issues/96227#issuecomment-1865326080) for more guidance.
+
 ### Shell scripts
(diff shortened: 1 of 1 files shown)


pjanotti (approved):

docs/README.md
```diff
@@ -157,6 +157,9 @@ When running your application, make sure to:
 > [!NOTE]
 > Some settings can be omitted on .NET. For more information, see [config.md](config.md#net-clr-profiler).
 
+> [!IMPORTANT]
+> Starting in .NET 8, the environment variable `DOTNET_EnableDiagnostics=0` disables all diagnostics, including the CLR Profiler facility which is needed to launch the instrumentation, if not using .NET Startup hooks. Ensure that `DOTNET_EnableDiagnostics=1`, or if you'd like to limit diagnostics only to the CLR Profiler, you may set both `DOTNET_EnableDiagnostics=1` and `DOTNET_EnableDiagnostics_Profiler=1` while setting other diagnostics features to 0. See this [issue](https://github.com/dotnet/runtime/issues/96227#issuecomment-1865326080) for more guidance.
```
Kielek, line 161: Should we change PowerShell scripts to verify if these variables are set correctly?
Or maybe set them by default to required values?
Other option, extend our RuleEngine.

Kielek (approved):

docs/README.md
nrcventura: Perhaps a change to the RuleEngine would be good in a separate PR. I do not know how often the scripts we include are used, so the RuleEngine could probably catch the cases where the scripts are not used.

nrcventura (approved):

rajkumar-rangaraj (approved):

docs/README.md
zacharycmontoya: Sounds good, I like the idea of updating the RuleEngine, so I can do that separately

Issue #3513, opened 2024-07-16
`OTEL_{SIGNAL}_EXPORTER` to handle comma-separated list


pellared:
## What

From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#exporter-selection:

> The implementation MAY accept a comma-separated list to enable setting multiple exporters.

## Why

Needed for https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3289 so that a user e.g. select an OTLP and console exporter at the same time using only `OTEL_{SIGNAL}_EXPORTER`.

Issue #3289, opened 2024-02-23
Use `OTEL_{SIGNAL}_EXPORTER` to handle console exporter


Kielek:
# Feature Request

**Are you requesting automatic instrumentation for a framework or library? Please describe.**

- 1.4.0 and current main

**Is your feature request related to a problem? Please describe.**

Spec changed 3 months ago: https://github.com/open-telemetry/opentelemetry-specification/commit/d72b0da79645432c62593f074c64ae44778680aa

**Describe the solution you'd like**

Support `console` for `OTEL_TRACES_EXPORTER`, `OTEL_METRICS_EXPORTER`, and `OTEL_LOGS_EXPORTER`.

**Describe alternatives you've considered**

N/A

**Additional context**

What to do with `OTEL_DOTNET_AUTO_TRACES_CONSOLE_EXPORTER_ENABLED`, `OTEL_DOTNET_AUTO_METRICS_CONSOLE_EXPORTER_ENABLED`, and `OTEL_DOTNET_AUTO_LOGS_CONSOLE_EXPORTER_ENABLED`.

pellared:
> What to do with OTEL_DOTNET_AUTO_TRACES_CONSOLE_EXPORTER_ENABLED, OTEL_DOTNET_AUTO_METRICS_CONSOLE_EXPORTER_ENABLED, and OTEL_DOTNET_AUTO_LOGS_CONSOLE_EXPORTER_ENABLED.

I suggest that before dropping support we should first make a minor release where we  we mark it as deprecated in docs and emitting a warning (e.g. to console) when anybody uses any of these env vars that they are deprecated and are going to be removed in next minor release.

PR #3523, opened 2024-07-19
Exporter selection env vars to handle comma separated list

ysolomchenko:
## Why

Needed for https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3289 so that a user e.g. select an OTLP and console exporter at the same time using only OTEL_{SIGNAL}_EXPORTER

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3513

## What

The implementation accept a comma-separated list to enable setting multiple exporters.

## Tests

Updated and added new test cases, including smoke tests and additional checks in SettingsTests.cs.

## Checklist


pjanotti (commented):
Thanks @ysolomchenko!

The comments made for logs apply to the other signals also.

docs/config.md
```diff
@@ -262,9 +262,9 @@ Exporters output the telemetry.
...
-| `OTEL_TRACES_EXPORTER`  | Traces exporter to be used. The value can be one of the following: `zipkin`, `otlp`, `none`.      | `otlp`        | [Stable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
-| `OTEL_METRICS_EXPORTER` | Metrics exporter to be used. The value can be one of the following: `otlp`, `prometheus`, `none`. | `otlp`        | [Stable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
-| `OTEL_LOGS_EXPORTER`    | Logs exporter to be used. The value can be one of the following: `otlp`, `none`.                  | `otlp`        | [Stable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
+| `OTEL_TRACES_EXPORTER`  | Traces exporter to be used. The value can be one of the following: `zipkin`, `otlp`, `none`, or a comma-separated list of values.       | `otlp`        | [Stable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
```
pellared, line 265: I propose simply something like

```suggestion
| `OTEL_TRACES_EXPORTER`  | Comma-separated list of propagators. Supported options: `otlp`, `zipkin`, `none. See [the OpenTelemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.35.0/specification/configuration/sdk-environment-variables.md#exporter-selection) for more details. | `otlp`        | [Stable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
```

src/OpenTelemetry.AutoInstrumentation/Configurations/LogSettings.cs
```diff
@@ -18,9 +18,9 @@ internal class LogSettings : Settings
     public bool LogsEnabled { get; private set; }
 
     /// <summary>
-    /// Gets the logs exporter.
+    /// Gets the list of enabled logs exporters.
```
pellared, line 21:
```suggestion
    /// List of logs exporters to be used.
```

based on https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/fc1af32094d792221d63ad5b9f31f453d73e671f/src/OpenTelemetry.AutoInstrumentation/Configurations/SdkSettings.cs#L17

test/IntegrationTests/SmokeTests.cs
```diff
@@ -218,6 +218,47 @@ public void ZipkinExporter()
         }
     }
 
+    [Fact]
+    [Trait("Category", "EndToEnd")]
+    public void ZipkinAndOtlpTracesExporter()
```
pellared, line 223: @open-telemetry/dotnet-instrumentation-approvers, what do you think `ZipkinTracesExporter` and `OtlpTracesExporter` to reduce the execution of integration tests? Basically this test combines both tests.

test/IntegrationTests/SmokeTests.cs
```diff
@@ -267,6 +308,58 @@ public void PrometheusExporter()
     }
 #endif
 
+#if NETFRAMEWORK
+    [Fact]
+    [Trait("Category", "EndToEnd")]
+    public void PrometheusAndOtlpMetricsExporter()
```
pellared, line 314: @open-telemetry/dotnet-instrumentation-approvers, what do you think `PrometheusMetricsExporter` and `OtlpMetricsExporter` to reduce the execution of integration tests? Basically this test combines both tests.

pellared (commented):
My comments applies to all signals

src/OpenTelemetry.AutoInstrumentation/Configurations/LogSettings.cs
Kielek: I think that the `Gets` predix is enforced by some code analysis. Worth to double check before commiting.

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
 
 ### Added
 
+- Added support for OTEL_TRACES_EXPORTER, OTEL_METRICS_EXPORTER, OTEL_LOGS_EXPORTER
+ to handle comma-separated list.
```
Kielek, line 13: nit:

```suggestion
- Added support for OTEL_TRACES_EXPORTER, OTEL_METRICS_EXPORTER, OTEL_LOGS_EXPORTER
  to handle comma-separated list.
```

docs/config.md
```diff
@@ -262,9 +262,15 @@ Exporters output the telemetry.
...
 |-------------------------|---------------------------------------------------------------------------------------------------|---------------|-----------------------------------------------------------------------------------------------------------------------------|
-| `OTEL_TRACES_EXPORTER`  | Traces exporter to be used. The value can be one of the following: `zipkin`, `otlp`, `none`.      | `otlp`        | [Stable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
-| `OTEL_METRICS_EXPORTER` | Metrics exporter to be used. The value can be one of the following: `otlp`, `prometheus`, `none`. | `otlp`        | [Stable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
-| `OTEL_LOGS_EXPORTER`    | Logs exporter to be used. The value can be one of the following: `otlp`, `none`.                  | `otlp`        | [Stable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
+<<<<<<< HEAD
```
Kielek, line 265: You have some issue with solving conflicts.

src/OpenTelemetry.AutoInstrumentation/Configurations/LogSettings.cs
```diff
@@ -18,9 +18,9 @@ internal class LogSettings : Settings
...
     /// </summary>
-    public LogExporter LogExporter { get; private set; }
+    public IReadOnlyList<LogExporter> LogExporters { get; private set; } = new List<LogExporter>();
```
pjanotti, line 23: Given that the property is set on `OnLoad` do we really need to create an instance here? Similar Q for the other signals.

src/OpenTelemetry.AutoInstrumentation/Configurations/TracerSettings.cs
```diff
@@ -89,32 +89,56 @@ protected override void OnLoad(Configuration configuration)
...
+            switch (exporterName)
+            {
+                case Constants.ConfigurationValues.Exporters.Otlp:
+                    exporters.Add(TracesExporter.Otlp);
+                    break;
+                case Constants.ConfigurationValues.Exporters.Zipkin:
+                    exporters.Add(TracesExporter.Zipkin);
+                    break;
+                case Constants.ConfigurationValues.None:
+                    if (!hasExporter)
```
pjanotti, line 121: Why we need this check? The `Add` avoids duplicates, the current code adds `None` if there is more than one name on the list, but, not if there is only `none` on the env var. Perhaps, we could actually just `break` here and never add the `none`.

src/OpenTelemetry.AutoInstrumentation/Configurations/TracerSettings.cs
```diff
@@ -89,32 +89,56 @@ protected override void OnLoad(Configuration configuration)
...
+                default:
+                    if (configuration.FailFast)
+                    {
+                        var message = $"Traces exporter '{exporterName}' is not supported.";
+                        Logger.Error(message);
+                        throw new NotSupportedException(message);
+                    }
+
+                    Logger.Error($"Traces exporter '{exporterName}' is not supported. Defaulting to '{Constants.ConfigurationValues.Exporters.Otlp}'.");
+                    exporters.Add(TracesExporter.Otlp);
```
pjanotti, line 136: I would rather ignore the entry in this case instead of adding OTLP.

test/IntegrationTests/SmokeTests.cs
pjanotti: Hum I like an integration test that we create multiple exporters and check their outputs.

src/OpenTelemetry.AutoInstrumentation/Configurations/LogSettings.cs
ysolomchenko: At the moment, this needs to remain as is to avoid nullable issues. However, we can refactor this in a separate PR in the future.

src/OpenTelemetry.AutoInstrumentation/Configurations/TracerSettings.cs
ysolomchenko: In the updated version, the `None` exporter is always initially added to the set. When an `Otlp` or `Prometheus` exporter is encountered, the set is cleared of `None` if it was present. This means that if there are invalid exporters specified, `None` will be included. However, if valid exporters like `Otlp` or `Prometheus` are specified, they will replace `None` in the final result.

src/OpenTelemetry.AutoInstrumentation/Configurations/EnvironmentConfigurationMetricHelper.cs
```diff
@@ -53,13 +53,18 @@ private static MeterProviderBuilder SetExporter(this MeterProviderBuilder builde
...
-            MetricsExporter.Prometheus => Wrappers.AddPrometheusHttpListener(builder, pluginManager),
-            MetricsExporter.Otlp => Wrappers.AddOtlpExporter(builder, settings, pluginManager),
-            MetricsExporter.None => builder,
-            _ => throw new ArgumentOutOfRangeException($"Metrics exporter '{settings.MetricExporter}' is incorrect")
-        };
+            builder = metricExporter switch
+            {
+                MetricsExporter.Prometheus => Wrappers.AddPrometheusHttpListener(builder, pluginManager),
+                MetricsExporter.Otlp => Wrappers.AddOtlpExporter(builder, settings, pluginManager),
+                MetricsExporter.None => builder,
```
pellared, line 62: Why do we even return `MetricsExporter.None`?

src/OpenTelemetry.AutoInstrumentation/Configurations/MetricSettings.cs
```diff
@@ -62,32 +62,54 @@ protected override void OnLoad(Configuration configuration)
...
+
+                    Logger.Error($"Metric exporter '{exporterName}' is not supported.");
+                    break;
+            }
+        }
+
+        if (exporters.Count == 0)
+        {
+            exporters.Add(MetricsExporter.None);
+        }
```
pellared, line 111: Related to https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3523/files#r1687773496

src/OpenTelemetry.AutoInstrumentation/Configurations/MetricSettings.cs
```diff
@@ -62,32 +62,54 @@ protected override void OnLoad(Configuration configuration)
...
+            return new List<MetricsExporter> { MetricsExporter.Otlp }.AsReadOnly();
         }
+
+        var exporters = new HashSet<MetricsExporter>();
+        var exporterNames = metricsExporterEnvVar!.Split(',')
+                                                  .Select(e => e.Trim())
+                                                  .Where(e => !string.IsNullOrEmpty(e))
+                                                  .ToList();
+
+        exporters.Add(MetricsExporter.None);
```
pellared, line 80: Related to https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3523/files#r1687773496

src/OpenTelemetry.AutoInstrumentation/Configurations/EnvironmentConfigurationMetricHelper.cs
Kielek: We can probably return empty collection when `none` is used.

src/OpenTelemetry.AutoInstrumentation/Configurations/TracerSettings.cs
```diff
@@ -89,32 +89,47 @@ protected override void OnLoad(Configuration configuration)
...
 
-    private static TracesExporter ParseTracesExporter(Configuration configuration)
+    private static IReadOnlyList<TracesExporter> ParseTracesExporter(Configuration configuration)
     {
-        var tracesExporterEnvVar = configuration.GetString(ConfigurationKeys.Traces.Exporter)
-            ?? Constants.ConfigurationValues.Exporters.Otlp;
+        var tracesExporterEnvVar = configuration.GetString(ConfigurationKeys.Traces.Exporter);
 
-        switch (tracesExporterEnvVar)
+        if (string.IsNullOrWhiteSpace(tracesExporterEnvVar))
```
pellared, line 96: My preference would be to use

```suggestion
        if (string.IsNullOrEmpty(tracesExporterEnvVar))
```

it may be good to catch if a user sets some strange whitespaces

src/OpenTelemetry.AutoInstrumentation/Configurations/TracerSettings.cs
```diff
@@ -89,32 +89,47 @@ protected override void OnLoad(Configuration configuration)
...
-
-                return TracesExporter.Otlp;
+            return new List<TracesExporter> { TracesExporter.Otlp }.AsReadOnly();
         }
+
+        var exporters = new HashSet<TracesExporter>();
+
+        var exporterNames = tracesExporterEnvVar!.Split(',')
+                                                 .Select(e => e.Trim())
+                                                 .Where(e => !string.IsNullOrEmpty(e))
```
pellared, line 105: This is different from how we are parsing propagators. We are not trimming the whitespaces and also we are not getting rid of empty entries.

See `SdkSettings.cs`.

The OpenTelemetry specification does not say anything about trimming whitespaces: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md

test/IntegrationTests/SmokeTests.cs
pellared: I mean. Keep this test and remove the "old ones" :D

src/OpenTelemetry.AutoInstrumentation/Configurations/TracerSettings.cs
pellared: Reference: https://blog.ploeh.dk/2014/11/18/the-isnullorwhitespace-trap/

test/IntegrationTests/SmokeTests.cs
ysolomchenko: Let's keep as it is, i can remove in separate PR

pellared (approved):

ysolomchenko:
I've opened an issue regarding the handling of comma-separated lists.
https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3534

ysolomchenko:
I think this is a random error. Could you please rerun it?

PR #3538, opened 2024-07-25
Ensure contentFiles are unobtrusive in code editors

Mpdreamz:
## Why
 As seen here by default today we inject two contentFiles into every project in the solution navigator.

image

## What

Include a props in the NuGet package that updates their inclusion to have the `<Visible>false</Visible>` flag set.

The props file is only necessary until https://github.com/NuGet/Home/issues/4856 gets resolved to do this more natively.

This ensures the instrumentation scripts are not visible in users's projects when they reference `OpenTelemetry.AutoInstrumentation`.

This makes the manual installation route slightly less unobtrusive as we don't clutter anyones editor with instrumentation.sh and instrumentation.cmd.

The `contentFiles` will still be published to output and publish directories.

See: https://learn.microsoft.com/en-us/nuget/concepts/msbuild-props-and-targets for more background on how $PROJECT.props file affect NuGet installations.

## Tests

Validated manually that the NuGet package still includes all relevant files and that the build file gets put in the correct place using https://nuget.info 
image

Installation of the local package validates the instrumentation scripts still get copied to output and publish folder.

Side note: nuget.info ❤️ 

## Checklist


nrcventura (approved):
I think this provides a cleaner experience.

Kielek (approved):
LGTM.

Thanks for your contribution! Looking for more improvements :)

PR #3553, opened 2024-07-31
Merge instrument.sh versions into one

Mpdreamz:
- expand instrument.sh to be able to run from RFID paths (nuget deployment)
- remove nuget instrument.sh temporarily
- rename instrument.sh from root to nuget contentFile
- update build scripts to reference instrument.sh from nuget folder


pellared (approved):
Should we also merge `instrument.sh` and `script-templates/otel-dotnet-auto-install.sh.template` as a next step?

RassK:
> update build scripts to reference instrument.sh from nuget folder

Shouldn't it be another way around? `instrument.sh` is the main script, nuget should reference the script instead.

pjanotti:
Hi @Mpdreamz thanks for the PR. Per SIG meeting: it would be good to follow @RassK suggestion:
> Shouldn't it be another way around? instrument.sh is the main script, NuGet should reference the script instead.

Typically we prefer follow-up PRs, but, in this case it will make the history much better, as it is the diff looks like a big addition and a corresponding deletion. Following the suggestion it will be clear what was changed on the "big" script.

Mpdreamz:
Aye I did a proper `git mv` in this commit: [`bd5237a` (#3553)](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3553/commits/bd5237aff8bff3941e06c795ee11e7a41093e041)

However this seems to get lost in the proposed merge commit 😿. 

The main reason for keeping the nuget script path since it required no changes to the `.nuspec` file. 

Will follow up with a PR that does the reverse, this might not happen this week though.

Mpdreamz:
> Should we also merge instrument.sh and script-templates/otel-dotnet-auto-install.sh.template as a next step?

@pellared this is an interesting proposal I have not considered. I think we might get into some argument conflicts since currently we just take all arguments as the command to execute. 

I could see it work if we did something like

`opentelemetry.sh instrument <args>`
`opentelemtery.sh install`

or somewhat more terse:

`otel.sh attach <args>`
`otel.sh install`

wdyt?

PR #3558, opened 2024-08-06
Add console sink to the diagnostic logging

### LINKED ISSUE #3552: Console Logger for internal diagnostic logs
### What

I'm thinking to add switch that passes diagnostic logs also to console (currently there is only a file output).
This should simplify diagnostic logging for containerized apps, like k8s pods.

Any opinions on this?

RassK:
## Why

Fixes #3552

## What

Adds console sink to managed and native diagnostic logging.

- Defaults to `file` sink if unspecified or invalid configuration (not new)
- When `file` sink doesn't work, fallbacks to `none` (NoOp) sink (not new)
- Uses `OTEL_DOTNET_AUTO_LOGGER` to set a sink (new)
- Possible options are `file`, `console`, `none` (new)

## Tests

Unit tests added for setting the sink and testing loggers.

## Checklist


rajkumar-rangaraj:
Sorry, I couldn't review the issue created earlier. Are you planning to enable this by default? There are console scrapers that capture information from the console and convert it into logs. In such environments, this change could result in additional logs. I recommend keeping this feature off by default, as logging to the console can also increase startup time in some scenarios.

RassK:
> Sorry, I couldn't review the issue created earlier. Are you planning to enable this by default? There are console scrapers that capture information from the console and convert it into logs. In such environments, this change could result in additional logs. I recommend keeping this feature off by default, as logging to the console can also increase startup time in some scenarios.
> 
> Adding a detailed summary to this PR on the behavior would help, as we plan to customize it in one of the environments.

More like described [here](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3552#issuecomment-2261411446). Half way there currently.

src/OpenTelemetry.AutoInstrumentation.Native/logger_impl.h
```diff
@@ -97,11 +104,11 @@ template <typename TLoggerPolicy>
...
-        // By writing into the stderr was changing the behavior in a CI scenario.
-        // There's not a good way to report errors when trying to create the log file.
-        // But we never should be changing the normal behavior of an app.
-        // std::cerr << "LoggerImpl Handler: " << msg << std::endl;
-    });
+            // By writing into the stderr was changing the behavior in a CI scenario.
+            // There's not a good way to report errors when trying to create the log file.
+            // But we never should be changing the normal behavior of an app.
+            // std::cerr << "LoggerImpl Handler: " << msg << std::endl;
+        });
```
pjanotti, line 111: nit: is this whitespace intentional?

src/OpenTelemetry.AutoInstrumentation.Native/logger_impl.h
```diff
@@ -131,28 +138,41 @@ LoggerImpl<TLoggerPolicy>::LoggerImpl()
...
+            m_fileout = spdlog::rotating_logger_mt(logger_name, GetLogPath(file_name_suffix), file_size, 10);
+        }
+        catch (...)
+        {
+            // By writing into the stderr was changing the behavior in a CI scenario.
+            // There's not a good way to report errors when trying to create the log file.
+            // But we never should be changing the normal behavior of an app.
+            // std::cerr << "LoggerImpl Handler: Error creating native log file." << std::endl;
+            m_fileout = spdlog::null_logger_mt(logger_name);
+        }
```
pjanotti, line 171: Now that this is inside a conditional block it seems that it deserves a separate function, or at least to be last branch of the if statement.

src/OpenTelemetry.AutoInstrumentation.Native/logger_impl.h
```diff
@@ -131,28 +138,41 @@ LoggerImpl<TLoggerPolicy>::LoggerImpl()
...
-    catch (...)
+    else if (configured_log_sink == log_sink_console)
     {
-        // By writing into the stderr was changing the behavior in a CI scenario.
-        // There's not a good way to report errors when trying to create the log file.
-        // But we never should be changing the normal behavior of an app.
-        // std::cerr << "LoggerImpl Handler: Error creating native log file." << std::endl;
-        m_fileout = spdlog::null_logger_mt(logger_name);
+        m_fileout = spdlog::stdout_logger_mt(logger_name);
     }
```
pjanotti, line 176: it should log a warning if didn't matching any and fallback to the default (file).

src/OpenTelemetry.AutoInstrumentation/Logging/OtelLogging.cs
```diff
@@ -89,37 +123,53 @@ internal static long GetConfiguredFileSizeLimitBytes()
...
+    {
+        if (_configuredLogSink == LogSink.NoOp)
         {
-            var logDirectory = GetLogDirectory();
-            if (logDirectory != null)
+            return new NoopSink();
+        }
+        else if (_configuredLogSink == LogSink.File)
+        {
+            try
```
pjanotti, line 144: Similar to the native one: ideally move this code here to a method or at least make it the last branch in the check.

src/OpenTelemetry.AutoInstrumentation/Logging/OtelLogging.cs
```diff
@@ -89,37 +123,53 @@ internal static long GetConfiguredFileSizeLimitBytes()
...
         {
-            // unable to configure logging to a file
+            return new ConsoleSink(suffix);
         }
 
-        sink ??= new NoopSink();
-
-        return new InternalLogger(sink, ConfiguredLogLevel.Value);
+        // Default to NoopSink
+        return new NoopSink();
```
pjanotti, line 172: Is this consistent with the native default?

Mpdreamz: Wondering this too: reading https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3558/files#diff-6576a72e1be921af76d4e4e9525cb42dff458c51b1a1439807de36371a31fa1aR296 makes me think `file` is the default fallback in the profiler.

RassK: File is the default sink, the default fallback sink is NoOp. This might cause the confusion at first.

src/OpenTelemetry.AutoInstrumentation.Native/logger_impl.h
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3558/commits/b434bd17e4b10337d1142c5b6e008f15bafa87a8

src/OpenTelemetry.AutoInstrumentation.Native/logger_impl.h
RassK: removed, seems unintentional. https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3558/commits/dc5caf6b437d523d8ed87c81ff3ca3395d2860c3

src/OpenTelemetry.AutoInstrumentation.Native/logger_impl.h
RassK: Something seems off regarding to this comment and diff 🤔
Could you check it again after some refactoring was done?

pjanotti: `configured_log_sink` is passed to this function with whatever value was on the environment variable, defaulting to `log_sink_file` if the environment variable was not specified, it seems that `m_fileout` is not set to an explicit fallback, and just reading the code around here I can't tell what happens in this case. Typically for log mechanisms we fallback to the "default" one if the configuration is incorrect. I'm opening to considering changing the fallback to console instead of file. That said reading just the code around here I can't tell what will happen if there was a typo on the value of the environment variable.

src/OpenTelemetry.AutoInstrumentation/Logging/OtelLogging.cs
pjanotti: I'm open to considering changing the default, but, the question is to ensure that we are consistent between defaults and fallbacks for the logs on both native and managed. What is the current code and the change doing in this regard?

src/OpenTelemetry.AutoInstrumentation.Native/logger_impl.h
RassK: > That said reading just the code around here I can't tell what will happen if there was a typo on the value of the environment variable.

Ah that's a good point, entirely missing. I need to refactor that.

RassK: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3558/commits/9b042ce36291835079025a8d1c613ec594e958ea

src/OpenTelemetry.AutoInstrumentation/Logging/OtelLogging.cs
RassK: I updated the "What" section about what is the current logic, defaults, fallbacks and what's added.

pjanotti (approved):

Issue #3534, opened 2024-07-24
Enhanced Handling of Comma-Separated Lists


ysolomchenko:
# Feature Request Enhanced Handling of Comma-Separated Lists

**Is your feature request related to a problem? Please describe.**
Currently, when handling comma-separated lists, the system does not trim spaces and allows duplicate entries. This can lead to confusion and inefficiencies. For instance, now the input should only be “otlp,zipkin”, which is not ideal.

**Describe the solution you'd like**

The system should:

- Trim trailing spaces around list elements.
- Remove duplicate entries.
- log a warning when duplicates are found.
- log a warning when an empty element is detected due to trailing commas or consecutive commas.

Kielek:
@ysolomchenko,
specification is pretty strict, we should support only comma-separated values. On the other hand, we should not fail if the fail fast is not enabled.

| desctiption| input example |Fail Fast disabled | fail fast enabled |
|-|-|-|-|
|Spaces around list elements.|`otlp, console`|Log warning, that the value ` console is not supported`, only OTLP protocol is used | Log warning + crash application|
|Duplicated entries.|`otlp,console,otlp`|Log warning about duplicated entries|Log warning about duplicated entries, Crash applcation|
|Empty element|`otlp, ,console`|Treat it as any non-supported exporter. Warn log only| Same as fail fast disabled + crash application |

Please let me know if it makes sense to you.

ysolomchenko:
> @ysolomchenko, specification is pretty strict, we should support only comma-separated values. On the other hand, we should not fail if the fail fast is not enabled.
> 
> desctiption	input example	Fail Fast disabled	fail fast enabled
> Spaces around list elements.	`otlp, console`	Log warning, that the value ` console is not supported`, only OTLP protocol is used	Log warning + crash application
> Duplicated entries.	`otlp,console,otlp`	Log warning about duplicated entries	Log warning about duplicated entries, Crash applcation
> Empty element	`otlp, ,console`	Treat it as any non-supported exporter. Warn log only	Same as fail fast disabled + crash application
> Please let me know if it makes sense to you.

What about the duplicated entries? Should we remove them and log a warning about duplicated entries, crash the application (fail fast), or keep the duplicates?
And what should we do with OTEL_DOTNET_AUTO_[SIGNAL]_CONSOLE_EXPORTER_ENABLED if it is enabled and present in the list?

pellared:
> What about the duplicated entries?

[...]

PR #3560, opened 2024-08-07
Enhanced handling of comma separated lists

ysolomchenko:
## Why

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3534

## What

| Description                     | Input Example         | Fail Fast Disabled                                                | Fail Fast Enabled                                                  |
|---------------------------------|-----------------------|-------------------------------------------------------------------|---------------------------------------------------------------------|
| Spaces around list elements.    | otlp, console         | Log warning, that the value `console` is not supported, only OTLP protocol is used | Log warning + crash application                                   |
| Duplicated entries.              | otlp,console,otlp    | Log warning about duplicated entries. Remove duplicated records. | Log warning about duplicated entries, Crash application            |
| Empty element                    | otlp, ,console        | Treat it as any non-supported exporter. Warn log only             | Same as fail fast disabled + crash application                      |

Also refactor Sdk settings

## Tests

SettingsTests.cs

## Checklist

### DIFF
--- a/test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/SettingsTests.cs
+++ b/test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/SettingsTests.cs
@@ -152,12 +152,27 @@ public class SettingsTests : IDisposable
     }
 
     [Theory]
+    [InlineData("non-supported", new TracesExporter[0])]
+    [InlineData("otlp,otlp", new[] { TracesExporter.Otlp })]
+    [InlineData("otlp, ,console", new[] { TracesExporter.Otlp, TracesExporter.Console })]
+    [InlineData("otlp, , console", new[] { TracesExporter.Otlp })]
+    [InlineData("otlp,,,", new[] { TracesExporter.Otlp })]
+    internal void TracesExporter_FailFastOff(string tracesExporters, TracesExporter[] expectedTracesExporters)
+    {
+        Environment.SetEnvironmentVariable(ConfigurationKeys.Traces.Exporter, tracesExporters);
+
+        var settings = Settings.FromDefaultSources<TracerSettings>(false);
+
+        settings.TracesExporters.Should().Equal(expectedTracesExporters);
+    }
+
+    [Theory]
     [InlineData("non-supported")]
     [InlineData("otlp,otlp")]
     [InlineData("otlp, ,console")]
     [InlineData("otlp, console")]
     [InlineData("otlp,,,")]
-    internal void TracesExporter_FailFast(string tracesExporters)
+    internal void TracesExporter_FailFastOn(string tracesExporters)
     {
         Environment.SetEnvironmentVariable(ConfigurationKeys.Traces.Exporter, tracesExporters);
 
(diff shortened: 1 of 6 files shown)


test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/SettingsTests.cs
```diff
@@ -142,19 +142,39 @@ internal void SdkSettings_DefaultValues()
...
 
-    [Fact]
-    internal void TracesExporter_FailFast()
+    [Theory]
+    [InlineData("non-supported", new TracesExporter[0])]
+    [InlineData("otlp,otlp", new[] { TracesExporter.Otlp })]
+    [InlineData("otlp, ,console", new[] { TracesExporter.Otlp, TracesExporter.Console })]
+    [InlineData("otlp, , console", new[] { TracesExporter.Otlp })]
+    [InlineData("otlp,,,", new[] { TracesExporter.Otlp })]
+    internal void TracesExporter_FailFastOff(string tracesExporters, TracesExporter[] expectedTracesExporters)
```
pellared, line 160: Why not adding test cases to `TracesExporter_SupportedValues`? It already has a similar case `[InlineData("non-supported,none", new TracesExporter[0])]`

test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/SettingsTests.cs
```diff
@@ -142,19 +142,39 @@ internal void SdkSettings_DefaultValues()
...
+        settings.TracesExporters.Should().Equal(expectedTracesExporters);
+    }
+
+    [Theory]
+    [InlineData("non-supported")]
+    [InlineData("otlp,otlp")]
+    [InlineData("otlp, ,console")]
+    [InlineData("otlp, console")]
+    [InlineData("otlp,,,")]
+    internal void TracesExporter_FailFastOn(string tracesExporters)
```
pellared, line 175: I think we can keep the old name `TracesExporter_FailFast`.

test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/SettingsTests.cs
ysolomchenko: Because it’s not supported values, this cases have different errors or warnings, but it can work when failfast is disabled

pellared: The test logic is exact the same. We can also rename the `TracesExporter_SupportedValues` to something else. Maybe even `internal void TracesExporter` (without the suffix).

ysolomchenko: Okey, i make it with renaming

pjanotti (approved):
LGTM, minus the test suggestions pointed by @pellared

pellared (approved):

nrcventura (approved):

PR #3601, opened 2024-09-04
[nuget] add publish warning when RuntimeIdentifier is not specified

lachmatt:
## Why

Warn when an app uses autoinstrumentation NuGet package, and is published without specifying runtime identifier. 
In such scenario, native profiler libraries for all of the supported platforms are copied to the publish output directory, which results in increased disk usage.

Fixes #

## What

- Add a MSBuild target that runs before the publish, and issues a warning when runtime identifier is not specified
- Allow the warning to be suppressed by setting `DisableAutoInstrumentationCheckForRuntimeIdentifier` property

I considered making target run before build instead of publish, and issue an Error instead of a Warning (would be consistent with other target), but decided it would be to noisy/restrictive. Let me know if you think I should change the approach.

## Tests

Included in PR.

## Checklist


src/OpenTelemetry.AutoInstrumentation.BuildTasks/OpenTelemetry.AutoInstrumentation.BuildTasks.targets
```diff
@@ -30,7 +30,16 @@
...
-
   </Target>
 
+ <Target Name="VerifyIfRuntimeIdSpecified" BeforeTargets="Publish" Condition="'$(RuntimeIdentifier)' == '' AND $(DisableAutoInstrumentationCheckForRuntimeIdentifier) != 'true'">
+   <ItemGroup>
+    <Lines Include="RuntimeIdentifier (RID) is not set. Consider setting it to avoid copying native libraries for all of the platforms supported by the OpenTelemetry.AutoInstrumentation package." />
+    <Lines Include="See the docs at https://opentelemetry.io/docs/zero-code/net/nuget-packages/#using-the-nuget-packages for details." />
+    <Lines Include="In order to suppress this warning, set DisableAutoInstrumentationCheckForRuntimeIdentifier property to true." />
+   </ItemGroup>
+   <Warning Text="@(Lines, ' ')">
```
lachmatt, line 41: This is the best workaround I've found to avoid very long line here and a clean warning message in dotnet publish results.

RassK (approved):

test/NuGetPackagesTests/InstrumentationTargetTests.cs
```diff
@@ -71,52 +168,23 @@ public void InstrumentApplication()
...
+
+                                      public static class Program
+                                      {
+                                          public static void Main()
+                                          {
+                                              System.Console.WriteLine("Hello World!");
+                                          }
+                                      }
+
+                                      """;
```
pjanotti, line 185: 👍🏼

pjanotti (approved):

Kielek (commented):
IMO worth to add changelog entry.

Kielek (approved):

Issue #3595, opened 2024-08-29
.NET SDK consistency check


Kielek:
As a maintainer I would like to have .NET SDK consistency check between docker files and pipeline definitions.

The goal is to prevent merges such as #3573 and #3577 without updating pipeline definitions #3588

The check should cover
 - `actions/setup-dotnet` - dotnet versions in yml files
 - Docker base images
 - dotnet versions installed by dotnet-install.sh

Kielek:
@pjanotti, FYI

Kielek:
While working on this, you can consider creating automatic script to update all versions based on https://dotnet.microsoft.com/en-us/download/dotnet

pjanotti:
Thanks @Kielek!

PR #3727, opened 2024-10-25
nuke target for verifying .NET SDK versions used

lachmatt:
## Why

Towards #3595 

## What

- Console application for verifying .NET SDK versions used in action workflows and dockerfiles.
- Nuke target running the app
- New workflow

## Tests

Local smoke tests.

## Checklist

### DIFF
--- a/tools/SdkVersionVerifier/ActionWorkflowVerifier.cs
+++ b/tools/SdkVersionVerifier/ActionWorkflowVerifier.cs
@@ -1,0 +1,118 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using YamlDotNet.RepresentationModel;
+
+namespace SdkVersionVerifier;
+
+internal static class ActionWorkflowVerifier
+{
+    public static DotnetSdkVersion? GetExpectedSdkVersionFromSampleWorkflow(string root)
+    {
+        var defaultWorkflow = Path.Combine(GetWorkflowsDirectory(root), "build.yml");
+        return ExtractDotnetSdkVersions(File.ReadAllText(defaultWorkflow)).FirstOrDefault();
+    }
+
+    public static bool VerifyVersions(string root, DotnetSdkVersion expectedDotnetSdkVersion)
+    {
+        var workflowsDir = GetWorkflowsDirectory(root);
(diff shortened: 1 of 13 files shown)


tools/SdkVersionVerifier/VersionComparer.cs
```diff
@@ -0,0 +1,36 @@
...
+internal static class VersionComparer
+{
+    public static bool CompareVersions(DotnetSdkVersion expectedDotnetSdkVersion, DotnetSdkVersion version)
+    {
+        return CompareVersions(expectedDotnetSdkVersion, version.Net6SdkVersion, version.Net7SdkVersion, version.Net8SdkVersion);
+    }
+
+    public static bool CompareVersions(
+        DotnetSdkVersion expectedDotnetSdkVersion,
+        string? net6SdkVersion,
```
RassK, line 15: could be more dynamic, so new sdk versions will not cause issues?

tools/SdkVersionVerifier/ActionWorkflowVerifier.cs
```diff
@@ -0,0 +1,118 @@
...
+        //   7.0.410
+        //   8.0.403
+
+        string? sdk6Version = null;
+        string? sdk7Version = null;
+        string? sdk8Version = null;
+
+        foreach (var version in dotnetVersionNode.ToString().Split())
+        {
+            if (version.StartsWith('6'))
```
RassK, line 88: for example group by first digit?

Kielek (approved):

tools/SdkVersionVerifier/ActionWorkflowVerifier.cs
lachmatt: Do you mean to sort by the first digit?
That'd work as well. Note there are workflows where single .NET SDK version is used, we'd probably assume there that single version is of .NET SDK 8. I'm not sure if suggested approach is better.
I'd prefer to include modifications/improvements in a separate PR.

tools/SdkVersionVerifier/VersionComparer.cs
lachmatt: I considered other approaches, but decided to start simple. I'd prefer to include improvements in a separate PR, unless they are urgent.

RassK (approved):

PR #3796, opened 2024-11-20
.NET9 support

Kielek:
## Why

Towards #3766

## What

Adds support for .NET9.
Drops support for .NET6 and .NET7.
Other related cleanup.
Bumping .NET SDK to `8.0.404 `
`LangVersion` updated to 13 (version supported by .net 9). previous version was taken from .NET7 (needed for some genrated, grpc code).

## Tests

CI

## Checklist

## Notes

`dotnet format` is failing. Not related to changes.
PR is huge, IMO worth to review commit by commit.

### DIFF
--- a/test/OpenTelemetry.AutoInstrumentation.Tests/TransientDependenciesTests.cs
+++ b/test/OpenTelemetry.AutoInstrumentation.Tests/TransientDependenciesTests.cs
@@ -1,8 +1,8 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-// This test is defined in NET8.0 because the tool is written in .NET 8.0
-// The actual test is testing .NET 462 context.
+// This test is defined in NET 9.0 because the tool is written in .NET 9.0
+// The actual test is testing .NET Framework 4.6.2 context.
 #if NET9_0_OR_GREATER
 
 using System.Reflection;


Kielek:
@stevejgordon, fyi

tools/SdkVersionAnalyzer/Program.cs
```diff
@@ -1,12 +1,15 @@
...
 
+using System.Diagnostics;
+
 namespace SdkVersionAnalyzer;
 
 internal static class Program
 {
     public static int Main(string[] args)
     {
+        Debugger.Launch();
```
RassK, line 12: should not be necessary any more?

Kielek, line 12:
```suggestion
    {
```

Kielek: Fixed.

tools/SdkVersionAnalyzer/DockerfileAnalyzer.cs
```diff
@@ -56,8 +55,8 @@ private static bool VerifySdkVersions(string content, DotnetSdkVersion expectedD
         // && echo "SHA256: $(sha256sum dotnet-install.sh)" \
         //     && echo "de4957e41252191427a8ba0866f640b9f19c98fad62305919de41bd332e9c820  dotnet-install.sh" | sha256sum -c \
         //     && chmod +x ./dotnet-install.sh \
-        //     && ./dotnet-install.sh -v 6.0.427 --install-dir /usr/share/dotnet --no-path \
-        //     && ./dotnet-install.sh -v 7.0.410 --install-dir /usr/share/dotnet --no-path \
+        //     && ./dotnet-install.sh -v 8.0.403 --install-dir /usr/share/dotnet --no-path \
```
RassK, line 58:
```suggestion
        //     && ./dotnet-install.sh -v 8.0.404 --install-dir /usr/share/dotnet --no-path \
```

tools/SdkVersionAnalyzer/ActionWorkflowAnalyzer.cs
```diff
@@ -79,14 +79,13 @@ private static Scalar GetScalarWithExpectedFormatting()
...
-        var val = $"{newDotnetSdkVersion.Net6SdkVersion!}{separator}{newDotnetSdkVersion.Net7SdkVersion!}{separator}{newDotnetSdkVersion.Net8SdkVersion!}{separator}";
+        var val = $"{newDotnetSdkVersion.Net8SdkVersion!}{separator}{newDotnetSdkVersion.Net9SdkVersion!}{separator}";
 
         // Use ctor with default values, apart from ScalarStyle.
         // Use ScalarStyle.Literal to get dotnet-version with value similar to below:
         //  dotnet-version: |
-        //    6.0.437
-        //    7.0.420
-        //    8.0.413
+        //    8.0.403
```
RassK, line 87:
```suggestion
        //    8.0.404
```

test/IntegrationTests/IntegrationTests.csproj
```diff
@@ -1,10 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
-    <!-- No warn is needed for autogenerated code by protobuf: CS8981 - The type name only contains lower-cased ascii characters. -->
-    <NoWarn Condition="'$(TargetFramework)' == 'net8.0' OR '$(TargetFramework)' == 'net7.0'">CS8981</NoWarn>
```
RassK, line 5: What changed here with .NET 8 that NoWarn is no longer needed?

Kielek: I think that it was leftover from previous grpc toolking upgrade. In some versions it stopped using internally bad names.

tools/SdkVersionAnalyzer/DockerfileAnalyzer.cs
Kielek: It is example of parsing. Not needed to upgrade version.

tools/SdkVersionAnalyzer/Program.cs
```diff
@@ -22,9 +22,9 @@ public static int Main(string[] args)
                     return VerifyDotnetSdkVersions(directoryRoot);
                 }
 
-            case "--modify" when args.Length == 5:
+            case "--modify" when args.Length == 4:
```
RassK, line 25: Marking this location for myself. I'll take a look how to improve, no array size changes would not waste time to debug this extra.

test/OpenTelemetry.AutoInstrumentation.Tests/TransientDependenciesTests.cs
```diff
@@ -3,7 +3,7 @@
 
 // This test is defined in NET8.0 because the tool is written in .NET 8.0
 // The actual test is testing .NET 462 context.
-#if NET8_0_OR_GREATER
+#if NET9_0_OR_GREATER
```
zacharycmontoya, line 6: Was this change intentional? If not, we should revert it back

```suggestion
#if NET8_0_OR_GREATER
```

zacharycmontoya: Actually it does look intentional, as the tool `tools/DependencyListGenerator/DependencyListGenerator.csproj` is being updated to be built with `net9.0`. If that's correct, can you confirm and close this?

test/test-applications/integrations/TestApplication.Azure/TestApplication.Azure.csproj
```diff
@@ -2,7 +2,5 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Storage.Blobs" VersionOverride="$(LibraryVersion)" />
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Condition=" '$(TargetFramework)' == 'net6.0' " />
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" VersionOverride="8.0.0" Condition=" '$(TargetFramework)' == 'net462' and '$(LibraryVersion)' != '' and '$(LibraryVersion)'&lt;'12.18.0' " />
```
zacharycmontoya, line 6: Is there a reason we no longer need the `net462` conditional PackageReference? I assumed we would still need it

build/LibraryVersions.g.cs
```diff
@@ -34,7 +34,7 @@ public static partial class LibraryVersion
             new List<PackageBuildInfo>
             {
                 new("6.0.35"),
-                new("8.0.10", supportedFrameworks: new string[] {"net8.0"}),
+                new("8.0.10"),
```
zacharycmontoya, line 37: I don't understand why lines 37, 46, and 47 were updated to remove the `supportedFrameworks` parameter but lines 123, 124, and 174 were updated to keep the `supportedFrameworks` and a new `net9.0` TFM. Why is that?

src/OpenTelemetry.AutoInstrumentation/DuckTyping/DuckType.cs
```diff
@@ -328,7 +328,11 @@ private static ModuleBuilder CreateTypeAndModuleBuilder(Type typeToDeriveFrom, T
             // If the proxy type definition is an interface we create an struct proxy
             // If the proxy type definition is an struct then we use that struct to copy the values from the target type
             parentType = typeof(ValueType);
-            typeAttributes = TypeAttributes.Public | TypeAttributes.AnsiClass | TypeAttributes.BeforeFieldInit | TypeAttributes.SequentialLayout | TypeAttributes.Sealed | TypeAttributes.Serializable;
+            typeAttributes = TypeAttributes.Public | TypeAttributes.AnsiClass | TypeAttributes.BeforeFieldInit | TypeAttributes.SequentialLayout | TypeAttributes.Sealed
+#pragma warning disable SYSLIB0050
+                             // TODO double check how to handle TypeAttributes.Serializable which is obsolete
```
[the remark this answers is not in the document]
zacharycmontoya: The usage of `TypeAttributes.Serializable` seems to have been used since its introduction into the Datadog automatic instrumentation. I've messaged my colleagues to understand if it's truly needed, and if it is then we'll need to suppress this compiler warning. Or perhaps if it's not needed for newer runtimes we can ifdef its usage, but I don't know if that's the case

zacharycmontoya (commented):
I've looked through and overall it looks good. I'll wait for resolution on existing comments before approving

nrcventura (commented):
There are many places in the code where we needed to change `NET60_OR_GREATER` to `NET80_OR_GREATER`. It seems like the intent of the directive in many, if not most of those locations was to use a version of the code that only applied to the .net build target and not the .net framework build target.

Where appropriate, should we change `NET80_OR_GREATER` to either `NET` or `!NETFRAMEWORK`, to make the intent clearer?

Kielek:
> There are many places in the code where we needed to change `NET60_OR_GREATER` to `NET80_OR_GREATER`. It seems like the intent of the directive in many, if not most of those locations was to use a version of the code that only applied to the .net build target and not the .net framework build target.
> 
> Where appropriate, should we change `NET80_OR_GREATER` to either `NET` or `!NETFRAMEWORK`, to make the intent clearer?

@nrcventura, I agree, but I do not want to make such changes in the same file. just kept the current convention. New issue created: #3798 (not a blocker to the release IMO, just normal maintenance task).

src/OpenTelemetry.AutoInstrumentation/DuckTyping/DuckType.cs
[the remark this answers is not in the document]
Kielek: I have created #3799.

It can be done after merge, it was working fine on .NET8 with .NET6 binaries for a while. No rush to changes.

test/OpenTelemetry.AutoInstrumentation.Tests/TransientDependenciesTests.cs
Kielek: Intentional, there is no need to execute it more than one. Comments updated in 6ac1c0472f60fdb07af01c2227fa1226d09d343c

build/LibraryVersions.g.cs
Kielek: Lines 123, 124, and 174: NServiceBus, and WCFClient are supported both on .NET Framework and .NET.
Some of the newer version of tested libraries are working only on .NET (currently 8 and 9) so I need to add target frameworks.

37, 46, and 47: EntityFrameworkCore and EntityFrameworkCore.Pomelo.MySql are supported only on .NET. As we dropped support for .NET6 and .NET7 there is no longer to explicitly mark that 8.* versions are supported only on .NET8+.  [Version 9.0 of EFC](https://www.nuget.org/packages/Microsoft.EntityFrameworkCore/9.0.0#supportedframeworks-body-tab) is also supporting .NET8+, so there will be changes in the coverage in near future. I have a plant update packages as a separate step.

test/test-applications/integrations/TestApplication.Azure/TestApplication.Azure.csproj
Kielek: It was probably needed when we were testing against older package versions

Currently we have in the generator:
```
            IntegrationName = "Azure",
            NugetPackageName = "Azure.Storage.Blobs",
            TestApplicationName = "TestApplication.Azure",
            Versions = new List<PackageVersion>
            {
                // new("12.13.0"), // all lower versions than 12.22.2 contains references impacted by https://github.com/advisories/GHSA-8g4q-xg66-9fp4
                new("12.22.2"),
                new("*")
            }
```

So it is no longer lower than 12.18.0. So the code is not needed. It is just kind of additional cleanup which technically could be done outside this PR.

test/IntegrationTests/BuildTests.DistributionStructure_alpine-linux-arm64.verified.txt
```diff
@@ -1,15 +1,13 @@
...
   /AdditionalDeps/shared/Microsoft.NETCore.App/8.0.0/OpenTelemetry.AutoInstrumentation.AdditionalDeps.deps.json,
+  /AdditionalDeps/shared/Microsoft.NETCore.App/9.0.0/OpenTelemetry.AutoInstrumentation.AdditionalDeps.deps.json,
   /LICENSE,
   /instrument.sh,
   /linux-musl-arm64/OpenTelemetry.AutoInstrumentation.Native.so,
   /net/Google.Protobuf.dll,
   /net/Grpc.Core.Api.dll,
   /net/Grpc.Net.Client.dll,
   /net/Grpc.Net.Common.dll,
-  /net/Microsoft.Extensions.Diagnostics.Abstractions.dll,
```
RassK, line 12: I guess this is now shipped by the framework? 

Could prob remove it here also then.
https://github.com/Kielek/opentelemetry-dotnet-instrumentation/blob/d992b5fd8518ceb7ae1aa9eaeb2fee2882caae88/LICENSE#L271

RassK (approved):
Some minor things and tweaks only. In overall tests are passing, seems like good to go.

test/IntegrationTests/BuildTests.DistributionStructure_alpine-linux-arm64.verified.txt
Kielek: It is shipped for .NET Fx -> should stay in Liecense file.

It is referenced in 8.0.0.0 version. I is included in .NET8 -> it is always available at least in this version. No need to shipping it.

build/LibraryVersions.g.cs
zacharycmontoya: Ah understood. I had not thought that the changes would be package-specific. Thanks for the follow-up!

zacharycmontoya (approved):
LGTM, thanks!

nrcventura (approved):

PR #3960, opened 2025-01-20
[log4net-bridge] limit number of cached loggers

lachmatt:
## Why

Limit number of cached loggers.


Kielek (approved):

RassK (approved):
This is simple way but to be sure to cache more frequently generated ones maybe cache with timeout function is more interesting one to use. So it's more probable that most used loggers are cached and low usage ones are not.

pjanotti (approved):

Kielek:
@RassK, only LRUCache I have found is in [AdaptiveExpressions.dll](https://learn.microsoft.com/en-us/dotnet/api/adaptiveexpressions.lrucache-2?view=botbuilder-dotnet-stable) / https://github.com/microsoft/botbuilder-dotnet/blob/bc17a5db2c717fec0d4c109d884bea88d29f75cb/libraries/AdaptiveExpressions/LRUCache.cs#L13

It is sad, that there is no part of the standard lib :(. Merging.

PR #3962, opened 2025-01-20
Fix macos arm64 local build

pjanotti:
Fixing building on an ARM mac.


RassK (approved):

Kielek (approved):
I do not have such machine to verify it, but changes LGTM

PR #3993, opened 2025-01-29
[Continuour Profiling] define Sample Type and Period Type

Kielek:
## Why

Towards profiling OTLP 1.4.0/1.5.0, values of Sample/Period Types consulted with @florianl.

## What

[Continuour Profiling] define Sample Type and Period Type

## Tests

CI + modified tests

## Checklist

### DIFF
--- a/test/test-applications/integrations/TestApplication.ContinuousProfiler/Exporter/OtlpOverHttpExporter.cs
+++ b/test/test-applications/integrations/TestApplication.ContinuousProfiler/Exporter/OtlpOverHttpExporter.cs
@@ -45,6 +45,7 @@ public class OtlpOverHttpExporter
                 var threadSample = threadSamples[i];
 
                 var sampleBuilder = CreateSampleBuilder(threadSample, extendedPprofBuilder);
+                sampleBuilder.SetValue(1);
 
                 extendedPprofBuilder.Profile.Sample.Add(sampleBuilder.Build());
             }
--- a/test/test-applications/integrations/TestApplication.ContinuousProfiler/Exporter/OtlpOverHttpExporter.cs
+++ b/test/test-applications/integrations/TestApplication.ContinuousProfiler/Exporter/OtlpOverHttpExporter.cs
@@ -1,7 +1,6 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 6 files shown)


test/test-applications/integrations/TestApplication.ContinuousProfiler/Exporter/OtlpOverHttpExporter.cs
```diff
@@ -33,13 +38,14 @@ public void ExportThreadSamples(byte[] buffer, int read, CancellationToken cance
...
             var timestampNanoseconds = threadSamples[0].TimestampNanoseconds; // all items in the batch have same timestamp
-            var extendedPprofBuilder = new ExtendedPprofBuilder("cpu", timestampNanoseconds);
+            var extendedPprofBuilder = new ExtendedPprofBuilder("samples", "count", "cpu", "nanoseconds", cpuPeriod, timestampNanoseconds);
 
             for (var i = 0; i < threadSamples.Count; i++)
             {
                 var threadSample = threadSamples[i];
 
                 var sampleBuilder = CreateSampleBuilder(threadSample, extendedPprofBuilder);
+                sampleBuilder.SetValue(1);
```
Kielek, line 48: @florianl, I have some doubts about this, but this values is corresponding to "samples/count". Could you please confirm that it is expected for this cases?

florianl: Yes - setting `1` here is correct. There is no implicit assumption, that if a sample got reported it automatically correlates to a value.

E.g. the same sample can be recorded `x` times in a single interval. Instead of reporting `x` times the same sample, the value of the sample can be directly set to `x` - in such a case, `Sample.timestamps_unix_nano` is also often set `x` times to report the precise recording times for each of the `x` recordings of the same sample.

florianl (approved):

test/test-applications/integrations/TestApplication.ContinuousProfiler/Exporter/OtlpOverHttpExporter.cs
Kielek: Thank you!

pjanotti (approved):

PR #4060, opened 2025-02-25
[next-gen] Add mini-SDK

CodeBlanch:
## Changes

* Adds the initial code for the mini-SDK based on what was used in the dotnet-monitor proof-of-concept.

## Details

See [dotnet-monitor OTLP Egress](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/next-gen/docs/dotnet-monitor-otlp-egress.md)

This is the mini-SDK component. I call it "mini-SDK" because it is a (more or less) fully compliant implementation of the OTel SDK spec but only partially. It is basically processor/reader & exporter portions. This will be used by dotnet-monitor to forward telemetry collected out of process using OTLP (certainly possible to plug-in other exporters in the future).

There is probably going to be one more small project added which will house the IConfiguration/Options binding but I'll add that as a follow-up.

Feel free to drop review comments. I'll handle small[ish] things. Anything bigger will probably need to be spun out into follow-up work because I only have limited time to work on this. The plan is for @rajkumar-rangaraj to drive the effort from here.

Work I know needs to be done:

* IConfiguration/Options binding
* CI/Publish pipelines
* OTLP exporter code needs to be updated to use the manual serialization style so references can be dropped
* Tests


next-gen/src/OpenTelemetry.AutoInstrumentation.Sdk/.publicApi/PublicAPI.Unshipped.txt
Kielek, line 1: General question about public API.
Do we really need public or it can be all marked as internal?

Do you think that end-users will have any benefits from touching it?

next-gen/src/OpenTelemetry.AutoInstrumentation.Sdk/OpenTelemetry.AutoInstrumentation.Sdk.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <MinVerTagPrefix>sdk-</MinVerTagPrefix>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" />
+
+    <!-- Temporary until manual serialization is implemented -->
+    <PackageReference Include="Grpc.Net.Client" />
+    <PackageReference Include="Google.Protobuf" />
+    <PackageReference Include="Grpc.Tools" PrivateAssets="All" />
```
Kielek, line 16: Follow up PR?

Kielek (commented):
What is the plan for syncing it with the normal SDK?
I expect that some fixes will be occurring. I would like to avoid manual cherry-picking. 
Do you see any possibility to (semi)automatically _vendorize_ the code?

Or mark this as an git submodule and host everything in the main repository with some compilation flags?

pellared:
This PR is huge. Do we really expect that it is going to be reviewed and merged?

Maybe it should it be marked as draft and used a reference when creating smaller PRs?

next-gen/src/OpenTelemetry.AutoInstrumentation.Sdk/.publicApi/PublicAPI.Unshipped.txt
CodeBlanch: * Sharing with main repo...

  There really isn't a whole lot of overlap with the main SDK. This mini-SDK was written from scratch. I did use some things, like CircularBuffer and a few enums, but I wouldn't expect a lot of work to keep anything in sync.

* Public API...

  `dotnet-monitor` will need to call the API. We could make everything internally visible to `dotnet-monitor` but I really discourage that. It encourages bad practices and really increases the chances of breaking `dotnet-monitor` anytime changes are made. The public-API enforcement will help make sure that doesn't happen. Perhaps a better solution would be to not publish to NuGet but some other/private feed? Would need to sort that out with dotnet team but they might be open to it.

CodeBlanch:
> This PR is huge. Do we really expect that it is going to be merged?
> 
> Can it be marked as draft and used a reference when creating smaller PRs?
> 
> I am not against having this PR open for high-level reviewing purposes, but I do not want to have such a big PR merged.

Code bomb 💣 Sorry about that 😄

We just discussed on the SIG. The plan is to merge this PR. It is going into `next-gen` folder. @rajkumar-rangaraj and his team will work on moving it piece-by-piece into the main folder structure with smaller PRs adding tests as they go. Those can be reviewed more thoroughly.

pellared (dismissed):
> @rajkumar-rangaraj and his team will work on moving it piece-by-piece into the main folder structure with smaller PRs adding tests as they go. Those can be reviewed more thoroughly.

It still does not answer the question why not keeping this PR open and create PRs into the main folder structure with smaller PRs :shrug:  I would appreciate some reasons. 

I have reasons why I am against merging it. Since a 4 years I have been working in OTel I have never seen such practice with the exception of code donations which are processed by @open-telemetry/technical-committee. I think that merging this PR would be against [the OpenTelemetry policies](https://github.com/open-telemetry/community/blob/main/guides/contributor/donations.md).

> Large donations – or small donations that turn up complex issues during PR review – should be referred to the Technical Committee (TC) by filing an issue in this `community` repository and tagging `@open-telemetry/technical-committee`. 

As a maintainer I feel responsible to make sure that these policies are followed.

> Can it be marked as draft and used a reference when creating smaller PRs?

Here are some examples from other repository:
- https://github.com/open-telemetry/opentelemetry-go/pull/4725
- https://github.com/open-telemetry/opentelemetry-go/pull/4955

Going this way we would not break the OpenTelemetry policies and also minimize the "code traffic" in this repository. Personally, I find pushing PoC/Prototype code into a git repository a bad practice.

tigrannajaryan:
>This is the mini-SDK component. I call it "mini-SDK" because it is a (more or less) fully compliant implementation of the OTel SDK spec but only partially. It is basically processor/reader & exporter portions. This will be used by dotnet-monitor to forward telemetry collected out of process using OTLP (certainly possible to plug-in other exporters in the future).

This is interesting and appears similar to how we have multiple Collector distributions, each including a different subset (or all) components. Is this the goal? To have multiple (2 for now - full and minimal) SDK distributions with different sets of components included? I would like to understand this better and make sure we are consistent in what we call an SDK across languages and how the mini-SDK matches the compliance matrix that we publish. It is worth having this discussion with a  broader audience since it may be applicable to other language SDKs.


> We just discussed on the SIG. The plan is to merge this PR. It is going into `next-gen` folder. @rajkumar-rangaraj and his team will work on moving it piece-by-piece into the main folder structure with smaller PRs adding tests as they go. Those can be reviewed more thoroughly.

If the plan is to have smaller PRs that are reviewed more thoroughly and independently, what's the added value of having `next-gen` folder compared to simply keeping this PR as a draft? I can see downsides (e.g. merging code that is not thoroughly reviewed), but not sure what the upsides are.

BTW, I like that this PR exists and shows the full picture.

pellared:
> BTW, I like that this PR exists and shows the full picture.

I should also say this :+1:  I am not against the mini-SDK, just the process of introducing it the the repository :wink: 

I simply share @tigrannajaryan's  point of view:

> If the plan is to have smaller PRs that are reviewed more thoroughly and independently, what's the added value of having `next-gen` folder compared to simply keeping this PR as a draft? I can see downsides (e.g. merging code that is not thoroughly reviewed), but not sure what the upsides are.

rajkumar-rangaraj:
It would be simpler to keep this in the same repository in a temporary location (e.g., the next-gen directory), allowing multiple contributors to access and collaborate on the code easily. This was discussed in the SIG meeting, and there was general agreement on this approach, as it facilitates collective progress without fragmenting efforts across multiple repositories.

If merging this PR into the main branch is a concern, an alternative would be to create a dedicated next-gen branch. This would provide a structured way to continue development while keeping it separate from the primary codebase, allowing for iterative improvements and discussions before any final integration.

This approach keeps the work centralized while giving flexibility to evolve the implementation. Let me know if you have any strong objections or if you prefer a different way forward.

reyang:
> It would be simpler to keep this in the same repository in a temporary location (e.g., the next-gen directory), allowing multiple contributors to access and collaborate on the code easily. This was discussed in the SIG meeting, and there was general agreement on this approach, as it facilitates collective progress without fragmenting efforts across multiple repositories.

The fact that @pellared (who is a maintainer of this repo) blocked this PR is telling me that there is no such general agreement.

> If merging this PR into the main branch is a concern, an alternative would be to create a dedicated next-gen branch. This would provide a structured way to continue development while keeping it separate from the primary codebase, allowing for iterative improvements and discussions before any final integration.

That sounds to me like a very reasonable approach.

RassK:
I think also next-gen branch is more suitable. I think even if some of the tests might be same for both branches their infra will be much different anyway. So there are no unnecessary running long pipelines if committing to either location.

I went through some of the design docs, so to me it seems auto instrumentation (part that also writes / translates "events") is still necessary in different form. This will be a huge conflict in the main branch.

RassK:
> > This is the mini-SDK component. I call it "mini-SDK" because it is a (more or less) fully compliant implementation of the OTel SDK spec but only partially. It is basically processor/reader & exporter portions. This will be used by dotnet-monitor to forward telemetry collected out of process using OTLP (certainly possible to plug-in other exporters in the future).
> 
> This is interesting and appears similar to how we have multiple Collector distributions, each including a different subset (or all) components. Is this the goal? To have multiple (2 for now - full and minimal) SDK distributions with different sets of components included? I would like to understand this better and make sure we are consistent in what we call an SDK across languages and how the mini-SDK matches the compliance matrix that we publish. It is worth having this discussion with a broader audience since it may be applicable to other language SDKs.
> 

A bit dangerous path to have 2 versions of SDK. I'm thinking of separating the main SDK as read and write part (depends how other languages deal with a common telemetry pipe, like dotnet has in form of System.Diagnostics.DiagnosticSource). The read part should be simply reusable for in-proc and out of proc listeners. The writing part should slowly-slowly die off, since libs should natively write telemetry instead of hacking telemetry into libs. Although something may remain in form of "translators" / aggregators.

I kind of see the separation already but there are some inst. libs that break the clear separation.

rajkumar-rangaraj:
**Out-of-Process Solution Discussion:** We’ve been discussing an out-of-process solution for over a year within the OpenTelemetry Auto-Instrumentation SIG. Notably, last September, @CodeBlanch presented a foundational design and demo. I suggest we continue that conversation in an upcoming SIG meeting rather than in this PR. We can also invite the Technical Committee (TC) - @open-telemetry/technical-committee to ensure we cover all perspectives. 

**Clarification on the “Mini-SDK”:** The term “mini-SDK” has sparked some debate. To clarify, this component doesn’t create tracer, meter, or log providers—it primarily functions as an auto-instrumentation forwarder. @CodeBlanch referred to it as a “mini-SDK,” but we can finalize a more precise name later to avoid misunderstandings. Given the confusion, SIG should discuss and rename it (e.g., "Auto-Instrumentation Forwarder") to better reflect its intended role.

**Integration into the Repository:** There seems to be confusion about whether this is an alternative to the existing auto-instrumentation solution. The plan is to eventually replace the in-process model with this out-of-process approach, which is why it makes sense to integrate it into this repository. Keeping both approaches within the same repository allows for a structured transition and avoids fragmentation. This also ensures that existing contributors can easily collaborate on the migration.

**Limitations of the Current In-Process Model:** The current approach relies on AdditionalDeps and runtime hooks provided by .NET, which may not be supported in future .NET versions. Moreover, the .NET team has strongly recommended against relying on in-process auto-instrumentation due to potential stability and compatibility concerns. Given these limitations, continuing to invest in the in-process model poses long-term risks, making an out-of-process alternative a more sustainable choice.

**Benefits for AOT and Self-Contained Scenarios:** With newer .NET versions, scenarios like Ahead-Of-Time (AOT) compilation and self-contained applications are more prevalent. The current in-process solution doesn’t support these scenarios effectively. Transitioning to an out-of-process model will enable broader compatibility, including AOT and self-contained cases, ensuring our approach remains future-proof. Additionally, the ability to support lightweight and cloud-native environments makes this transition critical for OpenTelemetry’s adoption in modern .NET applications.

**dotnet-monitor:** The decision to support OTLP export from dotnet-monitor is based on practical considerations. dotnet-monitor already has an established Event Pipe integration, making it the most efficient path forward. Creating and maintaining another OpenTelemetry Collector receiver specifically for this would require additional effort without significant benefit. However, the mini-SDK (forwarder) is not exclusive to dotnet-monitor—it is part of a broader effort to replace in-process auto-instrumentation. This approach ensures better compatibility for AOT, self-contained, and cloud-native applications, extending its applicability beyond just dotnet-monitor. While dotnet-monitor is the first adopter, similar functionality could be implemented in other tools as well.

CodeBlanch:
> Do we need another SDK?
I also share @tigrannajaryan and @RassK concerns about having 2 public versions of OpenTelemetry SDK. Would it not simply be a v2-light? If so why not even hosting the code in https://github.com/open-telemetry/opentelemetry-dotnet? I think this is very confusing and the plan how it integrates with OpenTelemtry as a whole ecosystem (also how the community understands it) should be clarified.

Full disclosure, I haven't read all the comments on this PR. I don't view myself as having skin in the game or a dog in the fight (so to speak). I'm just handing off what I have for @rajkumar-rangaraj and team to carry it forward. Whatever is decided here is fine by me.

But I wanted to share some context.

I do feel naming here is important. I brought this up on the initial discussions. Having two SDKs isn't good for the community. But this doesn't need to be called "SDK" I like @rajkumar-rangaraj's idea to use "forwarder."

The real question is, why aren't we using the main SDK, and why do we need this "forwarder" thing?

This work is coming out of a proof-of-concept I did. While doing that PoC the first thing I did was utilize the main SDK. I think it is natural to do that. But there are lots of challenges there and we quickly pivoted off the main SDK.

* Tracing

   The main SDK uses `Activity` & `ActivitySource`. In this out-of-process model, there aren't either things. There is just data streaming in from EventPipe. In order to use the main SDK, you have to create `Activity` instances and feed them through some `ActivitySource`. There aren't good APIs for that. `Activity` doesn't let you set data on everything. `ActivitySource` doesn't let you feed it something ended. You have to use reflection hacks to make it work.

* Logging

  The main SDK doesn't have a stable Log Bridge. There is a preview/experimental one. Log Bridge would probably be a good fit but much work needs to be done there in the main SDK first. We could read the logs off the stream and re-submit them through `ILogger` \ `ILoggerFactory` but there are issues there too. For example you can't set the timestamp to what came in on the stream. And for correlation you need to actually start a trace (`Activity.Current`).

* Metrics

  The main SDK doesn't support `MetricProducer`. It would be a nice fit if it did. We have looked at supporting `MetricProducer` and it is a massive effort. Sadly in the main SDK metrics are tied deeply to the internal implementation. A massive overhaul would need to be done to support letting `MetricProducer`s feed in batches. There doesn't seem to be demand or any active effort going on here.

What this mini-SDK/forwarder does is give an API to submit log records. More or less a Logs Bridge API. And it gives you a `MetricProducer` to emit metrics. For tracing it let's you submit ended spans with all the data.

Could the main SDK be made to support all this? Sure. But who is going to do that work? The SDK users seem happy. Does it make sense to blow up everything, take people off of other work, and delay solving this out-of-proc problem in order to achieve some ideal state?

Could the main SDK be re-written to utilize this mini-SDK thing? Sure. But why? Forcing SDK off of `Activity` would be a breaking change. And there would be a perf hit taking `Activity` and converting it to some other DTO. Metrics would likely need to be totally redone. What benefit would there be to users? Seem these two things have different needs/goals.

pellared:
> We’ve been discussing an out-of-process solution for over a year within the OpenTelemetry Auto-Instrumentation SIG. Notably, last September, @CodeBlanch presented a foundational design and demo. I suggest we continue that conversation in an upcoming SIG meeting rather than in this PR.

Having discussions during SIG meetings is fine as long as the outcome is reflected in the project as issues, design docs, comments etc. We cannot expect everyone is able to join the SIG meetings or watch the whole recordings.   Moreover, almost none of it is noted down in the SIG meeting notes. 

Meetings are great but this is not the only tool  that should be for decision making, and communication in an open source project.

Is there any GitHub issue, GitHub project, design doc where the this vision is described? 

As a maintainer of this project I expect that our plans are clear and transparent. Transparent of the communication and decision making is not only important for us but for future contributors and even users of this project. 

Giving the reactions to my previous comments I think that I am not alone who does not see this PR being merged to be justified.

Reference: https://opensource.guide/best-practices/#documenting-your-processes

pellared:
> I'm just handing off what I have for @rajkumar-rangaraj and team to carry it forward.

Fine for me 👍 This prototype is really great.

I would like you to know that I am not against anyone. I just think that the process of introducing such changes should be improved. 

I also think that this is a bigger change that the TC should be aware of. 

I does not mean that I am blocking this new way of doing auto-instrumentation.

RassK:
I think the mini sdk describes it perfectly. If you strip instrumentations (telemetry generation), aggregators (processors) and fix exporter to OTLP, then this is what you get. nservice

> Could the main SDK be made to support all this? Sure. But who is going to do that work? The SDK users seem happy. Does it make sense to blow up everything, take people off of other work, and delay solving this out-of-proc problem in order to achieve some ideal state?
> 
> Could the main SDK be re-written to utilize this mini-SDK thing? Sure. But why? Forcing SDK off of `Activity` would be a breaking change. And there would be a perf hit taking `Activity` and converting it to some other DTO. Metrics would likely need to be totally redone. What benefit would there be to users? Seem these two things have different needs/goals.

The same is with mini sdk. Now the OTLP exporter has to be managed in 2 repos, any other exporter has to be. Telemetry generation still needs separate work, since after 4 years major libs[^1] are still not properly natively instrumented (or are very unfriendly towards auto-instrumentation by taking heavy deps). Telemetry generation (hacks) can only be in-proc. Every concept from the main sdk has to be ported and manually tracked, if there is a need to opt in to something. Reminding that we simply started also just by having S.D.DS conflict "only", which now is a major conflict zone. 

Dotnet monitor can block integration concepts the same way as dotnet runtime does. So it's a way more restrictive environment than porting the event pipe listener directly to next gen. 

[^1]:redis, rabbit, kafka, need to double check service meshes like nservicebus and masstransit. Sqlite driver has been stalled for over a year https://github.com/dotnet/efcore/issues/13832.

pellared:
> Creating and maintaining another OpenTelemetry Collector receiver specifically for this would require additional effort without significant benefit.

I have found the [`dotnetdiagnosticsreceiver`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/f343580ebbb61aa79e033de7dfded2742c69c449/receiver/dotnetdiagnosticsreceiver) which does looks to already have some code for listening on .NET Diagnostics IPC on Linux and macOS.

CC @pmcollins

PR #4076, opened 2025-03-06
allow wget in otel-dotnet-auto-install.sh

### LINKED ISSUE #4072: Use wget instead of curl in otel-dotnet-auto-install.sh
# Feature Request

I'd like otel-dotnet-auto-install.sh to use wget instead of curl

.net has an alpine runtime image, and I'd prefer to not install any packages to get otel-dotnet-auto-install.sh to complete

otel-dotnet-auto-install.sh should be possible to change to wget and in doing so make it zero hassle. that or check which is present and use wget when curl isn't.

right now, I temporarily install curl

**Additional context**

nope

codefromthecrypt:
## Why

I want to be able to use otel-dotnet-auto-install.sh on mcr.microsoft.com/dotnet/runtime:8.0-alpine which doesn't have curl, and without adding a package for it.

Fixes #4072

## What

This conditionally uses curl or wget in that order.

## Tests

I tested this by running the relevant shell script on mcr.microsoft.com/dotnet/runtime:8.0-alpine, which only has wget, and mcr.microsoft.com/dotnet/sdk:8.0-alpine which also has curl. Both worked and the latter used curl.

## Checklist


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
 
 ### Changed
 
+- otel-dotnet-auto-install.sh can optionally use wget instead of curl, useful on
+  mcr.microsoft.com/dotnet/runtime alpine images.
```
pellared, line 15:
```suggestion
- `otel-dotnet-auto-install.sh` now optionally uses `wget` instead of `curl`,
  improving compatibility with `mcr.microsoft.com/dotnet/runtime` Alpine images.
```

pellared (approved):

RassK (approved):

Kielek (approved):
@codefromthecrypt, thanks! Looking for more contribution!

Issue #4186, opened 2025-05-06
Loading this assembly would produce a different grant set from other instances


iskiselev:
# Bug Report
Use Asp.NET web appliction without setting LoaderOptimization.SingleDomain results in site error 500 with "Loading this assembly would produce a different grant set from other instances"

## Symptom

**Describe the bug**
Most IIS web-applications for .Net Framework using Otel.AutoInstrumentation are non-functional without applying 
LoaderOptimization.SingleDomain through registry settings.

**Expected behavior**
Web applications should continue to work without registry modifications

**Additional context**
Origanally reported in 3839, workaround (with registry change) suggested in 3996.

iskiselev:
First attempts to fix was in #4165, but it was not enough.
Current plan to fix it in stages:

1. Patch `appDomainSetup.LoaderOptimization = LoaderOptimization.SingleDomain` in `AppDomain.CreateDomain` and `AppDomainManager.CreateDomainHelper`.
Patch functions would be implemented in `OpenTelemetry.AutoInstrumentation.Loader` (it will help in next phase). For it, move `__DDVoidMethodType__` into `mscorlib`, split `OpenTelemetry.AutoInstrumentation` initialization call into separate method (loading assembly and resolving reflection in `.cctor` of `__DDVoidMethodType__`) and adding method that will call AppDomainSetup modification through delegate resolved with reflection in static ctor.
On that stage, it would not be required to do registry modification with the cost that all non-default AppDomains would be loaded with `LoaderOptimization.SingleDomain`.

1.  If OpenTelemetry.AutoInstrumentation is loaded from GAC, we have better approach. We will patch app/web.config for new app-domains, configuring assembly references for us and dependencies and codebases. It will allow to not change LoaderOptimization. If assemblies are not in GAC, it will effectively work as SingleDomain even when default IIS mode MultiDomainHostis used (as we are added as dependencies to all assemblies). .Some Net Framework assemblies may still be loaded in neutral app-domain multiple times, assembly redirection in default domain and other domains will be different.
After step (2), per my tests no issues with both cases: when assemblies are in GAC or just in OTEL folder.

iskiselev:
If we want to be able load without switching to `LoaderOptimization.SingleDomain`, we need generate proper assembly bindings.

[...]

PR #4187, opened 2025-05-06
Automatic LoaderOptimization.SingleDomain

iskiselev:
## Why

First phase for solving #4186. With this change, LoaderOptimization.SingleDomain would be automatically applied for all non-default app domains.

## What

`__DDVoidMethodType__`  injected to mscorlib.
`__DDVoidMethodType__.__DDPatchAppDomainSetup__` uses reflection-created delegate that calls `OpenTelemetry.AutoInstrumentation.Loader.AppConfigUpdater.ModifyConfig` to modify `AppDomainSetup` by setting `LoaderOptimization.SingleDomain`. That method called from `AppDomain.CreateDomain` and `AppDomainManager.CreateDomainHelper`.
`OpenTelemetry.AutoInstrumentation.Loader` assembly loaded from `__DDVoidMethodType__` static constructor.
Added helpers: `SignatureBuilder` and `MemberResolver`.
Added installer parameter to skip GAC registration
Removed LoaderOptimization=SingleDomain through registry change for ASP.Net tests.

## Tests

Added Asp.Net tests without registration assemblies in GAC. Now automatic tests validates that registry change is not required. Additonally validated that default ASP.NET template site works without issues with assemblies in GAC and not.

## Checklist


iskiselev:
@RassK suggested in previous iteration that we may need unit tests for SignatureBuilder.

src/OpenTelemetry.AutoInstrumentation.Loader/AppConfigUpdater.cs
```diff
@@ -0,0 +1,28 @@
...
+/// </summary>
+public static class AppConfigUpdater
+{
+    /// <summary>
+    /// Modify assembly bindings in appDomainSetup
+    /// </summary>
+    /// <param name="appDomainSetup">appDomainSetup to be updated</param>
+    public static void ModifyConfig(AppDomainSetup appDomainSetup)
+    {
+        appDomainSetup.LoaderOptimization = LoaderOptimization.SingleDomain;
```
iskiselev, line 25: We always force application domain to be `LoaderOptimization.SingleDomain`, as it is `LoaderOptimization.MultiDomainHost` does not work. It will be a good idea to check, if `LoaderOptimization.MultiDomain` works and not reset it in that case.
Further work would be done in phase 2, where LoaderOptimization would be changed only for non-GAC case by default.
We probably should add in phase 2/3 environment variable that would allow disable that modification, as `LoaderOptimization.SingleDomain` will work in most cases after phase 2.

test/test-applications/integrations/TestApplication.Owin.IIS.NetFramework/Dockerfile
```diff
@@ -1,14 +1,14 @@
 # escape=`
 
-FROM mcr.microsoft.com/dotnet/framework/aspnet:4.8-windowsservercore-ltsc2022@sha256:84079c734ab5aec702409ef7967ec47af9468c56ff4046882239cabacda78097
+ARG windowscontainer_version=ltsc2022
+FROM mcr.microsoft.com/dotnet/framework/aspnet:4.8-windowsservercore-${windowscontainer_version}
```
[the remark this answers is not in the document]
iskiselev: Ops. Looks like it was accidentally changed when I picked files from previous PR.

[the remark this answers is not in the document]
iskiselev: Fixed, same as next.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -1770,617 +1796,1418 @@ HRESULT CorProfiler::RunAutoInstrumentationLoader(const ComPtr<IMetaDataEmit2>&
...
+        }
+    }
+
+    return S_OK;
+}
+
+// clang-format off
+// This method will generate new type __DDVoidMethodType__ in target module.
+// C# code for created class:
+// public static class __DDVoidMethodType__
```
RassK, line 1978: `__DDVoidMethodType__` should we fix "brading here" from DD to OTel? Is it ok in the same PR or new one - I'm not exactly sure about the scope (how big is it)?

src/OpenTelemetry.AutoInstrumentation.Loader/AppConfigUpdater.cs
```diff
@@ -0,0 +1,28 @@
...
+#pragma warning disable RS0016
+#pragma warning restore RS0016
+
+/// <summary>
+/// Handles update of config files for non-default AppDomain
+/// </summary>
+public static class AppConfigUpdater
+{
+    /// <summary>
+    /// Modify assembly bindings in appDomainSetup
```
RassK, line 20: Since it's a complex call, may need to add who / where is the caller.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
iskiselev: It is small, but let me do it in separate PR right after this will be merged (or before this PR if you think it is important to not add any new DD prefixed methods).

CHANGELOG.md
```diff
@@ -11,11 +11,16 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
 
 - Support for [`Npgsql`](https://www.nuget.org/packages/Npgsql/)
   metrics instrumentation for versions `6.0.0`+.
+- In install script, Install-OpenTelemetryCore accepts optional argument RegisterAssembliesInGAC,
+  which is true by default. When set to false, assemblies would not be installed in GAC.
+- In install script, new function added: Register-AssembliesInGAC. It installs OpenTelemetry assemlbies
```
nrcventura, line 16:
```suggestion
- In install script, new function added: Register-AssembliesInGAC. It installs OpenTelemetry assemblies
```

src/OpenTelemetry.AutoInstrumentation.Loader/AppConfigUpdater.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace OpenTelemetry.AutoInstrumentation.Loader;
+
+#pragma warning disable RS0016
+#pragma warning restore RS0016
+
+/// <summary>
+/// Handles update of config files for non-default AppDomain
+/// </summary>
+public static class AppConfigUpdater
```
nrcventura, line 17: If we are calling this via reflection, does this still need to be part of our public API?

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -1770,617 +1796,1418 @@ HRESULT CorProfiler::RunAutoInstrumentationLoader(const ComPtr<IMetaDataEmit2>&
     return S_OK;
 }
 
-HRESULT CorProfiler::GenerateLoaderMethod(const ModuleID module_id, mdMethodDef* ret_method_token)
+HRESULT CorProfiler::ModifyAppDomainCreate(const ModuleID module_id, mdMethodDef patch_app_domain_setup_method)
```
nrcventura, line 1799: It would be nice to have some comments that explain what this method is trying to do, especially since it's a longer function and it is modifying IL.

src/OpenTelemetry.AutoInstrumentation.Native/member_resolver.cpp
```diff
@@ -0,0 +1,44 @@
...
+    {
+        tkResolutionScope = mdTokenNil;
+    }
+
+    if (tkResolutionScope == mdTokenNil || TypeFromToken(tkResolutionScope) == mdtTypeDef)
+    {
+        return metadaImport_->FindTypeDefByName(szName, tkResolutionScope, token);
+    }
+
+    // TODO: Prior to metadaEmit_->DefineTypeRefByName it may be good to try
```
nrcventura, line 26: Are these TODOs something that you plan to follow up on in a separate PR?

src/OpenTelemetry.AutoInstrumentation.Loader/AppConfigUpdater.cs
iskiselev: It does not need to be public, but making it private will make profiler-generated code that resolves it harder, as it will need use GetType/GetMethod overloads that acceps `BindingFlags` and load const with `BindingFlags` allowing private members.
On other hand, the type is defined in `OpenTelemetry.AutoInstrumentation.Loader`, which is not distributed separetely in .Net Framework (available only as profiler resource). Type is not defined in .Net version of assembly, so it still hard to access by any other means than reflection.
I can make it private if it's still desired.

src/OpenTelemetry.AutoInstrumentation.Native/member_resolver.cpp
iskiselev: I will remove them. Per my test CLR runtume does it for us already.

src/OpenTelemetry.AutoInstrumentation.Loader/AppConfigUpdater.cs
nrcventura: Since it is not that simple, I am fine with keeping it public. That will help keep the injected code easier to read.

src/OpenTelemetry.AutoInstrumentation.Native/member_resolver.h
```diff
@@ -0,0 +1,43 @@
...
+
+    /// <summary>
+    /// Uses IMetaDataImport::FindTypeDefByName if tkResolutionScope is mdTokenNill, TypeDef or Assembly
+    /// Uses IMetaDataEmit::DefineTypeRefByName otherwise (if tkResolutionScope is AssemblyRef or TypeRef)
+    /// </summary>
+    HRESULT GetTypeRefOrDefByName(mdToken tkResolutionScope, LPCWSTR szName, mdToken* token);
+
+    /// <summary>
+    /// Uses IMetaDataImport::FindTypeDefByName if tkResolutionScope is mdTokenNill, TypeDef or Assembly
+    /// Uses IMetaDataEmit::DefineTypeRefByName otherwise (if tkResolutionScope is AssemblyRef or TypeRef)
```
zacharycmontoya, line 36: I'm not sure if my suggestion is 100% accurate, can you follow-up here?

```suggestion
    /// Uses IMetaDataImport::FindMember if tkResolutionScope is TypeDef
    /// Uses IMetaDataEmit::DefineMemberRef otherwise (if tkResolutionScope is mdTokenNill, Assembly, AssemblyRef, or TypeRef)
```

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -1770,617 +1796,1423 @@ HRESULT CorProfiler::RunAutoInstrumentationLoader(const ComPtr<IMetaDataEmit2>&
...
-                                                              ELEMENT_TYPE_I4};
-
-    mdMemberRef interlocked_compare_member_ref;
-    hr =
-        metadata_emit->DefineMemberRef(interlocked_type_ref, WStr("CompareExchange"),
-                                       interlocked_compare_exchange_signature,
-                                       sizeof(interlocked_compare_exchange_signature), &interlocked_compare_member_ref);
-    if (FAILED(hr))
+    // TypeDef/TypeRef for System.Object
+    mdToken system_object_token;
```
zacharycmontoya, line 2079: For my education, what's the benefit of using scopes here? I'm by no means an expert at C++ so using scopes like this is not something I'm used to

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -1770,617 +1796,1423 @@ HRESULT CorProfiler::RunAutoInstrumentationLoader(const ComPtr<IMetaDataEmit2>&
...
+            {
+                Logger::Warn(
+                    "GenerateLoaderType: GetMemberRefOrDef System.Runtime.InteropServices.Marshal::Copy failed");
+                return hr;
+            }
+        }
+
+        // MethodRef/MethodDef for
+        // class [mscorlib]System.Reflection.Assembly [mscorlib]System.Reflection.Assembly::Load(uint8[], uint8[])
+        mdToken appdomain_load_member_ref;
```
zacharycmontoya, line 2382: This could be a good time to make a small correction to the variable name, up to you

```suggestion
        mdToken assembly_load_member_ref;
```

zacharycmontoya (approved):
Overall this looks very good! And the helpers you've introduced are really nice, thank you for the contribution! I've left comments/questions but consider them non-blocking

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
iskiselev: Very minor - it defines scope for variable, so that I will not mistakenly use them in other parts of that long function, help group code logically and make expand/collapse feature in IDE works, and hints that we may need some helper function to define as we do similar multistring-operation many time.

RassK (approved):
LGTM 👍 

Probably should remove the section about setting LoaderOptimization=1 also from the troubleshooting quite, if it's not applicable any more:
https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/troubleshooting.md#iis---loading-this-assembly-would-produce-a-different-grant-set-from-other-instances

src/OpenTelemetry.AutoInstrumentation.Loader/AppConfigUpdater.cs
Kielek: Can we work on this as a follow up?
Especially making method private/internal. I would like to prevent manual calls to this method.

Kielek (approved):
Thanks for the contribution. I need one more round to fully go thorough native code changes. It is not a blocker to me.

@iskiselev, please let me know if you are ready merge/ @openetelemetryopentelemetry-dotnet-maintainers feel free to click merge button if you are fine.

iskiselev:
> LGTM 👍
> 
> Probably should remove the section about setting LoaderOptimization=1 also from the troubleshooting guide, if it's not applicable any more: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/troubleshooting.md#iis---loading-this-assembly-would-produce-a-different-grant-set-from-other-instances

It is already removed in this PR: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4187/files#diff-373ebedf1806538a48b09135d964fe714cd54bd8c385e2ef97383324b759e82dL235

src/OpenTelemetry.AutoInstrumentation.Loader/.publicApi/net462/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,3 @@
+#nullable enable
```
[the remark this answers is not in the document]
iskiselev: It is not included in shipped for `OpenTelemetry.AutoInstrumentation.Loader`, so it should be included either here or in shipped. Currently `OpenTelemetry.AutoInstrumentation.Loader` has no public API.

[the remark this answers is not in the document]
iskiselev: In any case, I will remove it - as I'll make AppConfigUpdater internal.

src/OpenTelemetry.AutoInstrumentation.Loader/AppConfigUpdater.cs
iskiselev: Actually we can. I missed, that we don't need any binder for resolving internal type. So, making AppConfigUpdater interanl (still leaving method inside public).

iskiselev:
From my point of view it is ready to be merged. Looks like I applied all suggested changes.

lachmatt:
LGTM 👍, thanks @iskiselev

Issue #4132, opened 2025-04-08
Configuration based Bytecode Instrumentation


Kielek:
# Feature Request

**Are you requesting automatic instrumentation for a framework or library? Please describe.**

- Framework or library name: any/configurable
- Library type: any/configurable
- Library version: any/configurable

**Is your feature request related to a problem? Please describe.**

Possibility to instrument method without source code modification. The goal is to have it defined in some configuration file.

**Describe the solution you'd like**

Mimic solution created by @johnbley for Splunk Java distribution. https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/13590

Configuration methods and forma /using Otel Config/OpAMP support, will be adjusted to be more .NET friendly.

**Describe alternatives you've considered**

Implement feature in the Splunk distribution, but I would like to just directly donate it here.
It requires support for injecting bytecode instrumentation from plugins #2440

**Additional context**

It will be great to have it implemented and released withing ~3 next months.

pellared:
> The main question is to drop SDK or not to drop SDK. If you drop SDK, then you do the same thing twice.

Can you please elaborate? What do you mean by dropping the SDK? What things will be done twice?

PR #4205, opened 2025-05-12
Configuration based instrumentation - base functionalities

Kielek:
## Why

POC for #4132

## What

Possibility to instrument any method with <=9 o parameters by injecting configuration by the file.
~~To make the review easier and split discussion about file format separate I hardcoded some examples separately.~~
Included file based configuration.

Performance - if we decide to keep code directly in the main project, we could avoid playing with method name fetching and pass it from the native code.
Documentation + Changelog

Topics not covered, raised on Java side:
- How to enable such configuration, is it enough to have one switch to this (lack of configuration)? Alternatively, we could have plugin to contains all this stuff.
- Possibility to setting up static span-attributes,
- Possibility to setting up static span-kind,
- Possibility to calculate span-name, span-kind, and span-attributes dynamically. Some evaluator is needed to avoid security issues.

> [!WARNING]
> It will be not working for inlined methods.

## Tests

CI + Added new

## Checklist


RassK:
Seems very security concerning feature for me even by having it off by default.

Just an idea, that the config could be signed and verified by a cert installed to trusted store.

brsmith080:
For most of the applications in our environment we use newrelic's dotnet agent https://github.com/newrelic/newrelic-dotnet-agent for instrumentation. I believe this change would let us achieve something like this https://docs.newrelic.com/docs/apm/agents/net-agent/custom-instrumentation/create-transactions-xml-net/? 

if so, that would be great. We use this a lot from the newrelic agent and is one of the question marks we have while looking into moving off of the newrelic agent and its api and onto open telemetry across the board.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegration0.cs
```diff
@@ -0,0 +1,29 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+using OpenTelemetry.AutoInstrumentation.CallTarget;
+
+namespace OpenTelemetry.AutoInstrumentation.Instrumentations.NoCode;
+
+/// <summary>
+/// NoCode calltarget instrumentation
+/// </summary>
+public static class NoCodeIntegration0
```
nrcventura, line 11: I think this approach shows how no code instrumentation can be implemented, but needing to stub out these NoCodeIntegration classes for different parameter counts will limit the types of methods that can be instrumented.

I wonder if there is a way to add some indirection to the implementation. That is, all NoCode instrumentation will have the exact same injected code, instead of injecting different code to call methods with different numbers of parameters. If we can do that, I think the implementation can be simplified. If we do not need access to the method parameters, I think this indirect approach can be simple. If we do need access to the parameters we may need to leverage arrays with boxing or Tuples with boxing in order to provide access to the parameters when necessary.

Kielek: I have checked all existing bytecode instrumentation in this repository and the max number in instrumented method is 5 + returned value. It is rare to have more attributes,
I think that for the first release, if documented correctly (separate PR), it will be enough to have coverage for methods with up to 9 parameters.

src/OpenTelemetry.AutoInstrumentation/Configurations/NoCodeSettings.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        var instrumentedMethods = new List<NoCodeInstrumentedMethod>(noCodeEntries.Length);
+
+        foreach (var noCodeEntry in noCodeEntries)
+        {
+            var noCodeTarget = noCodeEntry.Target;
+
+            if (noCodeTarget == null)
+            {
+                Log.Debug("No code target option is null. Skipping this entry.");
```
RassK, line 59: Shouldn't this be warning? `this entry` would be unidentifiable marker to find the broken line.

Kielek: I didn't implemented any specific switch to enable this functionality.
It hope that it will be empty for most cases when user switched to the file based configuration.

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/NoCodeEntry.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace OpenTelemetry.AutoInstrumentation.Configurations.FileBasedConfiguration;
+
+internal class NoCodeEntry
+{
+    [YamlMember(Alias = "target")]
+    public NoCodeTarget? Target { get; set; }
+
+    [YamlMember(Alias = "span")]
+    public NoCodeSpan? Span { get; set; }
+}
```
ysolomchenko, line 15: This is a subjective opinion. The way its implemented now is also fine, since there is no specification for it, but shouldn’t the `span` be inside the `target`? After all, its kind of the output of that `target`

test/test-applications/integrations/TestApplication.NoCode/NoCodeTestingClass.cs
```diff
@@ -0,0 +1,274 @@
...
+    public void TestMethod()
+    {
+        // This method is intentionally left empty.
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public void TestMethodA()
+    {
+        // This method is intentionally left empty.
+    }
```
ysolomchenko, line 26: Are two identical methods needed for the test?

test/test-applications/integrations/TestApplication.NoCode/NoCodeTestingClass.cs
```diff
@@ -0,0 +1,274 @@
...
+    public async Task TestMethodAsync()
+    {
+        await Task.Yield();
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public async Task TestMethodAAsync()
+    {
+        await Task.Yield();
+    }
```
ysolomchenko, line 192: Here is also two identical methods

test/test-applications/integrations/TestApplication.NoCode/NoCodeTestingClass.cs
ysolomchenko, line 1: I think you should add test cases for the following types:
- string
- some Custom class

test/test-applications/integrations/TestApplication.NoCode/config.yaml
ysolomchenko, line 1: NIT: Arrange `targets` in the same order as the test methods in `test/test-applications/integrations/TestApplication.NoCode/NoCodeTestingClass.cs` and the checks from `test/IntegrationTests/NoCodeTests.cs`. Currently, they are in the same order as the checks from `test/IntegrationTests/NoCodeTests.cs`

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/NoCodeEntry.cs
Kielek: Target describes what to instrument. Span describe how to do it.
I do not see any benefits to putting spans under the target.

test/test-applications/integrations/TestApplication.NoCode/NoCodeTestingClass.cs
Kielek: Just for testing purposes. If methods with the same set of parameters can be distinguished.

test/test-applications/integrations/TestApplication.NoCode/NoCodeTestingClass.cs
Kielek: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4205#discussion_r2378943434

test/test-applications/integrations/TestApplication.NoCode/NoCodeTestingClass.cs
Kielek: I think that all this cases are in the same class equivalence class at least for current implementation.

Added to avoid potential regressions dbeb0ec83401d3430f7f3326777974d26a4e3f18

test/test-applications/integrations/TestApplication.NoCode/config.yaml
Kielek: 7d2872d678ae350f77b9870ea5969da5df9b36b5

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/NoCodeSignature.cs
```diff
@@ -0,0 +1,15 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+using Vendors.YamlDotNet.Serialization;
+
+namespace OpenTelemetry.AutoInstrumentation.Configurations.FileBasedConfiguration;
+
+internal class NoCodeSignature
+{
+    [YamlMember(Alias = "return_type")]
+    public string? ReturnedType { get; set; }
```
ysolomchenko:
```suggestion
    [YamlMember(Alias = "return_type")]
    public string? ReturnType { get; set; }
```

Kielek: fffb91a9cb7899c70d41b001aceacb87419ffd9c

lachmatt (commented):
Overall, LGTM. I think we can work on simplification mentioned in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4205/files#r2248964625 in a follow-up.

lachmatt (approved):

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegrationHelper.cs
```diff
@@ -0,0 +1,140 @@
...
+        var parameters = method?.GetParameters()!;
+        var noCodeEntry = NoCodeEntries.Single(x =>
+            x.Definition.TargetType == typeName &&
+            x.Definition.TargetMethod == methodName &&
+            x.Definition.TargetAssembly == assemblyName &&
+            CheckParameters(x.SignatureTypes, parameters));
+
+        // TODO Span kind and attributes from configuration
+        // TODO Consider execute some dynamic code to build name/span/attributes based on config and taking data from method parameters
+        var activity = Source.StartActivity(name: noCodeEntry.SpanName, kind: ActivityKind.Internal, tags: [new("TODO", "attributes-from-configuration")]);
```
nrcventura, line 35: Is this `TODO` tag a carry-over from testing?

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegrationHelper.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            var genericReturnType = returnType.GetGenericTypeDefinition();
+            if (typeof(Task).IsAssignableFrom(returnType))
+            {
+                // The type is a Task<>
+                return new CallTargetReturn<TReturn>(returnValue);
+            }
+#if NET
+
+            if (genericReturnType == typeof(ValueTask<>))
```
nrcventura, line 58: ValueTasks can be used in net framework apps if the appropriate libraries are added to the application. Why is this check in a NET compiler directive?

Kielek: In the same way we have implemented it in the standard bytecode instrumentation. I just mimic this behavior. https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/283fbb7149a3ece491c2099e66a35a307718cfe3/src/OpenTelemetry.AutoInstrumentation/CallTarget/Handlers/EndMethodHandler%601.cs#L34-L63

If needed, we should update it in both places.

Do you think that we need this at this moment/schedule some issue for follow up/vNext?

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegrationHelper.cs
Kielek: Nope, intentionally left it here, as the comments 2 lines before.
I have a plan to add this 2 missing functionalities in the follow up, if you think that we need to remove this TODO from the code, I also can do this.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegrationHelper.cs
nrcventura: I think that it will be fine to track it as a follow up issue. In practice I see ValueTask used more in library code than in applications, and I suspect that this configuration based instrumentation will be used more for application instrumentation.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegrationHelper.cs
nrcventura: I think that we only need to remove the tag if someone is going to start using this code after we merge. If it is not ready to use, I am fine with waiting until we address the TODOs in the comments.

nrcventura (approved):

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegrationHelper.cs
Kielek: For sure, it will be removed before documentation/release will be in palce

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegrationHelper.cs
nrcventura: And by follow up issue, I mean an issue that can be scheduled for a later release (vnext is fine).

test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/FileBased/Files/NoCodeFile.yaml
```diff
@@ -0,0 +1,18 @@
+file_format: "1.0-rc.1"
+
+no_code/development:
+  targets:
+    - target:
+        assembly:
+          name: TargetAssembly
+          # TODO potenatil minimal and max version
+        type: TargetType
```
zacharycmontoya: nit: I know this is a minimal example for unit tests, but should we include a namespace as well to make it more clear that this is the full name? Like change `TargetType` to `TargetNamespace.TargetType`

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/NoCodeEntry.cs
zacharycmontoya: I agree with @Kielek, target is the input while span is the output

test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/FileBased/Files/NoCodeFile.yaml
Kielek: 12960a2f72f4a1e5f8383a0149420e1df650d099

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegrationHelper.cs
Kielek: #4502

Issue #2018, opened 2023-01-20
GAC operations don't work with pwsh shell


pjanotti:
My default shell on Windows is `pwsh` that by default runs on .NET, not on .NET Framework like the `Windows Powershell`. Due to this, it fails to correctly install assemblies on the GAC since it can't find a native file needed for the operation: `fusion.dll`. This is the error message:

```PowerShell
Install-OpenTelemetryCore: Could not setup OpenTelemetry .NET Automatic Instrumentation! Exception calling "GacInstall" with "1" argument(s): "C:\Program Files\WindowsApps\Microsoft.PowerShell_7.3.1.0_x64__8wekyb3d8bbwe\fusion.dll"
```

I don't think there is much that we can do besides document the issue and make clear that the user should use `Windows Powershell` not the newer versions. Anyway, let's take a quick look when we have a chance to see if it is possible to cheaply workaround or not.

pjanotti:
At a minimum, we could use a test like `$PSVersionTable.CLRVersion -eq $null` to determine if the module is running on the proper shell or not and give the message instead of erroring out.

pellared:
We could also use `$PSVersionTable.PSEdition` which should be `Desktop`. Probable the best would be to check it using the module manifest: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_powershell_editions?view=powershell-7.3#the-compatiblepseditions-module-manifest-field

However, I feel that having it just documented could be good enough.

pellared:
Fixed per https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2078

mrwensveen:
I stumbled on this while following https://opentelemetry.io/docs/zero-code/net/getting-started/. There is no mention of Windows PowerShell, which I feel would have been useful.

Kielek:
@mrwensveen, would you like to corbibute to the documentation? I will be happy to review

martincostello:
Just ran into this myself. Have opened a docs PR to try and make it so fewer users fall into the pit of failure with this here: https://github.com/open-telemetry/opentelemetry.io/pull/7140

PR #4257, opened 2025-06-18
Docs: Make PowerShell 5.1 usage more prominent

martincostello:
## Why

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2018#issuecomment-2983383073

Relates to #2018 and https://github.com/open-telemetry/opentelemetry.io/pull/7140.

## What

Make requirement to use Windows PowerShell to install zero-code instrumentation for .NET on Windows more prominent.

## Tests

Manual check that the directive causes the script to fail in pwsh and not in PowerShell.

```pwsh
#Requires -PSEdition Desktop
echo "Success"
```

With pwsh:

```terminal
❯ .\test.ps1
.\test.ps1: The script 'test.ps1' cannot be run because it contained a "#requires" statement for PowerShell editions 'Desktop'. The edition of PowerShell that is required by the script does not match the currently running PowerShell Core edition.
```

With PowerShell:

```terminal
PS D:\Coding> .\test.ps1
Success
```

## Checklist

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -240,9 +240,17 @@ uses environment variables as parameters:
 
 On Windows, you should install OpenTelemetry .NET Automatic Instrumentation
 and instrument your .NET application using the provided PowerShell module.
+
+> [!WARNING]
+> The PowerShell module works only on PowerShell 5.1
+which is the one installed by default on Windows.
+
 Example usage (run as administrator):
 
 ```powershell
+# PowerShell 5.1 is required
+#Requires -PSEdition Desktop
+
 # Download the module
 $module_url = "https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/releases/download/v1.11.0/OpenTelemetry.DotNet.Auto.psm1"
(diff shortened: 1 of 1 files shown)


RassK:
May need to bring up again in SIG. Limiting pwsh is not ideal but in the other hands we don't do partial install currently without netfx support.

martincostello:
It's fair to aim to not have this restriction long term, but right now if the user uses the wrong edition of PowerShell then the install just fails with an error.

Adding an explicit restriction to the script surfaces the mistake immediately with an error message that tells them it doesn't work with PowerShell Core and that PowerShell Desktop is required to continue and aids to guide the user into the pit of success.

RassK:
I think it's convincing enough to go forward with this PR after also looking into https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2018. This fix seems like a proper follow up.

zacharycmontoya (approved):

Kielek (approved):

PR #4263, opened 2025-06-18
[Infra] Test with Windows Server 2025

martincostello:
## Why

Windows Server 2025 images [are now available for GitHub Actions](https://github.com/actions/runner-images/issues/11228).

Test compatibility using the new hosted runner image.

## What

Test on Windows Server 2025.

## Tests

None.

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,6 +18,8 @@ jobs:
         include:
           - machine: windows-2022
             log-dir: "/c/ProgramData/OpenTelemetry .NET AutoInstrumentation/logs"
+          - machine: windows-2025
+            log-dir: "/c/ProgramData/OpenTelemetry .NET AutoInstrumentation/logs"
           - machine: ubuntu-22.04
             log-dir: "/var/log/opentelemetry/dotnet"
           - machine: macos-13
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,8 +18,6 @@ jobs:
         include:
           - machine: windows-2022
             log-dir: "/c/ProgramData/OpenTelemetry .NET AutoInstrumentation/logs"
-          - machine: windows-2025
-            log-dir: "/c/ProgramData/OpenTelemetry .NET AutoInstrumentation/logs"
           - machine: ubuntu-22.04
             log-dir: "/var/log/opentelemetry/dotnet"
           - machine: macos-13
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,10 +52,13 @@ jobs:
       - run: cd bin-ubuntu-22.04 ; zip -qq -r ../opentelemetry-dotnet-instrumentation-linux-glibc-x64.zip . * ; cd ..
       - run: cd bin-ubuntu-22.04-arm ; zip -qq -r ../opentelemetry-dotnet-instrumentation-linux-glibc-arm64.zip . * ; cd ..
       - run: cd bin-windows-2022 ; zip -qq -r ../opentelemetry-dotnet-instrumentation-windows.zip . * ; cd ..
+      - run: cd bin-windows-2025 ; zip -qq -r ../opentelemetry-dotnet-instrumentation-windows.zip . * ; cd ..
       - run: cd bin-macos-13 ; zip -qq -r ../opentelemetry-dotnet-instrumentation-macos.zip . * ; cd ..
       - run: cd bin-nuget-packages ; zip -qq -r ../opentelemetry-dotnet-instrumentation-nuget-packages.zip . * ; cd ..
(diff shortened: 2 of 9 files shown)


.github/workflows/format-native.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        machine: [ windows-2022, ubuntu-22.04, macos-13-xlarge ]
+        machine: [ windows-2025, ubuntu-22.04, macos-13-xlarge ]
```
RassK: note: seems here all images could use the latest version

.github/workflows/release.yml
```diff
@@ -52,10 +52,13 @@ jobs:
       - run: cd bin-ubuntu-22.04 ; zip -qq -r ../opentelemetry-dotnet-instrumentation-linux-glibc-x64.zip . * ; cd ..
       - run: cd bin-ubuntu-22.04-arm ; zip -qq -r ../opentelemetry-dotnet-instrumentation-linux-glibc-arm64.zip . * ; cd ..
       - run: cd bin-windows-2022 ; zip -qq -r ../opentelemetry-dotnet-instrumentation-windows.zip . * ; cd ..
+      - run: cd bin-windows-2025 ; zip -qq -r ../opentelemetry-dotnet-instrumentation-windows.zip . * ; cd ..
```
RassK: is there a need for 2025 publish separately?

martincostello: > Open question on which changes should be additional vs. replacing 2022. Possibly a mix of both.

I'm not sure yet. I figured I'd start with just adding 2025, see what broke if anything, and the update anything based on whether the tests should upgrade from 2022 to 2025, or _add_ 2025.

martincostello: Looks like nothing fails on 2025 at least!

.github/workflows/format-native.yml
martincostello: Will update. The required statuses will need changing.

I'll do that in a separate PR and decouple them from the specific version (e.g. `windows` not `windows-2022`).

martincostello: #4266

.github/workflows/build.yml
```diff
@@ -18,6 +18,8 @@ jobs:
         include:
           - machine: windows-2022
             log-dir: "/c/ProgramData/OpenTelemetry .NET AutoInstrumentation/logs"
+          - machine: windows-2025
+            log-dir: "/c/ProgramData/OpenTelemetry .NET AutoInstrumentation/logs"
```
Kielek: I think that it is not necessary, We are not shipping artifacts build on windows-2025 . Further Windows-2025 steps should based on the windows-2022 binaries. Other parts, LGTM

martincostello: Thanks - I was trying to work out exactly what was and wasn't needed and figured failing builds would find them.

martincostello:
```suggestion
```

martincostello: Looks like the ci and build jobs are tied together and the test assumes that there's a corresponding build leg.

For the aim of _"test that things work on a Windows 2025 machine"_, which parts need to be updated? Either it needs building on an extra platform so it can be tested on it, or there's some assumptions that need breaking.

Kielek: I think that it should be possible to take pre-build dependencies from other runners. It will require some configuration adjustments (new property/configuration field.

Kielek: Not tested., You could try something like this. Similar changes are probably needed for other failing steps.. Maybe you have better name for the new field?

```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index abdf79da8..caaa25098 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,14 +45,19 @@ jobs:
             machine: ubuntu-22.04-arm
         include:
           - machine: windows-2022
+            source-artifacts: windows-2022 
             containers: none
           - machine: windows-2025
+            source-artifacts: windows-2022
             containers: none
           - machine: ubuntu-22.04
+            source-artifacts: ubuntu-22.04
             containers: linux
           - machine: macos-13
+            source-artifacts: macos-13
             containers: none
           - machine: ubuntu-22.04-arm
+            source-artifacts: ubuntu-22.04-arm
             containers: linux
     runs-on: ${{ matrix.machine }}
     steps:
@@ -79,7 +84,7 @@ jobs:
       - name: Download Artifacts from build job
         uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # tag: v4.3.0
         with:
-          name: bin-${{ matrix.machine }}
+          name: bin-${{ matrix.source-artifacts }}
           path: bin/tracer-home
 
       - name: Install SQL Server (localdb)
```

Kielek (approved):

RassK (approved):

PR #4275, opened 2025-06-26
[internal-logging] remove flush to disk after write

### LINKED ISSUE #4261: [internal-logging] flush to disk after every write
# Bug Report

## Symptom

Currently, flush to disk is called after every write when `FileSink` is used. This is inefficient and not needed.

`FileSink` with flush to disk after write was initially used in Loader, where logging volume was low and flushing to disk after write [made sense](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/e81ae556e3e2297e08c2cf6cd73a1c3c5360a632/src/OpenTelemetry.AutoInstrumentation.Loader/LoaderLogger.cs#L30-L34). It was [later reused](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/456) for logging in other places.

Autoinstrumentation usually logs at startup (e.g. default log level of `info`, no warnings/errors), so inefficiency in logger will increase the startup time (I counted 16 calls in default configuration based on a quick smoke test).
Problem will be more severe when `debug` log level is used, when additional events from SDK event source are logged to a file (e.g. related to activity starting/stopping).

## Possible solution

Explicit flush to disk could instead be ran periodically, e.g. as in [Serilog's code](https://github.com/serilog/serilog-sinks-file/blob/53280f51d74deaa3e1bcde686a1653737832422d/src/Serilog.Sinks.File/Sinks/File/PeriodicFlushToDiskSink.cs).

lachmatt:
## Why

Reuses the solution from `Serilog` - writes are flushed to disk based on configured interval,  using a `Timer`.
Loggers are closed explicitly when no longer needed, or closed when application exits.

Note: disposal of a file sink by default doesn't trigger an immediate flush to disk. Let me know if you think this should be changed.

After logger with given suffix is closed:
- consecutive attempts to get logger from `OtelLogging` with the same suffix will result in `NoopLogger` being returned (which does nothing)
- consecutive attempts to log with the previously obtained instance will result in log message being forwarded to a `NoopSink` (but there will be some overhead, e.g. formatting a log message and writing to event source)

I think above is acceptable considering there shouldn't be any such attempts if loggers are closed at the right time.

I used a flush to disk interval of `5s`, but I'm open for suggestions. If we were to force a flush to disk when closing the logger, we could probably use a bigger interval. We could also make it configurable, in order for it to be decreased if needed in given scenario.
Most of the loggers should be closed early (e.g. `StartupHook`, `Loader` if `Debug` is not enabled, `AspNetCoreBootstrapper`).
The longest living logger will be `Managed` logger, closed on app exit.

Fixes #4261 

## What

Removes flush to disk after every write.
Uses `PeriodicFlushToDiskSink` vendored from `Serilog`.

## Tests

Verified manually. 
Automated tests added in the PR.

## Checklist


lachmatt:
There was a test [failure](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/15925472707/job/44922226905?pr=4275#step:7:892) which seemed possibly related to the changes from the PR, but I haven't found a connection yet. I also have a hard time reproducing the issue locally. Rerun helped. Let me know if you think root cause of that failure has to be found before proceeding with this PR, or should I create a separate issue for a possibly a flaky test.

Kielek (approved):

Kielek:
@nrcventura, could you please double check changes?

From internal communication with @lachmatt, the failing tests cases is related to the wrong dispose in the https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/b6eb9e3c4e24af0e84b5792ea9640faabe595860/test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/PluginManagerTests.cs#L124, will be fixed in separate PR.

lachmatt:
Thanks for the feedback @nrcventura 

> Should we introduce a disposable logger type that can handle the flushing logic by simply declaring the logger with a using keyword? Something like using var disposableLogger = GetLogger("LoggerName");

I think it would be straightforward to implement in simple scenarios (e.g. `BootstrapperHostingStartup`) but more problematic in others (e.g. `Managed` logger). 
At the moment, `OtelLogging` ensures there is single logger instance for given suffix (i.e. to a given file), so logger can be disposed only after it's last user finished using it.
`Managed` logger will live for the lifetime of the app, so would be disposed (and flushed) only on the app exit.

> Would using https://github.com/serilog/serilog-sinks-async be simpler and not require us to handle the manual flushing to disk using the try-finally blocks or a disposable type?

I looked a little bit into it, and my understanding is that after changing from flushing to disk after each write, to simply flushing should be fast enough. Based on a simple benchmark, simple flush is 2 orders of magnitude faster than flush to disk. It is also a [default configuration](https://github.com/serilog/serilog-sinks-file/blob/23c732a8658a0df2a5434fe69b0011800b14f0da/src/Serilog.Sinks.File/Sinks/File/FileSink.cs#L125-L126) used when using `FileSink` with `Serilog`.

It is my understanding that flushing to disk on some interval might still be beneficial in some cases (as described briefly [here](https://nblumhardt.com/2016/10/october-serilog-update/)).

Let me know if that makes sense, I can create a follow-up PR implementing suggestions if you disagree.

nrcventura:
> > Would using https://github.com/serilog/serilog-sinks-async be simpler and not require us to handle the manual flushing to disk using the try-finally blocks or a disposable type?
> 
> I looked a little bit into it, and my understanding is that after changing from flushing to disk after each write, to simply flushing should be fast enough. Based on a simple benchmark, simple flush is 2 orders of magnitude faster than flush to disk. It is also a [default configuration](https://github.com/serilog/serilog-sinks-file/blob/23c732a8658a0df2a5434fe69b0011800b14f0da/src/Serilog.Sinks.File/Sinks/File/FileSink.cs#L125-L126) used when using `FileSink` with `Serilog`.
> 
> It is my understanding that flushing to disk on some interval might still be beneficial in some cases (as described briefly [here](https://nblumhardt.com/2016/10/october-serilog-update/)).

The intent of my question is about determining if it simplifies the code that we need to maintain. It feels awkward to need to require a try-finally block to ensure that our log messages are persisted. My hope, without knowing the implementation details, is that using something like `async` sink could help with these scenarios and let the separate thread automatically handle the flushing to disk. If leveraging that sink does not help with simplifying the code then we probably should not use it. I do not know if adding that sink will improve the performance enough to use it in addition to the PeriodicFlushToDisk sink.

PR #4277, opened 2025-06-30
Consume .NET installation script from repo

### LINKED ISSUE #4256: Update dotnet-install verification step
Many of our dockerfiles in the /dockerfile folder have a step where we install dotnet by downloading the install-dotnet shell script, and then comparing its SHA to a known good SHA. These steps are done primarily for change notification tracking. We should document why we have these steps in the dockerfile along with the process for updating the sha.

1. Review the output of the failing run to grab updated sha from the output.
2. Update the dockerfiles to use the new SHA

In addition to this, we can add a verification step to these dockerfiles that validates the signature of the script.

We can add the public signing keys to our repo, and follow the steps in https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script#signature-validation-of-dotnet-installsh to validate the signature of the script. We just need to skip the step of downloading the signature verification key, because it will already be in the repo.

Then when the signature verification step fails, we can see if it was due to a problem, or the key expiring. If there's a problem, then we can't trust the script and should not run it. If the key expired we need to get the updated key. 

When we need to update the key we should require someone to take the following steps. These steps should also be documented.

1. Download the install-dotnet script and key
2. Review the contents of that script to verify that it is safe to use.
3. Verify the signature of the script using the downloaded key
4. Update our repo to use the new key

martincostello:
## Why

Fixes #4256

Avoid build-breaks caused by updates to the .NET install scripts (e.g. #4262) that need manual updates to the SHA.

## What

~~Use the steps described in [_Signature validation of dotnet-install.sh_](https://learn.microsoft.com/dotnet/core/tools/dotnet-install-script#signature-validation-of-dotnet-installsh) to verify the GPG signature of the .NET install script.~~

Add the .NET installation script to the repository to use in the Dockerfiles, plus a GitHub Actions workflow that checks for updates on a monthly basis and raises a pull request if there are any changes.

## Tests

None.

## Checklist

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -20,4 +20,4 @@ jobs:
         run: sudo apt update && sudo apt install --assume-yes shellcheck
 
       - name: Run shellcheck
-        run: find . -name \*.sh -o -name \*.sh.template  | xargs shellcheck --severity=warning
+        run: find . -name \*.sh -o -name \*.sh.template | grep -v scripts/dotnet-install.sh | xargs shellcheck --severity=warning


Kielek (approved):
💯

Kielek:
@RassK bring one comment in private channel.
It will ensure that it is published by MS, but it is not longer _pinned_ to the exact version. We might need both.

martincostello:
I'm not aware that you can pin when it's sourced from `dot.net` - it's a single file resource that's unversioned (as far as the consumer is concerned) that's whatever version they tested/validated and deployed. If there is a way to do that, then would need to know how.

If you want it to be pinned, then you'll need to consume it from source in GitHub (which isn't signed) using the Git SHA, e.g. `https://raw.githubusercontent.com/dotnet/install-scripts/46016b203522f96044b82a9dd8a52cfa8a92e216/src/dotnet-install.sh`

The options are:

1. Stay current and validate the signature. No updates required.
2. Switch to the unsigned file in GitHub and pin by SHA. Needs to be manually updated.
3. Do nothing, the CI breaks whenever an updated script is deployed and someone has to manually fix it.

RassK:
See https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/4256#issuecomment-2984241856
The sha itself is a "version" and protects from auto bumps. Not sure currently in which document it was stated to use for version pinning.

martincostello:
That's the source of the CI instability - if they update the script hosted at `https://dot.net/v1/dotnet-install.sh` then the CI breaks until someone goes and manually updates the Dockerfiles.

If you want to pin, then that's not hashing the content and validating it matches the hard-coded digest because `https://dot.net/v1/dotnet-install.sh` is not immutable by design - it's evergreen. That's why I believe they document validating the signature rather than the content, because otherwise you'd have the problem I'm trying to solve here of the rug being pulled from under you and breaking your automation.

For stability, the file needs to be gotten from the immutable location in GitHub (unless there's an undocumented way to do it via `dot.net`) - `https://raw.githubusercontent.com/dotnet/install-scripts/46016b203522f96044b82a9dd8a52cfa8a92e216/src/dotnet-install.sh`. That file however isn't signed.

The aim of this PR was to remove the instability. If you want pinning I can switch to use the content from GitHub, but I don't think you can have both because the script gets signed by an opaque (to non-Microsoft people) process before it gets served from `dot.net`.

RassK:
All correct, just the breaking was intentional so far. The idea was to avoid auto dependency bumps. The script is unfortunate one because it points to the latest path.

martincostello:
> The idea was to avoid auto dependency bumps.

I'm curious as to what the "best" outcome expected is.

- If it changes from under you and you validate the SHA, then your CI breaks.
- If it changes from under you and you validate the signature, then the CI would only break if there was a bug in the script after a change. Apart from in the case of an _extremely_ severe security incident with the CDN and signing key being compromised, if the script is signed and served from .NET's CDN, then it's an intentional and (presumably) validated change on the part of the .NET team.

I would have thought a _possible_ break is better than a definite break.

Switching to the GitHub raw URL would be a more reliable way to have pinning IMHO as it won't break without prior notice.

Alternatively, you could just copy the script into this repo directly and have its acquisition entirely "offline". That's what [actions/setup-dotnet does](https://github.com/actions/setup-dotnet/tree/main/externals).

If you still want a manual process to update, I think it's better to have one that doesn't break the CI.

pellared:
> Alternatively, you could just copy the script into this repo directly and have its acquisition entirely "offline". That's what [actions/setup-dotnet does](https://github.com/actions/setup-dotnet/tree/main/externals).

> If you still want a manual process to update, I think it's better to have one that doesn't break the CI.

I think this is the way. We can have some automation for updating the script.

RassK:
👍 having a local copy also needs hashing and gpg validation only once, so small optimization also.
👎 nothing any more notifies about the version changes.

pellared:
> nothing any more notifies about the version changes.

I think we would need to add some scheduled GitHub Action workflow that creates a PR once a new version of the script is published.

martincostello:
Automation would cover the 👎 - a GitHub Actions workflow can run on a schedule of some period and create a PR if the content no longer matches that checked into the repo.

martincostello:
Here's an example of such a workflow: https://github.com/martincostello/costellobot/blob/main/.github/workflows/update-bank-holidays.yml - content is downloaded from a URL and if it doesn't match the content in the repo, a PR is created for manual review to update it.

RassK:
That seems ever better than I had in my mind. Need to make sure that we didn't strip write rights from the workflow. Ideally random branches write (only main / long running branches should be protected) + PR should be fine.

martincostello:
Ideally there's a GitHub app's secrets in the repo with write permissions to open the PR as, otherwise, if GITHUB_TOKEN is used CI won't run when the PR is opened and a human has to close and re-open it to trigger it (which is why the example doesn't).

pellared:
This can be helpful: https://github.com/open-telemetry/community/blob/main/assets.md#otelbot

RassK:
yes forgot about CLA, so now there should be all missing pieces?

martincostello:
I'll take a look tomorrow to rework this into the in-repo + update workflow approach. As the workflow would need merging to test and I have to work from a fork it may take a few PRs to get fully working.

scripts/dotnet-install.sh
martincostello, line 1: This is intentionally [this previous version](https://github.com/dotnet/install-scripts/blob/2d1e2e7bc9fa1caa8e32d16531f1081756fdfb89/src/dotnet-install.sh) of the script, not [the latest](https://github.com/dotnet/install-scripts/blob/46016b203522f96044b82a9dd8a52cfa8a92e216/src/dotnet-install.sh), so post-merge the workflow can be run manually to verify it works and produces a pull request.

martincostello:
Could someone run [update-dotnet-install-script](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/workflows/update-dotnet-install-script.yml) manually to check it works?

Kielek:
> Could someone run [update-dotnet-install-script](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/workflows/update-dotnet-install-script.yml) manually to check it works?

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/16001051804/job/45135917802

martincostello:
> Failed to create token for "opentelemetry-dotnet-instrumentation" (attempt 1): The permissions requested are not granted to this installation.

Looks like `otelbot[bot]` doesn't have contents and/or pull request write access to this repo.

Kielek:
Whats about account used in this PR? https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4280

martincostello:
I can look to change it next time I'm at my laptop, unless the bot _should_ have those permissions and it just needs updating to have the right access.

RassK:
Does anyone see if secrets (`OTELBOT_PRIVATE_KEY`) are setup for the bot? I can see the secrets tab but it's empty, I guess maintainers may see more?
CC: @Kielek 

edit: the same with var `OTELBOT_APP_ID`

```
    - name: Generate GitHub application token
      id: generate-app-token
      uses: actions/create-github-app-token@df432ceedc7162793a195dd1713ff69aefc7379e # v2.0.6
      with:
        app-id: ${{ vars.OTELBOT_APP_ID }}
        permission-contents: write
        permission-pull-requests: write
        private-key: ${{ secrets.OTELBOT_PRIVATE_KEY }}
```

martincostello:
According to [the docs](https://github.com/open-telemetry/community/blob/main/assets.md#otelbot) these are organization-level secrets and vars.

I think the app just needs to be given permissions to access this repo as I'm _assuming_ it hasn't been granted access to all repos in the org by default.

trask:
Check out https://github.com/open-telemetry/community/issues/2763

PR #4337, opened 2025-07-25
fix: Refactor AssemblyResolve callbacks to avoid potential deadlock

### LINKED ISSUE #4269: Deadlock: AssemblyResolve_ManagedProfilerDependencies should be moved out of Loader class
This issue concerns the OpenTelemetry.AutoInstrumentation.Loader.Loader class. It has a static constructor, which means that while the static constructor is executing, no threads can enter any of the class's static methods. This can lead to a deadlock due to the presence of an assembly loading hook: AssemblyResolve_ManagedProfilerDependencies().

Deadlock Scenario:
1. The main thread enters the static constructor. The assembly resolve hook is installed, and additional initialization work begins, which may involve loading plugins.

2. Some of the initialization code within the static constructor spawns a new thread.

3. The main thread waits for the newly spawned worker thread to complete.

4. The worker thread attempts to load a new assembly, triggering the AssemblyResolve_ManagedProfilerDependencies() hook.

5. The hook is blocked because it requires access to the static context of the Loader class, which is still executing on the main thread.

6. Deadlock occurs.

zacharycmontoya:
## Why

Fixes #4269

## What

This moves the AssemblyResolve callbacks from being stored in the `OpenTelemetry.AutoInstrumentation.Loader.Loader` class and places them in the `OpenTelemetry.AutoInstrumentation.Loader.AssemblyResolver` class.

Currently, the static constructor of the `OpenTelemetry.AutoInstrumentation.Loader.Loader` type is responsible for both registering the AssemblyResolve callbacks AND executing all of the SDK initialization code, before returning. In the referenced issue, the SDK initialization initiates an HTTP call which then triggers instrumentation and, ultimately, an assembly resolve event. This requires a secondary thread to execute the callback methods stored in the `OpenTelemetry.AutoInstrumentation.Loader.Loader` type, but this creates a deadlock because the first thread still hasn't finished the SDK initialization. This solves the problem by avoiding all reentrancy into the Loader type.

An alternative approach, should this approach not work, is to move the SDK initialization into a static method of the `Loader` type rather than the static constructor. This would prevent the runtime from holding a lock on the `Loader` type since it will have already been successfully initialized. However, this would require updating the IL instructions that we emit into the process in our native components, so I'd rather test a managed only change first.

## Tests

The existing integration tests serve as a regression test that the existing functionality is not affected. This will be tested manually, and if we can find a deterministic way to test this in a test application we can add that to our integration tests as well.

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyResolver.Net.cs
+++ b/src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyResolver.Net.cs
@@ -10,50 +10,13 @@ namespace OpenTelemetry.AutoInstrumentation.Loader;
 /// <summary>
 /// A class that attempts to load the OpenTelemetry.AutoInstrumentation .NET assembly.
 /// </summary>
-internal partial class Loader
+internal partial class AssemblyResolver
 {
     internal static System.Runtime.Loader.AssemblyLoadContext DependencyLoadContext { get; } = new ManagedProfilerAssemblyLoadContext();
 
     internal static string[]? StoreFiles { get; } = GetStoreFiles();
 
-    private static string ResolveManagedProfilerDirectory()
-    {
-        string tracerFrameworkDirectory = "net";
-        string tracerHomeDirectory = ReadEnvironmentVariable("OTEL_DOTNET_AUTO_HOME") ?? string.Empty;
-
-        return Path.Combine(tracerHomeDirectory, tracerFrameworkDirectory);
-    }
-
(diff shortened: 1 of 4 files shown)


lachmatt:
I noticed an unrelated error in tests related to frequent sampling, I reran the test and will investigate the issue.

Kielek (commented):
@zacharycmontoya, as the issue was reported by external user, it will be great to include information in the CHANGELOG.

Kielek:
Closing. It will be fixed in scope of #4431

PR #4359, opened 2025-08-01
[frequent-sampling] resolve symbols outside of suspension

lachmatt:
## Why

Fixes: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/4302

Reduce time spent when runtime is paused.

## What

Limit the work done when capturing stack snapshots when runtime is suspended.
Capture only `FunctionIdentifier`s which are then used to resolve frames outside of runtime suspension.
Utilizes a temporary buffer.

Approach to locking stays the same as [previously](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/6f753a8b90445dfbf85ef14c763b3b2bc9aa76b8/src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp#L44-L75).

For [simple application](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/tree/main/examples/playground/AspNetCoreMvc) in idle state (e.g. 1rps) suspension time is decreased by 10-15%.

## Tests

Existing tests.

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
@@ -706,7 +706,7 @@ static HRESULT __stdcall FrameCallback(_In_ FunctionID         func_id,
     // Selective sampling is expected to be much more frequent, e.g. every 20ms compared to 10s for continuous
     // profiling.
     const unsigned ratio = prof->threadSamplingInterval.value() / prof->selectedThreadsSamplingInterval.value();
-    if (iteration % ratio != 0)
+    if (iteration != ratio)
     {
         return SamplingType::SelectedThreads;
     }
@@ -913,6 +913,12 @@ static void PauseClrAndCaptureSamples(ContinuousProfiler*                       
     // I don't have any proof but I sure hope that if suspending fails then it's still ok to ask to resume, with no
     // ill effects
     hr = info12->ResumeRuntime();
+
+    const auto end            = std::chrono::steady_clock::now();
+    const auto elapsed_micros = std::chrono::duration_cast<std::chrono::microseconds>(end - start).count();
+
+    prof->stats_.micros_suspended = static_cast<int>(elapsed_micros);
+
     if (FAILED(hr))
     {
         trace::Logger::Error("Could not resume runtime? HRESULT=0x", std::setfill('0'), std::setw(8), std::hex, hr);
@@ -928,10 +934,6 @@ static void PauseClrAndCaptureSamples(ContinuousProfiler*                       
         return;
     }
 
-    const auto end            = std::chrono::steady_clock::now();
-    const auto elapsed_micros = std::chrono::duration_cast<std::chrono::microseconds>(end - start).count();
-
-    prof->stats_.micros_suspended = static_cast<int>(elapsed_micros);
     prof->stats_.num_threads      = static_cast<int>(size);
 
     if (samplingType == SamplingType::Continuous)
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -919,39 +969,46 @@ unsigned int GetSleepTime(const ContinuousProfiler* const prof)
...
+
+        const auto samplingType = GetNextSamplingType(prof, iteration);
+
+        PauseClrAndCaptureSamples(prof, info12, samplingType, threadStacksBuffer);
+
+        // In order to avoid the need to recreate the vectors each iteration,
+        // for the most of the iterations, instead of clearing the map, clear only the vectors.
+        if (iteration == 100)
+        {
+            threadStacksBuffer.clear();
```
lachmatt, line 1039: Map init can be improved by reserving the capacity for the threads already being tracked (e.g. in `managed_tid_to_state_`)

nrcventura (approved):

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.h
```diff
@@ -35,6 +35,64 @@ extern "C"
...
+{
+    std::size_t operator()(const continuous_profiler::FunctionIdentifier& k) const noexcept
+    {
+        using std::hash;
+        using std::size_t;
+        using std::string;
+
+        const std::size_t h1 = std::hash<mdToken>()(k.function_token);
+        const std::size_t h2 = std::hash<ModuleID>()(k.module_id);
+
```
eftiquar, line 75: We also need to include bool is_valid flag. Two equal objects should produce same hash. If two objects with same token and module ID but different bool member will produce same hash.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.h
```diff
@@ -35,6 +35,64 @@ extern "C"
...
+    FunctionID  function_id;
+    COR_PRF_FRAME_INFO frame_info;
+
+    bool operator==(const FunctionIdentifierResolveArgs& p) const
+    {
+        return function_id == p.function_id && frame_info == p.frame_info;
+    }
+};
+}
+
```
eftiquar, line 63: The hash needs to be revisited 
1. Simple xor will not give good distribution. 
2. Also, XOR being commutative following two will produce same hash value FunctionIdentifier{token=0x06000123, moduleId=0x00000001, is_valid=true}
FunctionIdentifier{token=0x00000001, moduleId=0x06000123, is_valid=true}
2. Ignoring is_valid will also produce collisions 
3. We can borrow boost's hash impl

eftiquar: Something like this 

template <typename... Args>
std::size_t hash_combine(const Args&... args) {
    std::size_t seed = 0;
    
    (..., (seed ^= std::hash<Args>{}(args)+0x9e3779b9 + (seed << 6) + (seed >> 2)));
    return seed;
}

Uses left shift, right shift to prevent commutative collisions. That magic number is, well math magic to make hash values well spread.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.h
lachmatt: This is an existing code that had to be moved. Let me try to double check if the flag was omitted here by mistake or intentionally.

PR #4391, opened 2025-08-14
Bump .NET SDK to 9.0.304/8.0.413

RassK:
## What

Bump dotnet sdk from 9.0.303 to 9.0.304
Bump dotnet sdk from 8.0.412 to 8.0.413


.github/workflows/build-nuget-packages.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         with:
           dotnet-version: | 
             8.0.412
```
Kielek: 8.0.413 was released in the same day. Could you please update 8.0.* also?

Issue #4394, opened 2025-08-19
Implement File-Based Configuration


ysolomchenko:
## Implement File-Based Configuration in YAML format
https://opentelemetry.io/docs/specs/otel/configuration/#declarative-configuration

# 
This would have to be split into several issues.

Kielek:
Base functionality implemented. Other improvements can be fixed in the further PRs/are tracked by separate issues.

PR #4403, opened 2025-08-19
Vendor YamlDotNet package

### LINKED ISSUE #4395: Vendor YamlDotNet package
### Description
We need to vendor a YamlDotNet for the File-Based config to avoid dependency conflicts and ensure builds on .NET Framework 4.6.2.

### Acceptance Criteria

- A `net462` target framework is added.
- The project builds successfully without errors.

ysolomchenko:
Fixes #4395
Towards #4394 

## What

Vendor YamlDotNet package for file-based configuration

## Checklist

- [x] A net462 target framework is added.
- [ ] The project builds successfully without errors.

- [ ] ~`CHANGELOG.md` is updated.~
- [ ] ~Documentation is updated.~
- [ ] ~New features are covered by tests.~


RassK:
I would not prefer source based vendoring, this already looks more like a fork + edits, which makes it even worse.

zacharycmontoya:
This is the first time we've considered source-vendoring a library, so I wonder why this was decided instead of referencing the library and trying to isolate the assembly loading.

For this proposed source-vendoring approach, I think it's looking good so far. For me to approve, I would like to see docs page added that documents the steps that were done to vendor the library and a list of vendored libraries (e.g. YamlDotNet) and at what version/tag they were vendored from. Ideally we would have a script that would do most of the steps so that we could limit the amount of manual steps needed, but as long as the manual steps are documented I think it will be sufficient.

Kielek:
@zacharycmontoya, there are 2 reasons for vendoring this code
1. I think that it is very first package binged directly by auto-instrumentation, instead of indirectly dependency from the `OpenTelemetry*` packages.
2. YamlDotNet is targeted for .NET Framework 4.7+ and .NET Standard 2.0+. @nrcventura mentioned that such libraries can cause issues during runtime.

I agree that we need some level of automation for vendoring/updating such packages. I think that it can be deferred for further PRs/improvements. `YamlDotNet` is pretty stable, last changes was applied 6+ months ago.

LICENSE
```diff
@@ -282,4 +282,9 @@ Copyright (c) .NET Foundation and Contributors under the MIT License (MIT)
 Component Serilog compiled into OpenTelemetry.AutoInstrumentation, OpenTelemetry.AutoInstrumentation.Loader, OpenTelemetry.AutoInstrumentation.StartupHook
 is under the following copyright:
 Copyright 2013-2017 Serilog Contributors under Apache License Version 2.0
-(<https://github.com/serilog/serilog-sinks-file/blob/main/LICENSE>).
\ No newline at end of file
+(<https://github.com/serilog/serilog-sinks-file/blob/main/LICENSE>).
+
+Component YamlDotNet compiled into OpenTelemetry.AutoInstrumentation
+is under the following copyright:
+Copyright 2013-2017 Serilog Contributors under the MIT License (MIT)
```
rajkumar-rangaraj: Should this be Serilog Contributors or Copyright (c) Antoine Aubry and contributors

Kielek: I will fix it before merge

docs/internal/vendored-code.md
```diff
@@ -0,0 +1,38 @@
...
+This means that source code from external repositories is copied into the `/src/OpenTelemetry.AutoInstrumentation/Vendors/{LibName}`
+folder.
+
+To make vendored code fully internal, apply the following changes:
+
+1. Prefix all namespaces with `Vendors.` (e.g., change `YamlDotNet` to `Vendors.YamlDotNet`).
+2. Update all using directives to reference the vendored namespaces
+   (e.g., change `using YamlDotNet` to `using Vendors.YamlDotNet`).
+3. Mark whole public contract as internal.
+4. To suppress static analysis and enable nullable reference types,
```
rajkumar-rangaraj: Can we add a Directory.Build.props file at the vendored folder level to suppress the error instead of updating multiple files?

LICENSE
Kielek: ef7ecce7222f62b5a1400befad5d3ea1ca3eeadd

docs/internal/vendored-code.md
Kielek: Good call. Partially fixed in 0bc08816ae364e4794d006042da8c37264df23c5

We still have manual modification of the files (usings/namespace), but there is no need to put the custom headers.

I do not see any possibility to avoid changes in public contract, I would like to avoid also keeping original namespaces.

martincostello:
> I agree that we need some level of automation for vendoring/updating such packages.

I've seen some prior art for dealing with this in the .NET repos (e.g. they vendor Newtonsoft.Json) where they pull the code in as-is, then use `.patch` files to apply customisations on top so things are easier to keep in sync.

rajkumar-rangaraj (approved):

Kielek:
@martincostello, I will create follow up issue for this. It will be great if you can share example.

The alternative (some auto-transkfrmation) approach is in Datadog agent repository.

martincostello:
@Kielek Examples can be found in the [dotnet/source-build-reference-packages](https://github.com/dotnet/source-build-reference-packages) repo.

1. [Vendored libraries as submodules](https://github.com/dotnet/source-build-reference-packages/tree/main/src/externalPackages/src) including Newtonsoft.Json
2. [Custom project file to build it](https://github.com/dotnet/source-build-reference-packages/blob/4565124516b2a0fc895587bf8b42bec51fb4e99e/src/externalPackages/projects/newtonsoft-json.proj)
3. [Example `.patch` file](https://github.com/dotnet/source-build-reference-packages/blob/main/src/externalPackages/patches/newtonsoft-json/0001-Newtonsoft.Json-project-updates.patch)

PR #4405, opened 2025-08-20
[continuos-profiler] special handling of activity stoppage

lachmatt:
## Why && What

Thread to span context association on the native side is tracked based on [async local change notifications](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/5e2cf883f9fff06957f4cc498a3ebc255cd1f2da/src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/SampleExporter.cs#L71-L88).
This poses a problem, when in the context of activity (e.g. handling a request in a web service), some additional work is started on a supporting thread (by e.g. running Task.Run). In such cases, execution context will flow to the supporting thread (and span context with it), but might never be cleared on that supporting thread.
For a real world example, see [microservices from OTel .NET SDK repository](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/examples/MicroserviceExample/WebApi/Controllers/SendMessageController.cs) - controller has dependency on  `MessageSender` which starts connection in it's ctor, and `MessageSender` is injected through DI container. This results in `MessageSender` instance being created as a part of handling a request, with activity already started. This results in execution context (containing current activity) eventually flowing into work queued to the thread pool. In such case, span context associated with the thread pool thread won't be cleared.
This PR attempts to compensate for such scenarios by adding a special handling for activity stoppage. When stoppage is detected, native side of profiler is notified, and span context is cleared for all the threads that were previously associated with the activity that was stopped.

## Tests

Test application is based on template, and mimics the situation from the microservices example.

## Notes
In order to optimize handling of activity stoppage, I added a map that tracks span context to thread association (so that all the threads associated with activity being stopped can be efficiently identified).
Tests for one of the runs failed with `pure virtual method called` error, but I haven't found an issue with changes from this PR specifically.
It seems it might be connected with incorrect handling of shutdown in sampling loop, will create a separate issue for investigation.


src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -86,6 +86,8 @@ static std::vector<unsigned char> selective_sampling_buffer;
 
 static std::mutex                                                             thread_span_context_lock;
 static std::unordered_map<ThreadID, continuous_profiler::thread_span_context> thread_span_context_map;
+static std::unordered_map<continuous_profiler::thread_span_context, std::unordered_set<ThreadID>>
+    span_context_thread_map;
```
lachmatt: I will extract it to a class.

Kielek (approved):

PR #4420, opened 2025-08-27
[continuous profiler] handling shutdown

### LINKED ISSUE #4416: [continuous-profiler] notify continuous sampling loop of profiler shutdown
Continuous sampling loop should be notified of profiler shutdown.
See failure in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/17124220077/job/48572853099

lachmatt:
## Why

Fixes #4416

## What

Notify continuous profiler when profiler is notified of shutdown.
When handling shutdown in continous profiler:
- set a future value to notify thread sampling loop of shutdown
- wait to acquire exclusive lock on `profiling_lock` so that any thread/allocation sampling in progress finishes
- set shutdown requested flag
- stop allocation sampling session, if one was started
- wait for sampling loop to finish

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
@@ -1355,7 +1355,7 @@ void ContinuousProfiler::StopAllocationSampling()
     if (FAILED(hr))
     {
         trace::Logger::Error("Could not disable allocation sampling: session pipe error. HRESULT=0x",
-                           std::setfill('0'), std::setw(8), std::hex, hr);
+                             std::setfill('0'), std::setw(8), std::hex, hr);
     }
     session_ = 0;
 
--- a/src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
@@ -1354,8 +1354,8 @@ void ContinuousProfiler::StopAllocationSampling()
     HRESULT hr = this->info12->EventPipeStopSession(session_);
     if (FAILED(hr))
     {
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -1074,9 +1068,13 @@ DWORD WINAPI SamplingThreadMain(_In_ LPVOID param)
         if (sleepTime == 0)
         {
             trace::Logger::Warn("Unexpected sampling interval configured, exiting sampling thread.");
-            return 1;
+            return;
+        }
+
+        if (future.wait_for(std::chrono::milliseconds(sleepTime)) != std::future_status::timeout)
```
eftiquar, line 1074: The code is fully functional. I have two concerns regarding the shutdown signaling mechanism:

1. Unconventional std::future/promise usage: The current implementation uses std::future/promise for simple shutdown signaling, which is typically reserved for passing values between threads. For boolean state coordination, a condition variable with a predicate is the more idiomatic approach.

2. Performance overhead in sampling loop: The future.wait_for() call occurs frequently in the sampling loop (every sleep interval), multiplying any per-call overhead. std::future implementations typically involve additional synchronization primitives and shared state management beyond what's needed for simple shutdown coordination.

Suggested improvement: Replace the future/promise mechanism with a dedicated shutdown mutex paired with a boolean flag, using a condition variable with predicate for waiting:
// Replace future/promise with:
mutable std::mutex shutdown_mutex_; //Question - can we use profiling_lock itself? If yes, we can safely make shutdown_requested_ non atomic.
std::condition_variable shutdown_cv_;
std::atomic_bool shutdown_requested_{false}; 

// In sampling loop:
std::unique_lock<std::mutex> lock(shutdown_mutex_);
if (shutdown_cv_.wait_for(lock, sleep_duration, 
                         [this] { return shutdown_requested_; })) {
    return; // Shutdown requested
}

PR #4431, opened 2025-08-30
(fix) Issue 4269: Deadlock — Move AssemblyResolve_ManagedProfilerDependencies out of Loader class

### LINKED ISSUE #4269: Deadlock: AssemblyResolve_ManagedProfilerDependencies should be moved out of Loader class

lukeina2z:
Fixes #4269 

The Loader’s static constructor blocks all its static methods while it is running. This can cause a deadlock if the profiler attempts to call the assembly resolver registered by Loader.

The fix is safe and limited in scope: the assembly resolve hook is moved to its own class, avoiding the critical section introduced by Loader’s static constructor.

Testing:
Manual testing confirmed the fix resolves the deadlock. Windows: ran .\build.ps1 with both Debug and Release configurations. macOS: ran .\build.sh.

Note for reviewers:
To simplify code review, I intentionally kept the new classes in the existing files. StyleCop warnings are explicitly suppressed. Formatting changes will be addressed in a follow-up PR.


Kielek:
Fixes the same issue as #4337.
@zacharycmontoya, could you please review here/decide which way to go?
In previous PR some tests were missing?

zacharycmontoya:
Yes this supersedes the work I started in #4337 . I'm happy to accept this contribution instead, as I did not make time to locally test the failing scenario and ensure the proposed fix worked. Both my previous PR and this PR do the smallest fix possible to avoid the deadlock scenario so this approach looks good to me.

@lukeina2z could you add more detail in the PR description of the application set up that you used for local testing? Once you do that, I'm happy to approve this. And is there any way for us to add an automated test case?

lukeina2z:
> Yes this supersedes the work I started in #4337 . I'm happy to accept this contribution instead, as I did not make time to locally test the failing scenario and ensure the proposed fix worked. Both my previous PR and this PR do the smallest fix possible to avoid the deadlock scenario so this approach looks good to me.
> 
> @lukeina2z could you add more detail in the PR description of the application set up that you used for local testing? Once you do that, I'm happy to approve this. And is there any way for us to add an automated test case?

Thank you for looking into this @Kielek  and @zacharycmontoya. I have included the repro steps and my test in this PR's description part.

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
 
 ### Fixed
 
+- Deadlock: AssemblyResolve_ManagedProfilerDependencies should be moved out
```
Kielek:
```suggestion
- Fixed deadlock occurring in Assembly Resolver.
  See [#4269](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/4269)
  for details.
```

Kielek:
@lukeina2z, thank you!

PR #4439, opened 2025-09-03
(File rename only): Rename Loader.xxx.cs to AssemblyResolver.xxx.cs

lukeina2z:
This pull request follows up on PR #4431. To keep the previous review focused, that PR did not include a file rename to match the updated class name.

The sole change in this PR is the file rename, ensuring consistency between the file name and the class name. This change is low risk and does not introduce any functional modifications.


src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyResolver.Net.cs
zacharycmontoya, line 1: Can you also remove the `SuppressMessage` attributes?

lukeina2z: > Can you also remove the `SuppressMessage` attributes?

It will be in the next PR. I want to keep this PR as 'file name change only'.

zacharycmontoya: Ah, Git is smart enough to associate both a rename and modifications to a file in a single commit so I'd be okay with those both happening in this PR. But if you'd like to do them separately that's also okay

Kielek: It should be fine to remove suppressions. Git(hub) is handling such cases smoothly

lukeina2z: Sure. Created another iteration.

My initial thought is to remove all suppressions in an isolated PR. Because doing that will introduce a large chunk of code move/change.

PR #4462, opened 2025-09-16
[FileBasedConfiguration] Resources

Kielek:
## Why

Fixes #

## Checklist


src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceSettings.cs
```diff
@@ -0,0 +1,90 @@
...
+
+    /// <summary>
+    /// Gets or sets the list of enabled resources.
+    /// </summary>
+    public IReadOnlyList<KeyValuePair<string, object>> Resources { get; set; } = [];
+
+    /// <summary>
+    /// Gets or sets a value indicating whether environmental variables resource detector is enabled.
+    /// </summary>
+    public bool EnabledEnvironmentalVariablesDetector { get; set; } = true;
```
RassK: `EnabledEnvironmentalVariablesDetector` -> `IsEnabledEnvironmentalVariablesDetector` ?

RassK:
Will you address the TODOs here or later?

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/Parser/Parser.cs
```diff
@@ -0,0 +1,23 @@
...
+    public static YamlConfiguration ParseYaml(string filePath)
+    {
+        var deserializer = new DeserializerBuilder()
+            .WithNodeDeserializer(existing => new ConditionalDeserializer(existing), s => s.InsteadOf<NullNodeDeserializer>())
+            .WithTypeConverter(new EnvVarTypeConverter())
+            .IgnoreUnmatchedProperties()
+            .Build();
+
+        var yaml = File.ReadAllText(filePath);
+        var config = deserializer.Deserialize<YamlConfiguration>(yaml);
```
RassK, line 20: Would stream based TextReader perform better here?

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceSettings.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        foreach (var kv in resourceAttributesWithPriority)
+        {
+            if (!merged.ContainsKey(kv.Key))
+            {
+                merged[kv.Key] = kv.Value;
+            }
+        }
+
+        foreach (var kv in additionalResources)
```
RassK: Maybe consider IEnumerable if you call this only once and do post filtering.

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/Parser/Parser.cs
Kielek: I would left this check for follow up issue - #4463.

Kielek:
> Will you address the TODOs here or later?

Follow up PRs. I would like to avoid making this PR bigger.

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceSettings.cs
Kielek: Could you please be more specipic? I am not sure how to understand this comment.

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceSettings.cs
Kielek: I will double check naming style in the settings, tommorow.

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceSettings.cs
RassK: `List<KeyValuePair<string, object>> ParseResourceAttributes` -> `IEnumerable<KeyValuePair<string, object>> ParseResourceAttributes` so you don't create unnecessary dictionary.

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceSettings.cs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private static List<KeyValuePair<string, object>> ParseResourceAttributes(string? resourceAttributes)
+    {
+        if (string.IsNullOrEmpty(resourceAttributes))
+        {
+            return [];
+        }
+
+        var attributeListSplitter = ',';
```
RassK:
```suggestion
        const char attributeListSplitter = ',';
```

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceSettings.cs
```diff
@@ -0,0 +1,90 @@
...
+        var attributeKeyValueSplitter = '=';
```
RassK:
```suggestion
        const char attributeKeyValueSplitter = '=';
```

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceSettings.cs
RassK: ... or maybe consider working with spans, my tests show that spans do perform better but they do not work with IEnumerable (yield).

```csharp
    public List<KeyValuePair<string, object>> ParseResourceAttributes()
    {
        var results = new List<KeyValuePair<string, object>>();

        if (string.IsNullOrEmpty(resourceAttributesText))
        {
            return results;
        }

        var resourceAttributes = resourceAttributesText.AsSpan();

        const char attributeListSplitter = ',';
        const char attributeKeyValueSplitter = '=';

        int start = 0;
        while (start < resourceAttributes.Length)
        {
            int listSplitter = resourceAttributes[start..].IndexOf(attributeListSplitter);
            var segment = listSplitter == -1
                ? resourceAttributes[start..]
                : resourceAttributes.Slice(start, listSplitter);

            start += (listSplitter == -1 ? segment.Length : listSplitter + 1);
            if (segment.IsEmpty)
            {
                continue;
            }

            int valueSplitter = segment.IndexOf(attributeKeyValueSplitter);
            if (valueSplitter == -1)
            {
                continue;
            }

            var key = segment[..valueSplitter].Trim();
            var value = segment[(valueSplitter + 1)..].Trim();

            results.Add(new KeyValuePair<string, object>(key.ToString(), value.ToString()));
        }

        return results;
    }
```

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/ResourceConfiguration.cs
```diff
@@ -0,0 +1,39 @@
...
+    [YamlMember(Alias = "attributes")]
+    public List<ResourceAttribute>? Attributes { get; set; }
+
+    /// <summary>
+    /// Gets or sets the attributes list for the resource.
+    /// </summary>
+    [YamlMember(Alias = "attributes_list")]
+    public string? AttributesList { get; set; }
+
+    public List<KeyValuePair<string, object>> ParseAttributes()
```
RassK, line 22: Any reason `ParseResourceAttributes` and `ParseAttributes` are in different locations although their idea almost similar?

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceSettings.cs
Kielek: 3fda793fea8966ec07f366c648bfa63b29674da4

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceSettings.cs
Kielek: 3fda793fea8966ec07f366c648bfa63b29674da4

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/ResourceConfiguration.cs
Kielek: Probably I overlooked it while working on extraction form bigger PR.

Merged to one method in 5aff60873d5218b88936d66a17f010853516a115

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceSettings.cs
Kielek: Spans are promising, but your construction will not work on .NET Framework.
I doubt that it is worth to keep 2 implementation at this time.

Fully removed separate method per your suggestion in 5aff60873d5218b88936d66a17f010853516a115

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceSettings.cs
Kielek: Renamed `EnabledEnvironmentalVariablesDetector` to `EnvironmentalVariablesDetectorEnabledin`.
There is not pattern with Is in the settings classes. We have `ProfilingEnambed`, `MetricsEnabled`, etc.
8138c241559dfbff7023d95d18f8a6623071d629

RassK: Seems true, unfortunately no good pattern. I think this change is good enough 👍

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceSettings.cs
RassK: May want to separate it to a new issue then. Since .NET Fx requires some polyfills it would make this PR bigger, but worth looking into, there is something like 2x in performance left on the table.

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/ResourceConfiguration.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        if (AttributesList != null)
+        {
+            const char attributeListSplitter = ',';
+            const char attributeKeyValueSplitter = '=';
+
+            var rawAttributes = AttributesList.Split(attributeListSplitter);
+            foreach (var rawKeyValuePair in rawAttributes)
+            {
+                var keyValuePair = rawKeyValuePair.Split(attributeKeyValueSplitter);
```
RassK: Nit: May cause issues with base64 values. Not sure we use any, specially here. Easy fix would be (since keys never use `=`):

```suggestion
                var keyValuePair = rawKeyValuePair.Split([attributeKeyValueSplitter], count: 2);
```

Maybe worth adding `StringSplitOptions.RemoveEmptyEntries`.

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceSettings.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    /// <summary>
+    /// Gets or sets the list of enabled resource detectors.
+    /// </summary>
+    public IReadOnlyList<ResourceDetector> EnabledDetectors { get; set; } = [];
+
+    /// <summary>
+    /// Gets or sets the list of enabled resources.
+    /// </summary>
+    public IReadOnlyList<KeyValuePair<string, object>> Resources { get; set; } = [];
```
RassK, line 18: for the future, may want to use union ref struct not to cause object boxing for simple types.

RassK (approved):
In general seems fine as follow ups seem to come anyway and fix some of the parts per time.

src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceSettings.cs
Kielek: #4465

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/ResourceConfiguration.cs
Kielek: 2cefa3f1f1ede47c12674b4e1c1807d5a46c97cc

src/OpenTelemetry.AutoInstrumentation/Configurations/ConfigurationKeys.cs
```diff
@@ -51,6 +51,33 @@ internal partial class ConfigurationKeys
...
+    {
+        /// <summary>
+        /// Configuration key for enabling file based configuration.
+        /// </summary>
+        public const string Enabled = "OTEL_EXPERIMENTAL_FILE_BASED_CONFIGURATION_ENABLED";
+
+        /// <summary>
+        /// Configuration key for enabling file based configuration.
+        /// Default is <c>"config.yaml"</c>.
+        /// </summary>
```
ysolomchenko:
```suggestion
        /// <summary>
        /// Configuration key for the path to the configuration file.
        /// Default is <c>"config.yaml"</c>.
        /// </summary>
```

src/OpenTelemetry.AutoInstrumentation/Configurations/Settings.cs
```diff
@@ -1,31 +1,71 @@
...
+    protected abstract void OnLoadEnvVar(Configuration configuration);
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="Settings"/> class
+    /// using the specified <see cref="YamlConfiguration"/> to initialize values.
+    /// </summary>
+    /// <param name="configuration">The <see cref="YamlConfiguration"/> to use when retrieving configuration values.</param>
+    protected virtual void OnLoadFile(YamlConfiguration configuration)
+    {
+    }
```
ysolomchenko, line 59: I think this method should also be abstract in the future. It’s worth adding a `TODO: that once the implementation is done in all settings, it should be changed to abstract.`

Kielek:
@rajkumar-rangaraj, @nrcventura, @zacharycmontoya. I am going to merge to unblock further work on this tasks and NoCode feature.
Please let me know if you have any comments. I will try to address in the follow ups.

PR #4470, opened 2025-09-19
.NET Framework `SqlCommand` IL rewrite for `SqlClient` instrumentation

### LINKED ISSUE #4343: Bytecode instrumentation of (legacy) SqlClient code
# Feature Request

When using auto-instrumentation, the contrib SQL instrumentation is added. We have a customer stuck on an older version of System.Data.SqlClient, who reported that span names are very sparce (just the server name) and that the sql command is not available in the attributes. This [issue](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/2287) describes the problem which is a known issue due to the underlying implementation.

As a potential solution, we made a POC that switches to using a bytecode instrumentation for SQL, when using auto-instrumentation. This allows for richer (sem-con compliant) spans to be created.

We would like to discuss whether there would be an appetite to switch to a bytecode instrumentation to improve the experience for customers with legacy applications? We have the POC for this, and would be happy to tidy that up in order to make a PR to contribute this. Currently it focuses on just `System.Data.SqlClient` but removes the contrib instrumentation to avoid duplicated spans. We may need to instrument all libraries, including `Microsoft.Data.SqlClient` using bytecode.

Limited attributes and span naming

Add new bytecode instrumentation. Likely behind a config switch so it would be opt-in.

If/when autoinstrumentation supports plugin of custom bytecode instrumentations, use that mechanism to support opting into the bytecode for SQL.

**Additional context**
N/A

stevejgordon:
## Why

Details in #4343. In short, on .NET Framework `System.Data` doesn't emit the `CommandText` (except for stored procedures) in its events. This prevents the `SqlClient` instrumentation from enriching with the DB query text and summary.

Fixes #4343

## What

As discussed in the issue. This rewrites the `WriteBeginExecuteEvent` to ensure the `CommandText` is always present. The implementation currently enables this by default, but provides configuration to disable this behaviour if required.

## Tests

The `SqlClientSystemTests` integration test was updated to to the new `OTEL_DOTNET_AUTO_SQLCLIENT_NETFX_ILREWRITE_ENABLED` environmnet variable.

## Checklist

cc @alanwest, @martincostello


src/OpenTelemetry.AutoInstrumentation.Native/environment_variables.h
```diff
@@ -69,6 +69,9 @@ const WSTRING netfx_assembly_redirection_enabled = WStr("OTEL_DOTNET_AUTO_NETFX_
 // Enable the fail fast mode.
 const WSTRING fail_fast_enabled = WStr("OTEL_DOTNET_AUTO_FAIL_FAST_ENABLED");
 
+// Enable the assembly version redirection when running on the .NET Framework.
```
zacharycmontoya: Looks like a copy/paste typo 😅 Can you make sure to update this with appropriate text?

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -3234,6 +3242,139 @@ HRESULT CorProfiler::GenerateLoaderType(const ModuleID module_id,
     return S_OK;
 }
 
+HRESULT CorProfiler::RewriteILSystemDataCommandText(const ModuleID module_id)
```
zacharycmontoya, line 3234: Can you add some code comments here to describe the overall goal of rewriting this particular method's IL? Something to the effect of ensuring that the CommandText is propagated to the EventSource and downstream Diagnostic Source listeners.

test/IntegrationTests/SqlClientSystemTests.cs
```diff
@@ -37,23 +35,32 @@ public static TheoryData<string, bool> GetData()
     [Trait("Category", "EndToEnd")]
     [Trait("Containers", "Linux")]
     [MemberData(nameof(GetData))]
-    public void SubmitTraces(string packageVersion, bool dbStatementForText)
+    public void SubmitTraces(string packageVersion, bool enableIlRewrite)
```
zacharycmontoya: We should still test that by default we do not add the `db.statement` attribute. What do you think of keeping the `SubmitTraces` test method unchanged, but adding a new test with your new logic that is only defined for `#if NETFRAMEWORK`?

docs/config.md
```diff
@@ -414,9 +415,10 @@ Important environment variables include:
...
+| `OTEL_DOTNET_AUTO_PLUGINS`               | Colon-separated list of OTel SDK instrumentation plugin types, specified with the [assembly-qualified name](https://docs.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname?view=net-6.0#system-type-assemblyqualifiedname). *Note: This list must be colon-separated because the type names may include commas.* See more info on how to write plugins at [plugins.md](plugins.md). |               | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) | | 
+`OTEL_DOTNET_AUTO_SQLCLIENT_NETFX_ILREWRITE_ENABLED`        | Enables IL rewriting of `SqlCommand` on .NET Framework to ensure `CommandText` is present for `SqlClient` instrumentation, which is required for `db.query.text` and `db.query.summary` to be populated. | `true` | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
```
Kielek: There is a separate section for instrumentation specific env-vars. Please move it to `### Instrumentation options`

stevejgordon:
@zacharycmontoya and @Kielek - Thanks for the review. I am back from vacation and will update this ASAP.

src/OpenTelemetry.AutoInstrumentation.Native/environment_variables.h
stevejgordon: Fixed!

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
stevejgordon: I've added some comments

test/IntegrationTests/SqlClientSystemTests.cs
stevejgordon: Updated tests as suggested

docs/config.md
stevejgordon: Moved to the correct section.

test/IntegrationTests/SqlClientSystemTests.cs
```diff
@@ -41,6 +54,39 @@ public void SubmitTraces(string packageVersion)
...
+            collector.Expect("OpenTelemetry.Instrumentation.SqlClient", span => span.Attributes.All(attr => attr.Key != "db.statement"));
+        }
+
+        RunTestApplication(new()
+        {
+            Arguments = $"{_sqlServerFixture.Password} {_sqlServerFixture.Port}",
+            PackageVersion = packageVersion
+        });
+
+        collector.Expect("OpenTelemetry.Instrumentation.SqlClient", span => span.Attributes.Any(attr => attr.Key == "db.statement" && !string.IsNullOrWhiteSpace(attr.Value?.StringValue)));
```
nrcventura: Is this assertion supposed to be here? It looks like the same assertion is conditionally applied.

Kielek:
@stevejgordon, do you need any help here or just more time?
It will be great to include it in the next release.

test/IntegrationTests/SqlClientSystemTests.cs
stevejgordon: I've fixed this test up.

stevejgordon:
@Kielek I may need a bit of help. Previously I had the test running in Visual Studio and it was passing. Now it doesn't! ~I looked at the logs and I don't see the rewrite occurring.~ UPDATE: I was looking in the wrong log. I see the IL rewrite has occurred as expected. So I'm not immediately sure why the assertion is failing.

In an isolated test application running in IIS, with the files from `tracer-home` copied over the existing install, I see the rewrite logging appear. There are some errors in the managed log where it looks like it's picking up the wrong DLLs which didn't used to occur when I last tested, but I've since been messing with other things that likely screwed that up. However, as I see the IL rewrite, that's the main thing.

Would you perhaps be able to review if I've missed something in this test and/or test locally yourself to see if this is working as expected? I'll continue to dig into it on my end too and try to figure it out.

stevejgordon:
@Kielek Could it be anything to do with the removal of the `OTEL_DOTNET_AUTO_SQLCLIENT_SET_DBSTATEMENT_FOR_TEXT` option?

Kielek:
@stevejgordon, sorry for delay. Either @zacharycmontoya will check it today or I will try to help tommorow.

zacharycmontoya:
> Would you perhaps be able to review if I've missed something in this test and/or test locally yourself to see if this is working as expected? I'll continue to dig into it on my end too and try to figure it out.

I was able to get your integration test running on my machine and it passes ✅ However, I'd also like to add this same testing to our test case that uses the .NET Framework built-in `System.Data.SqlCommmand` in the `SqlClientSystemDataTests.cs` class. Can you add this or would it be fine if I added onto your PR?

stevejgordon:
@zacharycmontoya Thanks for checking. Ill have to try to figure out why they are failing on my machine. Feel free to duplicate this for the other scenarios and add to this PR in the mean time.

zacharycmontoya:
@stevejgordon I've added one test case in commit 355be8694af92db9da65014e48c5fc0d4e4712c0, which also passed locally on my machine. If CI passes, then I think it's good-to-go

docs/config.md
```diff
@@ -234,6 +235,7 @@ and [logs bridge](./log4net-bridge.md).
...
 | `OTEL_DOTNET_EXPERIMENTAL_ASPNET_DISABLE_URL_QUERY_REDACTION`                     | Whether the ASP.NET instrumentation turns off redaction of the `url.query` attribute value.                                                                                                                                                                                                          | `false`       | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
+| `OTEL_DOTNET_AUTO_SQLCLIENT_NETFX_ILREWRITE_ENABLED`        | Enables IL rewriting of `SqlCommand` on .NET Framework to ensure `CommandText` is present for `SqlClient` instrumentation, which is required for `db.query.text` and `db.query.summary` to be populated. | `true` | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
```
Kielek: I have some doubts here. As t might impact end-users, I would consider this as opt-in feature.

src/OpenTelemetry.AutoInstrumentation.Native/environment_variables_util.h
```diff
@@ -50,6 +50,7 @@ bool IsDumpILRewriteEnabled();
 bool IsAzureAppServices();
 bool IsFailFastEnabled();
 bool IsNetFxAssemblyRedirectionEnabled();
+bool IsSqlClientNetFxILRewriteEnabled();
```
Kielek, line 53: Ideally, it should be propagated from the managed coded. We are in the middle of introducing File based configuration and it need to be reworked in the near feature. Not a blocker for this PR IMO.

Ref: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4492

docs/config.md
zacharycmontoya: I am actually leaning towards on by default because the effect is that the user gets one more attribute on their spans, so overall this improves the tracing quality automatically and makes it more appealing to use our automatic package. If there are any side-effects from turning this on, then we should promptly identify and resolve them.

nrcventura: Unfortunately, I think that the side-effects that I'm most concerned about are outside of our control. The code that we are modifying causes the sql to get written to an EventSource. When things are written to an EventSource, that data can now be consumed by other things, including things that can result in data getting persisted outside of the process. Those things may not be sanitizing the sql, or may assume that the data was safe to begin with (just the stored procedure). 
I think having it on would be fine for the majority of users, but there are probably a few that would run into problems with having this data present by default.

zacharycmontoya: Thanks, I overlooked the fact that we're writing to an EventSource and not directly updating the span. It makes sense then to have users opt-in to the feature

Kielek: I fixed it in a3031d00b029afda81971bd5e7ce20d272d5510f

src/OpenTelemetry.AutoInstrumentation.Native/environment_variables_util.h
Kielek: #4561 created

Kielek (approved):
@zacharycmontoya, @nrcventura I have checked default behavior to disabled. Also I have added some explanation in  f8a0cad1df42d13749746ef85902ea17fafa01fc

Other parts of code LGTM. Please double check if we can merge it.

PR #4270, opened 2025-06-24
Implement file based configuration

ysolomchenko:
Demo for #4394 

## What
Current status of implementation of file-based configuration.

| ENV VAR                                                                         | Layer                                                                                     | Property name                                                                                         | Works now? |
| ------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------- | ---------- |
| Global settings                                                                 |                                                                                           |                                                                                                       |            |
| OTEL_DOTNET_AUTO_HOME                                                           | OpenTelemetry.AutoInstrumentation.Loader<br>OpenTelemetry.AutoInstrumentation.Native      | Out of Specification                                                                                  | ❌          |
| OTEL_DOTNET_AUTO_EXCLUDE_PROCESSES                                              | OpenTelemetry.AutoInstrumentation.StartupHook<br>OpenTelemetry.AutoInstrumentation.Native | Out of Specification                                                                                  | ❌          |
| OTEL_DOTNET_AUTO_FAIL_FAST_ENABLED                                              | OpenTelemetry.AutoInstrumentation<br>OpenTelemetry.AutoInstrumentation.Native             | fail_fast:                                                                                 | ✔       |
| OTEL_DOTNET_AUTO_LOGGER                                                         | OpenTelemetry.AutoInstrumentation<br>OpenTelemetry.AutoInstrumentation.Native             | Out of Specification                                                                                  | ❌          |
| OTEL_LOG_LEVEL                                                                  | OpenTelemetry.AutoInstrumentation<br>OpenTelemetry.AutoInstrumentation.Native             | log_level:                                                                                            | ✔?         |
|                                                                                 |                                                                                           |                                                                                                       |            |
| Resources                                                                       |                                                                                           |                                                                                                       |            |
| OTEL_RESOURCE_ATTRIBUTES                                                        | SDK                                                                                       | resource:<br>    attributes:<br>    or attributes_list:                                               | ✔          |
| OTEL_SERVICE_NAME                                                               | SDK                                                                                       | resource:<br>    attributes:<br>    or attributes_list:                                               | ✔          |
|                                                                                 |                                                                                           |                                                                                                       |            |
| Resource detectors                                                              |                                                                                           |                                                                                                       |            |
| OTEL_DOTNET_AUTO_RESOURCE_DETECTOR_ENABLED                                      | OpenTelemetry.AutoInstrumentation                                                         | resource:<br>     detection/development:                                                              | ✔          |
[...]

PR #4492, opened 2025-10-01
[FileBasedConfiguration] Instrumentation Configuration

ysolomchenko:
## Why

Towards https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/4394
Extracted and refactored from https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4270

## What

File based configuration for Instrumentations

## Tests

Ci

## Checklist

- [ ] ~`CHANGELOG.md` is updated.~

### DIFF
--- a/docs/file-based-configuration.md
+++ b/docs/file-based-configuration.md
@@ -63,3 +63,92 @@ resource:
       process:         # Detects process-level attributes (process.*)
       processruntime:  # Detects process runtime attributes (process.runtime.*)
 ```  
+
+## Instrumentation Configuration
+
+You can configure traces, metrics, and logs instrumentations.
+For more details and updates, see: [Instrumentation list and documentation](config.md#instrumentations)
+
+``` yaml
+instrumentation/development:
+  dotnet:
+    traces:
+      aspnet:              # ASP.NET (.NET Framework) MVC/WebApi [Framework only]
+      aspnetcore:          # ASP.NET Core [Core only]
+      azure:               # Azure SDK [Core & Framework]
+      elasticsearch:       # Elastic.Clients.Elasticsearch [Core & Framework]
+      elastictransport:    # Elastic.Transport (>=0.4.16) [Core & Framework]
+      entityframeworkcore: # Entity Framework Core (>=6.0.12) [Core only]
+      graphql:             # GraphQL (>=7.5.0) [Core only]
+      grpcnetclient:       # Grpc.Net.Client (>=2.52.0 & <3.0.0) [Core & Framework]
+      httpclient:          # System.Net.Http.HttpClient [Core & Framework]
+      kafka:               # Confluent.Kafka (>=1.4.0 & <3.0.0) [Core & Framework]
+      masstransit:         # MassTransit (>=8.0.0) [Core only]
+      mongodb:             # MongoDB.Driver (>=2.7.0 <4.0.0) [Core & Framework]
+      mysqlconnector:      # MySqlConnector (>=2.0.0) [Core only]
+      mysqldata:           # MySql.Data (>=8.1.0) [Core only]
+      npgsql:              # Npgsql (>=6.0.0) [Core only]
+      nservicebus:         # NServiceBus (>=8.0.0 & <10.0.0) [Core & Framework]
+      oraclemda:           # Oracle.ManagedDataAccess (>=23.4.0) [Core only]
+      rabbitmq:            # RabbitMQ.Client (>=6.0.0) [Core & Framework]
+      quartz:              # Quartz (>=3.4.0, not supported < .NET Framework 4.7.2)
+      sqlclient:           # Microsoft.Data.SqlClient & System.Data.SqlClient [Core & Framework]
+      stackexchangeredis:  # StackExchange.Redis (>=2.6.122 & <3.0.0) [Core only]
+      wcfclient:           # WCF Client [Core & Framework]
+      wcfservice:          # WCF Service [Framework only]
+    metrics:
+      aspnetcore:          # ASP.NET Core metrics [Core only]
+      httpclient:          # HttpClient metrics [Core & Framework]
+      netruntime:          # .NET Runtime metrics [Core only]
+      nservicebus:         # NServiceBus metrics [Core & Framework]
+      process:             # Process metrics [Core & Framework]
+      sqlclient:           # SQL Client metrics [Core & Framework]
+    logs:
+      ilogger:             # Microsoft.Extensions.Logging (>=9.0.0) [Core & Framework]
+      log4net:             # log4net (>=2.0.13 && <4.0.0) [Core & Framework]
+```
+
+## Instrumentation options
+
+``` yaml
+instrumentation/development:
+  dotnet:
+    traces:
+      entityframeworkcore:
+        # Whether the Entity Framework Core instrumentation can pass SQL statements through the db.statement attribute. Queries might contain sensitive information. If set to false, db.statement is recorded only for executing stored procedures.
+        # Default is false
+        set_db_statement_for_text: false
+      graphql:
+        # Whether the GraphQL instrumentation can pass raw queries through the graphql.document attribute. Queries might contain sensitive information.
+        # Default is false
+        set_document: false
+      oraclemda: 
+        # Whether the Oracle Client instrumentation can pass SQL statements through the db.statement attribute. Queries might contain sensitive information. If set to false, db.statement is recorded only for executing stored procedures.
+        # Default is false
+        set_db_statement_for_text: false
+      sqlclient:
+        # Whether the SQL Client instrumentation can pass SQL statements through the db.statement attribute. Queries might contain sensitive information. If set to false, db.statement is recorded only for executing stored procedures. 
+        # Not supported on .NET Framework for System.Data.SqlClient.
+        # Default is false
+        set_db_statement_for_text: false
+      aspnet:
+        # A comma-separated list of HTTP header names. ASP.NET instrumentations will capture HTTP request header values for all configured header names.
+        capture_request_headers: "X-Key=Value"
+        # A comma-separated list of HTTP header names. ASP.NET instrumentations will capture HTTP response header values for all configured header names.
+        capture_response_headers: "X-Key=Value"
+      aspnetcore:
+        # A comma-separated list of HTTP header names. ASP.NET Core instrumentations will capture HTTP request header values for all configured header names.
+        capture_request_headers: "X-Key=Value"
+        # A comma-separated list of HTTP header names. ASP.NET Core instrumentations will capture HTTP response header values for all configured header names.
+        capture_response_headers: "X-Key=Value"
+      httpclient:
+        # A comma-separated list of HTTP header names. HTTP Client instrumentations will capture HTTP request header values for all configured header names.
+        capture_request_headers: "X-Key=Value"
+        # A comma-separated list of HTTP header names. HTTP Client instrumentations will capture HTTP response header values for all configured header names.
+        capture_response_headers: "X-Key=Value"
+      grpcnetclient:
+        # A comma-separated list of gRPC metadata names. Grpc.Net.Client instrumentations will capture gRPC request metadata values for all configured metadata names.
+        capture_request_metadata: "X-Key=Value"
+        # A comma-separated list of gRPC metadata names. Grpc.Net.Client instrumentations will capture gRPC response metadata values for all configured metadata names.
+        capture_response_metadata: "X-Key=Value"
+```
--- a/src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/DotNetMetrics.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/DotNetMetrics.cs
@@ -1,0 +1,79 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Reflection;
+using Vendors.YamlDotNet.Serialization;
+
+namespace OpenTelemetry.AutoInstrumentation.Configurations.FileBasedConfiguration;
+
+internal class DotNetMetrics
+{
+#if NETFRAMEWORK
+    /// <summary>
+    /// Gets or sets the ASP.NET metrics instrumentation configuration.
+    /// </summary>
+    [YamlMember(Alias = "aspnet")]
+    public object? AspNet { get; set; }
+#endif
+
+#if NET
+    /// <summary>
+    /// Gets or sets the ASP.NET Core metrics instrumentation configuration.
+    /// </summary>
+    [YamlMember(Alias = "aspnetcore")]
+    public object? AspNetCore { get; set; }
+#endif
+
+    /// <summary>
+    /// Gets or sets the HttpClient metrics instrumentation configuration.
+    /// </summary>
+    [YamlMember(Alias = "httpclient")]
+    public object? HttpClient { get; set; }
+
+    /// <summary>
+    /// Gets or sets the .NET runtime metrics instrumentation configuration.
+    /// </summary>
+    [YamlMember(Alias = "netruntime")]
+    public object? NetRuntime { get; set; }
+
+    /// <summary>
+    /// Gets or sets the NServiceBus metrics instrumentation configuration.
+    /// </summary>
+    [YamlMember(Alias = "nservicebus")]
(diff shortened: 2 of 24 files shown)


docs/file-based-configuration.md
```diff
@@ -63,3 +63,93 @@ resource:
       process:         # Detects process-level attributes (process.*)
       processruntime:  # Detects process runtime attributes (process.runtime.*)
 ```  
+
+## Instrumentation Configuration
+
+You can configure traces, metrics, and logs instrumentations.
+For more details and updates, see: [Instrumentation list and documentation](config.md#instrumentations)
```
Kielek: Missing comment how to decide what to enable disable. Commenting-out/removing means that you it will be disabled.

docs/file-based-configuration.md
```diff
@@ -63,3 +63,93 @@ resource:
...
+        # Whether the Oracle Client instrumentation can pass SQL statements through the db.statement attribute. Queries might contain sensitive information. If set to false, db.statement is recorded only for executing stored procedures.
+        # Default is false
+        set_db_statement_for_text: false
+      sqlclient:
+        # Whether the SQL Client instrumentation can pass SQL statements through the db.statement attribute. Queries might contain sensitive information. If set to false, db.statement is recorded only for executing stored procedures. 
+        # Not supported on .NET Framework for System.Data.SqlClient.
+        # Default is false
+        set_db_statement_for_text: false
+      aspnet:
+        # A comma-separated list of HTTP header names. ASP.NET instrumentations will capture HTTP request header values for all configured header names.
```
Kielek, line 237: maybe you can put here couple of headers? It is applicable also for other similar options

src/OpenTelemetry.AutoInstrumentation/Configurations/LogSettings.cs
```diff
@@ -64,6 +65,11 @@ protected override void OnLoadEnvVar(Configuration configuration)
             enabledConfigurationTemplate: ConfigurationKeys.Logs.EnabledLogsInstrumentationTemplate);
     }
 
+    protected override void OnLoadFile(YamlConfiguration configuration)
+    {
+        EnabledInstrumentations = configuration.InstrumentationDevelopment?.DotNet?.Logs?.GetEnabledInstrumentations() ?? [];
```
Kielek, line 70: I see that you are covering exporters but hte separate PR, but you are missing here posisiblity to enable Log4Net Bridge and incldue formatted message.

src/OpenTelemetry.AutoInstrumentation/Configurations/MetricSettings.cs
```diff
@@ -66,6 +67,11 @@ protected override void OnLoadEnvVar(Configuration configuration)
         MetricsEnabled = configuration.GetBool(ConfigurationKeys.Metrics.MetricsEnabled) ?? true;
     }
 
+    protected override void OnLoadFile(YamlConfiguration configuration)
```
Kielek, line 70: Missing possibility to configure additionalSources.

docs/file-based-configuration.md
```diff
@@ -63,3 +63,93 @@ resource:
...
       processruntime:  # Detects process runtime attributes (process.runtime.*)
 ```  
+
+## Instrumentation Configuration
+
+You can configure traces, metrics, and logs instrumentations.
+For more details and updates, see: [Instrumentation list and documentation](config.md#instrumentations)
+
+``` yaml
+instrumentation/development:
```
zacharycmontoya, line 183: Is this `instrumentation` key and the language-specific `dotnet` subkey already an established standard for configuring instrumentations?

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/SetDocumentConfiguration.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using OpenTelemetry.AutoInstrumentation.Configurations.FileBasedConfiguration.Parser;
+using Vendors.YamlDotNet.Serialization;
+
+namespace OpenTelemetry.AutoInstrumentation.Configurations.FileBasedConfiguration;
+
+[EmptyObjectOnEmptyYaml]
+internal class SetDocumentConfiguration
```
zacharycmontoya: It seems strange to me to name the configuration based off one capability. If possible, I'd prefer to name it GraphQL configuration in case we ever need to expand the configuration options

test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/FileBased/Files/TestInstrumentationConfigurationFile.yaml
```diff
@@ -0,0 +1,22 @@
...
+  dotnet:
+    traces:
+      entityframeworkcore:
+        set_db_statement_for_text: true
+      graphql:
+        set_document: true
+      oraclemda: 
+        set_db_statement_for_text: true
+      aspnet:
+        capture_request_headers: "X-Key"
```
zacharycmontoya: Can we make this accept a list of string values rather than one csv string value?

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/DotNetTraces.cs
zacharycmontoya, line 1: Since we have some instrumentations that are only for .NET Framework and some that are only for .NET Core, it might be easier to make `DotNetTraces` a partial class and then split up the framework-specific properties into `DotNetTraces.Framework.cs` and `DotNetTraces.Core.cs`. What do you think?

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/DotNetMetrics.cs
zacharycmontoya, line 1: My suggestion for `DotNetTraces.cs` also applies here, what do you think of making this a partial class to split up the framework-specific properties?

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/DotNetTraces.cs
```diff
@@ -0,0 +1,175 @@
...
+    /// Returns the list of enabled traces instrumentations.
+    /// </summary>
+    public IReadOnlyList<TracerInstrumentation> GetEnabledInstrumentations()
+    {
+        var enabled = new List<TracerInstrumentation>();
+        var properties = typeof(DotNetTraces).GetProperties(BindingFlags.Instance | BindingFlags.Public);
+
+        foreach (var prop in properties)
+        {
+            var value = prop.GetValue(this);
```
zacharycmontoya, line 163: Does this mean that if a user wants to use file-based configuration, they will have to update their file to list every single instrumentation they want to enable? Can we instead add a bool `enabled` property to each entry, that way individual instrumentations can easily be disabled by setting `enabled: false` but if an instrumentation is not listed then it's on-by-default? That feels much more user-friendly

zacharycmontoya (commented):
Besides some small non-blocking comments I wrote, one main question I have is whether we should make users list all integrations they want to use, or make all integrations on by default but configurable with an `enabled` flag. I'm leaning towards the second one, but I'd like to get your thoughts

src/OpenTelemetry.AutoInstrumentation/Configurations/TracerSettings.cs
```diff
@@ -93,6 +94,13 @@ protected override void OnLoadEnvVar(Configuration configuration)
         InstrumentationOptions = new InstrumentationOptions(configuration);
     }
 
+    protected override void OnLoadFile(YamlConfiguration configuration)
```
Kielek, line 106: missing possibility to set additional sources and legacy sources

docs/file-based-configuration.md
ysolomchenko: Not stable, but compliant with the current specification.

test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/FileBased/Files/TestInstrumentationConfigurationFile.yaml
ysolomchenko: Yes, it is possible, but I prefer to keep comma separated list for backward compatibility with env var configuration.

Another possibility is to keep both (as we do for propagators and resources), But I think we can create an issue with the vNext milestone, and it can be implemented later when someone needs it.

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/DotNetTraces.cs
ysolomchenko: At the moment, it is implemented the same way as propagators and resource detectors, and this approach is compliant with the current specification. I believe we should keep the current approach until the specification changes.

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/DotNetMetrics.cs
ysolomchenko: Since in [TracerInstrumentation.cs](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/src/OpenTelemetry.AutoInstrumentation/Configurations/TracerInstrumentation.cs) for .NET Core and .NET Framework everything is in a single file, I don’t see the point of splitting out `DotNetTraces.cs`.

docs/file-based-configuration.md
ysolomchenko: fixed in [241ea19](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4492/commits/241ea19b3ffaf40a04ad8ba52150f09db92c65d2)

docs/file-based-configuration.md
ysolomchenko: changed in [548f7e4](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4492/commits/548f7e4dbd7ac6b854b7864fc3d1b045b202ddbc)

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/SetDocumentConfiguration.cs
ysolomchenko: changed in [2c36af4](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4492/commits/2c36af4ee84d9da63d5997871aeb045452231806)

src/OpenTelemetry.AutoInstrumentation/Configurations/LogSettings.cs
ysolomchenko: Log4Net bridge added in [65692bc](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4492/commits/65692bc9d5c0d31fabc4a5515e2072915de82995)

Formatted message will be included in logger provider configuration PR.

src/OpenTelemetry.AutoInstrumentation/Configurations/MetricSettings.cs
ysolomchenko: I think these changes enough for this PR, and additional sources will be added in a separate PR

src/OpenTelemetry.AutoInstrumentation/Configurations/TracerSettings.cs
ysolomchenko: I think these changes enough for this PR, and additional sources will be added in a separate PR

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/DotNetMetrics.cs
zacharycmontoya: Ok, sounds good to me 👍🏼

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/DotNetTraces.cs
zacharycmontoya: Yes, agreed that we should align with the specification. Do you have a link to the section describing the structure of those that I could look at?

ysolomchenko: Here is spec for detectors https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/kitchen-sink.yaml#L844

Here is spec for propagators https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/kitchen-sink.yaml#L541

Here is spec for instrumentations https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/kitchen-sink.yaml#L902

Created issue for disabled/enabled in spec https://github.com/open-telemetry/opentelemetry-configuration/issues/334

docs/file-based-configuration.md
```diff
@@ -172,6 +172,98 @@ propagator:
...
+
+You can configure traces, metrics, and logs instrumentations.
+For more details and updates, see: [Instrumentation list and documentation](config.md#instrumentations)
+To disable a instrumentation, comment out or remove its corresponding entry.
+
+``` yaml
+instrumentation/development:
+  dotnet:
+    traces:
+      aspnet:              # ASP.NET (.NET Framework) MVC/WebApi [Framework only]
```
Kielek: I think that you should put here a name from the config.md tables.
I would like to avoid 2 places to maintain same data.

docs/file-based-configuration.md
```diff
@@ -172,6 +172,98 @@ propagator:
...
+        set_document: false
+      oraclemda: 
+        # Whether the Oracle Client instrumentation can pass SQL statements through the db.statement attribute. Queries might contain sensitive information. If set to false, db.statement is recorded only for executing stored procedures.
+        # Default is false
+        set_db_statement_for_text: false
+      sqlclient:
+        # Whether the SQL Client instrumentation can pass SQL statements through the db.statement attribute. Queries might contain sensitive information. If set to false, db.statement is recorded only for executing stored procedures. 
+        # Not supported on .NET Framework for System.Data.SqlClient.
+        # Default is false
+        set_db_statement_for_text: false
```
Kielek: already dropped?

ysolomchenko: Fixed in ea047c0

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/DotNetTraces.cs
nrcventura: This behavior, even though it's in agreement with the spec does not seem like the desired behavior to me. In practice, I find that configuration files rarely get updated, and as new features are added they rarely get enabled. In this case I can see someone creating their configuration file with the current set of instrumentations. Then a couple of years later we offer more instrumentations out of the box, and the auto-instrumentation user wonders why they are not seeing that instrumentation being applied (especially if they are using the same or roughly the same configuration file template for all of their services). It's a departure from the behavior of our current environment variable, where if you specify nothing then you get all of the defaults. I'm not sure how to achieve that same behavior with this file based configuration approach. 
Is there some way to remain in compliance with the current spec and offer some other experimental config that follows either the presets described in the configuration issue, or the approach used by Java in that issue?

ysolomchenko: This [comment](https://github.com/open-telemetry/opentelemetry-configuration/issues/334#issuecomment-3415736068) quite accurately describes why this approach shouldn’t be used.

In short, the file-based configuration was created to ensure that as little as possible is hidden somewhere in the code, and that when we make changes or updates, everything remains as it was for end users.

Personally, I also like the idea of `presets`, maybe it will be possible to push it into the specification.
And as soon as this feature appears, I’ll definitely add it.

src/OpenTelemetry.AutoInstrumentation/Configurations/TracerSettings.cs
Kielek: #4570

Issue #3817, opened 2024-11-25
Automate detecting new transitive dependencies for .NET Framework


Kielek:
`TransientDependenciesTests.DefinedTransientDeps_Are_MatchingGeneratedDeps` is not able to detect new transitive dependencies.
Example: 

`OpenTelemetry.Exporter.Zipkin` references System.Text.Json 4.7.2 (without any reference to System.IO.Pipelines). The we upgrade this package to System.Text.Json 9.0.0. New version is referencing System.IO.Pipelines.

Mentioned tests was modified to manually handle this particular case. It should be automated to handle all future changes automatically.

iskiselev:
Should be closed now

PR #4505, opened 2025-10-07
WIP Use framework-specific dependencies for .Net Framework versions

iskiselev:
## Why
Currently OTEL registers in GAC assemblies, that was never intendend to be put in GAC on latest framework - and not registring netstandard on 4.6.2. It result in issues discussed at https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/4186#issuecomment-2979050475

## What
With this change, we pack different set of assemblies for supported .NET Frameworks:
- 4.6.2
- 4.7
- 4.7.1
- 4.7.2 and up

Only assemblies for current framework will be registered in GAC at install time or loaded from installation dir.
To minimize size set of assemblies deduplicated - with assemblies that used in all .NET Frameworks still stay at `netfx` folders. Assemblies used only in some versions stays in oldest TFM folder. For other TFM using same assembly .link file will be created that points to TFM with assembly.
To simplify buget packages tracking `CentralPackageTransitivePinningEnabled` was enabled - with it transient auto-generated packages were removed from `OpenTelemetry.AutoInstrumentation.csproj`.
.NET Framework version resolved though registry check in installation, profiler and managed code. It associated with some risk (proper rights). If version will not be resolved, it is fallback to 4.6.2, which should have same behaviour as before this change. It may be better nowdays to fallback to 4.7.2 - as most people probably have it.

## Tests
- validated OTEL work with loading assemblies from install folder on .NET 4.7.2. Manually validated that .link files are loaded correctly
- TODO: validate that GAC installation works properly
- TODO: validate that nuget installation was not affected

## Checklist


build/Build.Steps.cs
```diff
@@ -42,9 +44,17 @@ partial class Build
             : new[] { MSBuildTargetPlatform.x86 };
 
     private static readonly IEnumerable<TargetFramework> TargetFrameworks = new[]
+    {
+       TargetFramework.NET8_0,
+       TargetFramework.NET462,
```
RassK, line 49: `TargetFramework.NET462` should this remain here? Becomes quite confusing between `TargetFrameworks` vs `TargetFrameworksNetFx `

iskiselev: Yes, it should remain here.
Probably we should find better name for `TargetFrameworksNetFx`.
`TargetFrameworks` - TFM used for AutoInstrumentation compilation
`TargetFrameworksNetFx` - TFM list used to pack AutoInstrumentation for .Net Framework (packing for .NET has different, unchanged mechanic).

build/Build.Steps.cs
```diff
@@ -726,7 +819,12 @@ private void RunBootstrappingTests()
 
     private string MapToFolderOutput(TargetFramework targetFramework)
     {
-        return targetFramework.ToString().StartsWith("net4") ? "netfx" : "net";
+        return targetFramework.ToString().StartsWith("net4") ? $"netfx" : "net";
```
RassK:
```suggestion
        return targetFramework.ToString().StartsWith("net4") ? "netfx" : "net";
```

build/TargetFramework.cs
```diff
@@ -9,12 +9,16 @@ public class TargetFramework : Enumeration
 {
     public static readonly TargetFramework NOT_SPECIFIED = new() { Value = string.Empty };
     public static readonly TargetFramework NET462 = new() { Value = "net462" };
+    public static readonly TargetFramework NET47 = new() { Value = "net47" };
+    public static readonly TargetFramework NET471 = new() { Value = "net471" };
+    public static readonly TargetFramework NET472 = new() { Value = "net472" };
```
martincostello, line 14: Do you need `net48` and `net481` too?

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -3643,4 +3650,98 @@ HRESULT STDMETHODCALLTYPE CorProfiler::EventPipeEventDelivered(EVENTPIPE_PROVIDE
...
+                frameworkVersion = 470; // 4.7
+            }
+            else if (releaseValue >= 394802)
+            {
+                frameworkVersion = 462; // 4.6.2
+            }
+            else
+            {
+                Logger::Warn(
+                    "DetectFrameworkVersionTableForRedirectsMap: Old .Net Framework detected, use 462 as fallback");
```
martincostello:
```suggestion
                    "DetectFrameworkVersionTableForRedirectsMap: Old .NET Framework detected, use 462 as fallback");
```

build/Build.Steps.cs
Kielek: `TargetFrameworksForNetFxRedirections`?

build/TargetFramework.cs
iskiselev: Currently we don't need net48 and net481, as all of our dependencies does not have difference between 4.7.2 and higher. I've added new `NetFrameworkDistroTests.ReferencedPackagesNoUnsupportedTfm` test that validates it. If any dependencies will do customizations for 4.8, 4.8.1 or potential newer .Net Framework - it will fail. There is comment in tests what should be done in that case.

build/Build.Steps.cs
nrcventura: `TargetFrameworksForNetFxPacking`?

nrcventura:
> Should we fallback to 4.7.2 instead of 4.6.2 if .Net Framework version not detected?

As was described in the SIG meeting, I think falling back to 4.6.2 makes sense to maintain the existing the behavior. However, I'm wondering if we can improve upon it for those using the install script? Maybe the install script is run with enough permissions to configure an environment variable to modify the fallback behavior if a newer version of .net framework is detected at install time. If so, something like that could be implemented separately.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -3643,4 +3650,98 @@ HRESULT STDMETHODCALLTYPE CorProfiler::EventPipeEventDelivered(EVENTPIPE_PROVIDE
...
 }
 
+#ifdef _WIN32
+void CorProfiler::DetectFrameworkVersionTableForRedirectsMap()
+{
+    // Default to 4.6.2 (462) if detection fails
+    int frameworkVersion = 462;
+
+    HKEY hKey   = nullptr;
+    LONG result = RegOpenKeyExW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\NET Framework Setup\\NDP\\v4\\Full\\", 0,
```
nrcventura: The registry check logic is in multiple locations. Is it possible to just have it checked in one place and made available in both locations?

nrcventura:
> More automation tests? Suggest what to cover.

One idea for a set of integration tests is having a test app that runs in windows containers with only a specific .net framework version available to make it easier to validate behavior for each specific .net framework version. I don't know what type of test app we need to have to ensure that the correct dependencies are in place. I also don't know if these tests will have enough value given how long it can take to download and run these containers.

iskiselev:
> > Should we fallback to 4.7.2 instead of 4.6.2 if .Net Framework version not detected?
> 
> As was described in the SIG meeting, I think falling back to 4.6.2 makes sense to maintain the existing the behavior. However, I'm wondering if we can improve upon it for those using the install script? Maybe the install script is run with enough permissions to configure an environment variable to modify the fallback behavior if a newer version of .net framework is detected at install time. If so, something like that could be implemented separately.

@nrcventura, I will do .Net Framework 4.7.2 detection based on presence of ICorProfilerInfo8. We will fall back to registry check only for 4.6.2, 4.7, 4.7.1. If we ever need support different packing for 4.8 or newer, we will need to reconsider it. On other hand, by that time we probably would not need anything older than 4.7.2 - as 4.6.2 - 4.7.1 probably will be out of support after Jan 12, 2027.
I will also use PInvoke into profiler code to not duplicate registry query.

iskiselev:
> One idea for a set of integration tests is having a test app that runs in windows containers with only a specific .net framework version available to make it easier to validate behavior for each specific .net framework version. I don't know what type of test app we need to have to ensure that the correct dependencies are in place. I also don't know if these tests will have enough value given how long it can take to download and run these containers.

@nrcventura, it may be hard to do it. To test older frameworks, docker should be based on Windows Server 2016 (as Windows Server 2019 has 4.7.2 pre-installed).  As I understand (I'm not sure about it), windows docker in Github CI should match host version - and looks like there is no Windows Server 2016 [runner](https://github.com/actions/runner-images?tab=readme-ov-file#available-images). I also have concerns about value of this tests and whether their development is proper time investment, with limited lifespan of frameworks older than 4.7.2 (both [.Net Framework 4.6.2](https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-framework) and [Windows Server 2016](https://learn.microsoft.com/en-us/lifecycle/products/windows-server-2016) extended support will be finished on January 12, 2027)

CHANGELOG.md
```diff
@@ -39,6 +39,19 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
 - .NET Framework only, following packages updated
   - `OpenTelemetry.Instrumentation.AspNet` from `1.12.0-beta.1` to `1.13.0-beta.2`.
 
+#### Dependencies on .Net Framework
```
Kielek:
```suggestion
#### Dependencies on .NET Framework
```

CHANGELOG.md
```diff
@@ -39,6 +39,19 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
...
+
+When OpenTelemetry .NET AutoInstrumentation is compiled for .NET Framework,
+it uses the net462 Target Framework Moniker (TFM). As a result, the ZIP archive
```
Kielek:
```suggestion
it uses the `net462` Target Framework Moniker (TFM). As a result, the ZIP archive
```

tools/DependencyListGenerator/DotNetOutdated/Services/DotNetPackageListService.cs
```diff
@@ -0,0 +1,26 @@
...
+        _fileSystem = fileSystem;
+    }
+
+    public RunStatus ListPackages(string projectPath)
+    {
+        string[] arguments = new[]
+        {
+            "list", $"\"{projectPath}\"", "package", "--include-transitive", "--format", "json", "--output-version",
+            "1"
+        };
```
Kielek, line 21:
```suggestion
        string[] arguments =
        [
            "list", $"\"{projectPath}\"", "package", "--include-transitive", "--format", "json", "--output-version", "1"
        ];
```

tools/DependencyListGenerator/DotNetOutdated/Models/PackageListModel.cs
```diff
@@ -0,0 +1,51 @@
+namespace DependencyListGenerator.DotNetOutdated.Models;
+
+using System.Text.Json.Serialization;
+
+public record PackageListModel
```
Kielek, line 5: .NET provides some wrappers to play with nugets, and other stuff.
Eg https://www.nuget.org/packages/NuGet.ProjectModel

Is there any chance that there is similar package for this case?

Kielek:
> [!NOTE]
> When merged, close also #3817

test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt
Kielek, line 1: Could you please share the new/old package sizes on Windows?

test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt
```diff
@@ -83,110 +82,220 @@
...
+  \netfx\net471\System.Globalization.Extensions.dll,
+  \netfx\net471\System.IO.Compression.dll,
+  \netfx\net471\System.Net.Http.dll,
+  \netfx\net471\System.Net.Sockets.dll,
+  \netfx\net471\System.Runtime.Serialization.Primitives.dll,
+  \netfx\net471\System.Security.Cryptography.Algorithms.dll,
+  \netfx\net471\System.Security.SecureString.dll,
+  \netfx\net471\System.Threading.Overlapped.dll,
+  \netfx\net471\System.Xml.XPath.XDocument.dll,
+  \netfx\net472\_._,
```
Kielek, line 202: This look strange. What that means?

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -3628,4 +3635,108 @@ HRESULT STDMETHODCALLTYPE CorProfiler::EventPipeEventDelivered(EVENTPIPE_PROVIDE
...
+            {
+                Logger::Warn("DetectFrameworkVersionTableForRedirectsMap: Failed to read Release value, using "
+                             "default version 462");
+            }
+            else
+            {
+                // Map release numbers to framework versions
+                // Based on Microsoft documentation:
+                // https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed
+                if (releaseValue >= 461308)
```
Kielek, line 3836: I would put here all framework versions, up to 4.8.1.
in if statement, we can put something like

```
// intentionally pinining to .NET Framework 4.7.2, there is no differences in the expected library set
frameworkVersion = 472;
```

or similiar.

Now, we understand that it was intentional, in the future we can avoid some archeological stuff.

script-templates/OpenTelemetry.DotNet.Auto.psm1.template
```diff
@@ -308,8 +307,33 @@ function Register-AssembliesInGAC() {
...
     [System.Reflection.Assembly]::Load("System.EnterpriseServices, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a") | Out-Null
-    $publish = New-Object System.EnterpriseServices.Internal.Publish 
-    $dlls = Get-ChildItem -Path $installDir\netfx\ -Filter *.dll -File
+    $publish = New-Object System.EnterpriseServices.Internal.Publish
+
+    $dlls = @()
+
+    $release = Get-ItemPropertyValue -LiteralPath 'HKLM:SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full' -Name Release
+    switch ($release) {
+        { $_ -ge 461808 } { $fxversion = 'net472'; break }
```
Kielek, line 318: I would put all versions also here, similar comment as in cor_profiler.cpp

docs/README.md
```diff
@@ -317,6 +317,14 @@ Unregister-OpenTelemetryForWindowsService -WindowsServiceName MyServiceName
 Uninstall-OpenTelemetryCore
 ```
 
+#### Update .Net Framework version
```
Kielek:
```suggestion
#### Update .NET Framework version
```

Kielek:
>  Should we fallback to 4.7.2 instead of 4.6.2 if .Net Framework version not detected?

Do you know what can be the most common case when this can happen? Do you see any other possibility to check what .NET Framework is executed?

> Decide if we need any replacement for deleted TransientDependenciesTests.DefinedTransientDeps_Are_MatchingGeneratedDeps. Do we need it once we start utilize CentralPackageTransitivePinningEnabled?

Historically, it prevented me at least once to merge wrong PR. If you see possibility to still have it (this is pretty new feature in .NET) it will be great. I think, it can be done in separate PR, if you can commit to finish it;

> We don't have reference/version pinning in global/src level of Directory.Packages.props for some transitive packages. 
Should we have it? Currently they are added as other transitive dependencies in OpenTelemetry.AutoInstrumentation.Assemblies.NetFramework/Directory.Packages.props. We can prevent it by restoring !dep.Name.StartsWith("OpenTelemetry") && dep.Name != "OpenTracing" in Generator, but pinning them through higher level  Directory.Packages.props may be better idea. Packages to discuss:
>  - OpenTelemetry.Api.ProviderBuilderExtensions
 > - OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule
 > - OpenTracing

OpenTelemetry.Api.ProviderBuilderExtensions always has the same version as OpenTelemetry.Api/OpenTelemetry
OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule is always released together with OpenTelemetry.Instrumentation.AspNet. Both of them are versioned together
[OpenTracing](https://www.nuget.org/packages/OpenTracing) is a dead package. I do not expect new release here

I do not have strong opinion here.

> Should we integrate OpenTelemetry.AutoInstrumentation.Assemblies.NetFramework/Directory.Packages.props with dependabot? Priory only dependencies in top-level Directory.Packages.props were processed.

I would say, that we should switch whole repository to renovate first, then make any additional inclusions.
Based on what I see, renovate is much more reliable than dependabot.

tools/DependencyListGenerator/DotNetOutdated/Models/PackageListModel.cs
iskiselev: It is defined at https://github.com/NuGet/NuGet.Client/tree/81733e7572ae9cffc2ff5b3b89d8b398095af864/src/NuGet.Core/NuGet.CommandLine.XPlat/ListPackage ([NuGet.CommandLine.XPlat](https://www.nuget.org/packages/NuGet.CommandLine.XPlat) package), but is is defined as `internal` - using our own types was least evil, especially as json format is versioned.

test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt
iskiselev: We check folder existance as marker of designeted framework. But the folder `net472` has no assemblies in it - all assemblies used on 4.7.2 and up are share across all frameworks. In some cases empty folder may be ommited when extracted, so to preserve this empty folder I put empty `_,_` file in it. The same does nuget in packages for TFMs that are supported but don't need any assemblies.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
iskiselev: With new approach to detect 4.7.2 we will not use this branch on 4.7.2 and up at all. There is comment about it at lines 3644-3646.
I can either:
- extend that comment, specifying that newer framework version detection should be added to registry check
- add other frameworks here as commented code with words that it should not be used due to `ICorProfilerInfo8` check
- add detection with pinnig to 4.7.1 or 4.7.2

nrcventura:
> > > Should we fallback to 4.7.2 instead of 4.6.2 if .Net Framework version not detected?
> > 
> > 
> > As was described in the SIG meeting, I think falling back to 4.6.2 makes sense to maintain the existing the behavior. However, I'm wondering if we can improve upon it for those using the install script? Maybe the install script is run with enough permissions to configure an environment variable to modify the fallback behavior if a newer version of .net framework is detected at install time. If so, something like that could be implemented separately.
> 
> @nrcventura, I will do .Net Framework 4.7.2 detection based on presence of ICorProfilerInfo8. We will fall back to registry check only for 4.6.2, 4.7, 4.7.1. If we ever need support different packing for 4.8 or newer, we will need to reconsider it. On other hand, by that time we probably would not need anything older than 4.7.2 - as 4.6.2 - 4.7.1 probably will be out of support after Jan 12, 2027. I will also use PInvoke into profiler code to not duplicate registry query.

There are no [published support end dates for 4.7 and 4.7.1](https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-framework), so at this point I don't know when we will be able to drop support for those versions ([some of the OSs tied to those versions may be supported through 2032](https://learn.microsoft.com/en-us/lifecycle/faq/dotnet-framework)). But I do suspect, that it will be harder to encounter those systems with only 4.7 or 4.7.1 installed as time goes on. I think the ICorProfilerInfo8 check is a nice workaround and simple check for systems running newer .net framework versions.

src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyResolver.NetFramework.cs
```diff
@@ -50,33 +41,91 @@ private static string ResolveManagedProfilerDirectory()
...
-                logger.Debug(ex, "Assembly.LoadFrom(\"{0}\") Exception: {1}", path, ex.Message);
+                _logger.Debug(ex, "Assembly.LoadFrom(\"{0}\") Exception: {1}", path, ex.Message);
             }
         }
 
         return null;
     }
+
+    [DllImport("OpenTelemetry.AutoInstrumentation.Native.dll")]
+    private static extern int GetDetectedNetFrameworkVersion();
```
nrcventura: Should we add a comment here that indicates that it will only detect versions up to 4.7.2?

test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt
iskiselev: Od size 7,415,386 bytes (from #4560 windows build)
New size 7,838,414 bytes (from windows build of this PR after latest master merged)

So, windows zip size is increasing by 6% or 0.4 MB.

test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt
iskiselev: It is generated at https://github.com/iskiselev/opentelemetry-dotnet-instrumentation/blob/8a75f87ed123a481f585aaa6c42caf6cab23f68d/build/Build.Steps.cs#L445

src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyResolver.NetFramework.cs
iskiselev: Renamed method and added comment.

script-templates/OpenTelemetry.DotNet.Auto.psm1.template
```diff
@@ -308,8 +307,36 @@ function Register-AssembliesInGAC() {
...
+    $publish = New-Object System.EnterpriseServices.Internal.Publish
+
+    $dlls = @()
+
+    $release = Get-ItemPropertyValue -LiteralPath 'HKLM:SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full' -Name Release
+    switch ($release) {
+        { $_ -ge 533320 } { $fxversion = 'net472'; break } # net481 or later, pin to 4.7.2
+        { $_ -ge 528040 } { $fxversion = 'net472'; break } # net48, pin to 4.7.2
+        { $_ -ge 461808 } { $fxversion = 'net472'; break }
+        { $_ -ge 461808 } { $fxversion = 'net472'; break }
```
Kielek:
```suggestion
        { $_ -ge 461808 } { $fxversion = 'net472'; break }
```

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
iskiselev: Updated comment in front of if/else block that does ICorProfilerInfo8 check for 4.7.2 and up.

iskiselev:
> > Should we fallback to 4.7.2 instead of 4.6.2 if .Net Framework version not detected?
> 
> Do you know what can be the most common case when this can happen? Do you see any other possibility to check what .NET Framework is executed?

@Kielek, I've reworked .NET Framework detection so that it is not using registry check for 4.7.2 (and up) check. That is most common case. Managed part of loader will use PInvoke to check resolved framework from profiler.
So, registry check is used only to distinguish between 4.6.2, 4.7 and 4.7.1 with fallback to 4.6.2 in runtime (it is still used at install time to detect 4.7.2).
Fallback is bad situation, as we never registered in GAC assemblies that may be used in lower TFM, but Loader should help in resolving them in most case - and in fact, we are talking about extremly rare sitation - only 4.7 and 4.7.1 with restrictions on registry access may be affected.

nrcventura:
I think the last remaining question on this PR is the following.

> > Decide if we need any replacement for deleted TransientDependenciesTests.DefinedTransientDeps_Are_MatchingGeneratedDeps. Do we need it once we start utilize CentralPackageTransitivePinningEnabled?
> 
> Historically, it prevented me at least once to merge wrong PR. If you see possibility to still have it (this is pretty new feature in .NET) it will be great. I think, it can be done in separate PR, if you can commit to finish it;

I don't care if it is a part of this PR or not. If we decide to handle it in a separate PR, I want an issue submitted to track it.

iskiselev:
> > Decide if we need any replacement for deleted TransientDependenciesTests.DefinedTransientDeps_Are_MatchingGeneratedDeps. Do we need it once we start utilize CentralPackageTransitivePinningEnabled?
> 
> Historically, it prevented me at least once to merge wrong PR. If you see possibility to still have it (this is pretty new feature in .NET) it will be great. I think, it can be done in separate PR, if you can commit to finish it;

@Kielek and @nrcventura. I've added new test `ValidateTransientDependenciesPackageVersionsDefined` and think that together with `GeneratorDiscoversTransientDependencies` we have not loss any coverage with `DefinedTransientDeps_Are_MatchingGeneratedDeps` removal.

Previously, every transisient dependency have to be declared twice: as `PackageReference` and `PackageVersion`. `PackageVersion` alone was not enough to change version of transient dependency. With `CentralPackageTransitivePinningEnabled` we don't need to add `PackageReference`, as `PackageVersion` alone enough even for transient dependency.

Removed test logic: 
(1) read list of transient dependency (`PackageReference`) from project.
(2) remove transient dependency (`PackageReference`) from project.
(3) build list of transient dependencies
(4) validate that list from (1) and (3) are the same.

So, test validated:
(1) that dependency list generator is working and produce some set of dependencies
(2) we have not forget to run build and all dependencies in fact listed as `PackageReference` in project

Now:
- `GeneratorDiscoversTransientDependencies` validates that `System.Memory` transient dependency was discovered - so our dependency generator works and produce some set of dependencies (1). 
- `ValidateTransientDependenciesPackageVersionsDefined` validates that for every dependency dicovered by generator we have defined `PackageVersion` in `Directory.Packages.props` hierarchy (2).

So, both things that were validated before are still validated now.

PR #4507, opened 2025-10-07
Configuration based instrumentation - attributes

Kielek:
## Why

Towards #4132

## What

Configuration based instrumentation - attributes

## Tests

CI + extended tests.

## Checklist


src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/NoCodeAttribute.cs
ysolomchenko, line 1: Why not reuse `ResourceAttribute.cs`? And why `name` or `value` can be `null`?

Kielek: I think that it can be unified, when Resources Attributes will be implementing type-parsing functionality. It is internationally out of scope of this PR.

`null` values - it is possible that users will define it in this way. It will be filter out by the Parse method.

nrcventura (approved):

PR #4515, opened 2025-10-09
Configuration based instrumentation - documentation

### LINKED ISSUE #4132: Configuration based Bytecode Instrumentation

Kielek:
## Why

Fixes #4132

## What

Provide documentation for Configuration based instrumentation 

## Tests

N/A

## Checklist

### DIFF
--- a/.cspell/other.txt
+++ b/.cspell/other.txt
@@ -49,6 +49,7 @@ OPERATINGSYSTEM
 ORACLEMDA
 OTEL_DOTNET_AUTO_FLUSH_ON_UNHANDLEDEXCEPTION
 OWIN
+parameterless
 pipefail
 portfile
 PROCESSRUNTIME
@@ -58,9 +59,9 @@ protos
 RABBITMQ
 Serilog
 spdlog
-srcs
 SQLCLIENT
 sqlserver
+srcs
 STACKEXCHANGEREDIS
 TMPDIR
 tracesexporter


docs/nocode-instrumentation.md
```diff
@@ -0,0 +1,441 @@
...
+Each target specifies a method to instrument:
+
+- **`assembly.name`**: The name of the assembly containing the target method
+- **`type`**: The fully qualified type name containing the method
+- **`method`**: The method name to instrument
+- **`signature`**: Method signature specification for precise targeting
+  - **`return_type`**: The method's return type (e.g., `System.Void`,
+    `System.String`, `System.Threading.Tasks.Task`)
+  - **`parameter_types`**: Array of parameter types (empty array for
+    parameterless methods)
```
ysolomchenko: I think it’s better to write comments with explanations directly inside the YAML. IMO it’s more readable when the description appears right in the usage example rather than separately, since otherwise you have to look for the explanation below instead of seeing it immediately.

docs/nocode-instrumentation.md
```diff
@@ -0,0 +1,441 @@
...
+
+When specifying return types in the signature, use these common .NET type names:
+
+- **Void methods**: `System.Void`
+- **Basic types**: `System.String`, `System.Int32`, `System.Boolean`, `System.Double`
+- **Task methods**: `System.Threading.Tasks.Task`
+- **Task with return**: `System.Threading.Tasks.Task\`1[ReturnType]`
+- **ValueTask methods**: `System.Threading.Tasks.ValueTask`
+- **ValueTask with return**: `System.Threading.Tasks.ValueTask\`1[ReturnType]`
+- **Custom classes**: `YourNamespace.YourClassName`
```
ysolomchenko, line 431: IMO, there are too many identical examples, just with different `return_types`. Maybe it would be better to make a single section about this, adding descriptions like the ones here?

The only thing missing in this section is `Generic methods`.

docs/nocode-instrumentation.md
```diff
@@ -0,0 +1,441 @@
...
+no_code/development:
+  targets:
+    - target:
+        assembly:
+          name: TestApplication.NoCode
+        type: TestApplication.NoCode.NoCodeTestingClass
+        method: TestMethodStatic
+        signature:
+          return_type: System.Void
+          parameter_types:
```
nrcventura, line 110: I think that it will be helpful if we also showed the corresponding method being instrumented in C#. I have seem many people get confused with  how to map code from C# or VB to the formal dotnet types that are required to be defined here (especially generic types). Maybe also including a note about using a tool such as ilspy (or similar tool) to view the IL type definition to help with knowing how to define these values.

Kielek: I have included some notes related to ILSpy, please let me know, if it is enough in your opinion.

docs/nocode-instrumentation.md
Kielek: All comments inline with the first entry.

docs/nocode-instrumentation.md
Kielek: I have reviewed examples. I agree that most of them is redundant, but I would keep them for easier reference from the end-users.

PR #4516, opened 2025-10-09
chore: Build and test macOS using macOS 14

zacharycmontoya:
## Why

[The GitHub macOS 13 runner is being shut down.](https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/)

Fixes #

## What

Stops building and testing on macOS 13 runners, and uses macOS 14 runners instead since that is the [next lowest macOS version support by .NET 8 and .NET 9](https://learn.microsoft.com/en-us/dotnet/core/install/macos).

This PR also stops producing osx-x64 assets and moving forward only produces osx-arm64 assets.

## Tests

The existing integration tests will validate the migration.

## Checklist


Kielek (commented):
I think that we need to put a note about this change in the CHANGELOG.
Last note about similar changes: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/9f3ee7800a456a1446b08fc3443e22df50a26cef/CHANGELOG.md#removed-2

Other parts LGTM

RassK (approved):

zacharycmontoya:
I've added a changelog entry in d2342cd1fe856cefee4a319a693ee66d4e6f6736 to indicate that we've removed support for macOS 13

Kielek (approved):

PR #4529, opened 2025-10-16
Inject startup hook assembly if DOTNET_STARTUP_HOOKS not configured

### LINKED ISSUE #4491: Make setting DOTNET_STARTUP_HOOKS optional
# Feature Request
To use the auto instrumentation customers currently have to set both the profiler environment variables as well as the DOTNET_STARTUP_HOOKS environment variable. This really does not add any additional value to setup and is just a redundant step. The profiler could just ensure that the auto instrumentation initialization occurs.

Since we are already loading the auto instrumentation profiler, it can hook into a well-known method and inject code to ensure that the exact same startup behavior happens, but without setting the DOTNET_STARTUP_HOOKS environment variable.
This could happen in, for instance
- [System.StartupHookProvider.ManagedStartup](https://github.com/dotnet/runtime/blob/main/src/coreclr/System.Private.CoreLib/src/System/StartupHookProvider.CoreCLR.cs): Modify diagnosticsStartupHooks parameter by adding OpenTelemetry assembly
- [System.StartupHookProvider.ProcessStartupHooks](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/StartupHookProvider.cs): Similar to above
- [System.AppContext.GetData](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/AppContext.cs): Modify returned data for "STARTUP_HOOKS"

There aren't really alternatives to this. Either the DOTNET_STARTUP_HOOKS setting is required or with this change it is not.

Rhynier:
Resolves #4491

## Why

To use the auto instrumentation customers currently have to set both the profiler environment variables as well as the DOTNET_STARTUP_HOOKS environment variable. This really does not add any additional value to setup and is just a redundant step. The profiler could just ensure that the auto instrumentation initialization occurs.

## Checklist

### DIFF
--- a/.cspell/dot-net.txt
+++ b/.cspell/dot-net.txt
@@ -2,6 +2,7 @@ aspnet
 ASPNETCORE
 ASSERTE
 autoinstrumentation
+Blazor
 Bootstrapper
 buildtasks
 callq
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
 ### Added
 
 - Support for .NET 10.
+- Support for [ASP.NET Core 10 metrics](https://learn.microsoft.com/en-us/aspnet/core/log-mon/metrics/built-in?view=aspnetcore-10.0).
+- Support for ASP.NET Core 10 Blazor traces from
+  `Microsoft.AspNetCore.Components`
+  and `"Microsoft.AspNetCore.Components.Server.Circuits`.
 - Experimental support for file-based configuration.
 - Experimental support for configuration based instrumentation.
 - IL rewrite for SqlCommand on .NET Framework to ensure `CommandText` is
--- a/next-gen/docs/getting-started-dotnet-monitor-metrics/README.md
+++ b/next-gen/docs/getting-started-dotnet-monitor-metrics/README.md
@@ -190,14 +190,14 @@ powerful visualizations for the metrics.
 
 ### Explore metrics using Grafana
 
-[Install Grafana](https://grafana.com/docs/grafana/latest/installation/).
+[Install Grafana](https://grafana.com/docs/grafana/latest/setup-grafana/installation/).
 
 Start the standalone Grafana server (`grafana-server.exe` or
 `./bin/grafana-server`, depending on the operating system). Then, use the
 browser to navigate to [http://localhost:3000/](http://localhost:3000/).
 
 Follow the instructions in the Grafana getting started
-[doc](https://grafana.com/docs/grafana/latest/getting-started/getting-started/#step-2-log-in)
+[doc](https://grafana.com/docs/grafana/latest/fundamentals/getting-started/first-dashboards/#sign-in-to-grafana)
 to log in.
 
 After successfully logging in, click on the explore option on the left panel of
--- a/src/OpenTelemetry.AutoInstrumentation/Configurations/EnvironmentConfigurationMetricHelper.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Configurations/EnvironmentConfigurationMetricHelper.cs
@@ -42,12 +42,21 @@ internal static class EnvironmentConfigurationMetricHelper
 
 #if NET
                 MetricInstrumentation.AspNetCore => builder
-                    .AddMeter("Microsoft.AspNetCore.Hosting")
-                    .AddMeter("Microsoft.AspNetCore.Server.Kestrel")
-                    .AddMeter("Microsoft.AspNetCore.Http.Connections")
-                    .AddMeter("Microsoft.AspNetCore.Routing")
-                    .AddMeter("Microsoft.AspNetCore.Diagnostics")
-                    .AddMeter("Microsoft.AspNetCore.RateLimiting"),
+                    .AddMeter(
+                        "Microsoft.AspNetCore.Hosting",
+                        "Microsoft.AspNetCore.Server.Kestrel",
+                        "Microsoft.AspNetCore.Http.Connections",
+                        "Microsoft.AspNetCore.Routing",
+                        "Microsoft.AspNetCore.Diagnostics",
+                        "Microsoft.AspNetCore.RateLimiting",
+                        // Following metrics added in .NET 10
+                        "Microsoft.AspNetCore.Components",
+                        "Microsoft.AspNetCore.Components.Server.Circuits",
+                        "Microsoft.AspNetCore.Components.Lifecycle",
+                        "Microsoft.AspNetCore.Authorization",
+                        "Microsoft.AspNetCore.Authentication",
+                        "Microsoft.AspNetCore.Identity",
+                        "Microsoft.AspNetCore.MemoryPool"),
 #endif
                 MetricInstrumentation.SqlClient => Wrappers.AddSqlClientInstrumentation(builder, lazyInstrumentationLoader, pluginManager),
                 _ => null,
--- a/src/OpenTelemetry.AutoInstrumentation/Configurations/EnvironmentConfigurationTracerHelper.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Configurations/EnvironmentConfigurationTracerHelper.cs
@@ -283,7 +283,11 @@ internal static class EnvironmentConfigurationTracerHelper
         public static TracerProviderBuilder AddAspNetCoreInstrumentation(TracerProviderBuilder builder, PluginManager pluginManager, LazyInstrumentationLoader lazyInstrumentationLoader, TracerSettings tracerSettings)
         {
             DelayedInitialization.Traces.AddAspNetCore(lazyInstrumentationLoader, pluginManager, tracerSettings);
-            return builder.AddSource("Microsoft.AspNetCore");
+            return builder.AddSource(
+                "Microsoft.AspNetCore",
+                // Blazor activities first added in .NET 10.0
+                "Microsoft.AspNetCore.Components",
+                "Microsoft.AspNetCore.Components.Server.Circuits");
         }
 
         public static TracerProviderBuilder AddGraphQLInstrumentation(TracerProviderBuilder builder, PluginManager pluginManager, LazyInstrumentationLoader lazyInstrumentationLoader, TracerSettings tracerSettings)
--- a/test/IntegrationTests/HttpTests.cs
+++ b/test/IntegrationTests/HttpTests.cs
@@ -45,6 +45,16 @@ public class HttpTests : TestHelper
             return true;
         });
 
+#if NET10_0_OR_GREATER
+        collector.Expect("Microsoft.AspNetCore.Components");
+        // Note: Microsoft.AspNetCore.Components.Server.Circuits metrics require the blazor.server.js
+        // JavaScript client to establish a proper Blazor Server circuit with component rendering.
+        // This cannot be reliably tested in a programmatic test without browser automation.
+        // The meter is registered in EnvironmentConfigurationMetricHelper but circuit creation
+        // requires the full Blazor SignalR protocol handshake performed by blazor.server.js.
+        // collector.Expect("Microsoft.AspNetCore.Components.Server.Circuits");
+#endif
+
         SetEnvironmentVariable("OTEL_PROPAGATORS", propagators);
         SetEnvironmentVariable("DISABLE_DistributedContextPropagator", "true");
         RunTestApplication();
@@ -138,6 +148,22 @@ public class HttpTests : TestHelper
         collector.Expect("Microsoft.AspNetCore.Routing");
         collector.Expect("Microsoft.AspNetCore.Diagnostics");
         collector.Expect("Microsoft.AspNetCore.RateLimiting");
+
+#if NET10_0_OR_GREATER
+        collector.Expect("Microsoft.AspNetCore.Components");
+        // Note: Microsoft.AspNetCore.Components.Server.Circuits metrics require the blazor.server.js
+        // JavaScript client to establish a proper Blazor Server circuit with component rendering.
+        // This cannot be reliably tested in a programmatic test without browser automation.
+        // The meter is registered in EnvironmentConfigurationMetricHelper but circuit creation
+        // requires the full Blazor SignalR protocol handshake performed by blazor.server.js.
+        // collector.Expect("Microsoft.AspNetCore.Components.Server.Circuits");
+        collector.Expect("Microsoft.AspNetCore.Components.Lifecycle");
+        collector.Expect("Microsoft.AspNetCore.Authorization");
+        collector.Expect("Microsoft.AspNetCore.Authentication");
+        collector.Expect("Microsoft.AspNetCore.Identity");
+        collector.Expect("Microsoft.AspNetCore.MemoryPool");
+#endif
+
         collector.ExpectAdditionalEntries(x => x.All(m => m.InstrumentationScopeName != "OpenTelemetry.Instrumentation.Http"));
 
         RunTestApplication();
--- a/test/test-applications/integrations/TestApplication.Http/App.razor
+++ b/test/test-applications/integrations/TestApplication.Http/App.razor
@@ -1,0 +1,8 @@
+<Router AppAssembly="@typeof(Program).Assembly">
+    <Found Context="routeData">
+        <RouteView RouteData="@routeData" DefaultLayout="@typeof(MainLayout)" />
+    </Found>
+    <NotFound>
+        <p>Sorry, there's nothing at this address.</p>
+    </NotFound>
+</Router>
--- a/test/test-applications/integrations/TestApplication.Http/InMemoryUserStore.cs
+++ b/test/test-applications/integrations/TestApplication.Http/InMemoryUserStore.cs
@@ -1,0 +1,100 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#if NET10_0_OR_GREATER
+using Microsoft.AspNetCore.Identity;
+
+namespace TestApplication.Http;
+
+/// <summary>
+/// In-memory user store for testing Identity metrics
+/// </summary>
+public class InMemoryUserStore : IUserStore<TestUser>, IUserPasswordStore<TestUser>
+{
+    private readonly Dictionary<string, TestUser> _users = new();
+    private readonly Dictionary<string, string> _passwords = new();
+
+    public Task<IdentityResult> CreateAsync(TestUser user, CancellationToken cancellationToken)
+    {
+        _users[user.UserName!] = user;
+        return Task.FromResult(IdentityResult.Success);
+    }
+
+    public Task<IdentityResult> DeleteAsync(TestUser user, CancellationToken cancellationToken)
+    {
+        _users.Remove(user.UserName!);
+        _passwords.Remove(user.Id);
+        return Task.FromResult(IdentityResult.Success);
+    }
+
+    public Task<TestUser?> FindByIdAsync(string userId, CancellationToken cancellationToken)
+    {
+        var user = _users.Values.FirstOrDefault(u => u.Id == userId);
+        return Task.FromResult(user);
+    }
+
+    public Task<TestUser?> FindByNameAsync(string normalizedUserName, CancellationToken cancellationToken)
+    {
+        _users.TryGetValue(normalizedUserName, out var user);
+        return Task.FromResult(user);
+    }
+
+    public Task<string?> GetNormalizedUserNameAsync(TestUser user, CancellationToken cancellationToken)
+    {
+        return Task.FromResult(user.NormalizedUserName);
+    }
+
+    public Task<string> GetUserIdAsync(TestUser user, CancellationToken cancellationToken)
+    {
+        return Task.FromResult(user.Id);
+    }
+
+    public Task<string?> GetUserNameAsync(TestUser user, CancellationToken cancellationToken)
+    {
+        return Task.FromResult(user.UserName);
+    }
+
+    public Task SetNormalizedUserNameAsync(TestUser user, string? normalizedName, CancellationToken cancellationToken)
+    {
+        user.NormalizedUserName = normalizedName;
+        return Task.CompletedTask;
+    }
+
+    public Task SetUserNameAsync(TestUser user, string? userName, CancellationToken cancellationToken)
+    {
+        user.UserName = userName;
+        return Task.CompletedTask;
+    }
+
+    public Task<IdentityResult> UpdateAsync(TestUser user, CancellationToken cancellationToken)
+    {
+        _users[user.UserName!] = user;
+        return Task.FromResult(IdentityResult.Success);
+    }
+
+    public void Dispose()
+    {
+    }
+
+    public Task SetPasswordHashAsync(TestUser user, string? passwordHash, CancellationToken cancellationToken)
+    {
+        if (passwordHash != null)
+        {
+            _passwords[user.Id] = passwordHash;
+        }
+
+        return Task.CompletedTask;
+    }
+
+    public Task<string?> GetPasswordHashAsync(TestUser user, CancellationToken cancellationToken)
+    {
+        _passwords.TryGetValue(user.Id, out var hash);
+        return Task.FromResult(hash);
+    }
+
+    public Task<bool> HasPasswordAsync(TestUser user, CancellationToken cancellationToken)
+    {
+        return Task.FromResult(_passwords.ContainsKey(user.Id));
+    }
+}
+#endif
--- a/test/test-applications/integrations/TestApplication.Http/Pages/Counter.razor
+++ b/test/test-applications/integrations/TestApplication.Http/Pages/Counter.razor
@@ -1,0 +1,16 @@
+@page "/counter"
+
+<h1>Counter</h1>
+
+<p>Current count: @currentCount</p>
+
+<button class="btn btn-primary" @onclick="IncrementCount">Click me</button>
+
+@code {
+    private int currentCount = 0;
+
+    private void IncrementCount()
+    {
+        currentCount++;
+    }
+}
--- a/test/test-applications/integrations/TestApplication.Http/Pages/_Host.cshtml
+++ b/test/test-applications/integrations/TestApplication.Http/Pages/_Host.cshtml
@@ -1,0 +1,16 @@
+@page "/blazor"
+@namespace TestApplication.Http.Pages
+@addTagHelper *, Microsoft.AspNetCore.Mvc.TagHelpers
+
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8" />
+    <title>Test Blazor</title>
+    <base href="~/" />
+</head>
+<body>
+    <component type="typeof(App)" render-mode="ServerPrerendered" />
+    <script src="_framework/blazor.server.js"></script>
+</body>
+</html>
--- a/test/test-applications/integrations/TestApplication.Http/Program.cs
+++ b/test/test-applications/integrations/TestApplication.Http/Program.cs
@@ -33,6 +33,25 @@ public class Program
         httpClient.DefaultRequestHeaders.Add("Custom-Request-Test-Header3", "Test-Value3");
         httpClient.GetAsync($"{dnsAddress}/test").Wait();
         httpClient.GetAsync($"{dnsAddress}/exception").Wait();
+
+        // Trigger authentication metrics for .NET 10+
+        httpClient.GetAsync($"{dnsAddress}/login").Wait();
+        httpClient.GetAsync($"{dnsAddress}/logout").Wait();
+
+        // Trigger authorization metrics for .NET 10+
+        httpClient.GetAsync($"{dnsAddress}/protected").Wait();
+
+#if NET10_0_OR_GREATER
+        // Trigger Blazor Components metrics for .NET 10+
+        // This will trigger Microsoft.AspNetCore.Components and Microsoft.AspNetCore.Components.Server.Circuits metrics
+        httpClient.GetAsync($"{dnsAddress}/blazor").Wait();
+
+        // Trigger Identity metrics for .NET 10+
+        // This will trigger Microsoft.AspNetCore.Identity metrics
+        httpClient.GetAsync($"{dnsAddress}/identity/create-user").Wait();
+        httpClient.GetAsync($"{dnsAddress}/identity/find-user").Wait();
+#endif
+
         var hubConnection = new HubConnectionBuilder().WithUrl($"{dnsAddress}/signalr").Build();
         hubConnection.StartAsync().Wait();
         hubConnection.StopAsync().Wait();
--- a/test/test-applications/integrations/TestApplication.Http/Shared/MainLayout.razor
+++ b/test/test-applications/integrations/TestApplication.Http/Shared/MainLayout.razor
@@ -1,0 +1,5 @@
+@inherits LayoutComponentBase
+
+<div>
+    @Body
+</div>
--- a/test/test-applications/integrations/TestApplication.Http/Startup.cs
+++ b/test/test-applications/integrations/TestApplication.Http/Startup.cs
@@ -2,6 +2,15 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System.Diagnostics;
+using System.Security.Claims;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Authentication.Cookies;
+using Microsoft.AspNetCore.Authorization;
+#if NET10_0_OR_GREATER
+using System.Diagnostics.Metrics;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+#endif
 
 namespace TestApplication.Http;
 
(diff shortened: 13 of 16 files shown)


src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -3450,6 +3568,301 @@ HRESULT CorProfiler::AddIISPreStartInitFlags(const ModuleID module_id, const mdT
     return S_OK;
 }
 
+// clang-format off
+// This method will generate new type __DDLoaderFixup__ in target module.
```
iskiselev: Should we use new prefixm not DD for generated code?

src/OpenTelemetry.AutoInstrumentation.Native/pal.h
```diff
@@ -132,6 +136,52 @@ inline WSTRING GetCurrentModuleFileName()
...
 
+inline WSTRING GetModuleFilePath(const WSTRING& moduleName)
+{
+    const WSTRING currentModuleFileName = GetCurrentModuleFileName();
+    if (currentModuleFileName == EmptyWStr)
+    {
+        return EmptyWStr;
+    }
+
+#ifdef _WIN32
```
iskiselev: Are we linking old stdlib on linux that doesn't have path?

src/OpenTelemetry.AutoInstrumentation.Native/pal.h
```diff
@@ -132,6 +136,52 @@ inline WSTRING GetCurrentModuleFileName()
     return EmptyWStr;
 }
 
+inline WSTRING GetModuleFilePath(const WSTRING& moduleName)
```
iskiselev: As I understand, there are at least 3 possible layouts:
```
ZIP:
OTEL_HOME/<RID>/<native_profler_lib>
OTEL_HOME/net/OpenTelemetry.AutoInstrumentation.StartupHook.dll
```

```
NuGet,  platform dependent deployment:
<APP_FOLDER>/<native_profler_lib>
<APP_FOLDER>/OpenTelemetry.AutoInstrumentation.StartupHook.dll
```

```
NuGet,  platform independent deployment: (I suspect this is not validated correctly in IsStartupHookValid)
<APP_FOLDER>/runtimes/<RID>/<native_profler_lib>
<APP_FOLDER>/OpenTelemetry.AutoInstrumentation.StartupHook.dll
```

I'm also not sure, what is behaviour if OTEL_HOME envirinment variable not set, especially in case of zip deployment.

In any case, once we calculated the path - we should probe that file exists.

iskiselev: Probably, we should apply manual add of profiler hook only if OpenTelemetry.AutoInstrumentation.StartupHook.dll is not present in environment variable at all. If it is there, but points to unexpected location, we should just log it.
I really hope that noone will rename the startup hook file :)

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -3450,6 +3568,301 @@ HRESULT CorProfiler::AddIISPreStartInitFlags(const ModuleID module_id, const mdT
...
+
+    const auto& metadata_import = metadata_interfaces.As<IMetaDataImport2>(IID_IMetaDataImport);
+    const auto& metadata_emit   = metadata_interfaces.As<IMetaDataEmit2>(IID_IMetaDataEmit);
+    const auto& assembly_emit   = metadata_interfaces.As<IMetaDataAssemblyEmit>(IID_IMetaDataAssemblyEmit);
+
+    MemberResolver resolver(metadata_import, metadata_emit);
+
+    mdAssemblyRef corlib_ref = mdTokenNil;
+    // We need assemblyRef only when we generate type outside of mscorlib
+    if (module_info.assembly.name != mscorlib_assemblyName)
```
iskiselev: You never need corlib_ref, as that modification should be done inside corlib, as I understand - similat to ModifyAppDomainCreate.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -3450,6 +3568,301 @@ HRESULT CorProfiler::AddIISPreStartInitFlags(const ModuleID module_id, const mdT
...
+        {
+            LPCWSTR path_separator_str = ENV_VAR_PATH_SEPARATOR_STR;
+            auto    path_separator_str_size = wcslen(path_separator_str);
+
+            hr = metadata_emit->DefineUserString(path_separator_str,
+                                                 (ULONG)path_separator_str_size,
+                                                 &path_separator_token);
+            if (FAILED(hr))
+            {
+                Logger::Warn("GenerateHookFixup: DefineUserString ; failed");
```
iskiselev: Ptobably typo "; "

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -3450,6 +3568,301 @@ HRESULT CorProfiler::AddIISPreStartInitFlags(const ModuleID module_id, const mdT
...
+                return hr;
+            }
+        }
+
+        // Generate IL for the method body
+        // C# equivalent:
+        // if (startupHooks == null)
```
iskiselev: It may be helpfull to add IL code you are genereting. When I used same body, I see a little bit different code:
```
IL_0000 | ldarg.0 |  
IL_0001 | ldind.ref |  
IL_0002 | brtrue.s | IL_000C
IL_0004 | ldarg.0 |  
IL_0005 | ldstr | "<startup_hook_dll_name>"
IL_000A | stind.ref |  
IL_000B | ret |  
IL_000C | ldarg.0 |  
IL_000D | ldarg.0 |  
IL_000E | ldind.ref |  
IL_000F | ldstr | "<path_separator><startup_hook_dll_name>"
IL_0014 | call | String.Concat (String, String)
IL_0019 | stind.ref |  
IL_001A | ret
```

It may be eisier to verify if you target IL would be also visible.

src/OpenTelemetry.AutoInstrumentation.Native/otel_profiler_constants.h
```diff
@@ -66,6 +66,11 @@ const WSTRING skip_assemblies[]{WStr("mscorlib"),
 const WSTRING mscorlib_assemblyName = WStr("mscorlib");
 const WSTRING system_private_corelib_assemblyName = WStr("System.Private.CoreLib");
 const WSTRING opentelemetry_autoinstrumentation_loader_assemblyName = WStr("OpenTelemetry.AutoInstrumentation.Loader");
+#ifdef _WIN32
+const WSTRING opentelemetry_autoinstrumentation_startuphook_filepath = WStr("net\\OpenTelemetry.AutoInstrumentation.StartupHook.dll");
+#else
+const WSTRING opentelemetry_autoinstrumentation_startuphook_filepath = WStr("net/OpenTelemetry.AutoInstrumentation.StartupHook.dll");
```
iskiselev: Should we use that constant in IsStartupHookValid if we define it?

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -2048,6 +2078,94 @@ HRESULT CorProfiler::ModifyAppDomainCreate(const ModuleID module_id, mdMethodDef
...
+        if (FAILED(hr))
+        {
+            Logger::Warn("ModifyProcessStartupHooks: ILRewriter.Import System.StartupHookProvider::ProcessStartupHooks failed");
+            return hr;
+        }
+
+        ILInstr* pFirstInstr = rewriter.GetILList()->m_pNext;
+        ILInstr* pNewInstr   = NULL;
+
+        // ldarga.s 0  ; Load the address of the first argument (startupHooks string)
```
iskiselev: Would it be easier to do all checks/modifications here, instead of creating additional class / method?

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
Rhynier: I thought __DD was some kind of convention, not arbitrarily chosen. If the idea is to have a random pair of letters as prefix I can change it.

src/OpenTelemetry.AutoInstrumentation.Native/pal.h
Rhynier: This functionality is definitely just a placeholder for now in order to test that the overall idea works. I will update it to be more robust and take different layouts, etc. into consideration.

Regarding the second question, that is already done in the PR as it is right now. The `startup_fix_required` class member variable was added for that purpose.

src/OpenTelemetry.AutoInstrumentation.Native/pal.h
Rhynier: I don't know. Will have to look into this. Are you saying that older stdlib versions returned null or that they returned the module name without path?

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
Rhynier: I'll will add the comment, making sure it matches the generated code.

src/OpenTelemetry.AutoInstrumentation.Native/otel_profiler_constants.h
Rhynier: Good point. I will update the code to use the same constants.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
Rhynier: In theory, might we not have issues with method body resizing if we add too much code? I know there are things to be aware of but don't know the ins and outs as well as you do.

Other than that it is a good idea. I had originally planned to have more logic in the generated method but then realized I can resolve the assembly name in C++ and just hardcode it in the IL.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
iskiselev: DD=DataDog

src/OpenTelemetry.AutoInstrumentation.Native/pal.h
iskiselev: It was a question, why can't we use std:path on Linux also?

src/OpenTelemetry.AutoInstrumentation.Native/pal.h
iskiselev: I've seen it. Right now startup_fix_required is based on IsStartupHookValid. We probably need it to be even more flexible.
There are situations, that we may emit warning, as we have not identified hook path in environment varaible, but don't want to apply patch, as we still not sure - like we have dll name in env, but in unexpected path. Keep in mind, on windows check should be case insensitive.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
iskiselev: I'm not aware of any issues with changing method body size.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -177,7 +177,8 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
         const auto startup_hooks = GetEnvironmentValues(environment::dotnet_startup_hooks, ENV_VAR_PATH_SEPARATOR);
         if (!IsStartupHookValid(startup_hooks, home_path))
         {
-            FailProfiler(Error, "The required StartupHook was not configured correctly. No telemetry will be captured.")
+            Logger::Info("The StartupHook was not configured. Will patch ProcessStartupHooks.");
```
RassK, line 168: Is it too late to inject process level env in this point?
Just looking all that code that is necessary if there are simpler workarounds.

Rhynier: Unfortunately the CLR loads environment variables and caches them before it initializes the profiler. Injection is the only way we can overcome that.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
Rhynier: It is used in GenerateLoaderType to generate code. I am not quite following why it isn't needed to generate the startup hook patch method.

src/OpenTelemetry.AutoInstrumentation.Native/pal.h
Rhynier: Before this change, if `IsStartupHookValid` returns `false`, the profiler would not be loaded and an error message would be logged. I don't think we should change the behavior to be more flexible than that.
In other words, we check `IsStartupHookValid` and, only if it is `false` do we try to inject the startup hook patch. And if we cannot find a suitable startup hook DLL based on any of the 3 patterns above, the profiler load should fail with an error message logged.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
Rhynier: Had a conversation and addressed this.

src/OpenTelemetry.AutoInstrumentation.Native/stub_generator.cpp
```diff
@@ -0,0 +1,449 @@
...
+        ILInstr* pFirstInstr = rewriter.GetILList()->m_pNext;
+        ILInstr* pNewInstr   = NULL;
+
+        // ldarga.s 0  ; Load the address of the first argument (startupHooks string)
+        pNewInstr           = rewriter.NewILInstr();
+        pNewInstr->m_opcode = CEE_LDARGA_S;
+        pNewInstr->m_Arg8   = 0;
+        rewriter.InsertBefore(pFirstInstr, pNewInstr);
+
+        // call __DDLoaderFixup__::__DDPatchStartupHookValue__
```
nrcventura:
```suggestion
        // call __OTLoaderFixup__::__OTPatchStartupHookValue__
```

Rhynier:
> These changes look good to me other than a minor naming fix in a comment to match the names used for the injected type and method name.

Thank you very much, Chris, for reviewing the PR.

src/OpenTelemetry.AutoInstrumentation.Native/file_utils.h
```diff
@@ -0,0 +1,27 @@
...
+
+#ifndef OTEL_CLR_PROFILER_FILE_UTILS_H_
+#define OTEL_CLR_PROFILER_FILE_UTILS_H_
+
+#include <type_traits>
+
+#include "string_utils.h"
+
+#ifdef _WIN32
+#define PATH_TO_WSTRING(path_expr) (path_expr).wstring()
```
iskiselev, line 14: Can we use std::filesystem::path functions instead: https://en.cppreference.com/w/cpp/filesystem/path/append.html

Rhynier: Removing unnecessary definitions

src/OpenTelemetry.AutoInstrumentation.Native/file_utils.h
```diff
@@ -0,0 +1,23 @@
...
+#define ENV_VAR_PATH_SEPARATOR WStr(';')
+#define ENV_VAR_PATH_SEPARATOR_STR WStr(";")
+#else
+#define PATH_TO_WSTRING(path_expr) ToWSTRING((path_expr).string())
```
iskiselev: I think it should be:
```
#define PATH_TO_WSTRING(path_expr) (path_expr).u16string()
```

iskiselev: But nobody will remember that we have this helper, so direct usage may be preffered.

src/OpenTelemetry.AutoInstrumentation.Native/pal.h
```diff
@@ -132,6 +125,36 @@ inline WSTRING GetCurrentModuleFileName()
...
 
+inline WSTRING GetHomePath()
+{
+    std::filesystem::path home_path;
+
+    // First choice is the environment variable
+    const auto env_home_path = GetEnvironmentValue(environment::profiler_home_path);
+    if (!env_home_path.empty())
+    {
+        home_path = std::filesystem::absolute(env_home_path);
```
iskiselev: better use overload that accepts `std::error_code` - it should not throw excpetions on no access.

src/OpenTelemetry.AutoInstrumentation.Native/pal.h
```diff
@@ -132,6 +125,36 @@ inline WSTRING GetCurrentModuleFileName()
...
+        const auto module_filename = GetCurrentModuleFileName();
+        if (module_filename.empty())
+        {
+            return EmptyWStr;
+        }
+
+        home_path = std::filesystem::path(module_filename).parent_path();
+    }
+
+    if (!std::filesystem::exists(home_path))
```
iskiselev: Same - use eror_code overload

src/OpenTelemetry.AutoInstrumentation.Native/startup_hook.h
```diff
@@ -12,20 +12,24 @@
...
         return false;
     }
 
-    const auto expected_deployment_startuphook_path =
-        std::filesystem::path(home_path)
-        / "net" / "OpenTelemetry.AutoInstrumentation.StartupHook.dll";
-    const auto expected_nuget_startuphook_path = 
-        std::filesystem::path(home_path)
-        / "OpenTelemetry.AutoInstrumentation.StartupHook.dll";
-    if (!std::filesystem::exists(expected_deployment_startuphook_path) && !std::filesystem::exists(expected_nuget_startuphook_path))
```
iskiselev, line 28: interesting, that ec not used here, but used few lines bellow. Would it be better to use in both places?

src/OpenTelemetry.AutoInstrumentation.Native/startup_hook.h
```diff
@@ -34,14 +38,59 @@ inline bool IsStartupHookValid(const std::vector<WSTRING>& startup_hooks, const
...
+    }
+
+    // Find the StartupHook DLL based on the following possible folder structures:
+    // 1. ZIP:
+    //  <OTEL_HOME>/<RID>/<native_profiler_dll>
+    //  <OTEL_HOME>/net/OpenTelemetry.AutoInstrumentation.StartupHook.dll
+    auto startuphook_path =
+        std::filesystem::path(profiler_path).parent_path().parent_path()
+            / "net" / opentelemetry_autoinstrumentation_startuphook_filepath;
+    if (std::filesystem::exists(startuphook_path))
```
iskiselev: Same notice about error_code here and 2 places bellow.

src/OpenTelemetry.AutoInstrumentation.Native/pal.h
```diff
@@ -132,6 +125,36 @@ inline WSTRING GetCurrentModuleFileName()
     return EmptyWStr;
 }
 
+inline WSTRING GetHomePath()
```
iskiselev: Is it not used?

src/OpenTelemetry.AutoInstrumentation.Native/startup_hook.h
```diff
@@ -12,20 +12,24 @@
...
-        / "net" / "OpenTelemetry.AutoInstrumentation.StartupHook.dll";
-    const auto expected_nuget_startuphook_path = 
-        std::filesystem::path(home_path)
-        / "OpenTelemetry.AutoInstrumentation.StartupHook.dll";
-    if (!std::filesystem::exists(expected_deployment_startuphook_path) && !std::filesystem::exists(expected_nuget_startuphook_path))
+    auto startuphook_path = std::filesystem::path(home_path) / "net"
+                            / opentelemetry_autoinstrumentation_startuphook_filepath;
+    if (!std::filesystem::exists(startuphook_path))
+    {
+        startuphook_path = std::filesystem::path(home_path)
```
iskiselev: can we add comment that it is case of nuget installation - previosly it was self-explained in `expected_nuget_startuphook_path` variable name.

src/OpenTelemetry.AutoInstrumentation.Native/startup_hook.h
```diff
@@ -12,20 +12,24 @@
...
     }
 
-    const auto expected_deployment_startuphook_path =
-        std::filesystem::path(home_path)
-        / "net" / "OpenTelemetry.AutoInstrumentation.StartupHook.dll";
-    const auto expected_nuget_startuphook_path = 
-        std::filesystem::path(home_path)
-        / "OpenTelemetry.AutoInstrumentation.StartupHook.dll";
-    if (!std::filesystem::exists(expected_deployment_startuphook_path) && !std::filesystem::exists(expected_nuget_startuphook_path))
+    auto startuphook_path = std::filesystem::path(home_path) / "net"
```
iskiselev: Let's rename it to `expected_startup_hook_path`

src/OpenTelemetry.AutoInstrumentation.Native/startup_hook.h
```diff
@@ -34,14 +38,59 @@ inline bool IsStartupHookValid(const std::vector<WSTRING>& startup_hooks, const
     {
         const auto start_hook_path = std::filesystem::path(*i);
```
iskiselev: And this rename to `start_hook_path_candidate`

src/OpenTelemetry.AutoInstrumentation.Native/startup_hook.h
```diff
@@ -34,14 +38,59 @@ inline bool IsStartupHookValid(const std::vector<WSTRING>& startup_hooks, const
...
         std::error_code ec;
-        if (std::filesystem::equivalent(expected_deployment_startuphook_path, start_hook_path, ec) || std::filesystem::equivalent(expected_nuget_startuphook_path, start_hook_path, ec))
```
iskiselev, line 37: Hope we would not see situatations when we have zip-deployment extracted in app  folder with otel also installed as NuGet, so both path valid with hook environment variable selecting between two possible locations.

src/OpenTelemetry.AutoInstrumentation.Native/startup_hook.h
```diff
@@ -34,14 +38,59 @@ inline bool IsStartupHookValid(const std::vector<WSTRING>& startup_hooks, const
...
+    if (profiler_path == EmptyWStr)
+    {
+        return EmptyWStr;
+    }
+
+    // Find the StartupHook DLL based on the following possible folder structures:
+    // 1. ZIP:
+    //  <OTEL_HOME>/<RID>/<native_profiler_dll>
+    //  <OTEL_HOME>/net/OpenTelemetry.AutoInstrumentation.StartupHook.dll
+    auto startuphook_path =
```
iskiselev: Should we use OTEL_HOME environment variable here? Or better take root from outside, so that profiler would be responsible for it. I'm not sure if requirment of OTEL_HOME check is on purpose - and if managed part will work without it. It may be better idea to use OTEL_HOME for now, and make it support fallback in separate PR.
There are multiple ways to implement fallback - profiler may just set that environment variable for example - in that case HOOK assembly may also need to do it if it is not set for symmetry with no-profiler mode.

zacharycmontoya: In our setup it's `OTEL_DOTNET_AUTO_HOME` and I think it's required in every deployment to find managed assemblies, so I support the recommendation to rely on that env first (could be passed as an argument to this function) and only do a relative lookup as a fallback

src/OpenTelemetry.AutoInstrumentation.Native/stub_generator.cpp
```diff
@@ -0,0 +1,449 @@
...
+        // else
+        // {
+        //     startupHooks += "<path_separator>";
+        //     startupHooks += "<startup_hook_dll_name>";
+        // }
+
+        // clang-format off
+        // IL_0000: ldarg.0
+        // IL_0001: ldind.ref
+        // IL_0002: brtrue.s     IL_000c
```
zacharycmontoya, line 311: Is this actually safe to assume that a value of `0` is equivalent to null? I would have expected you to do `ldnull; ceq` and then branch off that

src/OpenTelemetry.AutoInstrumentation.Native/stub_generator.h
```diff
@@ -0,0 +1,46 @@
...
+    ~StubGenerator();
+
+    HRESULT PatchProcessStartupHooks(const ModuleID module_id, const WSTRING& startup_hook_assembly_path);
+
+private:
+    HRESULT ModifyProcessStartupHooks(const ModuleID module_id, mdMethodDef patch_startup_hook_method);
+    HRESULT GenerateHookFixup(const ModuleID module_id,
+                              mdTypeDef*     hook_fixup_type,
+                              mdMethodDef*   patch_startup_hook_method,
+                              const WSTRING& startup_hook_dll_name);
```
zacharycmontoya: My preference would be to make the `startup_hook_dll_name` the second argument, so that the "out" pointers are the last arguments in the list

zacharycmontoya (approved):
LGTM, with some very small feedback. Thanks!

src/OpenTelemetry.AutoInstrumentation.Native/startup_hook.h
Rhynier: That is a customer configuration issue.

src/OpenTelemetry.AutoInstrumentation.Native/file_utils.h
Rhynier: Direct usage would require #define _WIN32 everywhere path is converted to WSTRING. If someone else writes code with std::filesystem and has to convert it to WSTRING they will find this as soon as they get compiler errors between Linux and Windows usages.

src/OpenTelemetry.AutoInstrumentation.Native/pal.h
Rhynier: Good catch. It was a leftover from a previous implementation I tested. I removed it.

src/OpenTelemetry.AutoInstrumentation.Native/startup_hook.h
Rhynier: Reworked the implementation to start with possible locations based on the `OTEL_DOTNET_AUTO_HOME` environment variable.

src/OpenTelemetry.AutoInstrumentation.Native/stub_generator.cpp
Rhynier: This IL generation was written based on doing ILSpy on compiled C# code matching the comment. I checked it multiple ways.

docs/config.md
```diff
@@ -492,6 +492,10 @@ environment variable are used to mitigate assembly version conflicts in .NET.
 | `DOTNET_ADDITIONAL_DEPS` | `$INSTALL_DIR/AdditionalDeps`                                        | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
 | `DOTNET_SHARED_STORE`    | `$INSTALL_DIR/store`                                                 | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
 
+If the .Net CLR Profiler is used and the
```
Kielek:
```suggestion
If the .NET CLR Profiler is used and the
```

Kielek:
@Rhynier, I have pushed some trivial changes to your branch, then asked Copilot to review this. Please let me know if you think that it is worth to apply some suggestion from this review. Other part LGTM

src/OpenTelemetry.AutoInstrumentation.Native/stub_generator.cpp
zacharycmontoya: Sounds good, thanks for the follow-up 👍🏼

src/OpenTelemetry.AutoInstrumentation.Native/pal.h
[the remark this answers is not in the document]
Rhynier: Addressed all std::filesystem error handling based on review feedback from Igor.

src/OpenTelemetry.AutoInstrumentation.Native/startup_hook.h
```diff
@@ -34,14 +38,59 @@ inline bool IsStartupHookValid(const std::vector<WSTRING>& startup_hooks, const
...
+    //  <APP_FOLDER>/OpenTelemetry.AutoInstrumentation.StartupHook.dll
+    startuphook_path =
+        std::filesystem::path(profiler_path).parent_path().parent_path().parent_path()
+            / opentelemetry_autoinstrumentation_startuphook_filepath;
+    if (std::filesystem::exists(startuphook_path))
+    {
+        return PATH_TO_WSTRING(startuphook_path);
+    }
+
+    return EmptyWStr;
```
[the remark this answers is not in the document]
Rhynier: Addressed all std::filesystem error handling based on review feedback from Igor.

src/OpenTelemetry.AutoInstrumentation.Native/stub_generator.cpp
```diff
@@ -0,0 +1,448 @@
...
+        // C# equivalent:
+        // if (startupHooks == null)
+        // {
+        //     startupHooks = "<startup_hook_dll_name>";
+        // }
+        // else
+        // {
+        //     startupHooks += "<path_separator>";
+        //     startupHooks += "<startup_hook_dll_name>";
+        // }
```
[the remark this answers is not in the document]
Rhynier: For simpler IL generation and since the generated code will be executed only once during application startup, I am leaving the current code unchanged.

src/OpenTelemetry.AutoInstrumentation.Native/stub_generator.cpp
```diff
@@ -0,0 +1,448 @@
...
+        pNewInstr->m_Arg32  = string_concat_token;
+        rewriter.InsertBefore(pFirstInstr, pNewInstr);
+
+        pNewInstr           = rewriter.NewILInstr();
+        pNewInstr->m_opcode = CEE_STIND_REF;
+        rewriter.InsertBefore(pFirstInstr, pNewInstr);
+
+        pNewInstr           = rewriter.NewILInstr();
+        pNewInstr->m_opcode = CEE_RET;
+        rewriter.InsertBefore(pFirstInstr, pNewInstr);
```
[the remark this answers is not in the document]
Rhynier: If there was going to be a lot of code reuse here (additional code generation methods, etc.), this could turn out to be helpful. However, as it is, it is more helpful to have it like this with a clear view of how the IL in the comment above is crafted with this code.

docs/config.md
```diff
@@ -492,6 +493,15 @@ environment variable are used to mitigate assembly version conflicts in .NET.
...
 | `DOTNET_SHARED_STORE`    | `$INSTALL_DIR/store`                                                 | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
 
+If the .NET CLR Profiler is used and the
+[`DOTNET_STARTUP_HOOKS`](https://github.com/dotnet/runtime/blob/main/docs/design/features/host-startup-hook.md)
+environment variable is not set, the profiler looks for
+`OpenTelemetry.AutoInstrumentation.StartupHook.dll` in an appropriate directory
+relative to the `OpenTelemetry.AutoInstrumentation.Native.dll` file location.
+The folder structure can match the ZIP archive structure or the NuGet package
+structure (either platform dependent or independent). If the startup hook
+assembly is not found, the profiler loading will be aborted.
```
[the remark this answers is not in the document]
Rhynier: The `FailProfiler` macro does result in the profiler aborting and not being loaded.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -692,6 +694,27 @@ HRESULT STDMETHODCALLTYPE CorProfiler::ModuleLoadFinished(ModuleID module_id, HR
         }
 #endif
 
+        if (module_info.assembly.name == system_private_corelib_assemblyName && startup_fix_required)
+        {
+            WSTRING startup_hook_assembly_path = GetStartupHookPath(GetCurrentModuleFileName(), home_path);
+            if (startup_hook_assembly_path == EmptyWStr)
+            {
+                FailProfiler(
+                    Info,
```
iskiselev: Is there any reason behind why we use `Info` level and not `Error` here and few lines bellow? Previously missing og Hook reported with `Error` level.
FailProfiler may or may not thow exception based on other environment vairable. It may crash or not customer application depending on particular time of exception. So, there still may be sitautions when everything works - and no instrumentation applied, which would be nice to reoprt with higher visibility level.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.h
```diff
@@ -44,7 +44,9 @@ class CorProfiler : public CorProfilerBase
     std::mutex definitions_ids_lock_;
 
     // Startup helper variables
+    WSTRING home_path;
     bool first_jit_compilation_completed = false;
+    std::atomic_bool startup_fix_required = {false};
```
iskiselev: BTW, I'm not a fan of that change suggested by CoPilot. We set the value in Initialize, read it in ModuleLoad callback - there should be no race condition.
Previous field `first_jit_compilation_completed` - has much higher race condition chance. And now next reader would think why this is atomic_bool, but previous and next line is not.
If we really care about it, it may be better idea to address all of state variables in one commit.

src/OpenTelemetry.AutoInstrumentation.Native/startup_hook.h
```diff
@@ -8,40 +8,139 @@
...
-    if (!std::filesystem::exists(expected_deployment_startuphook_path) && !std::filesystem::exists(expected_nuget_startuphook_path))
+    // Check for StartupHook assembly in ZIP deployment path
+    auto expected_startuphook_path = std::filesystem::path(home_path) / "net"
+                            / opentelemetry_autoinstrumentation_startuphook_filepath;
+
+    std::error_code ec;
+    auto path_exists = std::filesystem::exists(expected_startuphook_path, ec);
+    if (ec)
+    {
+        Logger::Warn("Failed to access StartupHook path '",
```
iskiselev, line 34: I'd bet we need `expected_startuphook_path.string()` or `expected_startuphook_path.u8string()` - but no need to change. I'm not sure how it would encode string() for non-ANSI symbols on windows or how u8string will behave with logger.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
Rhynier: Agreed. This should be Error level.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.h
Rhynier: I'll change this back to be addressed separately as and if needed.

PR #4543, opened 2025-10-23
[frequent-sampling] Change API from span-based to trace-based 

lachmatt:
## Why&& What

In order to reduce number of calls from managed-side to native-side, change API for frequent profiling for selected traces from span-based to trace-based.


test/test-applications/integrations/TestApplication.SelectiveSampler/Plugins/FrequentSamplingProcessor.cs
```diff
@@ -22,20 +22,27 @@ public FrequentSamplingProcessor()
...
-        _stopSamplingDelegate = (Action<Activity>)Delegate.CreateDelegate(typeof(Action<Activity>), stopMethod!);
+        _startSamplingDelegate = (Action<ActivityTraceId>)Delegate.CreateDelegate(typeof(Action<ActivityTraceId>), startMethod!);
+        _stopSamplingDelegate = (Action<ActivityTraceId>)Delegate.CreateDelegate(typeof(Action<ActivityTraceId>), stopMethod!);
     }
 
     public override void OnStart(Activity data)
     {
-        _startSamplingDelegate?.Invoke(data);
+        // Native side API is trace-based, notifying only of root span is sufficient.
+        if (data.Parent == null)
```
nrcventura: Should this also include cases where there is a parent, but it is a remote parent?

lachmatt: In this specific case, this doesn't matter much, but generally yes.
This is test-only dependency, and case with remote parent is not verified in test.
On the other hand, secondary function of this code is documentation of sample usage, so I updated the code in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4543/commits/c365907613f169029d02466f3b5517ae62d84c5f

Kielek (approved):

test/test-applications/integrations/TestApplication.SelectiveSampler/Plugins/FrequentSamplingProcessor.cs
nrcventura: Yes, it was the documentation piece of it to show how this can be used in downstream services.

PR #4545, opened 2025-10-23
Drop re2 on Linux and MaOS

### LINKED ISSUE #2639: Update re2 to 2023-06-02 or newer.
Update re2 to 2023-06-02 or newer.

It can be separate issue.

_Originally posted by @Kielek in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2635#discussion_r1226348255_

Kielek:
## Why

Fixes #2639

## What

Drop re2 on Linux and MaOS.
All platforms relays on regex stdlib

## Tests

CI

## Checklist

### DIFF
--- a/scripts/dotnet-install.sh
+++ b/scripts/dotnet-install.sh
@@ -1169,11 +1169,11 @@ download() {
             exit 1
         fi
 
-        if [ "$failed" = false ] || [ $attempts -ge 3 ] || { [ ! -z $http_code ] && [ $http_code = "404" ]; }; then
+        if [ "$failed" = false ] || [ $attempts -ge 3 ] || { [ -n "${http_code-}" ] && [ "${http_code}" = "404" ]; }; then
             break
         fi
 
-        say "Download attempt #$attempts has failed: $http_code $download_error_msg"
+        say "Download attempt #$attempts has failed: ${http_code-} ${download_error_msg-}"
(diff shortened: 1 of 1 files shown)


src/OpenTelemetry.AutoInstrumentation.Native/regex_utils.cpp
```diff
@@ -0,0 +1,150 @@
...
+        return false;
+    }
+
+#ifdef _WIN32
+    static std::wregex re(WStr("PublicKeyToken=([a-fA-F0-9]{16})"));
+    std::wsmatch       match;
+
+    if (std::regex_search(str, match, re) && match.size() == 2)
+    {
+        for (int i = 0; i < 8; i++)
```
nrcventura: The length of the array pointed to by `data` is now defined in two locations: The declaration of the array being passed to this function, and in the this loop. Should we add the length of `data` as a parameter to this method, or perhaps define it in a constant or `#define` value so that it can be shared by the two locations?

Kielek: Fixed in ee27e33deaa779e25a745d49a2d66a59571a839c
Passed the length as a parameter to this method, as I do not want to put here reference to instrumentation.h header.
It is reusing already existing const.

zacharycmontoya (approved):

nrcventura (approved):

Issue #4552, opened 2025-10-27
[File Based Configuration] Add Additional Sources


ysolomchenko:
### Description

Add possibility for configuring 
- `OTEL_DOTNET_AUTO_TRACES_ADDITIONAL_SOURCES`
- `OTEL_DOTNET_AUTO_TRACES_ADDITIONAL_LEGACY_SOURCES`
- `OTEL_DOTNET_AUTO_METRICS_ADDITIONAL_SOURCES`

PR #4576, opened 2025-10-31
[FileBasedConfiguration] Additional Sources Configuration

ysolomchenko:
## Why

Towards https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/4394
Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/4552

## What

File Based Configuration for Additional Sources

## Tests

Ci

## Checklist

### DIFF
--- a/docs/nocode-instrumentation.md
+++ b/docs/nocode-instrumentation.md
@@ -46,24 +46,26 @@ The no-code instrumentation configuration is defined under the
 ```yaml
 file_format: "1.0-rc.1"
 
-no_code/development:
-  targets:
-    - target:                                             # Method target specification
-        assembly:                                         # Assembly information
-          name: TestApplication.NoCode                    # The name of the assembly containing the target method
-        type: TestApplication.NoCode.NoCodeTestingClass   # The fully qualified type name containing the method
-        method: TestMethod                                # The method name to instrument
-        signature:                                        # Method signature specification for precise targeting
-          return_type: System.Void                        # The method's return type (e.g., System.Void, System.String, System.Threading.Tasks.Task)
-          parameter_types:                                # Array of parameter types (empty array for parameterless methods)
-            - System.String
-      span:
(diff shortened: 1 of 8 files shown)


Kielek (approved):

docs/file-based-configuration.md
```diff
@@ -334,6 +334,32 @@ instrumentation/development:
...
 
+## Instrumentation Additional Sources
+
+``` yaml
+instrumentation/development:
+  dotnet:
+    traces:
+      # List of additional System.Diagnostics.ActivitySource names to be added to the tracer at the startup. 
+      # Use it to capture manually instrumented spans.
+      additional_sources:
```
zacharycmontoya, line 345: Is there some precedent that we're following for the naming of this `additional_sources` vs `additional_sources_list`? Maybe it's because I don't work with yaml and configurations often, but it seems confusing to me that the `additional_sources` key accepts a YAML List and the `additional_sources_list` key accepts a YAML string.

zacharycmontoya: This isn't a blocker, but I wanted to note a small confusion I had

zacharycmontoya: My concrete suggestion here would be to rename `additional_sources_list` to `additional_sources_env` to indicate it follows the environment variable comma-separated list of names. What do you think?

Kielek: `_list` suffix is common for the migration purposes ref https://github.com/open-telemetry/opentelemetry-configuration/blob/940e6c183c3a349004d9be2c63a7157816b78c7d/examples/sdk-migration-config.yaml#L58 and https://github.com/open-telemetry/opentelemetry-configuration/blob/940e6c183c3a349004d9be2c63a7157816b78c7d/examples/sdk-migration-config.yaml#L80

Technically, you can put there also comma separated list without env variable, but there is more convinient way - jus listing them as separate recors.

PR #4585, opened 2025-11-05
Improved GenerateNetFxTransientDependencies

### LINKED ISSUE #4584: GenerateNetFxTransientDependencies does not remove packages
# Bug Report

Nuke action `GenerateNetFxTransientDependencies` does not update dependencies correctly. When packages are updated and the updated version removes dependencies, `GenerateNetFxTransientDependencies` does not remove overleft package references.

## Reproduce

Use my branch https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4486 and update `OpenTelemetry.OpAmp.Client` from `0.1.0-alpha.1` to `0.1.0-alpha.2`

CC: @iskiselev

iskiselev:
## Why

Simplify upkeep of auto-generated transitive PackageVersions. Fixes #4584.

## What

Gather common PackageVersion under non-conditioned ItemGroup, remove unused PackageVersion from auto-generated ItemGroups.

## Tests

Manual tests only. Do we have any examples of Build steps tests?

## Checklist


build/Build.Steps.Windows.cs
```diff
@@ -1,6 +1,6 @@
 using DependencyListGenerator;
-using Extensions;
-using NuGet.Versioning;
+using JetBrains.Annotations;
```
pellared: We are not using these :wink:

build/TransientDependenciesGenerator.cs
```diff
@@ -0,0 +1,144 @@
...
+                            .First(it => GetPackageProperties(it) == pkg);
+                        group.RemoveChild(reference);
+                    }
+
+                    AddPackageReference(commonGroup, pkg.PackageName, pkg.PackageVersion);
+                }
+            }
+        }
+
+        // clea unused dependencies
```
RassK, line 134:
```suggestion
        // clean unused dependencies
```

build/TransientDependenciesGenerator.cs
```diff
@@ -0,0 +1,144 @@
...
+        var projectFrameworks = targetProject.GetTargetFrameworks();
+
+        if (projectFrameworks == null || projectFrameworks.Count == 0)
+        {
+            Log.Error(
+                $"Project {targetProject.Name} does not specify any target frameworks. Skipping {nameof(TransientDependenciesGenerator)} step.");
+            return;
+        }
+
+        // generate new transitive dependencies
```
RassK, line 65: nit: would extract to a method

build/TransientDependenciesGenerator.cs
```diff
@@ -0,0 +1,144 @@
...
+            foreach (var item in deps[framework])
+            {
+                if (!definedVersions.ContainsKey(item.Name))
+                {
+                    AddPackageReference(versionGroup, item.Name, item.Version);
+                }
+            }
+        }
+
+        // group dependencies
```
RassK, line 92: nit: would extract to a method

build/TransientDependenciesGenerator.cs
RassK, line 134: nit: would extract to a method

build/TransientDependenciesGenerator.cs
iskiselev: I though about it, but it would require either pass multiple parameters between methods - or change it to instance class where contructor init fields.
I was not sure whether it really worth it, especially as that is build only code.

RassK: sure, it's up to you if you notice something that makes it easier. For sure no point to make it more complex.

Issue #4402, opened 2025-08-19
[File Based Configuration] Add Documentation and Examples


ysolomchenko:
### Blocked by

- [x] #4397 
- [x] #4398 
- [x] #4399 
- [x] #4400 

### Description
Create documentation with examples of file-based configuration, demonstrating correct usage, supported formats, and environment variable overrides.

### Acceptance Criteria

- Documentation clearly explains the file-based configuration format.
- Environment variable substitution is demonstrated and explained.
- Documentation is reviewed and approved for accuracy and clarity.
---
- Added example files described in this [comment](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4469#discussion_r2374954829)

PR #4590, opened 2025-11-06
[FileBasedConfiguration] Add Example configuration files

ysolomchenko:
## Why

Towards #4394
Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/4402

## What

Updated documentation.

Added the following files:
- kitchen-sink.yaml (a full config with all supported fields)
- default-config.yaml (a config with the standard configurations that were used with env var configuration)
- sdk-migration-config.yaml (a config that uses env vars instead of configurations, like in the opentelemetry-configuration examples, but only with the implemented fields)

## Checklist


docs/file-based-configuration.md
```diff
@@ -96,10 +96,11 @@ tracer_provider:
 
     # Batch processor for OTLP gRPC
     - batch:
-        otlp_grpc:
-          # Configuration otlp_grpc is the same as otlp_http.
-          # if otlp_http is used it will override otlp_grpc.
-          # On .NET Framework, the grpc OTLP exporter protocol is not supported.
+        exporter:
+          # Configure exporter to be OTLP with gRPC transport.
+          otlp_grpc:
```
nrcventura, line 101: Should we use `otlp_http` in our examples especially if that is the version of otlp that we recommend by default?

nrcventura:
This is just an idea, and is something that could be done separately if we believe it has enough value. Should we add a test that ensures that our configuration parsing logic works with these new example files? The hope is that those tests will help us to keep these files up to date as we continue to make changes.

docs/file-based-configuration.md
ysolomchenko: In `default-config` and `sdk-migration-config`, I use only `otlp_http`. This is just documentation explaining that `otlp_grpc` is not supported on .NET Framework.

PR #4591, opened 2025-11-07
 Adding support for capturing NetFx call stacks

eftiquar:
# Stack Capture Engine: Design Overview

## Introduction

The Stack Capture Engine enables safe, deadlock-free continuous profiling of .NET Framework applications by capturing managed call stacks from running threads. This document describes the core design principles that make this possible, focusing on two critical mechanisms: the **Canary Thread** pattern for runtime safety detection and **RTL-based Stack Seeding** for accurate context preparation. The .NET Framework runtime cannot be suspended; therefore, the stack capture implementation must employ the safety mechanisms outlined below to prevent application deadlocks. 

1. First, the target thread is suspended to prevent it from progressing during the snapshot operation.
2. Then, a Canary Thread, which is a designated known thread not involved in the critical path of snapshotting, performs a probe (snapshot).
3. If the Canary Thread probe fails, it indicates that the target thread is likely in a critical path, so the stack capture is aborted, and the target thread is resumed to avoid deadlock.
4. Conversely, if the Canary Thread snapshot succeeds while the target thread remains suspended, it is inferred that the snapshot on the target thread will also succeed safely.
5. Thus, the Canary Thread acts as a first line of defense, performing safety checks to ensure that stack capture proceeds only when it is safe to do so.

### 1.1 The Core Problem

The CLR's `DoStackSnapshot` API can cause deadlocks when called during unsafe runtime states—such as during garbage collection, JIT compilation, or when critical runtime locks are held. Traditional approaches that directly suspend and snapshot threads risk:

- Suspending threads that hold critical CLR locks
- Triggering `CORPROF_E_STACKSNAPSHOT_UNSAFE` errors
- Causing application hangs or process termination

### 1.2 The Canary Solution

Rather than blindly attempting to capture stacks, the engine uses a **dedicated canary thread** as a safety sentinel. This is a known, controlled thread created by the application specifically for profiling purposes, identified by a configurable name prefix (default: `"OpenTelemetry Continuous"`).

**How It Works:**

Before capturing stacks from any production threads, the engine performs a **safety probe** using the canary thread:

1. **Suspend the target thread** - it is in arbitrary state, maybe holding locks that are in the path of snapshotting API
2.  **Suspend Canary thread** (which holds no application locks)
3. Execute a **safety probe** on a dedicated **worker thread**. 
   - This isolates the continuous profiler thread itself from deadlocking
   - Heap allocation is done (tests if heap locks are available)
   - RTL function lookup is done (tests if loader locks are available)
   - `DoStackSnapshot` called on the canary itself (tests CLR profiling API safety)
4. **Resume the canary thread**
5. **Analyze the results** with a timeout.
6. If the **canary probe** succeeds - proceed with snapshotting target thread 
7. Else **abort** snapshot operation; resume target thread via **RAII** mechanisms so that it is resumed unfailingly. 

If all probe operations complete successfully within the timeout (default 250ms), the runtime is considered safe for capturing production thread stacks. If the probe fails or times out, the engine skips the current capture cycle and waits for the next interval.

**Key Safety Properties:**

- **Timeout Protection**: The probe operations execute on a worker thread with a configurable timeout, preventing indefinite blocking
- **SEH Protection**: All probe operations are wrapped in Structured Exception Handling (`__try`/`__except`) to catch access violations gracefully
- **RAII Guarantees**: Thread suspension/resumption uses RAII patterns to ensure threads are always resumed, even in exception scenarios
- **Isolated Testing**: The canary thread performs no application work and holds no locks, making it safe to suspend and test

### 1.3 Canary Thread Lifecycle

The engine tracks all managed threads through profiler callbacks (`ThreadNameChanged`, `ThreadAssignedToOSThread`). When a thread with the canary name prefix is detected:

1. The engine registers both its managed thread ID and native OS thread ID
2. The capture loop is notified via condition variable
4. All subsequent capture cycles use this thread for safety probes

If the canary thread is destroyed, the engine clears its registration and waits for a new canary to be designated before resuming captures.

---

### 2.1 The Context Preparation Challenge

The CLR's `DoStackSnapshot` API requires a valid starting context pointing to **managed code**. However, when a thread is suspended for profiling, its instruction pointer (`RIP` on x64) may be in:

- Native Win32 APIs (e.g., `Sleep`, `WaitForSingleObject`)
- P/Invoke transitions
- COM/WinRT interop layers
- Native library code

Passing a context pointing to native code causes `DoStackSnapshot` to fail or return incomplete stacks. The engine must walk the native stack frames to locate the first managed frame before invoking `DoStackSnapshot`.

### 2.2 The RTL Function Solution

Windows provides low-level **Runtime Library (RTL)** functions for exception handling and stack unwinding:

- **`RtlLookupFunctionEntry`**: Retrieves unwind metadata for a given instruction pointer
- **`RtlVirtualUnwind`**: Simulates stack unwinding using function metadata, updating the context to the caller's frame

These functions are the same primitives the CLR uses internally for exception handling, making them reliable and accurate.

**The Seeding Algorithm:**

1. **Quick Check**: Test if the current instruction pointer is already managed code via `GetFunctionFromIP`
2. **Native Stack Walk**: If in native code, iterate through stack frames:
   - Use `RtlLookupFunctionEntry` to get unwind metadata for the current RIP
   - **If metadata exists** (non-leaf function): Extract the function's begin address and use `RtlVirtualUnwind` to update the context to the caller frame
   - **If no metadata** (leaf function): Manually read the return address from the stack pointer (`[RSP]`) and adjust RSP
   - After each frame transition, test the instruction pointer with `GetFunctionFromIP` to check if we've reached managed code
3. **Termination**: Stop when:
   - Managed code is found (success)
   - RIP becomes zero (end of stack)
   - Stack pointer stops progressing (corruption detection)
   - Maximum frame count exceeded (safety limit: 10,000 frames)
5. **Seed DoStackSnapshot**: Pass the prepared context (now pointing to a managed frame) to `DoStackSnapshot`

### 2.3 Critical Design Details

**Function Begin Address vs. Current RIP:**

The CLR's metadata associates managed functions with their **entry points** (begin addresses), not arbitrary instruction pointers mid-function. When unwind metadata is available, the engine uses `imageBase + runtimeFunction->BeginAddress` for managed detection, not the current RIP. This ensures reliable `GetFunctionFromIP` lookups.

**Leaf Function Handling:**

Leaf functions (functions with no stack frame) lack unwind metadata. The engine detects this case and manually pops the return address from the stack:
```
returnAddress = *[RSP]
RSP += 8  // Advance past return address
```

**Stack Corruption Detection:**

The engine tracks the stack pointer (RSP) across frames. If RSP fails to grow (or moves backward), the stack is considered corrupted and the walk terminates to prevent crashes.

**SEH Protection:**

All memory reads and RTL function calls are wrapped in Structured Exception Handling. If an access violation occurs (invalid memory, corrupted unwind metadata), the operation fails gracefully without crashing the application.

### 2.4 Why This Matters

Without accurate seeding:
- P/Invoke-heavy applications would have incomplete stack traces
- Threads blocked in native APIs would be skipped entirely
- Async/await state machine transitions might be missed
- Profiling data would have significant blind spots

With RTL-based seeding:
- Stacks are accurately captured regardless of where threads are suspended
- Native-to-managed transitions are handled correctly
- The engine works reliably across diverse application patterns (sync, async, P/Invoke, COM)

---

### 3.1 The Capture Flow

For each profiling cycle:

1. **Wait for Canary Availability**: Block until a canary thread is registered (or timeout)
2. **For Each Target Thread**:
   - Skip the canary thread itself
   - **Suspend the target thread** (RAII-based)
   - **Safety Probe**: Suspend canary → test operations → resume canary
   - **If probe succeeds**:
     - Capture target thread context
     - **Seed the context** (walk native frames to find managed code)
     - Invoke `DoStackSnapshot` with the seeded context
     - Collect managed stack frames
   - **If probe fails**: Skip this thread and continue to next
   - **Resume the target thread** (automatic via RAII)
3. **Repeat** at the configured interval (default: 1 second)

### 3.2 Layered Safety

The design provides **defense in depth**:

| **Layer** | **Protection** |
|-----------|----------------|
| **Canary Probe** | Detects unsafe runtime states before touching production threads |
| **Worker Thread + Timeout** | Prevents indefinite blocking in probe operations |
| **SEH Wrappers** | Catches memory access violations without process termination |
| **RAII Thread Suspend** | Guarantees thread resumption even on exceptions |
| **Stack Walk Limits** | Prevents runaway walks from corrupted stacks |
| **Context Validation** | Verifies managed code is found before calling DoStackSnapshot |

Each layer can independently fail without bringing down the application. The engine simply logs the failure and skips to the next capture cycle.

### 4.1 Windows x64 Only

The RTL functions (`RtlLookupFunctionEntry`, `RtlVirtualUnwind`) are Windows-specific APIs. The engine is currently limited to:
- **OS**: Windows
- **Architecture**: x64 (unwind metadata availability is guaranteed by the x64 ABI)

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
@@ -1535,23 +1535,9 @@ HRESULT STDMETHODCALLTYPE CorProfiler::JITCompilationStartedOnNetFramework(Funct
     auto valid_loader_callsite = true;
     if (is_desktop_iis)
     {
-        // For IIS: prefer injection into BuildManager.InvokePreStartInitMethods (application AppDomain)
-        // but also allow injection into default AppDomain for continuous profiling support
-        bool is_build_manager_injection = module_metadata->assemblyName == WStr("System.Web") &&
-                                          caller.type.name == WStr("System.Web.Compilation.BuildManager") &&
-                                          caller.name == WStr("InvokePreStartInitMethods");
-
-        bool is_default_appdomain = corlib_module_loaded && module_metadata->app_domain_id == corlib_app_domain_id;
-
-        valid_loader_callsite = is_build_manager_injection || is_default_appdomain;
-
-        if (is_default_appdomain && !is_build_manager_injection)
-        {
-            Logger::Info("JITCompilationStarted: Allowing loader injection into default AppDomain for IIS continuous "
-                         "profiling. ",
-                         "AppDomain: ", module_metadata->app_domain_id, ", Assembly: ", module_metadata->assemblyName,
-                         ", Method: ", caller.type.name, ".", caller.name);
-        }
+        valid_loader_callsite = module_metadata->assemblyName == WStr("System.Web") &&
+                                caller.type.name == WStr("System.Web.Compilation.BuildManager") &&
+                                caller.name == WStr("InvokePreStartInitMethods");
     }
     else if (module_metadata->assemblyName == WStr("System") ||
              module_metadata->assemblyName == WStr("System.Net.Http"))
--- a/src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.cpp
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.cpp
@@ -394,6 +394,22 @@ HRESULT StackCaptureEngine::ThreadDestroyed(ThreadID threadId)
         trace::Logger::Info("[StackCapture] Canary thread destroyed - ManagedID=", threadId,
                             ", NativeID=", canaryThread_.nativeId);
         canaryThread_.reset();
+        // threadNames_ has map of thread ID and names - find another canary if possible
+        for (const auto& [managedId, name] : threadNames_)
+        {
+            if (options_.IsCanaryThread(name))
+            {
+                auto osThreadIt = activeThreads_.find(managedId);
+                if (osThreadIt != activeThreads_.end())
+                {
+                    canaryThread_ = CanaryThreadInfo{managedId, osThreadIt->second};
+                    trace::Logger::Info("[StackCapture] New canary thread designated after destruction - ManagedID=",
+                                        managedId, ", NativeID=", osThreadIt->second, ", Name=", name);
+                    captureCondVar_.notify_all();
+                    break;
+                }
+            }
+        }
     }
 
     return S_OK;
@@ -412,18 +428,13 @@ HRESULT StackCaptureEngine::ThreadAssignedToOSThread(ThreadID managedThreadId, D
     auto nameIt = threadNames_.find(managedThreadId);
     if (nameIt != threadNames_.end())
     {
-        if (nameIt->second == options_.canaryThreadName)
+        if (options_.IsCanaryThread(nameIt->second))
         {
             canaryThread_ = CanaryThreadInfo{managedThreadId, osThreadId};
             trace::Logger::Info("[StackCapture] Canary thread designated via ThreadAssignedToOSThread - ManagedID=",
                                 managedThreadId, ", NativeID=", osThreadId, ", Name=", nameIt->second);
+            captureCondVar_.notify_all();
         }
-    }
-
-    // Notify waiting capture thread if canary was found
-    if (canaryThread_.isValid())
-    {
-        captureCondVar_.notify_one();
     }
 
     return S_OK;
@@ -435,20 +446,17 @@ HRESULT StackCaptureEngine::ThreadNameChanged(ThreadID threadId, ULONG cchName, 
         return S_OK;
 
     std::lock_guard<std::mutex> lock(threadListMutex_);
-    if (canaryThread_.isValid())
-    {
-        return S_OK;
-    }
 
     std::wstring threadName(name, cchName);
     threadNames_[threadId] = threadName;
     trace::Logger::Debug("[StackCapture] ThreadNameChanged - ManagedID=", threadId, ", Name=", threadName);
-    if (threadName == options_.canaryThreadName)
+    if (options_.IsCanaryThread(threadName))
     {
         auto osThreadIt = activeThreads_.find(threadId);
-        if (osThreadIt != activeThreads_.end())
+        if (osThreadIt != activeThreads_.end() && !canaryThread_.isValid())
         {
             canaryThread_ = CanaryThreadInfo{threadId, osThreadIt->second};
+            captureCondVar_.notify_all();
             trace::Logger::Info("[StackCapture] Canary thread designated via ThreadNameChanged - ManagedID=", threadId,
                                 ", NativeID=", osThreadIt->second, ", Name=", threadName);
         }
@@ -460,16 +468,10 @@ HRESULT StackCaptureEngine::ThreadNameChanged(ThreadID threadId, ULONG cchName, 
         }
     }
 
-    // Notify waiting capture thread if canary was found
-    if (canaryThread_.isValid())
-    {
-        captureCondVar_.notify_one();
-    }
-
     return S_OK;
 }
 
-bool StackCaptureEngine::SafetyProbe()
+bool StackCaptureEngine::SafetyProbe(const CanaryThreadInfo& canaryInfo)
 {
     if (!invocationQueue_)
         return true;
@@ -479,7 +481,7 @@ bool StackCaptureEngine::SafetyProbe()
 
     try
     {
-        ScopedThreadSuspend canaryThread(canaryThread_.nativeId);
+        ScopedThreadSuspend canaryThread(canaryInfo.nativeId);
 
         CONTEXT canaryCtx      = {};
         canaryCtx.ContextFlags = CONTEXT_FULL;
@@ -493,7 +495,7 @@ bool StackCaptureEngine::SafetyProbe()
             return false;
         }
 
-        auto canaryManagedId = canaryThread_.managedId;
+        auto canaryManagedId = canaryInfo.managedId;
 
         status = invocationQueue_->Invoke(
             [this, canaryManagedId, canaryCtx, &snapshotHr]()
@@ -607,34 +609,43 @@ HRESULT StackCaptureEngine::CaptureStackSeeded(ThreadID             managedThrea
     return hr;
 }
 
-bool StackCaptureEngine::WaitForCanaryThread(std::chrono::milliseconds timeout)
+CanaryThreadInfo StackCaptureEngine::WaitForCanaryThread(std::chrono::milliseconds timeout)
 {
     trace::Logger::Debug("[StackCapture] Waiting for canary thread (timeout=", timeout.count(), "ms)");
-    std::unique_lock<std::mutex> lock(threadListMutex_);
-    bool                         result =
-        captureCondVar_.wait_for(lock, timeout, [this]() { return stopRequested_.load() || canaryThread_.isValid(); });
-
-    if (!result)
+    CanaryThreadInfo canary;
     {
-        trace::Logger::Warn("[StackCapture] Canary thread wait timed out after ", timeout.count(), "ms");
+        std::unique_lock<std::mutex> lock(threadListMutex_);
+        bool                         result = captureCondVar_.wait_for(lock, timeout,
+                                                                       [this]() { return stopRequested_.load() || canaryThread_.isValid(); });
+
+        if (!result)
+        {
+            trace::Logger::Warn("[StackCapture] Canary thread wait timed out after ", timeout.count(), "ms");
+        }
+        else
+        {
+            canary = canaryThread_;
+            trace::Logger::Debug("[StackCapture] Canary thread ready - ManagedID=", canary.managedId,
+                                 ", NativeID=", canary.nativeId);
+        }
     }
 
-    return result;
+    return canary;
 }
 
 HRESULT StackCaptureEngine::CaptureStacks(std::unordered_set<ThreadID> const&                threads,
                                           continuous_profiler::StackSnapshotCallbackContext* clientData)
 {
-    bool canaryReady = WaitForCanaryThread();
+    auto canary = WaitForCanaryThread();
 
-    if (!canaryReady)
+    if (!canary.isValid())
         return E_FAIL;
 
     for (const auto& managedId : threads)
     {
         if (stopRequested_)
             break;
-        if (managedId == canaryThread_.managedId)
+        if (managedId == canary.managedId)
             continue;
         DWORD nativeId = 0;
         {
@@ -649,7 +660,7 @@ HRESULT StackCaptureEngine::CaptureStacks(std::unordered_set<ThreadID> const&   
         try
         {
             ScopedThreadSuspend targetThread(nativeId);
-            if (!SafetyProbe())
+            if (!SafetyProbe(canary))
             {
                 trace::Logger::Debug(
                     "[StackCapture] CaptureStacks - Skipping thread due to safety probe failure. ManagedID=", managedId,
--- a/src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.h
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.h
@@ -36,6 +36,9 @@ namespace ProfilerStackCapture {
     struct CaptureOptions {
         std::chrono::milliseconds probeTimeout = std::chrono::milliseconds(250);
         const wchar_t* canaryThreadName = L"OpenTelemetry Profiler Canary Thread";
+        bool                      IsCanaryThread(const std::wstring& threadName) const {
+            return threadName.find(canaryThreadName) == 0;
+        }
     };
 
     class IProfilerApi {
@@ -134,7 +137,7 @@ namespace ProfilerStackCapture {
     public:
         explicit StackCaptureEngine(std::unique_ptr<IProfilerApi> profilerApi, const CaptureOptions& options = {});
         ~StackCaptureEngine();
-        bool    WaitForCanaryThread(std::chrono::milliseconds timeout = std::chrono::milliseconds(0));
+        CanaryThreadInfo    WaitForCanaryThread(std::chrono::milliseconds timeout = std::chrono::milliseconds(0));
         HRESULT CaptureStacks(const std::unordered_set<ThreadID> &threads, continuous_profiler::StackSnapshotCallbackContext* clientData);
         void Stop();
         
@@ -145,7 +148,7 @@ namespace ProfilerStackCapture {
         
     private:
         HRESULT CaptureStackSeeded(ThreadID managedThreadId, HANDLE threadHandle, StackCaptureContext* stackCaptureContext);
-        bool SafetyProbe();
+        bool SafetyProbe(const CanaryThreadInfo& canaryInfo);
         
         std::unique_ptr<IProfilerApi> profilerApi_;
         CaptureOptions options_;
--- a/src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
@@ -285,21 +285,17 @@ internal static class Instrumentation
 
         NativeMethods.ConfigureNativeContinuousProfiler(threadSamplingEnabled, threadSamplingInterval, allocationSamplingEnabled, maxMemorySamplesPerMinute, selectiveSamplingInterval);
 #if NETFRAMEWORK
-        if (AppDomain.CurrentDomain.IsDefaultAppDomain())
+        // On .NET Framework, we need a dedicated canary thread for seeded stack walking
+        _canaryThreadManager = new CanaryThreadManager();
+        if (!_canaryThreadManager.Start(TimeSpan.FromSeconds(5)))
         {
-            // On .NET Framework, we need a dedicated canary thread for seeded stack walking
-            // host the canary thread manager in the default app domain only to avoid multiple canary threads
-            _canaryThreadManager = new CanaryThreadManager();
-            if (!_canaryThreadManager.Start(TimeSpan.FromSeconds(5)))
-            {
-                Logger.Error("Failed to start canary thread. Continuous profiling will not be enabled.");
-                _canaryThreadManager.Dispose();
-                _canaryThreadManager = null;
-                return;
-            }
-
-            Logger.Information("Canary thread started successfully for .NET Framework profiling.");
+            Logger.Error("Failed to start canary thread. Continuous profiling will not be enabled.");
+            _canaryThreadManager.Dispose();
+            _canaryThreadManager = null;
+            return;
         }
+
+        Logger.Information("Canary thread started successfully for .NET Framework profiling.");
 #endif
         _sampleExporter = _sampleExporterBuilder?.Build();
     }
--- a/src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/CanaryThreadManager.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/ContinuousProfiler/CanaryThreadManager.cs
@@ -43,7 +43,7 @@ internal sealed class CanaryThreadManager : IDisposable
 
         _canaryThread = new Thread(CanaryThreadProc)
         {
-            Name = CanaryThreadName,
+            Name = $"{CanaryThreadName} - {AppDomain.CurrentDomain.Id})",
             IsBackground = true,
         };
 


pellared:
From [the contributing docs](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/CONTRIBUTING.md): 

> If you would like to work on something that is not listed as an issue, please [request a feature](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/CONTRIBUTING.md#request-a-feature) first. It is best to do this in advance so that maintainers can decide if the proposal is a good fit for this repository. This will help avoid situations when you spend significant time on something that maintainers may decide this repo is not the right place for.

Given this a significant feature it should be discussed in an issue before it can be accepted.

Changing to this PR to draft so that you can reference it (e.g. as PoC) in the issue.

PS. Please sign the CLA :wink:

nrcventura:
The approach looks good to me. It looks like there is some commented out code that could use some cleanup. There should also be a test implemented for the netfx side of things.

eftiquar:
> The approach looks good to me. It looks like there is some commented out code that could use some cleanup. There should also be a test implemented for the netfx side of things.

Thank you for the feedback. Integration tests are under development. Sure - there may be some dead code or commented code, I will clean it up.

Kielek:
@eftiquar, tests are almost in place. When #4631 is merged, you should be able just to remove conditional compilation from `ContinuousProfilerContextTrackingTests`, `ContinuousProfilerSpanStoppageHandlingTests`, and `ContinuousProfilerTests`.

For now, all of them are red on .NET Fx on your branch locally compiled.

eftiquar:
> @eftiquar, tests are almost in place. When #4631 is merged, you should be able just to remove conditional compilation from `ContinuousProfilerContextTrackingTests`, `ContinuousProfilerSpanStoppageHandlingTests`, and `ContinuousProfilerTests`.
> 
> For now, all of them are red on .NET Fx on your branch locally compiled.

Thanks @Kielek I have merged your changes into my branch. I will verify the tests.

test/IntegrationTests/ContinuousProfilerTests.cs
```diff
@@ -17,6 +14,7 @@ public ContinuousProfilerTests(ITestOutputHelper output)
     {
     }
 
+#if NET // allocator tests are only supported on .NET
```
Kielek, line 17: Can we at least test that the application is not failing if env vars are set - just zero-code returning?

eftiquar: Added a test to assert that continuous profiler gracefully ignores allocation sampling and does not cause any undefined behavior or crash.

test/IntegrationTests/ContinuousProfilerContextTrackingTests.cs
```diff
@@ -45,19 +42,41 @@ private bool AssertAllProfiles(ICollection<ExportProfilesServiceRequest> profile
...
             Assert.True(samplesWithTraceContext.Count <= 1, "at most one sample in a batch should have trace context associated.");
-
+#endif
             totalSamplesWithTraceContextCount += samplesWithTraceContext.Count;
             if (samplesWithTraceContext.FirstOrDefault() is { } sampleWithTraceContext)
             {
-                managedThreadsWithTraceContext.Add(profile.AttributeTable[sampleWithTraceContext.AttributeIndices.Single()].Value.StringValue);
+                var threadId = GetThreadName(profile, sampleWithTraceContext);
+                Output.WriteLine($"Found trace context on thread: {threadId}");
+                managedThreadsWithTraceContext.Add(threadId!);
```
Kielek, line 52: I have reverted this file, except removing conditional compilation at the beginning of this file.
It still passes.
What is the reason behind changes?

test/IntegrationTests/SelectiveSamplerTests.cs
```diff
@@ -69,6 +64,8 @@ public void ExportThreadSamples()
         collector.AssertExpectations();
     }
 
+#if NET
+// TODO Implement better tests for .NET Framework, Continuous Profiler is not stopping all threads at once.
```
Kielek: @eftiquar, there is some problem when both continuous profiling and selective sampling is used.
There are no samples exporter by the continuous profiler.

test/IntegrationTests/ContinuousProfilerContextTrackingTests.cs
eftiquar: It  was for debugging purpose - we can revert it.

test/IntegrationTests/SelectiveSamplerTests.cs
eftiquar: Let me check .. thanks for spotting this.

eftiquar: @Kielek - @lachmatt and I discussed this. The test assumes selective profiling when a batch contains only a single sample, creating ambiguity about whether sample count alone distinguishes selective from continuous profiling. In .NET Fx profiling, stack capture may abort if the required seed for the DoStackSnapshot API is missing, even if all safety probes pass. This can result in single-sample batches during continuous profiling, explaining why the test observed an absence of continuous profiling. Captures that always succeed in .NET may not always succeed in .NET Fx.

src/OpenTelemetry.AutoInstrumentation.Native/stack_capture_strategy_factory.h
```diff
@@ -0,0 +1,44 @@
...
+public:
+    /// @brief Creates appropriate strategy based on runtime information
+    /// @param profilerInfo CLR profiler API (ICorProfilerInfo12 for .NET Core, ICorProfilerInfo7* for .NET FW)
+    /// refer to:cor_profiler.cpp - line 105 - line 124 for details on why it is safe to cast to ICorProfilerInfo12
+    /// @param runtimeInfo Runtime platform information
+    /// @return Heap-allocated strategy (caller owns), or nullptr on unsupported platform
+    static std::unique_ptr<IStackCaptureStrategy> Create(
+        ICorProfilerInfo2* profilerInfo,
+        const trace::RuntimeInformation& runtimeInfo);
+// {
```
iskiselev: We probably should remove commented implementation, as actual implementation is in cpp.

src/OpenTelemetry.AutoInstrumentation.Native/stack_capture_strategy_factory.h
```diff
@@ -0,0 +1,44 @@
...
+#include "stack_capture_strategy.h"
+
+namespace continuous_profiler {
+
+/// @brief Factory for creating platform-specific stack capture strategies
+class StackCaptureStrategyFactory {
+public:
+    /// @brief Creates appropriate strategy based on runtime information
+    /// @param profilerInfo CLR profiler API (ICorProfilerInfo12 for .NET Core, ICorProfilerInfo7* for .NET FW)
+    /// refer to:cor_profiler.cpp - line 105 - line 124 for details on why it is safe to cast to ICorProfilerInfo12
```
iskiselev: Line numbers tend to change soon. It is better either describe why it is safe - or extract logic to separate method and refer to specific method.

src/OpenTelemetry.AutoInstrumentation.Native/stack_capture_strategy_factory.h
```diff
@@ -0,0 +1,44 @@
...
+    /// @param runtimeInfo Runtime platform information
+    /// @return Heap-allocated strategy (caller owns), or nullptr on unsupported platform
+    static std::unique_ptr<IStackCaptureStrategy> Create(
+        ICorProfilerInfo2* profilerInfo,
+        const trace::RuntimeInformation& runtimeInfo);
```
iskiselev, line 21: Do we really need `runtimeInfo` here? Would it be better to test that if we have ICorProfilerInfo12*, we should use it, otherwise on windows use another logic?

src/OpenTelemetry.AutoInstrumentation.Native/stack_capture_strategy_factory.cpp
```diff
@@ -0,0 +1,35 @@
...
+{
+
+std::unique_ptr<IStackCaptureStrategy> StackCaptureStrategyFactory::Create(ICorProfilerInfo2* profilerInfo,
+                                                                           const trace::RuntimeInformation& runtimeInfo)
+{
+
+    if (runtimeInfo.is_desktop())
+    {
+#ifdef _WIN64
+        // trace::Logger::Info("StackCaptureStrategyFactory: Creating NetFxStackCaptureStrategy");
```
iskiselev: Why have we comment Logger messages?

src/OpenTelemetry.AutoInstrumentation.Native/stack_capture_strategy_factory.cpp
```diff
@@ -0,0 +1,35 @@
...
+namespace continuous_profiler
+{
+
+std::unique_ptr<IStackCaptureStrategy> StackCaptureStrategyFactory::Create(ICorProfilerInfo2* profilerInfo,
+                                                                           const trace::RuntimeInformation& runtimeInfo)
+{
+
+    if (runtimeInfo.is_desktop())
+    {
+#ifdef _WIN64
```
iskiselev: Our code is specifically X64, so better to check `_M_AMD64` - it will help if we ever create windows ARM64 build, which will use the same path otherwise.

src/OpenTelemetry.AutoInstrumentation.Native/stack_capture_strategy.h
```diff
@@ -0,0 +1,84 @@
...
+/// - .NET Framework: Per-thread suspension + seeded DoStackSnapshot
+/// </summary>
+
+    using StackSnapshotCallbackRaw = HRESULT(*)(
+        FunctionID funcId,
+        UINT_PTR ip,
+        COR_PRF_FRAME_INFO frameInfo,
+        ULONG32 contextSize,
+        BYTE context[],
+        void* clientData);
```
iskiselev: It is hard to understand, as we have to many wrappers that casted to `void*`. As that function is internal use only, it may help to declare type of `clientData*` here explicitly - even if it means to use forward declaration.
Are we not using `std::function` wrapper here due to perfomance cost? Otherwise it may have requred internal clientData in closure, instea of passing it outside.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -1180,9 +1185,24 @@ void CorProfiler::InternalAddInstrumentation(WCHAR* id, CallTargetDefinition* it
 
 bool CorProfiler::InitThreadSampler()
 {
+#ifdef _WIN64
+    // for net fx, the native thread ID is needed by stack capture
+    // the profiler callback, ThreadAssignedToOSThread is not invoked for main thread
+    // for the following macihnery to work,
```
iskiselev: macihnery -> machinery

test/IntegrationTests/ContinuousProfilerContextTrackingTests.cs
Kielek: 8b6f2e3f2f41c9326c154b306112bbde8bc61714

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -1200,6 +1220,8 @@ bool CorProfiler::InitThreadSampler()
 
     this->continuousProfiler = new continuous_profiler::ContinuousProfiler();
     this->continuousProfiler->SetGlobalInfo12(this->info12_);
+    this->continuousProfiler->SetGlobalInfo7(this->info_);
+    this->continuousProfiler->SetStackCaptureStrategy(stack_capture_strategy_.get());
```
iskiselev, line 1224: Would it be better separation if we compelety hide `stack_capture_strategy_` inside `continuousProfiler`?

test/IntegrationTests/SelectiveSamplerTests.cs
eftiquar: I will improve debug level logs for net fx stack capture to note dropped stack samples due to lack of seed or other safety probe failures.

src/OpenTelemetry.AutoInstrumentation.Native/dot_net_stack_capture_strategy.h
```diff
@@ -0,0 +1,109 @@
...
+                auto errorString = "SuspendRuntime failed with HRESULT=" + std::to_string(suspendResult);
+                //trace::Logger::Error(errorString); // Log before throwing
+                throw std::runtime_error(errorString);
+            }
+        }
+
+        ~RuntimeSuspensionGuard() {
+            
+            if (HRESULT resumeHr = profilerInfo_->ResumeRuntime(); FAILED(resumeHr)) {
+                /*trace::Logger::Error("DotNetStackCaptureStrategy: ResumeRuntime FAILED! HRESULT=",
```
iskiselev: It would be probably nice to log something here. I reall hope, that we always can `ResumeRuntime` - but if it failed, does it mean that we supended customers application forever without saying anything anywhere?

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.h
```diff
@@ -0,0 +1,244 @@
...
+#include <corprof.h>
+#include "stack_capture_strategy.h"
+
+namespace ProfilerStackCapture {
+
+    // Forward declarations
+    class ILogger;
+
+    /// @brief Log severity levels for diagnostic output
+    enum class LogLevel {
```
iskiselev: Is there is some purpose of alternative logging system - or can we reuse same logger that is used in other part of project?

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.h
```diff
@@ -0,0 +1,244 @@
...
+
+    struct StackFrame {
+        FunctionID functionId;
+        bool isNative;
+        StackFrame(FunctionID id, bool native = false) : functionId(id), isNative(native) {}
+    };
+
+    struct CaptureOptions {
+        std::chrono::milliseconds probeTimeout = std::chrono::milliseconds(250);
+        std::chrono::milliseconds snapshotInterval = std::chrono::milliseconds(1000);
```
iskiselev: Looks like it is not used, can it be removed?

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.h
```diff
@@ -0,0 +1,244 @@
...
+        std::atomic<bool> stop_{false};
+        std::unique_ptr<std::thread, ThreadJoinerOnDelete> worker_;
+        void WorkerLoop();
+    };
+
+    struct CanaryThreadInfo
+    {
+        ThreadID     managedId = 0;
+        DWORD        nativeId  = 0;
+        std::wstring name;
```
iskiselev: Looks like `name` is not used anywhere

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.cpp
```diff
@@ -0,0 +1,677 @@
...
+        UNWIND_HISTORY_TABLE historyTable = {};
+        DWORD64              imageBase    = 0;
+        RtlLookupFunctionEntry(canaryCtx.Rip, &imageBase, &historyTable);
+
+        // Test 3: DoStackSnapshot
+        auto probeCallback = [](FunctionID, UINT_PTR, COR_PRF_FRAME_INFO, ULONG32, BYTE[], void*) -> HRESULT
+        { return S_FALSE; };
+
+        result =
+            profilerApi->DoStackSnapshot(canaryManagedId, probeCallback, COR_PRF_SNAPSHOT_DEFAULT, nullptr, nullptr, 0);
```
iskiselev, line 111: We do non-seeded async stack snapshot. I beleive it should be safe at most time - but we should be very sure about current state of Canary thread waitng, and not doing any work.
If we are sure, worth commenting why - otherwise we may need do it seeded.

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.cpp
```diff
@@ -0,0 +1,677 @@
...
+        trace::Logger::Debug("[StackCapture] SafetyProbe: Enqueueing probe operations with timeout=",
+                             options_.probeTimeout.count(), "ms");
+
+        status = invocationQueue_->Invoke(
+            [this, canaryManagedId, canaryCtx, &snapshotHr]()
+            {
+                HRESULT hr = ExecuteProbeOperations(profilerApi_.get(), canaryManagedId, canaryCtx);
+                snapshotHr.store(hr);
+            },
+            options_.probeTimeout);
```
iskiselev, line 506: We probably need to document known issues with that approach:
- we wait for canary thread snapshot some constant time
- usually canary thread snapshot should be fast
- if it take longer - most probably we detected lock, so that snapshot would be skipped
- we can't wait too long - otherwise in case of lock, we would supend application for long time
- on other side, it is known that if customer app has high load and resource hungry, snapshot may take longer time - and we would not proceed despite there was no lock
- so, we may completlt miss snapshots when application is most busy - and it may be intersted to end user

We discussed with changing alorithm from constant sleep, to some number of yeilds in low-priority thread, that will work as dynamic timer. We have not come up with nice implementation yet.

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -326,9 +318,13 @@ private static bool TryInitializeContinuousSamplingExport(
             return false;
         }
 
+#if NET
         var threadSamplesMethod = exportThreadSamplesMethod.CreateDelegate<Action<byte[], int, CancellationToken>>(continuousProfilerExporter);
         var allocationSamplesMethod = exportAllocationSamplesMethod.CreateDelegate<Action<byte[], int, CancellationToken>>(continuousProfilerExporter);
-
+#else
+        var threadSamplesMethod = (Action<byte[], int, CancellationToken>)exportThreadSamplesMethod.CreateDelegate(typeof(Action<byte[], int, CancellationToken>), continuousProfilerExporter);
```
iskiselev, line 354: Probably easier to read if .NET Framework overload will be used for both.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -336,7 +336,14 @@ void ThreadSamplesBuffer::StartSample(ThreadID                   id,
...
     WriteByte(kThreadSamplesStartSample);
-    WriteString(state->thread_name_);
+    if (state->thread_name_.empty())
+    {
+        WriteString(trace::ToWSTRING(std::to_string(id)));
+    }
+    else
+    {
+        WriteString(state->thread_name_);
+    }
```
Kielek: Why this change? ThreadId != ThreadName. If it is not available, it should be just empty.
Was it for debugging purpises?

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
Kielek: ac55229114cdb4c279be57d884b9b77823bb6445

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.h
Kielek: +1 to this, If possible utilize existing infrastructure

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.h
Kielek: debugging purposes only? I agree that we can remove if not needed

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
eftiquar: on .Net thread-pool threads have names on net fx that is not guaranteed. We need to produce unique list of threads that have context. Thead ID is guaranteed to be always present.

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.h
eftiquar: Yes that is to be removed. I have yet to do final cleanup there is unused code which I will remove.

src/OpenTelemetry.AutoInstrumentation.Native/stack_capture_strategy_factory.h
eftiquar: There is a pending task on my to do list - removing dead code and comments. Thanks for looking at this.

src/OpenTelemetry.AutoInstrumentation.Native/stack_capture_strategy_factory.cpp
zacharycmontoya: Are you saying that the NetFxStackCaptureStrategy is intended only for NetFx X64? If so, maybe we could rename that strategy type to include X64 to make it more clear

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -1180,9 +1185,24 @@ void CorProfiler::InternalAddInstrumentation(WCHAR* id, CallTargetDefinition* it
 
 bool CorProfiler::InitThreadSampler()
 {
+#ifdef _WIN64
```
zacharycmontoya: Related to another one of Igor's comments, is this support for ARM64 as well (in the case we add support for .NET Framework ARM64)? Or should we limit this to X64 as Igor mentioned with `_M_AMD64`?

eftiquar: Thank you @zacharycmontoya for reviewing this. We should limit this to X64, which is what current implementation supports. Support for ARM 64 will be introduced later.

eftiquar:
> @eftiquar, thanks for the PR. Most part looks good to me (not C++ expert). I think that we have good enough tests coverage for this.
> 
> As a separate PR it will be great to include description provided in this PR. It should go to https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/internal/continuous-profiler.md
> 
> Looking to address/resolve comments from @iskiselev.

Sure thing. I am working on resolving the review feedback.

src/OpenTelemetry.AutoInstrumentation.Native/stack_capture_strategy_factory.h
eftiquar: If I understand the comment - if ICorProfilerInfo12 is available, use runtime suspension strategy, else use thread suspension strategy.
The strategy is function of runtime type - using ICorProfilerInfo version will technically work, but will be less cleaner. 
The assumption is, for net fx we will use thread suspension and for .Net we will use CLR suspension. I wanted to be consistent with existing method of distinguishing .Net against net fx. Let me know if there is any other overarching concern.

iskiselev: You correctly understand my comment.
Current approach is also working, but may be riskier in some edge cases - as we expect that the code will never be executed (prevented by version guard) in profiler if we run it on .NET Core (older than 5). Otherwise, it may crash on interface upcast.
I think that with only one parameter to check it will be harder to misuse it, though I'm not insisting on change here.

src/OpenTelemetry.AutoInstrumentation.Native/stack_capture_strategy.h
eftiquar: I have significantly simplified this area. I am hoping that it will be more clear now.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
eftiquar: Stack capture engine needs to start accumulating thread metadata much before continuous profiler is initialized. We will miss thread associations that precede cont. profiler initialization.

src/OpenTelemetry.AutoInstrumentation.Native/dot_net_stack_capture_strategy.h
eftiquar: It is defensive code- resumption of runtime is only attempted if suspension succeeds due to c-tor d-tor rules. 
So if it fails direct your blame to the .Net CLR :) 
This code is no worse than pre-existing implementation. In fact there was a comment which wondered if it is safe to resume runtime if we previously failed to suspend it. But that part is history now.

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.h
eftiquar: Relics of my protype which was later adapted here. I have removed it.

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.h
eftiquar: Dropping it ..

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.cpp
eftiquar: We now have dedicated canary which is in perpetual wait state at well known location. Hence no stack seeding needed.

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.cpp
eftiquar: I will update the design description that appears at the top.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
eftiquar: I have renamed the class to reflect explicit x64 support.

eftiquar:
> In general, this approach makes sense and the implementation looks reasonable. Thank you for the detailed description in the PR summary.
> 
> Parts of the PR that I didn't feel especially knowledgeable on and could use deeper review from others:
> 
> * Snapshot implementation / unwinding in `PrepareContextForSnapshot`
> * Composition of callbacks via casting of pointers

I have removed casting from the call back machinery. Looks much cleaner now. @iskiselev has looked into PrepareContextForSnapshot.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
Kielek: I have removed this change in scope of 62c8e065f38bde1c463bd92066395e4a97b57d51
I think that is was important for one of the tests. Backends should handle correctly cases when the thread name is not available. Lack of data is better than putting there obviously wrong data.

rajkumar-rangaraj (approved):

src/OpenTelemetry.AutoInstrumentation.Native/OpenTelemetry.AutoInstrumentation.Native.vcxproj
```diff
@@ -196,13 +197,17 @@
     <ClInclude Include="miniutfdata.h" />
     <ClInclude Include="module_metadata.h" />
     <ClInclude Include="netfx_assembly_redirection.h" />
+    <ClInclude Include="netfx_stack_capture_strategy.h" />
```
iskiselev:
```suggestion
    <ClInclude Include="netfx_stack_capture_strategy_x64.h" />
```

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.h
```diff
@@ -0,0 +1,170 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+#ifndef OTEL_PROFILER_STACK_CAPTURE_H_
+#define OTEL_PROFILER_STACK_CAPTURE_H_
+
+#if defined(_WIN32) && defined(_M_AMD64)
+
+#include <windows.h>
+#include <vector>
```
iskiselev: Looks like it is not used

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -219,7 +217,42 @@ public static void Initialize()
         }
     }
 
-#if NET
+    private static void TryInitializeContinuousProfiling()
+    {
+#if NETFRAMEWORK
+        // Check AppDomain FIRST, before touching the flag
+        if (!AppDomain.CurrentDomain.IsDefaultAppDomain())
```
iskiselev: I beleive on ASP.NET app we do not inited in default app domain, so that would not be executed - and we would not have canary thread.

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.cpp
```diff
@@ -0,0 +1,669 @@
...
+    else
+    {
+        trace::Logger::Debug("[StackCapture] CaptureStackSeeded - Successfully captured stack. ThreadID=",
+                             managedThreadId);
+    }
+
+    return hr;
+}
+
+bool StackCaptureEngine::IsManagedFunction(BYTE* ip) const
```
iskiselev: Is it used anywhere?

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.cpp
```diff
@@ -0,0 +1,669 @@
...
+            }
+            // no runtime function,  manually unwind to previous frame, adjust the RIP and RSP fields
+            pContext->Rip = returnAddress;
+            pContext->Rsp += sizeof(DWORD64);
+            instructionPointer = returnAddress;
+        }
+        else
+        {
+            // Has unwind info - use function begin address (critical for CLR detection)
+            instructionPointer = imageBase + runtimeFunction->BeginAddress;
```
iskiselev, line 192: Why do we need use begin of function and not just `pContext->Rip`?

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -1180,9 +1185,24 @@ void CorProfiler::InternalAddInstrumentation(WCHAR* id, CallTargetDefinition* it
 
 bool CorProfiler::InitThreadSampler()
 {
+#if defined(_WIN32) && defined(_M_AMD64)
```
iskiselev, line 1188: Do we really need `_M_AMD64` check here? We may forget to remove it when add other architectures strategies - and there should be no harm to call it even if it is not needed.

iskiselev: BTW, why don't we need it for CoreCLR? How we will now native<->managed map for main thread there? Or is it not used there?
Prebably we don't need any ifdef here.

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.cpp
```diff
@@ -0,0 +1,669 @@
...
+#include "profiler_stack_capture.h"
+
+#include <stdexcept>
+#include <algorithm>
+#include <cassert>
+#include <dbghelp.h>
+#include <atlbase.h>
+#include <atlcom.h>
+#include <sstream>
+#include <iomanip>
```
iskiselev: Resharper suggests sstream and iomanip are not used.

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -31,6 +26,7 @@ internal static class Instrumentation
     private static readonly Lazy<LoggerProvider?> LoggerProviderFactory = new(InitializeLoggerProvider, true);
 
     private static int _initialized;
+    private static int _continuousProfilingInitialized; // Process-wide guard for continuous profiling
```
iskiselev: It is not process-wide, but appdomain-wide.

In which cases in single app-domain scenario may we have it be called twice? Do we need it at all?

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
eftiquar: We need win32 here - native thread ID is needed only for the scenarios when we suspend thread which happens for Windows platform for net fx only. 
For .Net we do not need native thread ID as we do not suspend threads - we suspend CR itself. .Net capturing is untouched by new implementation. 
Yes AMD 64 check can be dropped. win32 check is needed. GetCurrentThreadId is Windows API. I am deliberately using win API as opposed to profiler API to get current thread ID.

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
eftiquar: That is good catch - I have resolved this by updating the check that skips injecting the loader for IIS desktop - which is essentially what the ASP .Net scenario is.

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.cpp
eftiquar: removed.

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
eftiquar: we dont need it.

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.h
eftiquar: dropping

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.cpp
eftiquar: dropping

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -1525,9 +1535,23 @@ HRESULT STDMETHODCALLTYPE CorProfiler::JITCompilationStartedOnNetFramework(Funct
...
-                                caller.name == WStr("InvokePreStartInitMethods");
+        // For IIS: prefer injection into BuildManager.InvokePreStartInitMethods (application AppDomain)
+        // but also allow injection into default AppDomain for continuous profiling support
+        bool is_build_manager_injection = module_metadata->assemblyName == WStr("System.Web") &&
+                                          caller.type.name == WStr("System.Web.Compilation.BuildManager") &&
+                                          caller.name == WStr("InvokePreStartInitMethods");
+
+        bool is_default_appdomain = corlib_module_loaded && module_metadata->app_domain_id == corlib_app_domain_id;
+
+        valid_loader_callsite = is_build_manager_injection || is_default_appdomain;
```
iskiselev: It may be better to have multiple Canary threads (one thread per app domain) than change established managed code initializaion.
Suggested modification may move Initialization in ASP.NET appdomains to earlier point, which would require investiogation if it is safe.

iskiselev (approved):

src/OpenTelemetry.AutoInstrumentation.Native/profiler_stack_capture.cpp
eftiquar: WHy use BeginAddress instead of current RIP for GetFunctionFromIP ?
  kernel32.dll!SleepEx+0x42         RIP points here (middle of SleepEx)
  clr.dll!TransitionStub+0x0        Entry point (Begin Address)
  MyApp.dll!DoWork+0x15             Managed method (mid-function)
  // Fails:  instructionPointer = somewhere inside SleepEx
  instructionPointer = currentRIP;
  GetFunctionFromIP(instructionPointer, &fid);  // Returns fid = 0 (not found)

  // Succeeds: instructionPointer = Entry point of TransitionStub
  instructionPointer = imageBase + runtimeFunction->BeginAddress;
  GetFunctionFromIP(instructionPointer, &fid);  // Returns valid fid!

PR #4610, opened 2025-11-17
.NET10 Support

### LINKED ISSUE #4068: .NET10 support
Bring support for .NET10

Kielek:
## Why

Fixes #4068 
Based on #4608 and #4607

## What

Add support for .NET10.
There nuke.build left on .NET9 - legacy Ubuntu is not supporting .NET10. What is more, there is some issues with working nuke on .NET10. I will create later steps to reproduce and check with nuke/.net team what we can do with this, - This is tracked under #4612. and should not be considered as a blocker.

## Tests

CI

## Checklist


Common.props
```diff
@@ -9,6 +9,11 @@
     <NuGetAuditLevel>low</NuGetAuditLevel>
   </PropertyGroup>
 
+  <!-- <PropertyGroup>
```
zacharycmontoya: Can we delete this now?

zacharycmontoya (approved):

Common.props
Kielek: 🤦 it should never be commited. Fixing.

PR #4655, opened 2025-11-26
Add -NoReset option to service registration functions (#4641)

### LINKED ISSUE #4641: Register-OpenTelemetryForWindowsService should allow not restarting the service
# Feature Request

- I want to setup .NET Autoinstrumentation on Windows machine. 
- I am using https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/script-templates/OpenTelemetry.DotNet.Auto.psm1.template to do this
- I have group of services that will start in a controlled sequence, with delays etc.
- After services have been registered, I want to call  Register-OpenTelemetryForWindowsService for each of them, and decide by myself if/when/how the services are started.
- Currently, this function does  Restart-Service -Name $WindowsServiceName -Force

I would like to be able to avoid Restart-Service call somehow.

This could be e.g. -NoRestart option to skip the restart, like.

Register-OpenTelemetryForWindowsService -WindowsServiceName Foo -OTelServiceName Foo -NoRestart

vivainio:
Affected functions:

## Why

Users want to perform service restart and iisreset manually, separate from registration

Fixes #4641 

## What

Added -NeReset for:

- Register-OpenTelemetryForIIS
- Register-OpenTelemetryForWindowsService
- Unregister-OpenTelemetryForWindowsService

## Tests

Didn't want to add the test for these in the docker test as they would disrupt the flow. I *could* add a call to registration with -NoReset and perform service restart manually, but this would remove the test for the default flow.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,14 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
 
 ### Added
 
+- Add `-NoReset` switch parameter to skip service restart in PowerShell registration functions.
+
 ### Changed
+
+- `-NoReset` parameter in `Unregister-OpenTelemetryForIIS` changed from
+  `[bool]` to `[switch]`. Use `-NoReset` instead of `-NoReset $true`.
+  Note: `-NoReset $true` still works, but `-NoReset $false` will not do what you expect.
+  You should omit the parameter for the expected behavior.
 
 #### Dependency updates
 
--- a/docs/iis-instrumentation.md
+++ b/docs/iis-instrumentation.md
@@ -24,7 +24,8 @@ occurred. See this [issue](https://github.com/open-telemetry/opentelemetry-dotne
 for further details.
 
 > [!WARNING]
-> `Register-OpenTelemetryForIIS` performs IIS restart.
+> `Register-OpenTelemetryForIIS` performs IIS restart by default.
+> Use `-NoReset` to skip the restart.
 
 ## Configuration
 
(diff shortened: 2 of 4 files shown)


vivainio:
I would prefer the syntax to be "-NoReset" instead of "-NoReset $true" (i.e. using powershell [switch] type, but "-NoReset $true" was used in existing code.  Let me know if I should change these new additions to use [switch]

RassK:
The switch type would be better, so the cmd would look more natural. Need to refator the uninstall part also for consitency.

vivainio:
@RassK I changed these functions to use [switch]. I have documented the compatibility break for Unregister-OpenTelemetryForIIS in Changelog: 
https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4655/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR16

RassK:
I'm not sure if it's technically a breaking change. PS should ignore the $true and treat no flag when $false is passed.

vivainio:
@RassK based on my check, when using [switch], `-NoReset $true` will work as before. But `-NoReset $false` will be interpreted the same as `-NoReset $true`, i.e. it will cause reset. 

So, there is a small change in the functionality, but this probably affects a very marginal and contorted use case where user called `-NoReset $false` double negative instead of omitting the argument whatsoever. 

I will document only this change in behavior in Changelog.

vivainio:
@RassK guidance on how to implement the test automation for this would be appreciated. Would it be something like calling these function with -NoReset once before calling it with Reset?

RassK:
I think these are not covered currently, only manual verificaton.

PR #4661, opened 2025-12-01
Enable Roslyn analysis - SourceGenerator

Kielek:
## Why &  What

Enable Roslyn analysis (`<AnalysisLevel>latest-All</AnalysisLevel>`) - SourceGenerator

## Tests

CI

## Checklist

### DIFF
--- a/docker/centos-stream9.dockerfile
+++ b/docker/centos-stream9.dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/centos/centos:stream9@sha256:8f23193019c25232ea9a657fcaf0cca2570e1bdb3be54d7bd56fbb05f3eeaa02
+FROM quay.io/centos/centos:stream9@sha256:e7094d722b6c1e8ecdd774cacc4769baa3ce6591adecee94a66043f20ffc8baf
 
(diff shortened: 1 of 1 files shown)


src/SourceGenerators/InstrumentationDefinitionsGenerator.cs
```diff
@@ -134,8 +132,10 @@ private static string GenerateInstrumentationDefinitionsPartialClass(
 
         var generatedMethodName = integrationKind == IntegrationKindDirect ? "GetDefinitionsArray" : "GetDerivedDefinitionsArray";
 
+#pragma warning disable SA1118 // Parameter should not span multiple lines
```
nrcventura: Is this needed because of the string constant in that parameter list? If so, should we just declare that string constant somewhere else and then just reference it in the parameter list?

Kielek: Fixed: 46cebf04fa6d9c031f209f31eedcd6e22b502aec

zacharycmontoya (approved):

PR #4665, opened 2025-12-02
Enable Roslyn analysis - StartupHook

Kielek:
## Why &  What

Enable Roslyn analysis (`<AnalysisLevel>latest-All</AnalysisLevel>`) - StartupHook

## Tests

CI

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/ApplicationInExcludeListRule.cs
+++ b/src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/ApplicationInExcludeListRule.cs
@@ -45,9 +45,7 @@ internal class ApplicationInExcludeListRule : Rule
         {
             return Process.GetCurrentProcess().MainModule.ModuleName;
         }
-#pragma warning disable CA1031 // Do not catch general exception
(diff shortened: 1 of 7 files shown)


src/OpenTelemetry.AutoInstrumentation.StartupHook/RulesEngine/ApplicationInExcludeListRule.cs
```diff
@@ -45,7 +45,9 @@ private static string GetProcessModuleName()
         {
             return Process.GetCurrentProcess().MainModule.ModuleName;
         }
+#pragma warning disable CA1031 // Do not catch general exception
```
nrcventura: Would it be better to disable this rule at the project level? It seems like this rule is not really applicable to this specific project, because I cannot think of situations where we want to avoid catching all errors.

Kielek: Done in scope of 0be1f4242a7da93330307964a78c9b1af531c546

I have kept couple of muted issues common for this project and main AutoInstrumentation project.

PR #4710, opened 2025-12-15
Enable Roslyn analysis - NuGetPackagesTests

Kielek:
## Why &  What

Enable Roslyn analysis (`<AnalysisLevel>latest-All</AnalysisLevel>`) - NuGetPackagesTests

Most of the changes is related to the shared code with other packages.

## Tests

CI

## Checklist

### DIFF
--- a/test/IntegrationTests/Helpers/HealthzHelper.cs
+++ b/test/IntegrationTests/Helpers/HealthzHelper.cs
@@ -11,10 +11,11 @@ namespace IntegrationTests.Helpers;
 
 internal static class HealthzHelper
 {
+    private static readonly HttpClient HttpClient = new();
+
     public static async Task TestAsync(string healthzUrl, ITestOutputHelper output)
     {
         output.WriteLine($"Testing healthz endpoint: {healthzUrl}");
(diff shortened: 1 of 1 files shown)


test/IntegrationTests/Helpers/HealthzHelper.cs
```diff
@@ -14,29 +14,31 @@ internal static class HealthzHelper
     public static async Task TestAsync(string healthzUrl, ITestOutputHelper output)
     {
         output.WriteLine($"Testing healthz endpoint: {healthzUrl}");
-        HttpClient client = new();
+        using HttpClient client = new();
```
nrcventura: We might not want to dispose this client immediately. We may want to consider having a shared HttpClient instance or have an HttpClientFactory that we can use instead.

Kielek: Switched to static HttpClient in this helper. I think that it will be fine to have even couple of intances across whole tests infrastructure. I suppose more occurs when we enable analysis in other projects. 1e96cb8252b5f379062686696c805e1ee88b01ec

Issue #4541, opened 2025-10-23
[File Based Configuration] Native code fixes


Kielek:
Native code is reading bunch of env. vars directly. Need to implement way to reading some/all these values from the file.

One of the options should be propagation data from managed code.

Ref: src/OpenTelemetry.AutoInstrumentation.Native/environment_variables.h

PR #4721, opened 2025-12-17
[FileBasedConfiguration] Support for IL rewrite for SqlClient

### LINKED ISSUE #4561: File based configuration - propagate config to native part
Ideally, it should be propagated from the managed coded. We are in the middle of introducing File based configuration and it need to be reworked in the near feature. Not a blocker for this PR IMO.

Ref: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4492

It affects not only newly created method. I think bunch of other should be also propagated from the managed code.

_Originally posted by @Kielek in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4470#discussion_r2453917272_

ysolomchenko:
## Why

Towards #4541
Fixes #4561

## What

Added support for file-based configuration of the IL rewrite for SqlClient instrumentation setting

## Tests

Ci

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation.Native/CMakeLists.txt
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/CMakeLists.txt
@@ -150,6 +150,7 @@ endif()
 add_library("OpenTelemetry.AutoInstrumentation.Native.static" STATIC
         class_factory.cpp
         clr_helpers.cpp
+	configuration.cpp
         continuous_profiler_clr_helpers.cpp
         continuous_profiler.cpp
         cor_profiler_base.cpp


src/OpenTelemetry.AutoInstrumentation.Native/environment_variables_util.h
Kielek, line 1: I have some doubts about file name here.
It is no longer related only to env vars.
I would consider moving it to better naming/creating similar for configuration delivered by managed code. There is a chance to fully remove this file, when all cases are covered iin the same way.

test/IntegrationTests/SqlClientSystemTests.cs
```diff
@@ -18,14 +18,16 @@ public SqlClientSystemTests(ITestOutputHelper output, SqlServerFixture sqlServer
         _sqlServerFixture = sqlServerFixture;
     }
 
-    public static TheoryData<string, bool> GetDataForIlRewrite()
+    public static TheoryData<string, bool, bool> GetDataForIlRewrite()
```
Kielek, line 21: Time to execute will be double on .NET Fx. Could you please check how many additional cases we have/how long it takes?

ysolomchenko: Previously there were 4 test cases, now there are 8.

PR #4724, opened 2025-12-19
chore(): Documents mTLS environment variables and usage notes

sandy2008:
## Why

The OpenTelemetry .NET SDK (version 1.14.0+) supports mTLS (mutual TLS) authentication for OTLP exporter via environment variables, but this feature is not documented in the OpenTelemetry .NET Auto Instrumentation repository. Users may not be aware that they can configure mTLS for secure communication with their collectors.

## What
Added documentation for mTLS configuration in `config.md`.

## Tests

- Documentation-only change, no code changes.

## Checklist

### DIFF
--- a/docs/file-based-configuration.md
+++ b/docs/file-based-configuration.md
@@ -91,6 +91,16 @@ file-based configuration to include these parameters.
 
 ---
 
+### OTLP
+
+| Environment variable                    | Description                                                                               |
+| --------------------------------------- | ----------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


Kielek:
@sandy2008, I think that it is not yet released with OTel .NET SDK. It will be included into 1.15.0. We should way with merge here.

sandy2008:
> @sandy2008, I think that it is not yet released with OTel .NET SDK. It will be included into 1.15.0. We should way with merge here.

@Kielek Got it, thank you!

PR #4728, opened 2025-12-24
Implemented assembly redirect for non-default AppDomain

iskiselev:
## Why
Exisitng solution #4187 force LoaderOptimization.SingleDomain for every additional app domains. It is not standard behaviour and may result in increased memory consumption and slower warm-up for process hosting multiple web-applications - and other processes that use additional AppDomains.

Implement second stage of fix for #4186.

Parts of implementation uses code generated by Codex with GPT-5.4 model

## What
Added environment variable `OTEL_DOTNET_AUTO_APP_DOMAIN_STRATEGY` that controls strategy of fixing assembly conflicts in non-default app domains.
Possible values:
- `None` - do not apply anything. Works with test app and OTEL 1.14, crash with #4186 on OTEL 1.13. May be useful for custom host that creates AppDomains.
- `LoaderOptimizationSingleDomain` - exisiting solution implemented in #4187 - all additional AppDomains use `LoaderOptimization.SingleDomain` - used by default
- `AssemblyRedirect` - new stategy - patch assemblyBindings/dependentAssembly/bindingRedirect in memory-loaded app.config. On test app docker requires app pool to loading user profile if assemblies loaded from folder (and has no issues when assemblies are registered in GAC). On my local computer works even without loading user profile for app pool.

## Tests
AspNetTests.SubmitsTraces now test additional modes.
AssemblyBindingUpdaterTests - unit test coverage
Tested all strategies with OTEL assemblies in GAC and not in GAC on web sites that does/does not use newer version of dependent assemblies. For dependent assembly with newer version System.Diagnostics.DignosticsSource 11 was used.

## Checklist
- ~~How new setting should be integrated with file-based config? - documented as not supported~~


iskiselev:
~~Some special processing required for MongoDB - should be validated. Probably we should not include MongoDB dll in our distro at all, if we are not loading it?~~

SOLVED: We do not use any special processing for MongoDB anymore and do not include it in distro

src/OpenTelemetry.AutoInstrumentation.Loader/AppConfigUpdater.cs
```diff
@@ -2,23 +2,193 @@
...
-    /// Will be called through reflection when new <see cref="System.AppDomain"/> created.
-    /// Call done using bytecode modifications for <see cref="AppDomain.CreateDomain(string,System.Security.Policy.Evidence,System.AppDomainSetup)"/>
-    /// and <see cref="AppDomainManager.CreateDomainHelper(string,System.Security.Policy.Evidence,System.AppDomainSetup)"/>.
+    /// Modify assembly bindings in appDomainSetup
     /// </summary>
     /// <param name="appDomainSetup">appDomainSetup to be updated</param>
     public static void ModifyConfig(AppDomainSetup appDomainSetup)
     {
-        appDomainSetup.LoaderOptimization = LoaderOptimization.SingleDomain;
+        var patchMode = Environment.GetEnvironmentVariable("OTEL_APP_DOMAIN_STRATEGY") ?? string.Empty;
```
iskiselev: Should we try to parse app.config when we check it? Which app.config should be used - creator domain or target domain?
How should be file-based config be integrated here?
For now, I plan to add it to the list of settings supported only by environment variables.

src/OpenTelemetry.AutoInstrumentation.Loader/AppConfigUpdater.cs
```diff
@@ -2,23 +2,223 @@
...
+                {
+                    existingNewVersion = new Version(versionString);
+                }
+                catch (Exception ex)
+                {
+                    Logger.Error(ex, $"Version {versionString} not parsed for {assemblyInfo.FullName.Name}. Treat it as 0.0.");
+                    existingNewVersion = new Version(0, 0);
+                }
+
+                if (existingNewVersion >= assemblyInfo.Version)
```
iskiselev: It would be bad if app uses newer version of  the same assembly that we need, but does not have binding redirect.
To solve it better, we may need implement scan of assembly folders and GAC to check if newer version exisists. I do not plan to do it in that PR, but can be done later if any requests will come.

docs/config.md
```diff
@@ -463,19 +464,20 @@ Important environment variables include:
...
+| `OTEL_DOTNET_AUTO_APP_DOMAIN_STRATEGY`              | Defines how AppDomains are handled by the Auto Instrumentation. Supported values: `LoaderOptimizationSingleDomain` - apply `LoaderOptimization.SingleDomain` for non-default AppDomain; `AssemblyRedirect` - modify bindingRedirect an app config; `None` - do not apply any workarounds. See [.NET Framework AppDomain Strategy](./netfx-appdomain-strategy.md) for details.                      | `LoaderOptimizationSingleDomain` | [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md) |
```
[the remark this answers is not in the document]
iskiselev: Fixed

src/OpenTelemetry.AutoInstrumentation/Util/ManagedProfilerLocationHelper.cs
```diff
@@ -13,6 +13,44 @@ internal static partial class ManagedProfilerLocationHelper
...
+            }
+
+            var targetDirPath = Path.Combine(runtimeDir, targetDirName);
+            var assemblyFileName = Path.GetFileNameWithoutExtension(linkPath);
+            var assemblyName = Path.GetFileNameWithoutExtension(assemblyFileName);
+
+            var path = Probe(targetDirPath, assemblyName);
+            if (path == null)
+            {
+                logger?.Error($"Linked assembly path \"{Path.Combine(targetDirPath, $"{assemblyFileName}.dll")}\" does not exist");
```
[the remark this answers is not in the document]
iskiselev: Fixed.

src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyCatalog.cs
```diff
@@ -0,0 +1,129 @@
...
+
+                        if (info.Version < assemblyName.Version)
+                        {
+                            logger.Warning(
+                                $"Multiple files for {assemblyName.Name}, using ${assemblyName.Version} from {file}");
+                        }
+                        else
+                        {
+                            logger.Warning(
+                                $"Multiple files  for {assemblyName.Name}, using ${info.Version} from {info.Path}");
```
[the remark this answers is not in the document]
iskiselev: Fixed

src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyBindingUpdater.cs
```diff
@@ -0,0 +1,712 @@
...
+                        return null;
+                    }
+
+                    return new(CandidateSource.Probe, assemblyName.Version, ReadFileVersion(candidatePath), new Uri(candidatePath).AbsoluteUri);
+                }
+                catch (Exception ex)
+                {
+                    logger.Debug(ex, $"Failed to inspect probing candidate {candidatePath} for {assemblyInfo.FullName.Name}");
+                    return null;
+                }
```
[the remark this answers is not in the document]
iskiselev: Behaviour match .NET Framework behaviour. Comment added.

test/test-applications/integrations/TestApplication.AspNet.NetFramework/Dockerfile
```diff
@@ -12,15 +12,19 @@ RUN Import-Module .\OpenTelemetry.DotNet.Auto.psm1 -Verbose; `
     Install-OpenTelemetryCore -LocalPath .\tracer.zip -RegisterAssembliesInGAC $false; `
     Register-OpenTelemetryForIIS;
 ENV OTEL_DOTNET_AUTO_LOG_DIRECTORY=C:\inetpub\wwwroot\logs `
-    OTEL_LOG_LEVEL=debug
+    OTEL_LOG_LEVEL=debug
```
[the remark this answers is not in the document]
iskiselev: Fixed

test/OpenTelemetry.AutoInstrumentation.Loader.Tests/AssemblyBindingUpdaterTests.cs
```diff
@@ -0,0 +1,818 @@
...
+    private static readonly XNamespace AsmNs = "urn:schemas-microsoft-com:asm.v1";
+    private readonly string _rootDirectory = Path.Combine(Path.GetTempPath(), "AssemblyBindingUpdaterTests", Guid.NewGuid().ToString("N"));
+
+    public AssemblyBindingUpdaterTests()
+    {
+        Directory.CreateDirectory(_rootDirectory);
+    }
+
+    [Fact]
+    public void ModifyAssemblyRedirectConfig_UsesProfilerAssembly_WhenNoCustomerCandidateExists()
```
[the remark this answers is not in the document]
iskiselev: Not an error. Customer changed to Client

docs/netfx-appdomain-strategy.md
```diff
@@ -0,0 +1,295 @@
...
+  than cross-domain sharing.
+- Choose `AssemblyRedirect` when you want to preserve the multi-domain
+  model and provide the CLR enough binding information to keep the
+  assembly graph consistent.
+- Choose `None` only as a last resort, after extensive testing, when the
+  other options are not suitable or introduce unacceptable issues for
+  the application. In that mode you accept the risk that secondary
+  `AppDomain` creation may make instrumentation incomplete or unstable.
+
+## Scope of this document
```
zacharycmontoya: Thanks for the great detail in this markdown. I think if you move this `## Scope of this document` section before the `## Why secondary AppDomains are special` section, that could be helpful

test/IntegrationTests/AspNetTests.cs
```diff
@@ -48,16 +48,29 @@ public enum AppPoolMode
...
+    // But it resulted in app crash with "Loading this assembly would produce a different grant set from other instances" with OTEL 1.13.0
     [WindowsAdministratorTheory]
     [Trait("Category", "EndToEnd")]
     [Trait("Containers", "Windows")]
-    [InlineData(AppPoolMode.Classic, Gac.UseGac)]
-    [InlineData(AppPoolMode.Classic, Gac.UseLocal)]
-    [InlineData(AppPoolMode.Integrated, Gac.UseGac)]
-    [InlineData(AppPoolMode.Integrated, Gac.UseLocal)]
-    public async Task SubmitsTraces(AppPoolMode appPoolMode, Gac useGac)
+    [InlineData(AppPoolMode.Integrated, Gac.UseGac, AppDomainWorkaround.LoaderOptimizationSingleDomain)]
```
zacharycmontoya, line 72: Is there a reason to not test `AppPoolMode.Integrated` with all possibilities of `Gac` and `AppDomainWorkaround`?

iskiselev: Saving of electric eneregy :)
I do not expect that changing app pool mode between ingerated and classic will chenge anything in assembly loading. So, checking it only once will work (or checking classic only once).
I will add a comment to explain it.

test/test-applications/integrations/TestApplication.AspNet.NetFramework/Dockerfile
```diff
@@ -17,10 +17,14 @@ WORKDIR /inetpub/wwwroot
 COPY bin/${configuration}/app.publish .
 
 FROM integrated-base AS integrated-nogac
+RUN Start-IISCommitDelay;`
```
zacharycmontoya, line 20: I don't understand what these changes are for, can you explain?

iskiselev: >  `AssemblyRedirect` - ... On test app docker requires app pool to loading user profile if assemblies loaded from folder... 

So, here I enable loading of user profile for both IIS when assemlbies are not installed in GAC to make it compatible with `AssemblyRedirect` mode

src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyBindingUpdater.cs
```diff
@@ -0,0 +1,714 @@
...
+        {
+            logger.Debug(
+                FormatProbeDirectoryResolutionMessage(appDomainSetup, applicationBase, disableApplicationBaseProbe, resolvedDirectories));
+        }
+
+        return resolvedDirectories;
+    }
+
+#pragma warning disable SA1202
+    protected virtual string? GetCreatorApplicationBase()
```
zacharycmontoya: Can we inline this where it's called?

src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyBindingUpdater.cs
```diff
@@ -0,0 +1,714 @@
...
+        => AppDomain.CurrentDomain.BaseDirectory;
+#pragma warning restore SA1202
+
+    private static XElement CreateDependentAssembly(AssemblyCatalog.AssemblyInfo assemblyInfo)
```
zacharycmontoya: Can you isolate the helper methods like this, creating versions, etc. into another C# file? This file is very complex so keeping only the necessary loading logic in this file would be very helpful

src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyBindingUpdater.cs
```diff
@@ -0,0 +1,714 @@
...
+                    return null;
+                }
+            }
+        }
+
+        logger.Debug($"No probing candidate found for {assemblyInfo.FullName.Name}.");
+        return null;
+    }
+
+    private List<string> GetProbeDirectories(XElement assemblyBinding, AppDomainSetup appDomainSetup)
```
zacharycmontoya, line 453: I had a hard time following this method (maybe because of the inner static methods) so I fed it into Claude and it gave me the following summary, which I found helpful for giving some high-level context first then it was easier to look into each step. Do you think any of this is worth adding as a high-level comment?

```
  GetProbeDirectories (line 388)

  Replicates Fusion's probe path order:
  1. If DisallowApplicationBaseProbing → return empty.
  2. Resolves ApplicationBase; if missing, falls back to GetCreatorApplicationBase() (virtual, returns AppDomain.CurrentDomain.BaseDirectory) and writes it back into the setup. (:430-445, 565-566)
  3. Adds the application base — unless PrivateBinPathProbe != null, which disables it (matches CLR semantics).
  4. For web.config, adds {base}/bin. (:534-537)
  5. Splits PrivateBinPath on ; — for rooted entries, only keeps them if they're under ApplicationBase (security: prevents probing outside the app), warns otherwise. Relative entries are joined to the base.
  (:466-481)
  6. Same treatment for <probing privatePath="..."> from the config.
  7. Dedupes by full path, case-insensitive. Logs the resolved set with full context (config file, base, flags). (:393-428)
```

src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyBindingUpdater.cs
zacharycmontoya: Oh I see that this is used for testing. Can you comment the virtual / internal methods that are intended to be used only for testing?

src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyBindingUpdater.cs
zacharycmontoya, line 1: This file was rather difficult to review because there's a lot of Fusion / probing logic that I wasn't already familiar with, so I have two requests to hopefully help other reviewers/maintainers:
- If there are any public docs or links that explain some of the behaviors, can you link to it?
- If reasonable, could you provide code comments to explain the high-level "what" is the goal/logic of a given method? No need to explain the "how" as that's in code.

When reviewing this file, I told Claude verbatim `Walk me through the changes introduced in src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyBindingUpdater.cs starting from ModifyAssemblyRedirectConfig` and it was surprisingly insightful and provided a helpful method-by-method breakdown. This could also be a good source of method-level comments so I'd be happy to share if you'd like

zacharycmontoya (commented):
I've reviewed everything but the AssemblyBindingUpdaterTests.cs, which I can continue tomorrow

test/OpenTelemetry.AutoInstrumentation.Loader.Tests/AssemblyBindingUpdaterTests.cs
```diff
@@ -0,0 +1,818 @@
...
+        var result = ApplyConfig(
+            updater,
+            new AppDomainSetup { ApplicationBase = _rootDirectory },
+            CreateConfig(
+                profilerAssembly,
+                $$"""
+                <dependentAssembly>
+                  <assemblyIdentity name="{{TestAssemblyName}}" publicKeyToken="{TOKEN}" culture="neutral" />
+                  <bindingRedirect oldVersion="0.0.0.0-{{lowerRedirectVersion}}" newVersion="{{lowerRedirectVersion}}" />
+                  <codeBase version="{{customerVersion}}" href="{{customerHref}}" />
```
zacharycmontoya, line 375: Is this typical? I'd expect the user config to only provide a codebase with the version matching the original redirect version

zacharycmontoya (approved):
Overall the changes and testing look good, thanks!

Rhynier (approved):

docs/netfx-appdomain-strategy.md
iskiselev: Update in 633eb0ec5f7763476f130b5e41a9606d38b3fe5b

test/IntegrationTests/AspNetTests.cs
iskiselev: Comment added in 633eb0ec5f7763476f130b5e41a9606d38b3fe5b

src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyBindingUpdater.cs
iskiselev: Added comments in 633eb0ec5f7763476f130b5e41a9606d38b3fe5b

src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyBindingUpdater.cs
iskiselev: Moved helper methods / classes to another file. Make outer class partial to not expose visibility to outer scope.

src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyBindingUpdater.cs
iskiselev: Comments added in 633eb0ec5f7763476f130b5e41a9606d38b3fe5b

src/OpenTelemetry.AutoInstrumentation.Loader/AssemblyBindingUpdater.cs
iskiselev: Added comments. Hope they are enough to get the idea easier.

test/OpenTelemetry.AutoInstrumentation.Loader.Tests/AssemblyBindingUpdaterTests.cs
iskiselev: Not typical, most probably error in user config - but usefull in test to document / confirm how our logic works - that we update binding refirect to highest version even if it is not our version and not origianl biniding redirect - based on the fact that config claim where to get newer version.

iskiselev: At least if we break customer application in that case - that test will confirm that it is design. Bad config -> bad result, even if it was not exposed prior.

nrcventura (approved):

PR #4731, opened 2025-12-30
Enable Roslyn analysis - Test Applications - final

Kielek:
## Why &  What

Enable Roslyn analysis (`<AnalysisLevel>latest-All</AnalysisLevel>`) - Test Applications - final

`await usings` is not well handled by analyzers. For now, using this construction is suggestion only. Lack of ConfigureAwait in such cases - warning/error. I have converted all usages of `await usings` to just usings. We can fix it later when https://github.com/dotnet/roslyn-analyzers/issues/7185 is fixed

## Tests

CI

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
 - Support for [`NLog`](https://www.nuget.org/packages/NLog/)
   logs instrumentation for versions `5.*` and `6.*` on .NET using duck typing
   for zero-config auto-injection.
+- Support for file-based configuration of the IL rewrite for
+  SqlClient instrumentation setting
 
 ### Changed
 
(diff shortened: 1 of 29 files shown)


test/test-applications/integrations/TestApplication.GraphQL/StarWars/Types/HumanType.cs
```diff
@@ -3,7 +3,9 @@
 
 namespace StarWars.Types;
 
-public class HumanType : ObjectGraphType<Human>
+#pragma warning disable CA1812 // Avoid uninstantiated internal classes. This class is instantiated by GraphQL.
```
RassK: Seems very repetitive pattern, could use more global way to exclude. Either `.editorconfig` or global suppression by target namespace.

test/test-applications/integrations/TestApplication.Logs/Startup.cs
```diff
@@ -13,13 +15,13 @@ public Startup(IConfiguration configuration)
     public IConfiguration Configuration { get; }
 
     // This method gets called by the runtime. Use this method to add services to the container.
-    public void ConfigureServices(IServiceCollection services)
+    public static void ConfigureServices(IServiceCollection services)
```
RassK, line 18: can these really be static? I do not find any source currently to validate.

Kielek: Yes, it can be static. Tested. I have commented out this method. It leads to tests failures.

Image

test/test-applications/integrations/TestApplication.GraphQL/StarWars/Types/HumanType.cs
Kielek: I have added NoWarn tag to GraphQL test app only. I think that we can kept disabled issue by class for other. Typically, 1/2 classes in other projects.

RassK (approved):

PR #4747, opened 2026-01-12
[Infra] Fix building native DLL in Visual Studio

martincostello:
## Why

Trying to build the native DLL in Visual Studio 2026 produces the following error:

```
RC2127: version WORDs separated by commas expected
```

From here:

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/9624d8ac0add9fdb534ef6f39f8232f9dfa65515/src/OpenTelemetry.AutoInstrumentation.Native/Resource.rc#L62

## What

Set default values for the version parts to fix compiler errors when building in Visual Studio, as they are otherwise blank.

## Tests

Compiled the project successfully in Visual Studio 2026 after the change.

## Checklist

- [ ] ~~`CHANGELOG.md` is updated.~~
- [ ] ~~Documentation is updated.~~
- [ ] ~~New features are covered by tests.~~

PR #3845, opened 2024-12-02
Bytecode instrumentation MongoDB 

ysolomchenko:
## Why

Transitioning to bytecode instrumentation because the `MongoDB.Driver` and `MongoDB.Driver.Core` libraries are being merged into a single `MongoDB.Driver` library starting from version 3.0.0

Fixes https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3738

## What

Bytecode instrumentation for MongoDB with following attributes:

- db.system
- db.collection.name
- db.namespace
- db.operation.name
- server.address
- server.port
- network.peer.address
- network.peer.port

The implementation targets `MongoDB.Driver` and `MongoDB.Driver.Core` versions from 2.7.0 to the latest.

## Tests

Manual + CI

## Checklist

- [x] `CHANGELOG.md` is updated.
- [x] Documentation is updated.
- [x] New features are covered by tests.

PR #4748, opened 2026-01-12
[Redis] Add support for .NET Framework

### LINKED ISSUE #2564: Support StackExchange.Redis on .NET Framework
It seems that the blockers that existed when the instrumentation was originally added are not an issue anymore, see https://cloud-native.slack.com/archives/C01NR1YLSE7/p1684769872359809

While I don't recall any request to support this instrumentation on .NET Framework I don't see any reason to not make it happen. The only downside coming to my mind is limited coverage in CI since we don't have Windows runners that can run Linux containers, however, this is the same for any .NET Fx instrumentation that needs Linux containers for integration tests.

martincostello:
## Why

Fixes #2564

## What

Changes to support Redis instrumentation when targeting .NET Framework.

## Tests

Updated integration tests to also run for netfx TFMs.

## Checklist

### DIFF
--- a/test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt
+++ b/test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt
@@ -91,7 +91,6 @@
   \netfx\System.Runtime.CompilerServices.Unsafe.dll,
   \netfx\System.Text.Encodings.Web.dll,
   \netfx\System.Text.Json.dll,
-  \netfx\System.Threading.Channels.dll,
   \netfx\System.Threading.Tasks.Extensions.dll,
   \netfx\net462\Microsoft.Win32.Primitives.dll,
   \netfx\net462\System.AppContext.dll,


src/OpenTelemetry.AutoInstrumentation.Native/OpenTelemetry.AutoInstrumentation.Native.DLL.vcxproj
```diff
@@ -211,6 +211,11 @@
     <ClInclude Include="dllmain.h" />
     <ClInclude Include="resource.h" />
   </ItemGroup>
+  <PropertyGroup>
+    <OTEL_AUTO_VERSION_MAJOR>$([MSBuild]::ValueOrDefault('$(OTEL_AUTO_VERSION_MAJOR)', '1'))</OTEL_AUTO_VERSION_MAJOR>
+    <OTEL_AUTO_VERSION_MINOR>$([MSBuild]::ValueOrDefault('$(OTEL_AUTO_VERSION_MINOR)', '0'))</OTEL_AUTO_VERSION_MINOR>
+    <OTEL_AUTO_VERSION_PATCH>$([MSBuild]::ValueOrDefault('$(OTEL_AUTO_VERSION_PATCH)', '0'))</OTEL_AUTO_VERSION_PATCH>
+  </PropertyGroup>
```
Kielek: I have some doubts about this. Ideally, It will be great to execute here MinVer here. I have never found a good time/good way to do this.

I prefer to have some VS build failures instead of sneaking 1.0.0 to the release process.

src/OpenTelemetry.AutoInstrumentation/OpenTelemetry.AutoInstrumentation.csproj
```diff
@@ -17,12 +17,12 @@
...
 
-  <!-- Separate dependencies are to be shipped in the distribution but not on the NuGet package -->
+  <!-- Separate dependencies are to be shipped in the distribution but not in the NuGet package -->
   <ItemGroup Condition="$(_IsPacking) != true">
-    <!-- Dependencies below are shipped on the distribution folder, but not included in the NuGet package. -->
+    <!-- Dependencies below are shipped in the distribution folder, but not included in the NuGet package. -->
     <!-- If necessary these packages are flagged by custom build tasks and manually added to the targeted  -->
     <!-- projects that are using the NuGet package. -->
-    <PackageReference Include="OpenTelemetry.Instrumentation.StackExchangeRedis" Condition=" '$(TargetFramework)' == 'net8.0' " />
+    <PackageReference Include="OpenTelemetry.Instrumentation.StackExchangeRedis" Condition=" '$(TargetFramework)' == 'net8.0' OR '$([MSBuild]::GetTargetFrameworkIdentifier(`$(TargetFramework)`))' == '.NETFramework'" />
```
Kielek: I think that it can be now unconditional. Both .NET and .NET Fx handles this package in the same way.

```suggestion
    <PackageReference Include="OpenTelemetry.Instrumentation.StackExchangeRedis"  />
```

src/CommonExcludedAssets.props
```diff
@@ -1,15 +1,12 @@
...
-    <!-- System.Drawing.Common is transitive dependency required by OpenTelemetry.Instrumentation.Wcf. -->
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net8.0'" Label="Transient dependencies of OpenTelemetry.Instrumentation.Wcf">
+    <PackageReference Include="Microsoft.Win32.SystemEvents" ExcludeAssets="all" />
     <PackageReference Include="System.Drawing.Common" ExcludeAssets="all" />
-    <!-- System.Windows.Extensions is transitive dependency required by OpenTelemetry.Instrumentation.Wcf. -->
     <PackageReference Include="System.Windows.Extensions" ExcludeAssets="all" />
-    <!-- Microsoft.Win32.SystemEvents is transitive dependency required by OpenTelemetry.Instrumentation.Wcf. -->
-    <PackageReference Include="Microsoft.Win32.SystemEvents" ExcludeAssets="all" />
+  </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net8.0' OR '$([MSBuild]::GetTargetFrameworkIdentifier(`$(TargetFramework)`))' == '.NETFramework' " Label="(Transient) dependencies of OpenTelemetry.Instrumentation.StackExchangeRedis">
```
Kielek: This group can be unconditional. Historically, there were for MongoDB: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3845/files#diff-6f6eafff31edc8dbb1fc4aec15ea6334e53f139bfb7cb6d035c1de6001c800ab

src/CommonExcludedAssets.props
Kielek: I would keep comments why packages are excluded. It is easier to check if it is really true than traversing package deendencies.

src/OpenTelemetry.AutoInstrumentation/OpenTelemetry.AutoInstrumentation.csproj
Kielek: Can we add here `System.Threading.Channels`?

It is not OpenTelemetry dependency, so the application should be responsible to bring it during runtime.

test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt
```diff
@@ -90,6 +91,7 @@
   \netfx\System.Runtime.CompilerServices.Unsafe.dll,
   \netfx\System.Text.Encodings.Web.dll,
   \netfx\System.Text.Json.dll,
+  \netfx\System.Threading.Channels.dll,
```
Kielek: If there will be no possibility to remove this dependency by CommonExcludedAssets, it should be added to the LICENSE file.

src/OpenTelemetry.AutoInstrumentation.Native/OpenTelemetry.AutoInstrumentation.Native.DLL.vcxproj
martincostello: #4747

src/OpenTelemetry.AutoInstrumentation/OpenTelemetry.AutoInstrumentation.csproj
martincostello: The build didn't seem to like it being in both files, so I added it to the CommonExcludedAssets file.

src/OpenTelemetry.AutoInstrumentation.Native/netfx_assembly_redirection.h
```diff
@@ -147,6 +148,7 @@ void CorProfiler::InitNetFxAssemblyRedirectsMap()
             { L"System.Text.Json", {10, 0, 0, 0} },
             { L"System.Text.RegularExpressions", {4, 1, 1, 0} },
             { L"System.Threading", {4, 0, 11, 0} },
+            { L"System.Threading.Channels", {5, 0, 0, 0} },
```
Kielek: You need to regenerate this file. `nuke BuildTracer` should do the trick

martincostello: For whatever reason when I ran `build.cmd` before making the previous commit it didn't regenerate the file as it had previously.

Kielek: `git clean -fxd`?
Probably dotnet does not refreshed build outputs correctly.

test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt
martincostello, line 1: This passes for me locally, so there must be something dirty/out-of-date in the build. Will need to investigate.

test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt
Kielek: It should be here, as you put it in CommonExceludedAssets.
```suggestion
```

test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt
martincostello: Whatever generates the file under test was using stale inputs/wasn't being re-generated.

Had to close all IDEs and run `git clean -fXd` then `build.cmd` to get the test to fail to see what was different (and locally edit the test to get a nice UI diff).

martincostello:
The two test failures appear to be flakes.

Kielek:
> The two test failures appear to be flakes.

At least one of them in the new way. Reruning.

PR #4740, opened 2026-01-09
CoreWCF.Primitives traces instrumentation

Kielek:
## Why

Fixes #2673

## What

CoreWCF.Primitives traces instrumentation
Configuration switch is called `WCFCORE` instead of `COREWCF`. I think that it will be easier to find by the end users.

## Tests

CI + added tests.
.NET WCF Client is executed against .NET Framework and CoreWCF.

## Checklist

- [x] `CHANGELOG.md` is updated.
- [x] Documentation is updated.
- [x] New features are covered by tests.

PR #4757, opened 2026-01-13
Skip tests that require Windows admin instead of fail

martincostello:
## Why

Improves local development experience.

## What

Skip tests that require Windows admin permissions, rather than fail.

## Tests

None.

## Checklist

### DIFF
--- a/test/IntegrationTests/WindowsAdministratorFactAttribute.cs
+++ b/test/IntegrationTests/WindowsAdministratorFactAttribute.cs
@@ -11,6 +11,9 @@ public sealed class WindowsAdministratorFactAttribute : FactAttribute
     public WindowsAdministratorFactAttribute()
         : base()
     {
-        Skip = EnvironmentTools.IsWindowsAdministrator() ? null : "This test requires Windows Administrator privileges.";
+        Skip = GetSkipReason();
     }
+
+    internal static string? GetSkipReason() =>
+        !EnvironmentTools.IsWindows() || EnvironmentTools.IsWindowsAdministrator() ? null : "This test requires administrative privileges on Windows.";
 }
--- a/test/IntegrationTests/WcfTestsBase.cs
+++ b/test/IntegrationTests/WcfTestsBase.cs
@@ -62,7 +62,6 @@ public abstract class WcfTestsBase : TestHelper, IDisposable
 
     protected async Task SubmitsTracesInternal(string clientPackageVersion, WcfServerTestHelperBase wcfServerTestHelperBase)
     {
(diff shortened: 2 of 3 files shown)


martincostello:
Failures look like unrelated flakes.

test/IntegrationTests/WindowsAdministratorFactAttribute.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace IntegrationTests;
+
+[AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+public sealed class WindowsAdministratorFactAttribute : FactAttribute
+{
+    public WindowsAdministratorFactAttribute()
+        : base()
+    {
+        Skip = EnvironmentTools.IsWindowsAdministrator() ? null : "This test requires Windows Administrator privileges.";
```
Kielek: I think that you need to check 2 things now, `!IsWindows && IsWindowsAdministrator`
The same in theory. See https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4740/changes#diff-8105f2edfc8dd16975de19070c5457ef32e2f9852c69f481102f736a9d8da58fR65

martincostello: The method already checks that:

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/16b8a60619e5a381463e109563e314f0e7da17f3/test/IntegrationTests/Helpers/EnvironmentTools.cs#L71-L76

Kielek: If I understand correctly WcfDotNEtTests will be skipped on Linux/macOS machines now. It should be executed.

Edit: checked in CI:  09:42:04 [DBG]   Skipped IntegrationTests.WcfDotNetTests.SubmitTraces [1 ms]

martincostello: Ah, I see what you mean. I'll change it, but I don't think that's a new problem as the old asserts relied on the same behaviour.

Kielek: Please check current main: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/16b8a60619e5a381463e109563e314f0e7da17f3/test/IntegrationTests/WcfTestsBase.cs#L65

It was merged just after you created this PR

martincostello: That condition is now redundant as the two callers are already covered by the new fact/theory.

PR #4774, opened 2026-01-15
Add SourceLink to native Windows DLL

martincostello:
## Why

Fixes [BinSkim](https://github.com/microsoft/binskim) warning I saw while working on #4768:

```text
OpenTelemetry.AutoInstrumentation.Native.dll: warning BA2027: The PDB for 'OpenTelemetry.AutoInstrumentation.Native.dll' does not contain SourceLink information, compromising frictionless source-driven debugging and increasing latency of security response.
Enable SourceLink by configuring necessary project properties and adding a package reference for your source control provider.
See https://aka.ms/sourcelink for more information.
```

## What

Configure [SourceLink](https://learn.microsoft.com/cpp/build/reference/sourcelink) for the native Windows DLL. Based on the example implemented provided [here](https://github.com/dotnet/designs/blob/main/accepted/2020/diagnostics/source-link.md#github).

This appears to be an MSVC-specific feature, so there isn't a macOS or Linux equivalent as far as I'm aware.

## Tests

Manually ran BinSkim on a locally compiled DLL and the warning isn't present.

Example content of the generated `source_link.json` file:

```json
{"documents": { "D:\\Coding\\open-telemetry\\opentelemetry-dotnet-instrumentation\\*" : "https://raw.githubusercontent.com/martincostello/opentelemetry-dotnet-instrumentation/8274215e86868e90c181aed15e97535c33e70cb3/*" }}
```

Will also verify against the CI build once completed.

## Checklist

### DIFF
--- a/build/Build.Steps.Windows.cs
+++ b/build/Build.Steps.Windows.cs
@@ -122,13 +122,16 @@ partial class Build
         {
             foreach (var architecture in ArchitecturesForPlatform)
             {
-                var source = NativeProfilerProject.Directory / "bin" / BuildConfiguration / architecture.ToString() /
-                             $"{NativeProfilerProject.Name}.dll";
+                var sourceDir = NativeProfilerProject.Directory / "bin" / BuildConfiguration / architecture.ToString();
+                var sourceDll = sourceDir / $"{NativeProfilerProject.Name}.dll";
+                var sourcePdb = sourceDir / $"{NativeProfilerProject.Name}.pdb";
                 var dest = TracerHomeDirectory / $"win-{architecture}";
 
-                Log.Information($"Copying '{source}' to '{dest}'");
+                Log.Information($"Copying '{sourceDll}' to '{dest}'");
+                sourceDll.CopyToDirectory(dest, ExistsPolicy.FileOverwrite);
 
-                source.CopyToDirectory(dest, ExistsPolicy.FileOverwrite);
+                Log.Information($"Copying '{sourcePdb}' to '{dest}'");
+                sourcePdb.CopyToDirectory(dest, ExistsPolicy.FileOverwrite);
             }
         });
 
--- a/test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt
+++ b/test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt
@@ -368,5 +368,7 @@
   \store\x86\net9.0\microsoft.extensions.primitives\8.0.0\lib\net10.0\Microsoft.Extensions.Primitives.dll,
   \store\x86\net9.0\system.diagnostics.diagnosticsource\10.0.0\lib\net10.0\System.Diagnostics.DiagnosticSource.dll,
   \win-x64\OpenTelemetry.AutoInstrumentation.Native.dll,
-  \win-x86\OpenTelemetry.AutoInstrumentation.Native.dll
+  \win-x64\OpenTelemetry.AutoInstrumentation.Native.pdb,
+  \win-x86\OpenTelemetry.AutoInstrumentation.Native.dll,
+  \win-x86\OpenTelemetry.AutoInstrumentation.Native.pdb
 ]


martincostello:
> Will also verify against the CI build once completed.

No warning, but there's no PDB in the output to verify.

Should the PDB/symbols for the binary files be output in the artifacts too?

Kielek:
> > Will also verify against the CI build once completed.
> 
> No warning, but there's no PDB in the output to verify.
> 
> Should the PDB/symbols for the binary files be output in the artifacts too?

I am not against, one more thing to check, how to correctly build .`snupkg` file for the native part. I am not sure what are recommendation/support for this.

martincostello:
> how to correctly build .`snupkg` file for the native part

Short answer is _"no idea"_ 😄

martincostello:
Potentially one way to do it is to just include the symbols in the package itself via `OpenTelemetry.AutoInstrumentation.Runtime.Native.nuspec`. I'll read into it to see if there's an easy way to create symbol packages when there's only a `.nuspec` file.

martincostello:
From [NuGet.org symbol package constraints](https://learn.microsoft.com/nuget/create-packages/symbol-packages-snupkg):

> [!NOTE]
> Native projects, such as C++ projects, produce Windows PDBs instead of Portable PDBs. These are not supported by NuGet.org's symbol server. Please use [Legacy Symbol Packages](https://learn.microsoft.com/nuget/create-packages/symbol-packages) instead.

Will look at [Legacy Symbol Packages](https://learn.microsoft.com/nuget/create-packages/symbol-packages) next...

martincostello:
> No warning, but there's no PDB in the output to verify.

Symbols included (and found) now, and there's no warning related to SourceLink:

```text
OpenTelemetry.AutoInstrumentation.Native.dll: warning BA2024: 'OpenTelemetry.AutoInstrumentation.Native.dll' was compiled with one or more modules that do not enable code generation mitigations for speculative execution side-channel attack (Spectre) vulnerabilities.
Spectre attacks can compromise hardware-based isolation, allowing non-privileged users to retrieve potentially sensitive data from the CPU cache.
To resolve this issue, ensure that all modules compiled into the binary are compiled with /Qspectre switch on cl.exe command-line.
You may need to install the 'C++ spectre-mitigated libs' component from the Visual Studio installer if you observe violations against C runtime libraries such as libcmt.lib, libvcruntime.lib, etc.
For VC projects use SpectreMitigation property with 'Spectre' value.
When using older MSVC versions pass /d2guardspecload in cases where your compiler supports this switch and it is not possible to update to a toolset that supports /Qspectre.
This warning should be addressed for code that operates on data that crosses a trust boundary and that can affect execution, such as parsing untrusted file inputs or processing query strings of a web request.
For mitigation to be effective, all dlls and libs that are part of an exe, should be compiled with /QSpectre, so decision should be made for the whole process.
Can cause some perf degradations, so should not be used when not needed.

Done. 1 files scanned.
Analysis completed successfully.
```

martincostello:
> Will look at [Legacy Symbol Packages](https://learn.microsoft.com/nuget/create-packages/symbol-packages) next...

I can produce `.symbols.nupkg` files for `OpenTelemetry.AutoInstrumentation` and `OpenTelemetry.AutoInstrumentation.Runtime.Native` by adding the following here:

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/45a3b23a3a7e3d228997b89e922b033791138e82/build/Build.NuGet.Steps.cs#L138-L142

```diff
- .SetOutputDirectory(NuGetArtifactsDirectory));
+ .SetOutputDirectory(NuGetArtifactsDirectory)
+ .EnableSymbols());
```

Other packages still produce `.snupkg` files.

However, the contents are effectively identical because the PDBs are already included in the NuGet packages anyway:

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/45a3b23a3a7e3d228997b89e922b033791138e82/nuget/OpenTelemetry.AutoInstrumentation.Runtime.Native/OpenTelemetry.AutoInstrumentation.Runtime.Native.nuspec#L25

With a change like below, the PDBs are missing from both packages:

```diff
- <file src="**\*" target="\" />
+ <file src="**\*.dll" target="\" />
+ <file src="**\*.dylib" target="\" />
+ <file src="**\*.so" target="\" />
```

There doesn't appear to be a way to make `<file>` elements conditional.

Options seem to be:

1. Leave as-is
1. Create the `.symbols.nupkg` files anyway
1. Exclude the `.pdb` files from the NuGet package.

I would go with 1 or 2.

martincostello:
> For the same reasons zip files does not include pdb files for the managed code.

This isn't the case:

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/45a3b23a3a7e3d228997b89e922b033791138e82/test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt#L9-L16

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/45a3b23a3a7e3d228997b89e922b033791138e82/test/IntegrationTests/BuildTests.DistributionStructure_windows.verified.txt#L62

martincostello:
ZIP artifact sizes:

| **Name** | **Before** | **After** | **% Delta** |
|:--|--:|--:|--:|
| `bin-alpine-arm64` | 4.09 MB | 4.09 MB | - |
| `bin-alpine-x64` | 4.14 MB | 4.14 MB | - |
| `bin-macos-14` | 3.28 MB | 3.28 MB | - |
| `bin-ubuntu-22.04` | 4.02 MB | 4.02 MB | - |
| `bin-ubuntu-22.04-arm` | 4 MB | 4 MB | - |
| `bin-ubuntu1604-native` | 1.24 MB | 1.24 MB | - |
| `bin-windows-2022` | 8.36 MB | 17 MB | x2.0 |
| `bin-nuget-packages` | 7.5 MB | 16 MB | x2.1 |

rajkumar-rangaraj (approved):

zacharycmontoya:
I just got around to looking at this PR. By including the PDB's we're doubling the size of the artifacts. Is that required for this to work or can we separate out the artifacts as Piotr initially mentioned?

martincostello:
Whilst they are 2x, the absolute difference isn't much.

As noted here https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4774#issuecomment-3759781958, PDBs are already included in the artifacts, so you could argue including extra ones is _more_ correct.

If you want to optimise for size, then IMHO that should be done comprehensively as a separate thing (i.e. to have separate "debug" artifacts).

PR #4775, opened 2026-01-15
Add version file to artifact

martincostello:
## Why

Sometimes ZIP files from the instrumentation get copied around and shared, and it isn't immediately obvious exactly what code was used to create it as the archive names don't contain the version.

For example, modern .NET includes a `.version` file containing similar metadata. On my computer right now, there's a `C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App\10.0.2\.version` file with the following content:

```text
44525024595742ebe09023abe709df51de65009b
10.0.2
```

## What

Add a `VERSION` file containing the GitHub ref name and SHA it was built from.

If the concept is welcome, happy to iterate on the name of the file and/or contents. The general gist is to have a file can be quickly opened and read directly from a ZIP, rather than require it to be fully extracted to, for example, look at the file properties in Windows Explorer or open a .NET assembly in ILSpy to look at the attributes.

## Tests

Local testing creates a file containing the version and `local-dev@{Git SHA-1}`.

Will inspect the build artifacts from CI to check that contain the relevant information for this PR.

## Checklist

### DIFF
--- a/build/Build.Steps.cs
+++ b/build/Build.Steps.cs
@@ -510,7 +510,6 @@ partial class Build
             if (Environment.GetEnvironmentVariable("GITHUB_ACTIONS") is "true")
             {
                 refName = Environment.GetEnvironmentVariable("GITHUB_REF_NAME");
-                gitSha = Environment.GetEnvironmentVariable("GITHUB_SHA");
             }
 
             var dest = TracerHomeDirectory / "VERSION";


martincostello:
> Will inspect the build artifacts from CI to check that contain the relevant information for this PR.

`bin-macos-14.zip` contains a `VERSION` file containing `4775/merge@7c750c8f2f6f0d3f68f2590997ea80a3da02d371`.

Kielek:
I think that you need to update verified files related to IntegrationTests.BuildTests.DistributionStructure

RassK:
Shouldn't it contain product version as well for easier identification? (like the original example)

martincostello:
> Shouldn't it contain product version as well for easier identification? (like the original example)

It wasn't immediately obvious to me how to get that in the build script, but in the case of proper releases it would have the tag, so it would (e.g. `v1.13.0`).

Kielek:
> > Shouldn't it contain product version as well for easier identification? (like the original example)
> 
> It wasn't immediately obvious to me how to get that in the build script, but in the case of proper releases it would have the tag, so it would (e.g. `v1.13.0`).

There is VersionHelper class in the build project if you need it.

martincostello:
With latest changes:

## Local

```text
1.13.1-alpha.0.113
local-dev@40a2da2e90ec69d71783492bf4416408aa1b50b5
```

## CI

```text
1.13.1-alpha.0.110
4775/merge@7bf22460d18fd421c8ae32574e13b1592f1d5820
```

build/Build.Steps.cs
```diff
@@ -497,6 +497,26 @@ bool FilesAreEqual(string filePath1, string filePath2)
...
+        .Executes(() =>
+        {
+            var version = VersionHelper.GetVersion();
+            var refName = "local-dev";
+            var gitSha = VersionHelper.GetCommitId();
+
+            if (Environment.GetEnvironmentVariable("GITHUB_ACTIONS") is "true")
+            {
+                refName = Environment.GetEnvironmentVariable("GITHUB_REF_NAME");
+                gitSha = Environment.GetEnvironmentVariable("GITHUB_SHA");
```
Kielek: I think that this is always redundant
```suggestion
```

martincostello: I figured it would always definitely defer to the authoritative SHA in CI in case for whatever reason it didn't get set in the build, but it can be removed if you want.

Kielek: I think that it is better to take it from MinVer. This is included in all managed libraries.
If there is no SHA, you can always throw here InvalidOperationException or something like this.

PR #4783, opened 2026-01-16
Revise assembly version conflict resolution on .NET to eliminate DOTNET_ADDITIONAL_DEPS/DOTNET_SHARED_STORE requirement

alexeypukhov:
## Why

OpenTelemetry.AutoInstrumentation for .NET often requires setting `DOTNET_ADDITIONAL_DEPS` and `DOTNET_SHARED_STORE` to resolve assembly version conflicts between the instrumented application and the instrumentation tool (see #4715). A more robust, unified approach is needed across all deployment modes.

## What

Leverage .Net `AssemblyLoadContext` to handle assembly conflict resolution across different deployments and reuse existing logic implemented for .NET Framework

**See [Assembly Conflict Resolution](https://github.com/alexeypukhov/opentelemetry-dotnet-instrumentation/blob/8e67ff08e9bdb6b66cd5eb769b8d5d63c91a65dd/docs/assembly-conflict-resolution.md)** for:
- How the .NET runtime's assembly resolution pipeline works (both .NET and .NET Framework)
- The three deployment strategies and how each leverages runtime mechanics to solve conflicts
- Known limitations and their customer impact

The strategies below are grounded in these runtime mechanics.

### Strategies

1. **NuGet package deployment** - Build-time resolution (no changes)
2. **Native Profiler deployment** - Assembly redirection + custom ALC for conflicts on .NET, no changes for .NET Framework
3. **StartupHook-only deployment** (.NET only) - Full isolation in custom ALC

### Implementation Highlights

- Native assembly redirection previously implemented for .NET Framework now extended to .NET
- Native assembly redirection on .NET narrows assembly resolution to just two cases where .NET runtime needs help: 1) non-TPA assemblies (our assemblies) - can be loaded to Default ALC and 2) TPA version conflicts - should be isolated into custom ALC
- StartupHook-only mode achieves best-effort isolation by running the entire customer app in a custom ALC

### Essential Changes

1. Extended and renamed the multi-target project (`OpenTelemetry.AutoInstrumentation.Assemblies.NetFramework.csproj`) to support .NET targets and added .NET-specific project references (Loader, AspNetCoreBootstrapper, StartupHook).
	- Enables `build/TransientDependenciesGenerator.cs` to process both .NET Framework and .NET dependencies, contributing .NET dependencies to `src/OpenTelemetry.AutoInstrumentation.Assemblies/Directory.Packages.props`.
[...]

PR #4799, opened 2026-01-21
[FileBasedConfiguration] Support for Netfx Redirection

ysolomchenko:
## Why

Towards #4541

## What

Added support for file-based configuration of the .NET Framework Assembly Redirect setting

## Tests

Ci

## Checklist

### DIFF
--- a/test/IntegrationTests/AssemblyRedirectionOnNetFrameworkTests.cs
+++ b/test/IntegrationTests/AssemblyRedirectionOnNetFrameworkTests.cs
@@ -34,7 +34,7 @@ public class AssemblyRedirectionOnNetFrameworkTests : TestHelper
 
         const string TestApplicationActivitySource = "AssemblyRedirection.NetFramework.ActivitySource";
         SetEnvironmentVariable("OTEL_DOTNET_AUTO_TRACES_ADDITIONAL_SOURCES", TestApplicationActivitySource);
-        SetEnvironmentVariable("OTEL_DOTNET_AUTO_NETFX", isNetFxRedirectEnabled.ToString());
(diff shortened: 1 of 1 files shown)


src/OpenTelemetry.AutoInstrumentation/NativeMethods.cs
```diff
@@ -57,6 +57,18 @@ public static void SetSqlClientNetFxILRewriteEnabled(bool enabled)
             NonWindows.SetSqlClientNetFxILRewriteEnabled(enabled);
         }
     }
+
+    public static void SetNetFxAssemblyRedirectionEnabled(bool enabled)
```
iskiselev, line 61: It may be to late to chenge this settings here - someassemblies were loaded already, .NET already done desicions whether load assembly in shared AppDomain or in separate.

iskiselev: the same setting would support assembly redirection on .net if #4783 would be merged (probably renamed), and it would be absolutly critical to do redirection much earlier, as we need to redirect System.Diagnostics.DiagnosticSource.

iskiselev:
I'm not sure whether that setting should be controlled by file based config at all.
It is more a fixed value based on the deployment model - we need it for zip depoyment and don't need it for nuget deployments.

If we disable it in zip deployment, most probably nothing would work. Even if it needs to be overriden by end user (which I doubd)  for some RCA if they have issues, we still need to weight if it should be done from config.

If we are not happy with environment variable for the setting, that we set from startup scrupt, we can implement in profiler some logic that would validate whether it is deployed as zip or nuget. We stll may have some safeguard to override it, but I think we should not require users to set it with file-based config by default.

PR #4855, opened 2026-02-17
Avoid invalid traces when EF Core and Npgsql instrumentations run together

### LINKED ISSUE #4252: Invalid trace when Npsql and EFCore instrumentation are enabled
See details in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/2792

This issue is here in case we need follow up work.

aviralgarg05:
## Why

Fixes #4252 – prevents invalid trace recursion when Entity Framework Core and Npgsql trace instrumentations are enabled together by removing the conflicting Npgsql instrumentation automatically.

## What

- Added `FilterConflictingInstrumentations` inside `TracerSettings` so that whenever EF Core tracing is enabled the Npgsql trace instrumentation is filtered out.
- Hooked the filter into both the environment-variable and file-based configuration paths and emitted a warning to explain the automatic disabling.

## Tests

- `dotnet test test/OpenTelemetry.AutoInstrumentation.Tests/OpenTelemetry.AutoInstrumentation.Tests.csproj -c Release` *(fails because `dotnet` is not available in this environment; please rerun locally once the SDK is installed.)*

## Checklist


nrcventura:
This approach, even though it works, has some tradeoff that may not be worth it. It seems to work by filtering the "enabled" instrumentation list. However, if someone is just using the default settings, and not using file based configuration to only opt-into the instrumentation that they care about, they will only get postgresql instrumentation if they are using EF Core to access postgres. The failing integration tests reflect that behavior.

Ideally I think we need to detect if an application is using both the EF Core and Posgres libraries and then disable one of them. Doing that is more complicated than just filtering things at configuration time, especially when there are no guaranties that both of the DB libraries will be loaded into memory at configuration time.

Maybe there's something that can be done via byte code injection into the method(s) used to register the postgres library with EF Core, and when that call runs we can update unsubscribe from the postgres ActivitySource.

nrcventura:
Can an integration test be added to help with preventing any regressions in this fix?

PR #4870, opened 2026-02-25
NoCode - support for dynamic attributes/span names/statuses

Kielek:
## Why

Internal request to add possibility to evaluate attributes, span names and statuses into no code functionality.

## What

NoCode - support for dynamic attributes/span names/statuses
Details described under updated documentation

## Notes

Should be merged after #4875.

Implementation fully relays on local  implementation of https://cel.dev/.
I was considering including external libraries:
- Cel.Net package brings a lot of indirect dependencies. It does not work for us.
- Roslyn, seems to be to heavy,, but potentially can execute any C# code.
- DynamicExpresso - brings dependency for Microsoft.CSharp, I doubt that it can be included.
- Flee - looks like abandoned project.

## Tests

CI + Extended

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegrationHelper.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegrationHelper.cs
@@ -78,7 +78,7 @@ internal static class NoCodeIntegrationHelper
 
                 if (!string.IsNullOrEmpty(dynamicName))
                 {
-                    spanName = dynamicName;
+                    spanName = dynamicName!;
                 }
                 else
                 {


docs/nocode-instrumentation.md
```diff
@@ -88,6 +91,273 @@ Supported attribute types and their formats:
...
+span:
+  name: DefaultRequest                                   # Fallback name
+  name_source: concat($arg1.HttpMethod, " ", $arg1.Path) # e.g., "GET /api/users"
+```
+
+### Status Configuration
+
+You can configure span status based on return values or other conditions using
+status rules. Rules are evaluated in order, and the first matching rule sets
+the status.
```
iskiselev: It may be good extension to support condition with attribute also - with first success for attribute defines it.

Kielek: Maybe, as a follow up PR. This one is big enough. What is more, if the value evaluate to null, it is skipped.

docs/nocode-instrumentation.md
```diff
@@ -33,6 +33,9 @@ collection for specified methods through YAML configuration.
 - **Span Customization**: Configure span names, kinds, and custom attributes
 - **Method Overload Support**: Target specific method overloads using precise
   signature matching
+- **Dynamic Expressions**: Use [CEL](https://cel.dev/) (Common Expression
```
nrcventura: If we are not fully compliant with CEL, maybe we should refer to it as a domain specific language (DSL) based on a subset of CEL.

docs/nocode-instrumentation.md
```diff
@@ -88,6 +94,389 @@ Supported attribute types and their formats:
...
+| `method`    | Method name                                | `method`               |
+| `type`      | Declaring type name                        | `type`                 |
+
+**Member Access** - Access properties and array elements:
+
+| Expression                     | Description                 |
+|--------------------------------|-----------------------------|
+| `arguments[0]`                 | First method argument       |
+| `arguments[1]`                 | Second method argument      |
+| `arguments[0].PropertyName`    | Property of first argument  |
```
nrcventura, line 125: Is there an initial need to support fields too?

docs/nocode-instrumentation.md
```diff
@@ -88,6 +94,389 @@ Supported attribute types and their formats:
...
+| `<=`, `>=` | Comparison          | `return.Score >= 50`                     |
+| `&&`       | Logical AND         | `arguments[0] != null && return.Success` |
+| `\|\|`     | Logical OR          | `return.Success \|\| return.Retryable`   |
+| `!`        | Logical NOT         | `!return.HasError`                       |
+| `? :`      | Ternary conditional | `return.Success ? "ok" : "failed"`       |
+
+> [!NOTE]  
+>
+> - Arguments are zero-indexed (`arguments[0]` to `arguments[8]`)
+> - Property access uses reflection and only works with public properties
```
nrcventura, line 146: I think restricting this to `public` properties is fine for now (and possibly for a long time), but we may need to consider expanding it to `internal` too. It just depends on which use cases we want to satisfy with this type of instrumentation.

For example, are we trying to help customers with writing custom nocode instrumentation? Or are we trying to provide a way for vendors to have custom sets of nocode instrumentation included with their distros?

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeExpression.cs
```diff
@@ -0,0 +1,195 @@
...
+using System.Text.RegularExpressions;
+using OpenTelemetry.AutoInstrumentation.Logging;
+
+namespace OpenTelemetry.AutoInstrumentation.Instrumentations.NoCode;
+
+/// <summary>
+/// Represents a parsed expression for extracting values from method context.
+/// Supports expressions like: $arg1, $arg2.Property, $instance.Property.Nested, $return, $method, $type
+/// </summary>
+internal sealed class NoCodeExpression
```
nrcventura: The code in this class seems to have a lot of overlap with the code in CelExpression. Is there a way to abstract things so that the logic does not need to be repeated?

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeFunctionExpression.cs
```diff
@@ -0,0 +1,373 @@
...
+using System.Text.RegularExpressions;
+using OpenTelemetry.AutoInstrumentation.Logging;
+
+namespace OpenTelemetry.AutoInstrumentation.Instrumentations.NoCode;
+
+/// <summary>
+/// Represents a function expression that can combine multiple expressions.
+/// Supports: concat(...), coalesce(...), substring(...), tostring(...)
+/// </summary>
+internal sealed class NoCodeFunctionExpression
```
nrcventura: This class seems to overlap with code in CelFunctionCallNode. Is there a way to get the classes to use the same logic?

nrcventura:
It seems like CEL is nice choice for basing the DSL that's necessary to evaluate these dynamic pieces of information. I think there is room to improve the performance of the implementation, but I don't think those improvements are necessary for the initial implementation.

docs/nocode-instrumentation.md
```diff
@@ -88,6 +94,389 @@ Supported attribute types and their formats:
...
+CEL functions are divided into two categories:
+
+- **[Standard]** - Part of the CEL specification, portable to other CEL implementations
+- **[Extension]** - Custom functions added for instrumentation scenarios
+
+#### Supported Functions
+
+| Function                          | Type      | Description                              | Example                                    |
+|-----------------------------------|-----------|------------------------------------------|--------------------------------------------|
+| `concat(...)`                     | Extension | Concatenate values into a string         | `concat(type, ".", method)`                |
```
iskiselev: CEL allows `+` on strings, so we probably don't need concat

docs/nocode-instrumentation.md
```diff
@@ -88,6 +94,389 @@ Supported attribute types and their formats:
...
+| `coalesce(...)`                   | Extension | Return first non-null value              | `coalesce(arguments[0].Name, "unknown")`   |
```
iskiselev: I suggest to start without non-standard `coalesce` function.
Users can use ternary operator to do it.
In future we may want to implement `optional` support - https://github.com/google/cel-spec/wiki/proposal-246  and `filter`
With it, stadard way to get first non-null argument or string "none" will be like:
`arguments.filter(x, x != null)[?0].orValue("none")`
or `[arguments[0], arguments[1], dyn("none")].filter(x, x != null)[?0].orValue("none")`
Or if we work with complex sub-expressions:
`args[?0].prop_in_first.or(args[?1].prop_in_second).orValue("none")`

It still may be beneficial to implement `coalesce`, but better to do it later - it is always easier to give more functions than remove something.

docs/nocode-instrumentation.md
Kielek: #4895

docs/nocode-instrumentation.md
Kielek: Concat and coalesce was removed during rebasing on #4875.

docs/nocode-instrumentation.md
Kielek: Concat and coalesce was removed during rebasing on #4875.

docs/nocode-instrumentation.md
Kielek: From my perspective, it is the last chance for the instrumentation and typically, it should not be used.

There are better options:
 - creating and listening activities, if you are able to modify code
 - writing bytecode instrumentation (for now, we do not have public extension point for this (except full recompilation). I hope that vendors will contribute here rather than writing some custom stuff in fortks.
 
The only good reason, is to instrument so rarely used, library without recompilation possibilities.

Request tracked under #4896

docs/nocode-instrumentation.md
Kielek: I would defer it to separate PR. For me, properties cover MVP requirements.

Tracked under #4897

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeExpression.cs
Kielek: Removed in 6599b1bdba0601ecb9487aaf426aa5b8e2181636. It was not used outside tests.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeFunctionExpression.cs
Kielek: Removed in 6599b1bdba0601ecb9487aaf426aa5b8e2181636. It was not used outside tests.

docs/nocode-instrumentation.md
Kielek: Wording updated in c2fc585b08a9457d374ae88bf44181a1f496ee43.
I hope it makes it clear.

Kielek:
@nrcventura, @iskiselev, I have dropped substring functionality, for now. CEL proposes better support for this by python like slicing mechanism:

#### String Slicing (Standard CEL Syntax)

| Syntax        | Description                           | Example                     |
|---------------|---------------------------------------|-----------------------------|
| `str[n:m]`    | Extract substring from index n to m   | `arguments[0][0:10]`        |
| `str[n:]`     | Extract from index n to end           | `arguments[0][5:]`          |
| `str[:m]`     | Extract from start to index m         | `arguments[0][:10]`         |

docs/nocode-instrumentation.md
```diff
@@ -65,8 +68,11 @@ instrumentation/development:
             kind: internal                                    # (Optional) Span kind: internal, server, client, producer, consumer (defaults to internal)
             attributes:                                       # Array of custom attributes to add to the span
               - name: custom.attribute                        # Attribute name
-                value: "attribute_value"                      # Attribute value
+                value: "attribute_value"                      # Static attribute value
                 type: string                                  # Attribute type (see Attribute Types section below)
+              - name: dynamic.attribute                       # Dynamic attribute name
```
iskiselev, line 73: Do we need to split attibutes for static / dynamic? Probably, otherwise it will be require users to define attributes as `'"attribute_value"'` - to make it proper CEL string, which is not backward compatible. We also pay some price for every dynamic attribute to parse it, so looks like descion to split them is proper.

docs/nocode-instrumentation.md
```diff
@@ -88,6 +94,365 @@ Supported attribute types and their formats:
...
+**Member Access** - Access properties and array elements:
+
+| Expression                     | Description                 |
+|--------------------------------|-----------------------------|
+| `arguments[0]`                 | First method argument       |
+| `arguments[1]`                 | Second method argument      |
+| `arguments[0].PropertyName`    | Property of first argument  |
+| `arguments[0].Nested.Property` | Nested property access      |
+| `instance.PropertyName`        | Property of instance object |
+| `return.Result`                | Property of return value    |
```
iskiselev: Let's change example. When I see `result.Result` I start thinking about `Task` - and for them such usage is route to dead-lock and thread starvation.

```suggestion
| `return.ResultProperty`        | Property of return value    |
```

iskiselev: BTW, for async methods, are we making result callback when Task created or when Task was complete? What about non-aync Task-returning methods?

docs/nocode-instrumentation.md
```diff
@@ -88,6 +94,365 @@ Supported attribute types and their formats:
...
+| `contains(str, substring)` | Check if string contains substring       | `contains(arguments[0].Message, "error")`  |
+
+#### Function Expression Examples
+
+**Concatenate values:**
+
+```yaml
+attributes:
+  - name: operation.id
+    source: type + "." + method)
```
iskiselev:
```suggestion
    source: type + "." + method
```

docs/nocode-instrumentation.md
```diff
@@ -88,6 +94,365 @@ Supported attribute types and their formats:
...
+    type: string
+  - name: order.key
+    source: arguments[0].CustomerId + "-" + arguments[0].OrderId)
```
iskiselev:
```suggestion
    source: arguments[0].CustomerId + "-" + arguments[0].OrderId
```

docs/nocode-instrumentation.md
iskiselev: I see, we do not split them now into static/dynamic.
We need to fix sample, as in yaml sting started with quote is quoted string, so quote will not be part of sting itself.
So, we need to use either: `'"attribute_value"'` or `"\"attribute_value\""`

docs/nocode-instrumentation.md
```diff
@@ -88,6 +94,365 @@ Supported attribute types and their formats:
...
+> - This DSL implements a subset of CEL; not all CEL features are supported
+
+#### Dynamic Attribute Examples
+
+Extract method argument values:
+
+```yaml
+attributes:
+  - name: order.id
+    source: arguments[0]              # First argument value
```
iskiselev: Are we using `source:` (as here) or `value:` (as at the beginning)?

docs/nocode-instrumentation.md
```diff
@@ -88,6 +94,365 @@ Supported attribute types and their formats:
...
+  name: DefaultOperation                                 # Fallback name
+  name_source: method + "-" + arguments[0].OperationType # e.g., "ProcessOrder-Express"
+```
+
+Use with nested properties:
+
+```yaml
+span:
+  name: DefaultRequest                                            # Fallback name
+  name_source: arguments[0].HttpMethod + " " + arguments[0].Path) # e.g., "GET /api/users"
```
iskiselev:
```suggestion
  name_source: arguments[0].HttpMethod + " " + arguments[0].Path # e.g., "GET /api/users"
```

docs/nocode-instrumentation.md
```diff
@@ -88,6 +94,365 @@ Supported attribute types and their formats:
...
+```
+
+Use conditional expressions:
+
+```yaml
+span:
+  name: DefaultOrder                                     # Fallback name
+  name_source: arguments[0].Amount > 1000 ? ("LargeOrder-" + arguments[0].Id) : ("Order-" + arguments[0].Id)
```
iskiselev: Bas example, as includes `.Id` which most probaly violate low-cardinality

```suggestion
  name_source: arguments[0].Amount > 1000 ? "LargeOrder" : "Order"
```

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/NoCodeSpan.cs
```diff
@@ -2,28 +2,145 @@
...
+
+            // Parse using CEL expression
+            var celExpression = CelExpression.Parse(attribute.Source);
+            if (celExpression == null)
+            {
+                Log.Debug("Failed to parse dynamic attribute expression '{0}' for attribute '{1}'. Skipping.", attribute.Source, attribute.Name);
+                continue;
+            }
+
+            dynamicAttributes.Add(new NoCodeDynamicAttribute(attribute.Name!, celExpression, attribute.Type));
```
iskiselev, line 79: Will we support fallback to `value` in attribute same as we have in span `name`? If we make it similar, we may rename `source` to `value_source` - as with `name` and `name_source`.

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/NoCodeSpan.cs
```diff
@@ -2,28 +2,145 @@
...
+            return [];
+        }
+
+        var statusRules = new List<NoCodeStatusRule>();
+
+        foreach (var rule in Status.Rules)
+        {
+            if (string.IsNullOrEmpty(rule.Condition))
+            {
+                Log.Debug("Status rule has no condition. Skipping.");
```
iskiselev, line 101: shall it be higher importance message? it is clearly incorrect configuration.

docs/nocode-instrumentation.md
```diff
@@ -88,6 +94,365 @@ Supported attribute types and their formats:
...
+> a string type.
+
+#### Dynamic Span Name Examples
+
+Create span names from argument values:
+
+```yaml
+span:
+  name: DefaultTransaction                   # Fallback name
+  name_source: "Transaction-" + arguments[0] # Dynamic name using first argument
```
iskiselev: yaml string can't start with `"`. You need use single or double quoted string.

docs/nocode-instrumentation.md
```diff
@@ -88,6 +94,365 @@ Supported attribute types and their formats:
...
+```
+
+Combine multiple values:
+
+```yaml
+span:
+  name: DefaultQuery                                        # Fallback name
+  name_source: "Query." + arguments[0] + "." + arguments[1] # e.g., "Query.ProductionDB.users"
```
iskiselev: yaml string can't start with ". You need use single or double quoted string.

docs/nocode-instrumentation.md
```diff
@@ -88,6 +94,365 @@ Supported attribute types and their formats:
...
+                source: arguments[0].OrderId
+                type: string
+              - name: customer.id
+                source: arguments[0].CustomerId
+                type: string
+              - name: order.total
+                source: arguments[0].TotalAmount
+                type: double
+              - name: operation.name
+                source: "ProcessOrder-" + arguments[0].OrderType
```
iskiselev: yaml string can't start with ". You need use single or double quoted string.

test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/FileBased/glogal.json
```diff
@@ -0,0 +1,6 @@
+{
```
iskiselev: File name probably should be `global.json`

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegrationHelper.cs
```diff
@@ -104,16 +162,109 @@ internal static CallTargetReturn OnMethodEnd(Exception? exception, in CallTarget
...
+        }
+
+        var activity = Source.StartActivity(name: spanName, kind: noCodeEntry.ActivityKind, tags: tags);
+
+        // Store state for OnMethodEnd (needed for status rules evaluation)
+        var noCodeState = new NoCodeCallTargetState(noCodeEntry, instance, arguments, methodName, typeName);
+        return new CallTargetState(activity, noCodeState);
+    }
+
+    private static void HandleActivity<TReturn>(TReturn? returnValue, Exception? exception, Activity activity, NoCodeCallTargetState? noCodeState)
```
iskiselev: OnMethodBegin, StartActivity and this has very similar method body. Can we have some helper for shared logic?

test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/FileBased/glogal.json
Kielek: 7bf3f8d67a624e684104ebb2417b1538ec7ee770

docs/nocode-instrumentation.md
Kielek: d5bd1baa05ed6adc598b2f05c82286192693b166

docs/nocode-instrumentation.md
Kielek: d5bd1baa05ed6adc598b2f05c82286192693b166

docs/nocode-instrumentation.md
Kielek: d5bd1baa05ed6adc598b2f05c82286192693b166

docs/nocode-instrumentation.md
Kielek: d5bd1baa05ed6adc598b2f05c82286192693b166

docs/nocode-instrumentation.md
Kielek: d5bd1baa05ed6adc598b2f05c82286192693b166

docs/nocode-instrumentation.md
Kielek: d8269965ff236970e53d9cae840675025b124f4a

docs/nocode-instrumentation.md
Kielek: d8269965ff236970e53d9cae840675025b124f4a

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/NoCodeSpan.cs
Kielek: It is consistently used in the FileBasedConfiguration folder:
image

Even for existing code. We can consider changing it as a follow up, if you think that is is worth.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegrationHelper.cs
Kielek: StartActivity - fully dropped c90f107d5628231c9645d4cddff756e099a0d9f9
HandleActivity renamted to StopActivity: b1e4ef81b0aa023b39333d3d6cf77ef6b4ad0307

I do not see there any repetitive part except creating NoCodeContext by constructor.

test/test-applications/integrations/TestApplication.NoCode/config-malformed.yaml
```diff
@@ -0,0 +1,186 @@
+file_format: "1.0-rc.1"
```
ysolomchenko:
```suggestion
file_format: "1.0"
```

Kielek: 7c0d16e1f6e5aa788f2dc957faed005acc131f78

src/OpenTelemetry.AutoInstrumentation/Configurations/FileBasedConfiguration/NoCodeSpan.cs
Kielek: For now, there is no such fallback. It can be fully achieved by ternary operator.
Named fixed in 8bcba47bd8dc91c4ec5bbec3b0eca568c7c87e84

docs/nocode-instrumentation.md
Kielek: With some help of our common friend, I was able to prepare this chart. It is the same what I know about this.
What is more, I have added additional tests: 919284fcfbca2a2a8cbca4c9c2472236b04cdc20

```
┌─────────────────────────────────────────────────────────────────────────────┐
│                    METHOD EXECUTION TIMELINE SCENARIOS                       │
└─────────────────────────────────────────────────────────────────────────────┘

═══════════════════════════════════════════════════════════════════════════════
1️⃣  SYNCHRONOUS METHOD (int, string, void, custom class)
═══════════════════════════════════════════════════════════════════════════════
   public int GetValue() { return 42; }

   ┌──────────────┐   ┌──────────────┐   ┌──────────────┐
   │OnMethodBegin │   │Execute Method│   │ OnMethodEnd  │
   │              │   │              │   │              │
   │• Start       │   │• Run code    │   │• Evaluate    │
   │  Activity    │──▶│• Return 42   │──▶│  status rules│
   │• Eval dynamic│   │              │   │• Stop Activity│
   │  attributes  │   │              │   │  IMMEDIATELY │
   └──────────────┘   └──────────────┘   └──────────────┘
        ▲                                      │
        └──────────────────────────────────────┘
         Activity Lifecycle: SYNCHRONOUS (immediate stop)

═══════════════════════════════════════════════════════════════════════════════
2️⃣  TRUE ASYNC METHOD (async Task, async Task<T>)
═══════════════════════════════════════════════════════════════════════════════
   public async Task ProcessAsync() { await Task.Delay(100); }

   ┌──────────────┐   ┌──────────────┐   ┌────────────┐   ┌───────────────┐
   │OnMethodBegin │   │Execute Method│   │OnMethodEnd │   │OnAsyncMethodEnd│
   │              │   │              │   │            │   │               │
   │• Start       │   │• Start async │   │• Detect    │   │• Task complete│
   │  Activity    │──▶│  operation   │──▶│  Task type │   │• Eval status  │
   │• Eval dynamic│   │• Return Task │   │• Return    │   │  rules with   │
   │  attributes  │   │  IMMEDIATELY │   │  early     │   │  $return      │
   └──────────────┘   └──────────────┘   │• NO STOP   │   │• Stop Activity│
        ▲                  │              └────────────┘   └───────────────┘
        │                  │                   │                    ▲
        │                  └───────────────────┘                    │
        │                        (Task running in background)       │
        │                                                           │
        │                  ⏳ ... 100ms delay ...                  │
        │                                                           │
        └───────────────────────────────────────────────────────────┘
         Activity Lifecycle: ASYNC (stops when Task completes)

═══════════════════════════════════════════════════════════════════════════════
3️⃣  NON-ASYNC TASK - ALREADY COMPLETED (Task.FromResult, Task.CompletedTask)
═══════════════════════════════════════════════════════════════════════════════
   public Task<int> GetCached() { return Task.FromResult(42); }

   ┌──────────────┐   ┌──────────────┐   ┌────────────────┐   ┌──────────────┐
   │OnMethodBegin │   │Execute Method│   │  OnMethodEnd   │   │OnAsyncMethodEnd│
   │              │   │              │   │                │   │              │
   │• Start       │   │• Create      │   │• Detect Task   │   │• Called      │
   │  Activity    │──▶│  completed   │──▶│• Check status  │──▶│  SYNCHRONOUSLY│
   │• Eval dynamic│   │  Task        │   │• Status =      │   │• Eval status │
   │  attributes  │   │• Return      │   │  RanToCompletion│   │  rules       │
   └──────────────┘   └──────────────┘   │• Call          │   │• Stop Activity│
        ▲                                 │  continuation  │   └──────────────┘
        │                                 │  SYNC          │          │
        └─────────────────────────────────┴──────────────────────────┘
         Activity Lifecycle: EFFECTIVELY SYNCHRONOUS
         (no async overhead, continuation runs immediately)

   📌 Key Code (TaskContinuationGenerator.cs:43-46):
      if (previousTask.Status == TaskStatus.RanToCompletion) {
          _continuation(instance, default, null, in state);  // ← SYNC call
          return returnValue;
      }

═══════════════════════════════════════════════════════════════════════════════
4️⃣  NON-ASYNC TASK - PENDING (Task.Delay, Task.Run without await)
═══════════════════════════════════════════════════════════════════════════════
   public Task ProcessBackground() { return Task.Delay(100); }

   ┌──────────────┐   ┌──────────────┐   ┌────────────────┐   ┌───────────────┐
   │OnMethodBegin │   │Execute Method│   │  OnMethodEnd   │   │OnAsyncMethodEnd│
   │              │   │              │   │                │   │               │
   │• Start       │   │• Create      │   │• Detect Task   │   │• Task complete│
   │  Activity    │──▶│  pending     │──▶│• Check status  │   │• Eval status  │
   │• Eval dynamic│   │  Task        │   │• Status ≠      │   │  rules with   │
   │  attributes  │   │• Return      │   │  RanToCompletion│   │  $return      │
   └──────────────┘   └──────────────┘   │• Attach async  │   │• Stop Activity│
        ▲                  │              │  continuation  │   └───────────────┘
        │                  │              └────────────────┘          ▲
        │                  └───────────────────┐                      │
        │                                      │                      │
        │                  ⏳ ... 100ms delay ...                    │
        │                                                             │
        └─────────────────────────────────────────────────────────────┘
         Activity Lifecycle: ASYNC (stops when Task completes later)

   📌 Key Code (TaskContinuationGenerator.cs:48):
      return ToTReturn(ContinuationAction(previousTask, instance, state));
                       └─ Async method waits for Task completion

═══════════════════════════════════════════════════════════════════════════════
```

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeStatusRule.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// Evaluates the condition against the provided context.
+    /// </summary>
+    /// <param name="context">The method execution context.</param>
+    /// <returns>True if the condition is met, false otherwise.</returns>
+    public bool EvaluateCondition(NoCodeExpressionContext context)
+    {
+        var result = Condition.Evaluate(context);
+
+        // Convert result to boolean
+        return result switch
```
iskiselev: Should we use `CelHelpers.IsTrue`? It will not do auto-parse of `false`, but it will work similar with inner engine.

Kielek: I have had prepared this change, and never committed.
I see benefits from both sides, but after reconsideration, common behavior might be better.

I have added also description with truthy value:
97518a1d07d74efaa123751b9401f4a86e614757

docs/nocode-instrumentation.md
```diff
@@ -88,6 +94,413 @@ Supported attribute types and their formats:
...
+In boolean contexts (such as conditions in status rules, ternary operators, and
+logical operators), values are evaluated for truthiness according to their type.
+
+| Type          | Truthy Condition            | Examples                                             |
+|---------------|-----------------------------|------------------------------------------------------|
+| `bool`        | The value itself            | `true` is truthy, `false` is falsy                   |
+| `string`      | Non-null and non-empty      | `"text"` is truthy, `""` is falsy                    |
+| `int`         | Non-zero                    | `1` is truthy, `0` is falsy                          |
+| `long`        | Non-zero                    | `100L` is truthy, `0L` is falsy                      |
+| `double`      | Absolute value > 0          | `1.23` is truthy, `0.0` is falsy                     |
```
RassK, line 164: Any reason for using different terms for the same thing? `Non-zero` vs `Absolute value > 0`, it's confusing if there are any differences compared to `int` and `long`.

Kielek: The response is double.Epislon and the comparison way in the code

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegrationHelper.cs
```diff
@@ -30,9 +31,66 @@ internal static CallTargetState OnMethodBegin()
...
+                    instance: instance,
+                    arguments: arguments,
+                    returnValue: null,
+                    methodName: noCodeEntry.Definition.TargetMethod,
+                    typeName: noCodeEntry.Definition.TargetType);
+
+                var dynamicNameValue = noCodeEntry.DynamicSpanName.Evaluate(context);
+                if (dynamicNameValue != null)
+                {
+                    spanName = dynamicNameValue.ToString() ?? noCodeEntry.SpanName;
```
RassK: `??` expression also fallbacks to static span name but this location does not debug log.

docs/nocode-instrumentation.md
```diff
@@ -88,6 +94,413 @@ Supported attribute types and their formats:
...
+
+#### Supported Functions
+
+| Function                   | Description                              | Example                                    |
+|----------------------------|------------------------------------------|--------------------------------------------|
+| `string(value)`            | Convert value to string                  | `string(arguments[0].Id)`                  |
+| `size(value)`              | Get length/count of string, list, or map | `size(arguments[0].Items)`                 |
+| `startsWith(str, prefix)`  | Check if string starts with prefix       | `startsWith(arguments[0].Path, "/api/")`   |
+| `endsWith(str, suffix)`    | Check if string ends with suffix         | `endsWith(arguments[0].FileName, ".json")` |
+| `contains(str, substring)` | Check if string contains substring       | `contains(arguments[0].Message, "error")`  |
```
lachmatt, line 265: Would it make sense to document details related to string comparison, e.g. case-sensitivity?

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegrationHelper.cs
Kielek: 8170d45debf0696a39058fc830ae47fd4d7aac85

docs/nocode-instrumentation.md
Kielek: 57e015bcad2494b5ab6e99b3b4fd861b9cd60d58

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/NoCodeIntegrationHelper.cs
```diff
@@ -30,9 +31,78 @@ internal static CallTargetState OnMethodBegin()
...
+                    }
+                    else
+                    {
+                        Log.Debug("Dynamic span name ToString() returned null, using static name '{0}'", noCodeEntry.SpanName);
+                    }
+                }
+                else
+                {
+                    Log.Debug("Dynamic span name evaluation returned null, using static name '{0}'", noCodeEntry.SpanName);
+                }
```
RassK:
```suggestion
                var dynamicName = noCodeEntry.DynamicSpanName.Evaluate(context)?.ToString();
                
                if(!string.IsNullOrEmpty(dynamicName))
                {
                    spanName = dynamicName;
                }
                else
                {
                    Log.Debug("Dynamic span name evaluation returned null, using static name '{0}'", noCodeEntry.SpanName);
                }
```

RassK: do you need to assign `spanName` also or it picks the static somewhere else?

Kielek: aa2178b969c414d9cb169fdb235725c40c50b810

nrcventura (approved):

PR #4875, opened 2026-02-27
NoCode - dynamic attributes - CEL

Kielek:
## Why

Utils extracted for #4870 to make PR easier to review.

## What

Implementation of https://cel.dev/
Publicly available implementation https://www.nuget.org/packages/Cel.NET brings a lot of redundant dependencies.

Implemetation supported by Copilot together with Claude Sonet 4.5 mdoel.

## Tests

CI.
Test coverage is around 95%. for this code.

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/Nodes/CelBinaryOperatorNode.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/Nodes/CelBinaryOperatorNode.cs
@@ -130,6 +130,11 @@ internal sealed class CelBinaryOperatorNode : CelNode
             return null;
         }
 
+        if (left is string leftStr && right is string rightStr)
+        {
+            return leftStr + rightStr;
+        }
+
         if (IsNumeric(left) && IsNumeric(right))
         {
             var leftNum = Convert.ToDouble(left, CultureInfo.InvariantCulture);
--- a/test/OpenTelemetry.AutoInstrumentation.Tests/Instrumentations/NoCode/Cel/CelExpressionTests.cs
+++ b/test/OpenTelemetry.AutoInstrumentation.Tests/Instrumentations/NoCode/Cel/CelExpressionTests.cs
@@ -494,10 +494,26 @@ public class CelExpressionTests
         Assert.Equal(expected, Convert.ToDouble(result, System.Globalization.CultureInfo.InvariantCulture), 5);
     }
 
-    [Fact]
-    public void Evaluate_AddOperator_WithNonNumericTypes_ReturnsNull()
+    [Theory]
+    [InlineData("\"hello\" + \"world\"", "helloworld")]
+    [InlineData("\"hello\" + \" \" + \"world\"", "hello world")]
+    [InlineData("\"\" + \"test\"", "test")]
+    [InlineData("\"test\" + \"\"", "test")]
+    [InlineData("'single' + 'quotes'", "singlequotes")]
+    public void Evaluate_AddOperator_WithStrings_ReturnsConcatenation(string expression, string expected)
     {
-        var expr = CelExpression.Parse("\"hello\" + \"world\"");
+        var expr = CelExpression.Parse(expression);
+        var context = new NoCodeExpressionContext(null, null, null, null, null);
+
+        var result = expr!.Evaluate(context);
+
+        Assert.Equal(expected, result);
+    }
+
+    [Fact]
+    public void Evaluate_AddOperator_WithMixedTypes_ReturnsNull()
+    {
+        var expr = CelExpression.Parse("\"hello\" + 123");
         var context = new NoCodeExpressionContext(null, null, null, null, null);
 
         var result = expr!.Evaluate(context);


ysolomchenko:
Is it a vendored part from CEL package, or is it just a self-implementation of the required parts of CEL?

Kielek:
> Is it a vendored part from CEL package, or is it just a self-implementation of the required parts of CEL?

Self-implemented together with our @coplilot friend.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelExpression.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        var property = PropertyCache.GetOrAdd(cacheKey, key =>
+        {
+            var prop = key.Item1.GetProperty(key.Item2, BindingFlags.Public | BindingFlags.Instance | BindingFlags.IgnoreCase);
+            if (prop == null)
+            {
+                Log.Debug("Property '{0}' not found on type '{1}'", key.Item2, key.Item1.FullName);
+            }
+
+            return prop;
```
nrcventura: Consider caching the property getter instead of the ProperInfo reference. Directly invoking the property getting method is faster than calling proertyInfo.GetValue if I'm remembering some previous microbenchmarking results correctly.

nrcventura (approved):
I think that this makes sense as a first pass/initial implementation. I think that there are multiple opportunities to make this perform better, but such improvements can be done later.

Some ideas include. Adding some caching at the node level so that certain evaluations do not need to repeated with each call (such as the string comparisons). A bigger change that might lead to a larger improvement in performance overall is to consider converting each celNode into a dotnet `Expression`, and then have the root node return a compiled `Expression.Lambda` that takes the calling context as input, which gives a Func<Context> instance that can be invoked within the nocode instrumentation.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelExpression.cs
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    internal static object? GetPropertyValue(object target, string propertyName)
+    {
+        var targetType = target.GetType();
+        var cacheKey = (targetType, propertyName);
+
+        var property = PropertyCache.GetOrAdd(cacheKey, key =>
+        {
+            var prop = key.Item1.GetProperty(key.Item2, BindingFlags.Public | BindingFlags.Instance | BindingFlags.IgnoreCase);
```
iskiselev, line 92: It is possible for override property to not define getter, despite base property has it. We may consider in that case to do proper search over type hierarchy - or just document it:
```c#
using System.Reflection;

Console.WriteLine(
	typeof(B)
	  .GetProperty("Prop", BindingFlags.Public | BindingFlags.Instance | BindingFlags.IgnoreCase)
	  ?.GetMethod
	);
var b = new B();
Console.WriteLine(b.Prop);

public class A{
public virtual int Prop 
{
 get => 99;
 set {}
}
}

public class B : A{
public override int Prop 
{
 set{}
}
}
```

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelExpression.cs
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        if (target is System.Collections.IList list)
+        {
+            if (index < 0 || index >= list.Count)
+            {
+                return null;
+            }
+
+            return list[index];
```
iskiselev: There is try/catch when we call property accessor, but not here - where we still call some method. We should consider add it here.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelExpression.cs
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        return null;
```
iskiselev: From practice, customers may want to also get data from:
- IList<T> / IReadOnlyList<T> - may be harder to implement as genereic involved;
- IEnumerable - by just skippin N items - especially often asks for first item from IEnumerable;
- Span ? - have not seen such request, but more and more API work with Span - but I'm not sure if they are able to pass though interception mechanics - probably not due to non-heap restrictions.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelExpression.cs
```diff
@@ -0,0 +1,143 @@
...
+
+namespace OpenTelemetry.AutoInstrumentation.Instrumentations.NoCode.Cel;
+
+/// <summary>
+/// CEL (Common Expression Language) compatible expression evaluator.
+/// Supports a minimal subset of CEL for performance-critical instrumentation scenarios.
+/// </summary>
+/// <remarks>
+/// Supported features:
+/// - Member access: instance.property, arguments[0].field
```
iskiselev, line 16: Very often user need to extract data from string-indexed dictionary-like objects, so may be nice candidate for future extension.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelLexer.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            if (c == '\\' && position + 1 < input.Length)
+            {
+                position++;
+                var next = input[position];
+                sb.Append(next switch
+                {
+                    'n' => '\n',
+                    'r' => '\r',
+                    't' => '\t',
```
iskiselev, line 123: If I correctly read CEL [language definition](https://github.com/google/cel-spec/blob/master/doc/langdef.md), only "\n" and "\r" has special meaning. There are no quotes-escape, but has `"""str"""` and `'''s'''`. 

Are there any other language agent that has dynamic property capability? Do we have something like it in Collector? Should we have common language for dynamic expressions across OTEL? Answers to it may help us undedrstand how far we may go from standard for increase convinient for .NET-centered people.

iskiselev: Was wrong. I find that it in fact has escape sequence: https://github.com/google/cel-spec/blob/master/doc/langdef.md#string-and-bytes-values

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelExpression.cs
Kielek: 4fa22ec7afb16758c538978ab44eba9614bb8cff

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelExpression.cs
Kielek: 4d96c9f83ebc1159127d9eee5016e798a2dc97e4

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelLexer.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            // String literals
+            if (c == '"' || c == '\'')
+            {
+                tokens.Add(ReadStringLiteral(input, ref position));
+                continue;
+            }
+
+            // Numbers
+            if (char.IsDigit(c) || (c == '-' && position + 1 < input.Length && char.IsDigit(input[position + 1])))
```
iskiselev: It will not parse correclty BinOp minus if it is followed by number.
Looks like the easiest fix will be to always treat numbers as positive only with unary minus to cover first negative.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelExpression.cs
Kielek: 798dee89c38e1105350557914bcf8f89e312370b

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelExpression.cs
Kielek: Deffered to #4887. It can be done as follow up, when the initial implementation is done/further releases

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelExpression.cs
Kielek: #4888

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelLexer.cs
Kielek: 561e2b2af4ff4acef449445074c89998cdd48f94

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelExpression.cs
```diff
@@ -0,0 +1,179 @@
...
+
+                    baseType = baseType.BaseType;
+                }
+
+                if (getMethod == null)
+                {
+                    Log.Debug("Property '{0}' on type '{1}' has no getter", key.Item2, key.Item1.FullName);
+                    return null;
+                }
+            }
```
iskiselev, line 124: It may be good idea to check number of arguments (should be 0) - to be sure that the property is not an indexer.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelExpression.cs
```diff
@@ -0,0 +1,179 @@
...
+            return _root.Evaluate(context);
+        }
+        catch (Exception ex)
+        {
+            Log.Debug("Failed to evaluate CEL expression '{0}': {1}", _rawExpression, ex.Message);
+            return null;
+        }
+    }
+
+    internal static object? GetPropertyValue(object target, string propertyName)
```
iskiselev, line 85: Right now we don't use any type information here.
In fact, in many cases dynamic attributes will be defined with very simple property path that is statically known based on instrumented method. That may allow greatly improve perfomance of such expressions - but it will have one quirk comparing with fully dynamic implementation:
- if dervied type hides property from static-known base type, dynamic-only implementation will select property from derived class, where type-aware implementation may prefer type from static-known base type.

I'm putting it here to think, if we can come up with implementation that will be type-aware from the beginnig? It will fallback to dynamic impelementation if property is not discovered with provided static type.

Kielek: #4890 together with @nrcventura message: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4875#pullrequestreview-3879271659

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelExpression.cs
Kielek: 2e379481e438152eed7a0c7bcc1bf36ebf83a658
It will be some perf-improvement, but in general behavior was not changed. Exception was mapped to the null already.

nrcventura (approved):

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelLexer.cs
```diff
@@ -0,0 +1,238 @@
...
+                var next = input[position];
+                sb.Append(next switch
+                {
+                    'n' => '\n',
+                    'r' => '\r',
+                    't' => '\t',
+                    '\\' => '\\',
+                    '"' => '"',
+                    '\'' => '\'',
+                    _ => next
```
iskiselev: Per https://github.com/google/cel-spec/blob/master/doc/langdef.md#string-and-bytes-values
The following constructions are syntactically invalid and will result in a parse error:
A backslash (\) outside of a valid escape sequence, e.g. \s.

Despite we do not parse all escape sequence, we probably should reject all sequence that we do not parse instead of just skipping "/"

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelLexer.cs
```diff
@@ -0,0 +1,238 @@
...
+
+        // Handle negative sign
+        if (input[position] == '-')
+        {
+            sb.Append('-');
+            position++;
+        }
+
+        // Read digits
+        while (position < input.Length && (char.IsDigit(input[position]) || input[position] == '.'))
```
iskiselev, line 154: Note: numbers with more than one dots will fail at Parse step, which should be ok. Same as too big or too small for int / double.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelParser.cs
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    private CelNode ParseTernary()
+    {
+        var expr = ParseLogicalOr();
+
+        if (Match(CelTokenType.Question))
+        {
+            Consume();
+            var trueExpr = ParseTernary();
```
iskiselev: True expression should use `ParseLogicalOr` per https://github.com/google/cel-go/blob/master/parser/gen/CEL.g4#L25

In [playground](https://playcel.undistro.io/): `true ? true ? 1 : 3 : 2` - fails, while it is parsed in our implementation

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelParser.cs
```diff
@@ -0,0 +1,278 @@
...
+        {
+            var op = Consume();
+            var right = ParseEquality();
+            left = new CelBinaryOperatorNode(left, op.Value, right);
+        }
+
+        return left;
+    }
+
+    private CelNode ParseEquality()
```
iskiselev: Nitpicking, but `ParseEquality` should be unified with `ParseRelational` as all relation operation will have same rank:
https://github.com/google/cel-go/blob/master/parser/gen/CEL.g4#L38

Current implentation will return true for: `false <= false == false`, while CEL return false: https://playcel.undistro.io/

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/Nodes/CelBinaryOperatorNode.cs
```diff
@@ -0,0 +1,229 @@
...
+            var rightNum = Convert.ToDouble(right, CultureInfo.InvariantCulture);
+            return leftNum < rightNum;
+        }
+
+        if (left is string leftStr && right is string rightStr)
+        {
+            return string.CompareOrdinal(leftStr, rightStr) < 0;
+        }
+
+        return false;
```
iskiselev, line 96: CEL support all relation over boolean. If we omit it, my comment about higher rank of equility can be ignored.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/Nodes/CelBinaryOperatorNode.cs
```diff
@@ -0,0 +1,229 @@
...
+            "+" => Add(left, right),
+            "-" => Subtract(left, right),
+            "*" => Multiply(left, right),
+            "/" => Divide(left, right),
+            "%" => Modulo(left, right),
+            _ => null
+        };
+    }
+
+    private static bool AreEqual(object? left, object? right)
```
iskiselev, line 48: We probably should support equlity over boolean. But it can be a separate ticket.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/Nodes/CelTernaryNode.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        _condition = condition;
+        _trueValue = trueValue;
+        _falseValue = falseValue;
+    }
+
+    public override object? Evaluate(NoCodeExpressionContext context)
+    {
+        var condition = _condition.Evaluate(context);
+        var isTrue = condition switch
```
iskiselev: Should we use `CelBinaryOperatorNode.IsTrue` to make sure our bool conversion will be in-sync?

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/Nodes/CelUnaryOperatorNode.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        return _operator switch
+        {
+            "!" => !IsTrue(operand),
+            "-" => Negate(operand),
+            _ => null
+        };
+    }
+
+    private static bool IsTrue(object? value)
```
iskiselev: Same. Should we use `CelBinaryOperatorNode.IsTrue` to make sure our bool conversion will be in-sync?

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelParser.cs
Kielek: a498334fc49200735def3823c62c651524affd6a

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/Nodes/CelBinaryOperatorNode.cs
Kielek: 796b055c95ca01bd1df67b68847a73cf095f2ddc

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/Nodes/CelUnaryOperatorNode.cs
Kielek: 7932520a4e27a1da330543cbfc1731e0e5b1e7a0

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/Nodes/CelTernaryNode.cs
Kielek: 7932520a4e27a1da330543cbfc1731e0e5b1e7a0

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelParser.cs
Kielek: b41a9af6ef2454fc582727e746b73e4c969be172

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelLexer.cs
Kielek: c8a4ab414c1ffa57a4524670c853d3bfd55de5e4

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/Nodes/CelFunctionCallNode.cs
```diff
@@ -0,0 +1,205 @@
...
+
+    public CelFunctionCallNode(string functionName, CelNode[] arguments)
+    {
+        _functionName = functionName;
+        _arguments = arguments;
+    }
+
+    public override object? Evaluate(NoCodeExpressionContext context)
+    {
+        return _functionName.ToUpperInvariant() switch
```
iskiselev: CEL is case-sensitive, so i suggest use case-sensitive check with lower case constants.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/Nodes/CelFunctionCallNode.cs
```diff
@@ -0,0 +1,205 @@
...
+        {
+            "CONCAT" => EvaluateConcat(context),
```
iskiselev: CEL supports "+" operator for string concatenation, so add same support and remove non-standard concat function.
We can add it later if needed.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelExpression.cs
```diff
@@ -0,0 +1,184 @@
...
+
+/// <summary>
+/// CEL (Common Expression Language) compatible expression evaluator.
+/// Supports a minimal subset of CEL for performance-critical instrumentation scenarios.
+/// </summary>
+/// <remarks>
+/// Supported features:
+/// - Member access: instance.property, arguments[0].field
+/// - Operators: ==, !=, &gt;, &lt;, &gt;=, &lt;=, &amp;&amp;, ||, !
+/// - Functions: concat(), coalesce(), substring(), string(), size()
```
iskiselev: See my comment about concat/colesce. If we remove it, it should be removed here also.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/Nodes/CelFunctionCallNode.cs
Kielek: d35cffd8986f20d5d76419519deeedb0ecdfbf32

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/CelExpression.cs
Kielek: d35cffd8986f20d5d76419519deeedb0ecdfbf32, no more references to concat ahd  coalesce (except integration test with ternary operator)

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/Nodes/CelFunctionCallNode.cs
Kielek: 52cbd9f15d561cf58b49e3c7f5bb7c44e5fd82e8

src/OpenTelemetry.AutoInstrumentation/Instrumentations/NoCode/Cel/Nodes/CelBinaryOperatorNode.cs
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        if (IsNumeric(left) && IsNumeric(right))
+        {
+            var leftNum = Convert.ToDouble(left, CultureInfo.InvariantCulture);
+            var rightNum = Convert.ToDouble(right, CultureInfo.InvariantCulture);
+            return leftNum + rightNum;
+        }
+
+        return null;
```
iskiselev, line 145: We should add string concatenation here instead of concat.

Kielek: 8f3bb4f8bf87bef09b6ac2a8f5b56b5cbbdeb826
You are right.
Integration tests were in the place, but for some reasons it was not refreshed in my local unit tests session.

PR #4894, opened 2026-03-05
[FileBasedConfiguration] Validate File Existence

ysolomchenko:
## What

Add file existence validation for file-based configuration.

## Tests

Manual

## Checklist

### DIFF
--- a/test/IntegrationTests/docker/rabbitmq.Dockerfile
+++ b/test/IntegrationTests/docker/rabbitmq.Dockerfile
@@ -1,1 +1,1 @@
-FROM rabbitmq:4.2.4@sha256:8a76a8d5c418faf260abba10580f983890ad1df069fe27e5f1cfeab8e327453d
+FROM rabbitmq:4.2.4@sha256:132dbf078b6b188bcf21da09796270e440fb4beb22b44e8aeee311404d1caf88


src/OpenTelemetry.AutoInstrumentation/Configurations/Settings.cs
```diff
@@ -82,7 +82,11 @@ private static YamlConfiguration ReadYamlConfiguration()
             configFile = "config.yaml";
         }
 
-        // TODO validate file existence
+        if (!File.Exists(configFile))
+        {
+            Logger.Error($"Configuration file '{configFile}' was not found.");
+            throw new FileNotFoundException($"Configuration file '{configFile}' was not found.", configFile);
```
RassK: shouldn't it also use failfast settings?

ysolomchenko: No, failfast cannot be used in that place, because it only becomes available after the file is read.

RassK: Ah sure, that's correct. Does it throw if you input a random or malformed yaml?

Ideally I'm thinking here, if those exceptions should be wrapped with a single high level exception that's pointing out the root of the issues (so here we know that the root is always configuration file). eg: FileConfigurationException("There was an issue with configuration file."), then internal exception provides the details (either FileNotFound, InvalidYaml, InvalidAccess etc).

ysolomchenko: If the file is a valid YAML file, no exception will be thrown and the default behavior will be used. This is intentional we ignore unknown fields so the file based can be used across different distributions, where some of them may provide additional functionality.

If the file is not in YAML format or corrupted, an exception will be thrown at the deserializer level.

nrcventura: If someone configured the fail fast environment variable should we respect it? It seems like always throwing this exception could be problematic.

ysolomchenko: > If someone configured the fail fast environment variable should we respect it? It seems like always throwing this exception could be problematic.

It would be strange if fail fast is configured from a file, but we respect the fail fast from an environment variable and only in this one place.

RassK: > If someone configured the fail fast environment variable should we respect it? It seems like always throwing this exception could be problematic.

Weird situation here... spec does not allow it unless it's defined in the yaml itself. So theoretically you could read it from the env but in the same time, you do not know if it was defined (since file is missing) 😄

nrcventura: Does the specification provide guidance on what to do here? What do the other auto instrumentation implementations do?

Kielek:
SIG note:
consider no-op version, just log ERROR with information that the file is not available.
Fail-fast should not be an option as you cannot read it from the file.

but Chris note is also valid;
> Does the specification provide guidance on what to do here? What do the other auto instrumentation implementations do?

src/OpenTelemetry.AutoInstrumentation/Configurations/Settings.cs
ysolomchenko: > Does the specification provide guidance on what to do here? What do the other auto instrumentation implementations do?

The specification states (https://github.com/open-telemetry/opentelemetry-specification/blob/5712620c7840d17a3e358ddb50e2d6029724f09a/specification/configuration/sdk.md?plain=1#L233) that parsing should return an error.

In Java, if a non-existent file is specified, an exception is thrown and the application starts without the agent

nrcventura: Based on that, I think the decision described in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4894#issuecomment-4040636516 makes sense. This should not trigger an error in the customer application, but we should log an error.

src/OpenTelemetry.AutoInstrumentation/Configurations/Settings.cs
```diff
@@ -82,7 +82,10 @@ private static YamlConfiguration ReadYamlConfiguration()
             configFile = "config.yaml";
         }
 
-        // TODO validate file existence
+        if (!File.Exists(configFile))
+        {
+            Logger.Error($"Configuration file '{configFile}' was not found.");
+        }
```
Kielek, line 89: I think that you should return noop-configuration here. Why you are trying to parse non-existing file?

ysolomchenko: 1995409

PR #4901, opened 2026-03-06
[infra] Update C# LangVersion to 14.0

Kielek:
## Why & What

[infra] Update C# LangVersion to 14.0

_build.csproj needs to stay on 13.0, as it is used to build native code on legacy Ubuntu with net9.0. The same is for SourceGenerator.

Also affected are aour applications build for IIS. Under the hood local MSBuild/VS version is used. It means VS2022 on CI, so it cannot be updated.

## Tests

CI

## Checklist


Kielek:
@alexeypukhov, FYI

PR #4904, opened 2026-03-06
Clarify config behavior when running multiple services in the same Application Pool

matt-hensley:
## Why

Documents a non-obvious config behavior for users hosting multiple ASP.NET/.NET Framework applications in the same IIS App Pool.

When services share an AppPool they're running inside a single w3wp.exe process, and the OTel SDK is initialized only once per process.
  
All applications in the pool inherit the configuration of whichever application starts first. web.config `OTEL_*` settings in later-starting applications are ignored without any error or warning.

### DIFF
--- a/docs/iis-instrumentation.md
+++ b/docs/iis-instrumentation.md
@@ -33,7 +33,6 @@ for further details.
 > Remember to restart IIS after making configuration changes.
 > You can do it by executing `iisreset.exe`.
 
-
 > [!WARNING]
 > When multiple ASP.NET (.NET Framework) applications share the same
 > Application Pool, they run inside a single `w3wp.exe` process. `OTEL_*`


iskiselev:
Sounds strange.
Different web-application in same AppPool should have different AppDomains. So, it will share environment varaibles, but not web.config settings.
It sounds like a bug for me. I'll need to check it. Can I use any setting to reproduce it? For example `OTEL_SERVICE_NAME`?

RassK:
> Sounds strange. Different web-application in same AppPool should have different AppDomains. So, it will share environment varaibles, but not web.config settings. It sounds like a bug for me. I'll need to check it. Can I use any setting to reproduce it? For example `OTEL_SERVICE_NAME`?

Yes,  `OTEL_SERVICE_NAME` is a good test for this. The problem is that `web.config` settings are converted to envs. OTel Auto reads from (web/app).config but SDK does not do it, that's why those settings are converted to envs.

That time we did not want to create all options instances in auto side, so SDK would just silently read envs. I think declarative config introduced anyway all available options types to auto side, maybe it could be fixed by going over all options.

PR #4906, opened 2026-03-09
Fix test application console buffer deadlocks by ensuring consistent process stream draining

alexeypukhov:
## Why

Some tests may suffer a deadlock when their console output increases (just enough to overflow the buffer which by default is not that big, even a line can cause it)

Discovered while working on https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4783 but is not required for this PR

## What

Introduced IntegrationTestProcess to start reading output streams immediately upon launch. Updated ProcessHelper to reuse existing readers where available, preventing console buffer deadlocks across all test patterns.

## Tests

Existing Integration tests will validate it

## Checklist

### DIFF
--- a/test/IntegrationTests/Helpers/InstrumentedProcessHelper.cs
+++ b/test/IntegrationTests/Helpers/InstrumentedProcessHelper.cs
@@ -30,6 +30,26 @@ internal static class InstrumentedProcessHelper
         startInfo.RedirectStandardInput = false;
         startInfo.StandardOutputEncoding = Encoding.Default;
 
-        return Process.Start(startInfo);
+        // 1. Create the specialized process object
+        var process = new IntegrationTestProcess
+        {
+            StartInfo = startInfo
+        };
+
+        // 2. Start it
+        process.Start();
+
+        // 3. IMMEDIATELY attach the helper.
+        // This calls BeginOutputReadLine()/BeginErrorReadLine()
+        // and prevents the process hang if it has a lot of output that no one reads
+        process.AttachedHelper = new ProcessHelper(process);
+
+        return process;
+    }
+
+    internal sealed class IntegrationTestProcess : Process
+    {
+        // This holds the helper that is already draining the process
+        public ProcessHelper? AttachedHelper { get; set; }
     }
 }
--- a/test/IntegrationTests/Helpers/ProcessHelper.cs
+++ b/test/IntegrationTests/Helpers/ProcessHelper.cs
@@ -26,6 +26,18 @@ internal sealed class ProcessHelper : IDisposable
             return;
         }
 
+        // If this is our special process type and it already has a helper...
+        if (process is InstrumentedProcessHelper.IntegrationTestProcess { AttachedHelper: not null } it)
+        {
+            // "Adopt" the state of the existing helper instead of starting a new one
+            Process = it.AttachedHelper.Process;
+            _outputBuffer = it.AttachedHelper._outputBuffer;
+            _errorBuffer = it.AttachedHelper._errorBuffer;
+            _outputMutex = it.AttachedHelper._outputMutex;
+            // Note: In this case, don't call BeginOutputReadLine()/BeginErrorReadLine() again!
+            return;
+        }
+
         Process = process;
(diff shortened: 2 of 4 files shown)


test/IntegrationTests/Helpers/InstrumentedProcessHelper.cs
```diff
@@ -30,6 +30,26 @@ internal static class InstrumentedProcessHelper
...
+            StartInfo = startInfo
+        };
+
+        // 2. Start it
+        process.Start();
+
+        // 3. IMMEDIATELY attach the helper.
+        // This calls BeginOutputReadLine()/BeginErrorReadLine()
+        // and prevents the process hang if it has a lot of output that no one reads
+        process.AttachedHelper = new ProcessHelper(process);
```
iskiselev, line 45: Should we subscribe (create helper) before `process.Start();`

test/IntegrationTests/Helpers/ProcessHelper.cs
```diff
@@ -26,6 +26,18 @@ public ProcessHelper(Process? process)
             return;
         }
 
+        // If this is our special process type and it already has a helper...
+        if (process is InstrumentedProcessHelper.IntegrationTestProcess { AttachedHelper: not null } it)
```
iskiselev, line 30: Why not always create at and drop attach path?

test/IntegrationTests/Helpers/InstrumentedProcessHelper.cs
alexeypukhov: I agree and thank you for your comment. This reminded me that the change here was a quick fix which I originally intended to make as a part of https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4783 so I tried to keep it as short as possible.

Now that we extracted the change into a separate PR and it doesn't even block my original PR we can work on a better design.

---
My plan was to introduce a Process subclass: `class IntegrationTestProcess : Process` that owns its output buffer which it starts draining inside its own `Start()` override - so draining is always guaranteed to begin before control returns to any caller. This directly addresses your "subscribe before Start" concern above.

`StartTestApplication` and `RunTestApplication` then be dealing with `IntegrationTestProcess` instead of base `Process` and `ProcessHelper` then becomes a thin wrapper that simply borrows the already-running buffer from that subclass or starts a buffer if happened to receive the base `Process`.

The impact on existing tests should be minimal since the subclass inherits `Process` and `ProcessHelper` keeps the same constructor signature.
There are three distinctive patterns I've noticed in the tests and all of them align well with the new design:
1. PATTERN 1: Fire and Forget - `NServiceBusTests.cs`
```
// The process is automatically draining in the background. No hang.
using var process = StartTestApplication(new TestSettings { PackageVersion = packageVersion });
```
2. PATTERN 2: Interactive / Scoped - `GraphQLTests.cs`
```
// ProcessHelper now just proxies the existing buffer. Direct access is also possible.
using var process = StartTestApplication(new TestSettings { PackageVersion = packageVersion });
using var helper = new ProcessHelper(process); 
try
{
}
finally
{
    if (process is { HasExited: false })
    {
        process.Kill();
        // Wait for exit and then ensure buffer is drained
        process.WaitForExit();
        process.Buffer.WaitForDrain(TimeSpan.FromSeconds(1));
    }

    Output.WriteLine(process.Buffer.StandardOutput);
}
```
3. PATTERN 3: Run and Check - `HttpTests.cs`
```
// The simplest pattern. RunTestApplication returns the captured strings.
var result = RunTestApplication(new() { PackageVersion = packageVersion });
Output.WriteLine(result.StandardOutput);
```

We can definitely revise and get rid of some patterns in tests. For example pattern 2 with `ProcessHelper` may become not so needed if the heavy-lifting is delegated to `IntegrationTestProcess`. That will simplify the refactor as well. The most used pattern based on my observation is pattern 3

test/IntegrationTests/Helpers/ProcessHelper.cs
alexeypukhov: With the new design idea that I posted in the comment above, the `ProcessHelper` becomes a thin wrapper that just borrows the already-running buffer so I believe it should address your concern here

PR #4914, opened 2026-03-11
Instrumentation stability proposal for Auto Instrumentation

### LINKED ISSUE #4846: We need guidelines to determine when we should mark instrumentation as stable
At this point in time, all of the instrumentation listed in our documentation is marked as stable. Since then some libraries have released stable instrumentation. Before we can update our documentation to declare that instrumentation stable we need to define what "stable" means for auto instrumentation.

Consider an instrumentation library that auto instrumentation pulls in from the contrib repo. Auto instrumentation has the ability to pull in specific versions of that instrumentation as a dependency giving us greater control over when to roll out changes to the instrumentation. Contrast this with a library like HttpClient or RabbitMQ where the instrumentation is built directly into the library and auto instrumentation simply listens to an ActivitySource. In that scenario, auto instrumentation is does not have any control over which "version" of the instrumentation is being used.

Some of the things that we need to decide.

1. What does it mean to declare instrumentation stable within auto instrumentation? Does auto-instrumentation need to ensure that the instrumentation is producing data following the appropriate stable semantic conventions?
2. Do our integration tests need to be more robust and validate that the stable semantic conventions are being followed?
3. Do we need different guidelines for native (built-in) instrumentation in a library, library instrumentation (instrumentation provided in a separate instrumentation library), and zero-code instrumentation (instrumentation maintained directly by us)?

nrcventura:
## Why

Resolves #4846.

## What

This PR creates a document that defines what we need to do before marking a particular instrumentation as stable. I'm submitting a PR because I think it will be easier to comment and make changes to a concrete proposal instead of just discussing it using comments in the original issue.

## Checklist

### DIFF
--- a/docs/internal/instrumentation-stability.md
+++ b/docs/internal/instrumentation-stability.md
@@ -28,7 +28,7 @@ An example of this type of instrumentation is a library such as `Npgsql`, which 
 code. An end user, or auto instrumentation does not need to bring in additional dependencies to make the instrumentation available. The maintenance of this
 code is outside of the scope of auto instrumentation. The ability to control which version of the instrumentation is used, is also outside of the control
 of auto instrumentation. For example, if an end user is using `Npgsql` version 6 in their application, they will be tied to the semantic conventions
-produced by that version of the library. Auto instrumentation will should not upgrade or download the `Npgsql` version just to enforce that a specific 
+produced by that version of the library. Auto instrumentation will not upgrade or download the `Npgsql` version just to enforce that a specific 
 semantic convention version is used between minor releases of auto instrumentation.
 
 Auto instrumentation does not need to test for semantic convention compliance for natively instrumented code. Semantic convention compliance is defined 


docs/internal/instrumentation-stability.md
```diff
@@ -0,0 +1,71 @@
...
+
+Each of these instrumentation types have their own maintainers that may be different than the maintainers of auto instrumentations.
+
+### Natively instrumented code
+
+An example of this type of instrumentation is a library such as `Npgsql`, which is used by an end user, include OpenTelemetry instrumentation directly in its
+code. An end user, or auto instrumentation does not need to bring in additional dependencies to make the instrumentation available. The maintenance of this
+code is outside of the scope of auto instrumentation. The ability to control which version of the instrumentation is used, is also outside of the control
+of auto instrumentation. For example, if an end user is using `Npgsql` version 6 in their application, they will be tied to the semantic conventions
+produced by that version of the library. Auto instrumentation will should not upgrade or download the `Npgsql` version just to enforce that a specific
```
iskiselev:
```suggestion
produced by that version of the library. Auto instrumentation will not upgrade or download the `Npgsql` version just to enforce that a specific 
```

docs/internal/instrumentation-stability.md
```diff
@@ -0,0 +1,71 @@
...
+instrumentation API stability.
+
+## Instrumentation types and stability
+
+Auto instrumentation has the ability to manage multiple forms of instrumentation.
+
+1. Natively instrumented code
+2. Instrumentation libraries
+3. Byte-code instrumentation
+4. No-code instrumentation
```
ysolomchenko, line 21: I’m not sure that no-code instrumentation should be here. It’s more like manual instrumentation and should be treated as a feature from a stability perspective.

nrcventura: I'm mostly including it for completeness, and to have something that we can point to if someone asks a question about it.

Issue #4924, opened 2026-03-13
UnsafeAccessorType loads DiagnosticSource to Default ALC breaking StartupHook isolation on .NET 10+


alexeypukhov:
# Bug Report

### Summary
The `UnsafeAccessorTypeAttribute` (introduced in .NET 10) loads `System.Diagnostics.DiagnosticSource` into Default ALC during .NET runtime initialization, bypassing StartupHook isolation and causing type/state drift across ALCs.

**This issue is already reproducible on .NET 10 and affects StartupHook isolation deployments on .NET 10+.**

### Background: StartupHook Isolation Mode

Our assembly conflict resolution approach (introduced in [PR #4783](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4783)), when the native profiler is not installed, uses StartupHook isolation to avoid assembly conflicts and works as follows:
- The entire customer application is loaded into a custom (isolated) AssemblyLoadContext
- The isolated context is set to be the `CurrentContextualReflectionContext` (via `customALC.EnterContextualReflection()`) to support runtime reflection redirection
- All dependencies the customer app brings go through isolated ALC overriden `Load()` method picking up the highest version (ours or app)
- Only assemblies that **must** remain in Default ALC (like `System.Private.CoreLib`, which loads before our StartupHook) are excluded from isolation

**Goal:** Maximize isolation to prevent version conflicts between customer dependencies and our agent dependencies.

### The Problem: UnsafeAccessor Bypasses Isolation

.NET 10 introduced a new JIT-level reflection mechanism via `UnsafeAccessorTypeAttribute`. Unlike traditional managed reflection, it performs JIT-level type resolution that sticks to the ALC context of the **requesting assembly**, bypassing `CurrentContextualReflectionContext`

**Why UnsafeAccessorType breaks StartupHook isolation:** When the requesting assembly is in Default ALC, the runtime try to load it to Default ALC. As a result the requesting assembly and all its dependencies may leaked to Default ALC.

**General impact:** Any assembly in Default ALC using `UnsafeAccessorType` will cause a leak to Default ALC.

### Why This Is Especially Critical for OpenTelemetry

The .NET runtime uses this during initialization to access types from DiagnosticSource:

```csharp
[UnsafeAccessor(UnsafeAccessorKind.StaticMethod, Name = "GetInstance")]
[...]

PR #4931, opened 2026-03-16
UnsafeAccessorType native redirection

### LINKED ISSUE #4923: UnsafeAccessorType will bypass Native Profiler assembly redirection for DiagnosticSource (v11+) on .NET 10+
# Bug Report

### Summary
The `UnsafeAccessorTypeAttribute` (introduced in .NET 10) bypasses our native profiler's assembly redirection mechanism, which will cause type and state drift on .NET 10+ when OpenTelemetry upgrades to System.Diagnostics.DiagnosticSource 11.

**This issue is not currently reproducible but will manifest once we switch to DS 11 after .NET 11 is released.**

### Background: Native Profiler Assembly Redirection

Our assembly conflict resolution approach (introduced in [PR #4783](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4783)) works as follows:
- **Native profiler** (`cor_profiler.cpp`) redirects `AssemblyRefs` for key dependencies using the redirection map in `assembly_redirection_net.h`
- **Managed AssemblyResolver** subscribes to `DefaultALC.Resolving` event to handle redirected dependencies  
- When a **higher version** than the app originally referenced is needed, we load it into a **custom ALC** to avoid runtime crashes

**Goal:** Prevent version conflicts by ensuring all code paths reference our redirected (higher) dependency versions.

### The Problem: UnsafeAccessor Bypasses Redirection

.NET 10 introduced `UnsafeAccessorTypeAttribute`, a JIT-level reflection mechanism:
- Native redirection of `AssemblyRef` does not affect assembly resolution during reflection
- Unlike traditional managed reflection, it performs JIT-level type resolution that sticks to the ALC context of the **requesting assembly**, bypassing `CurrentContextualReflectionContext` (so `customALC.EnterContextualReflection()` has no effect)

**Why UnsafeAccessorType breaks Native Assembly redirection:** When the runtime is requested to load an assembly from Default ALC and can find it (e.g. in TPA list), it loads immediately to Default ALC—there is no opportunity to interfere.

**General impact:** Any assembly in Default ALC using `UnsafeAccessorType` to reference our redirected dependencies will cause drift when our version is higher than what originally referenced.

### Why This Is Especially Critical for OpenTelemetry

The .NET runtime itself (**System.Private.CoreLib**) uses `UnsafeAccessorType` during initialization to access types from DiagnosticSource:

```csharp
[UnsafeAccessor(UnsafeAccessorKind.StaticMethod, Name = "GetInstance")]
[return: UnsafeAccessorType("System.Diagnostics.Metrics.MetricsEventSource, System.Diagnostics.DiagnosticSource")]
static extern object GetInstance(
    [UnsafeAccessorType("System.Diagnostics.Metrics.MetricsEventSource, System.Diagnostics.DiagnosticSource")] object? _);
```

[Source: EventSourceInitHelper in System.Private.CoreLib](https://github.com/dotnet/runtime/blob/v10.0.0/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs#L3923C13-L3926C135)

**Why this is critical:** OpenTelemetry initialization triggers this runtime code path. `DiagnosticSource` is foundational to OpenTelemetry and must be loaded exactly once.

### Current State and Future Break Scenario

**Why not reproducible now:**
For .NET 10, we have **no DiagnosticSource redirection** in `assembly_redirection_net.h`. This is intentional: DS 10 ships with .NET 10 and satisfies OpenTelemetry's requirements, loading to Default ALC without conflict.

**What breaks when .NET 11 releases:**
When .NET 11 ships with DS 11, OpenTelemetry will upgrade accordingly. All .NET versions—including .NET 10—will receive a DS 11 redirection.

**For .NET 10 applications:**
1. Runtime (System.Private.CoreLib) loaded in Default ALC uses `UnsafeAccessorType` to load `MetricsEventSource` from DS
2. Our AssemblyRef redirection does not apply, and JIT reflection ignores contextual reflection—DS 10 is resolved from TPA and loaded immediately to Default ALC without raising the Resolving event
3. Our `AssemblyResolver` (via other code paths) loads DS 11 to custom ALC
4. **Result:** Two DS versions across ALCs → type drift and state drift

### Proposed Solution

Extend the native profiler to update **`UnsafeAccessorTypeAttribute`** assembly references (not just AssemblyRefs) when bumping dependency versions. This ensures UnsafeAccessor resolution respects redirections and triggers the `Resolving` event handler.

**Status:** I have a POC that successfully redirects UnsafeAccessorType definition which results in triggering the Resolving event.

### Related Issue

This issue is related to: [Issue #4924 - UnsafeAccessorType breaks StartupHook isolation](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/4924). Both stem from `UnsafeAccessorType` bypassing our assembly conflict resolution mechanisms, but affect differently different deployment modes.

### Runtime environment
- OpenTelemetry Automatic Instrumentation version: Discovered during work on [PR #4783](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/4783)
- OS: All platforms
- .NET version: .NET 10+ (will manifest when .NET 11 is released)
- Deployment mode: Native Profiler deployment

alexeypukhov:
## Why

**Fixes #4923**

On .NET 10+, assembly-qualified type names stored in `UnsafeAccessorTypeAttribute` metadata are resolved independently of normal `AssemblyRef` metadata. Consequently, rewriting only `AssemblyRef` entries leaves a path that can load an older redirected dependency into the wrong `AssemblyLoadContext`.

This change closes that gap for native-profiler deployments before dependencies such as `System.Diagnostics.DiagnosticSource` require a higher redirected version.

## What

- Detect whether the runtime provides `UnsafeAccessorTypeAttribute` before searching modules for metadata to rewrite. This limits the additional processing to supported .NET runtimes.
- Find complete unsafe-accessor declarations: methods using `UnsafeAccessorAttribute` together with `UnsafeAccessorTypeAttribute` on a return value or parameter.
- Rewrite assembly versions in `UnsafeAccessorTypeAttribute` values using the existing framework-specific assembly redirection map:
  - add a missing version;
  - replace a lower version with the mapped version;
  - preserve equal or higher versions;
  - participate in the existing highest-version selection behavior - symmetrical to AssemblyRef redirection.
- Re-encode modified custom-attribute blobs safely, including compressed string-length transitions and trailing metadata.
- Extend test-build automation to pass arbitrary MSBuild properties to generated application variants.
- Add focused native unit tests and a .NET 10 end-to-end test application.
- Document how `UnsafeAccessorTypeAttribute` redirection participates in assembly conflict resolution.

**NOTE:** This applies only to native-profiler deployments. StartupHook-only isolation remains subject to the limitation tracked by #4924.

The implementation redirects the outer assembly qualification of the unsafe-accessor type. Assembly identities nested inside generic type arguments are intentionally left unchanged.

## Tests

- Focused native unit-test suites:
  - `UnsafeAccessorTypeNameRewriterTest`
  - `UnsafeAccessorTypeAttributeBlobRewriterTest`
- Focused integration tests for .NET 10+ runtimes covering:
  - a versionless assembly name redirection;
  - a lower version redirection;
  - an equal version preservation;
  - a higher version preservation;
  - no redirection on an incomplete declaration without `UnsafeAccessorAttribute`.

## Checklist

### DIFF
--- a/docs/assembly-conflict-resolution.md
+++ b/docs/assembly-conflict-resolution.md
@@ -160,6 +160,18 @@ controlled by a version map compiled into the native profiler (see
 and
 [`assembly_redirection_netfx.h`](../src/OpenTelemetry.AutoInstrumentation.Native/assembly_redirection_netfx.h)).
 
+On .NET 10 and later, the profiler also rewrites the outer assembly qualifier
+of type names stored in `UnsafeAccessorType` attributes when that assembly
+appears in the same version map. This includes attributes in
+`System.Private.CoreLib` and in application or library modules. It considers
+only parameter and return-value attributes whose declaring method also has
+`UnsafeAccessor`, because that pairing is what gives `UnsafeAccessorType` its
+native-reflection meaning. A missing or lower `Version` is raised to the mapped
+version; an equal or higher version is preserved. As with `AssemblyRef`, a
+higher version seen before any rewrite becomes the target for later references.
+Only the outer type's assembly qualifier is handled; assembly qualifiers nested
+inside generic type arguments are left unchanged.
+
 > **NOTE**: On .NET, the instrumentation ships the baseline versions of its
 > dependencies for each target framework (for example, for `net8.0` the
 > 8.0.0 versions, for `net9.0` - 9.0.0, etc.) unless known issues
@@ -302,11 +314,13 @@ This can happen in the following scenarios:
   `AssemblyLoadContext.Default.LoadFromAssemblyPath` or `Assembly.LoadFrom`
   to load an assembly into the Default ALC.
 
-- **`UnsafeAccessorType` reflection (.NET 10+)**: Code uses
+- **StartupHook-only: `UnsafeAccessorType` reflection (.NET 10+)**: Code uses
   `UnsafeAccessorType` attributes in assemblies loaded in the Default ALC;
   in this case, types bind to the requesting assembly's load context.
   Once the runtime resolves a type for an `UnsafeAccessor` method in the
   Default ALC, it cannot interact with a different version loaded elsewhere.
+  StartupHook-only deployment cannot rewrite this metadata because the native
+  profiler is not attached.
 
 ### StartupHook-only: the customer application is loaded twice
 
@@ -352,12 +366,16 @@ architectures.
 
 ### Native profiler: conflicting version ordering
 
-The native profiler processes assembly references in module-load order.
-If an earlier module has been redirected to the instrumentation's version
-and a later module references a **higher** version than the
-instrumentation ships, the profiler cannot reconcile the two. This is
-logged as an error. In practice this is rare because the instrumentation
-ships the latest versions of its dependencies.
+The native profiler processes assembly references and `UnsafeAccessorType`
+attribute values in module-load order. CoreLib attributes are considered during
+CoreLib loading; in each later module, `AssemblyRef` entries are considered
+before attribute values. If an earlier reference has been
+redirected to the instrumentation's version and a later reference requests a
+**higher** version than the instrumentation ships, the profiler cannot safely
+revise the earlier metadata. This is logged as an error. An attribute in
+`System.Private.CoreLib` can be that first redirection because CoreLib loads
+before application modules. In practice this is rare because the
+instrumentation ships recent versions of its dependencies.
 
 ### Unexpected resolution request for a higher version than available
 
@@ -403,8 +421,9 @@ starts.
    `COREHOST_TRACEFILE=corehost_verbose_tracing.log` to capture the
    runtime's own assembly-loading decisions.
 3. **Review the native profiler log** — look for
-   `RedirectAssemblyReferences` entries to confirm that IL rewriting
-   happened and which versions were involved.
+   `RedirectAssemblyReferences` and `UnsafeAccessorTypeAttributeUpdater`
+   entries to confirm which metadata was rewritten and which versions were
+   involved.
 
 ### Last resort: `DOTNET_ADDITIONAL_DEPS` and the runtime store
 
--- a/src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
@@ -1,0 +1,519 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "unsafe_accessor_type_attribute_updater.h"
+
+#include <algorithm>
+#include <cctype>
+#include <cstdint>
+#include <limits>
+
+#include "logger.h"
+#include "module_metadata.h"
+#include "string_utils.h"
+
+namespace trace
+{
+
+namespace
+{
+
+const WSTRING unsafe_accessor_type_attribute_name = WStr("System.Runtime.CompilerServices.UnsafeAccessorTypeAttribute");
+const WSTRING unsafe_accessor_attribute_name      = WStr("System.Runtime.CompilerServices.UnsafeAccessorAttribute");
+
+// A custom-attribute SerString prefixes its UTF-8 bytes with an ECMA-335 compressed unsigned integer. Use the CLR's
+// codec for that standard 1/2/4-byte representation, but guard its reader because it examines the first byte before
+// validating the supplied length.
+bool TryReadPackedLength(const BYTE* data, size_t size, ULONG& length, size_t& bytes_read)
+{
+    if (data == nullptr || size == 0)
+    {
+        return false;
+    }
+
+    uint32_t   decoded_length = 0;
+    uint32_t   encoded_size   = 0;
+    const auto result         = CorSigUncompressData(data, static_cast<DWORD>(size), &decoded_length, &encoded_size);
+    length                    = static_cast<ULONG>(decoded_length);
+    bytes_read                = encoded_size;
+    return SUCCEEDED(result);
+}
+
+bool WritePackedLength(std::vector<BYTE>& blob, size_t length)
+{
+    constexpr size_t max_packed_length = 0x1FFFFFFF;
+    if (length > max_packed_length)
+    {
+        return false;
+    }
+
+    BYTE       encoded_length[4]{};
+    const auto encoded_size = CorSigCompressData(static_cast<ULONG>(length), encoded_length);
+    blob.insert(blob.end(), encoded_length, encoded_length + encoded_size);
+    return true;
+}
+
+size_t SkipWhitespace(const std::string& value, size_t position)
+{
+    while (position < value.size() && std::isspace(static_cast<unsigned char>(value[position])) != 0)
+    {
+        position++;
+    }
+
+    return position;
+}
+
+size_t TrimmedEnd(const std::string& value, size_t begin, size_t end)
+{
+    while (end > begin && std::isspace(static_cast<unsigned char>(value[end - 1])) != 0)
+    {
+        end--;
+    }
+
+    return end;
+}
+
+bool EqualsIgnoreCase(const std::string& left, const std::string& right)
+{
+    return left.size() == right.size() && std::equal(left.begin(), left.end(), right.begin(),
+                                                     [](char left_char, char right_char)
+                                                     {
+                                                         return std::tolower(static_cast<unsigned char>(left_char)) ==
+                                                                std::tolower(static_cast<unsigned char>(right_char));
+                                                     });
+}
+
+size_t FindAssemblySeparator(const std::string& type_name)
+{
+    // The first unescaped comma at bracket depth zero starts the outer assembly qualifier. Ignore commas inside generic
+    // arguments and escaped type-name commas.
+    size_t bracket_depth = 0;
+    bool   escaped       = false;
+
+    for (size_t i = 0; i < type_name.size(); i++)
+    {
+        const auto current = type_name[i];
+        if (escaped)
+        {
+            escaped = false;
+            continue;
+        }
+
+        if (current == '\\')
+        {
+            escaped = true;
+        }
+        else if (current == '[')
+        {
+            bracket_depth++;
+        }
+        else if (current == ']' && bracket_depth > 0)
+        {
+            bracket_depth--;
+        }
+        else if (current == ',' && bracket_depth == 0)
+        {
+            return i;
+        }
+    }
+
+    return std::string::npos;
+}
+
+bool TryParseVersion(const std::string& value, ASSEMBLYMETADATA& version)
+{
+    // Assembly metadata represents a version as exactly four unsigned 16-bit components.
+    constexpr auto max_component = static_cast<unsigned int>((std::numeric_limits<USHORT>::max)());
+    USHORT*        components[]  = {&version.usMajorVersion, &version.usMinorVersion, &version.usBuildNumber,
+                                    &version.usRevisionNumber};
+    size_t         position      = 0;
+
+    for (size_t i = 0; i < 4; i++)
+    {
+        const auto separator = value.find('.', position);
+        const auto end       = separator == std::string::npos ? value.size() : separator;
+        if (end == position || (i < 3 && separator == std::string::npos) || (i == 3 && separator != std::string::npos))
+        {
+            return false;
+        }
+
+        unsigned int component = 0;
+        for (size_t j = position; j < end; j++)
+        {
+            const auto current = value[j];
+            if (current < '0' || current > '9')
+            {
+                return false;
+            }
+
+            const auto digit = static_cast<unsigned int>(current - '0');
+            if (component > (max_component - digit) / 10)
+            {
+                return false;
+            }
+
+            component = (component * 10) + digit;
+        }
+
+        *components[i] = static_cast<USHORT>(component);
+        position       = end + 1;
+    }
+
+    return true;
+}
+
+std::string VersionString(const AssemblyVersionRedirection& redirect)
+{
+    return std::to_string(redirect.usMajorVersion) + "." + std::to_string(redirect.usMinorVersion) + "." +
+           std::to_string(redirect.usBuildNumber) + "." + std::to_string(redirect.usRevisionNumber);
+}
+
+bool IsUnsafeAccessorTypeAttribute(const ComPtr<IMetaDataImport2>& metadata_import, mdToken constructor)
+{
+    // A constructor is a MethodDef when the attribute is defined in this module and a MemberRef when it is imported.
+    mdToken parent = mdTokenNil;
+    HRESULT hr     = E_FAIL;
+
+    if (TypeFromToken(constructor) == mdtMethodDef)
+    {
+        hr = metadata_import->GetMethodProps(constructor, &parent, nullptr, 0, nullptr, nullptr, nullptr, nullptr,
+                                             nullptr, nullptr);
+    }
+    else if (TypeFromToken(constructor) == mdtMemberRef)
+    {
+        hr = metadata_import->GetMemberRefProps(constructor, &parent, nullptr, 0, nullptr, nullptr, nullptr);
+    }
+
+    if (FAILED(hr))
+    {
+        return false;
+    }
+
+    WCHAR attr_name[kNameMaxSize]{};
+    ULONG attr_name_length = 0;
+    if (TypeFromToken(parent) == mdtTypeDef)
+    {
+        hr = metadata_import->GetTypeDefProps(parent, attr_name, kNameMaxSize, &attr_name_length, nullptr, nullptr);
+    }
+    else if (TypeFromToken(parent) == mdtTypeRef)
+    {
+        hr = metadata_import->GetTypeRefProps(parent, nullptr, attr_name, kNameMaxSize, &attr_name_length);
+    }
+    else
+    {
+        return false;
+    }
+
+    return SUCCEEDED(hr) && attr_name_length > 0 && WSTRING(attr_name) == unsafe_accessor_type_attribute_name;
+}
+
+} // namespace
+
+bool HasUnsafeAccessorTypeAttribute(const ComPtr<IMetaDataImport2>& metadata_import)
+{
+    mdTypeDef attribute_type = mdTypeDefNil;
+    return metadata_import->FindTypeDefByName(unsafe_accessor_type_attribute_name.c_str(), mdTokenNil,
+                                              &attribute_type) == S_OK;
+}
+
+bool TryRewriteUnsafeAccessorTypeName(const std::string&                                       type_name,
+                                      std::unordered_map<WSTRING, AssemblyVersionRedirection>& assembly_redirects,
+                                      std::string&                                             rewritten_type_name,
+                                      WSTRING&                                                 redirected_assembly_name)
+{
+    const auto assembly_separator = FindAssemblySeparator(type_name);
+    if (assembly_separator == std::string::npos)
+    {
+        // A type without an assembly qualifier cannot participate in assembly redirection.
+        return false;
+    }
+
+    const auto assembly_begin = SkipWhitespace(type_name, assembly_separator + 1);
+    const auto assembly_end   = type_name.find(',', assembly_begin);
+    const auto assembly_name_end =
+        TrimmedEnd(type_name, assembly_begin, assembly_end == std::string::npos ? type_name.size() : assembly_end);
+    if (assembly_name_end == assembly_begin)
+    {
+        return false;
+    }
+
+    const auto assembly_name = type_name.substr(assembly_begin, assembly_name_end - assembly_begin);
+    auto       redirect      = assembly_redirects.find(ToWSTRING(assembly_name));
+    if (redirect == assembly_redirects.end())
+    {
+        redirect = std::find_if(assembly_redirects.begin(), assembly_redirects.end(),
+                                [&assembly_name](const auto& entry)
+                                { return EqualsIgnoreCase(ToString(entry.first), assembly_name); });
+    }
+    if (redirect == assembly_redirects.end())
+    {
+        return false;
+    }
+
+    auto qualifier = assembly_end;
+    while (qualifier != std::string::npos)
+    {
+        // Preserve qualifier spelling and order; only replace the value of an existing Version qualifier.
+        const auto qualifier_begin = SkipWhitespace(type_name, qualifier + 1);
+        const auto qualifier_end   = type_name.find(',', qualifier_begin);
+        const auto equals          = type_name.find('=', qualifier_begin);
+        if (equals != std::string::npos && (qualifier_end == std::string::npos || equals < qualifier_end))
+        {
+            const auto key_end = TrimmedEnd(type_name, qualifier_begin, equals);
+            if (EqualsIgnoreCase(type_name.substr(qualifier_begin, key_end - qualifier_begin), "Version"))
+            {
+                const auto value_begin = SkipWhitespace(type_name, equals + 1);
+                const auto value_end =
+                    TrimmedEnd(type_name, value_begin,
+                               qualifier_end == std::string::npos ? type_name.size() : qualifier_end);
+                ASSEMBLYMETADATA existing_version{};
+                if (!TryParseVersion(type_name.substr(value_begin, value_end - value_begin), existing_version))
+                {
+                    // Do not guess at malformed input; leave it for the runtime to reject or resolve.
+                    return false;
+                }
+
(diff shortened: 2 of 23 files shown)


src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,519 @@
...
+size_t FindAssemblySeparator(const std::string& type_name)
+{
+    // The first unescaped comma at bracket depth zero starts the outer assembly qualifier. Ignore commas inside generic
+    // arguments and escaped type-name commas.
+    size_t bracket_depth = 0;
+    bool   escaped       = false;
+
+    for (size_t i = 0; i < type_name.size(); i++)
+    {
+        const auto current = type_name[i];
```
eftiquar, line 97: Consider dropping the escaped state machine. When a \ is encountered, advance i to skip the escaped character and let the loop increment continue from the following character. This removes the extra parser state and makes the escape handling easier to follow.

if (current == '\\')
{
if (++i == type_name.size())
{
break;
}
 
continue;
}

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,519 @@
...
+{
+    while (position < value.size() && std::isspace(static_cast<unsigned char>(value[position])) != 0)
+    {
+        position++;
+    }
+
+    return position;
+}
+
+size_t TrimmedEnd(const std::string& value, size_t begin, size_t end)
```
eftiquar, line 68: This is minor suggestion - end can never become < begin, and hence this little change makes it cleaner - isspace returns bool 
 
while (end != begin &&
std::isspace(static_cast<unsigned char>(value[end - 1])))
{
--end;
}

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,519 @@
...
+        return false;
+    }
+
+    BYTE       encoded_length[4]{};
+    const auto encoded_size = CorSigCompressData(static_cast<ULONG>(length), encoded_length);
+    blob.insert(blob.end(), encoded_length, encoded_length + encoded_size);
+    return true;
+}
+
+size_t SkipWhitespace(const std::string& value, size_t position)
```
eftiquar, line 58: Same as below in TrimmedEnd - 
while (position != value.size() && std::isspace(static_cast<unsigned char>(value[position])))
{
++position;
}

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,519 @@
...
+{
+    if (data == nullptr || size == 0)
+    {
+        return false;
+    }
+
+    uint32_t   decoded_length = 0;
+    uint32_t   encoded_size   = 0;
+    const auto result         = CorSigUncompressData(data, static_cast<DWORD>(size), &decoded_length, &encoded_size);
+    length                    = static_cast<ULONG>(decoded_length);
```
eftiquar, line 39: length and bytes_read are updated even when CorSigUncompressData() fails. Please return false without touching out params upon failure.

eftiquar (changes requested):

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,519 @@
...
+    HCORENUM          attribute_enum  = nullptr;
+    mdCustomAttribute attributes[32];
+    ULONG             applicable_count = 0;
+    ULONG             updated_count    = 0;
+    HRESULT           hr               = S_OK;
+
+    while (hr == S_OK)
+    {
+        ULONG attribute_count = 0;
+        hr = metadata_import->EnumCustomAttributes(&attribute_enum, mdTokenNil, mdTokenNil, attributes, 32,
```
iskiselev, line 416: We probably can resolve ctor methodDef / methodRef first and after it EnumCustomAttributes using it only. It should be more effective than check each attribute individually.
Technically, there may be more than 1 TypeRefs, so it may require EnumTypeDefs / EnumTypeRefs first. It may require more thoughts which will be more effective.
We can try to use `FindTypeRef`, but it may be tricky to identfy resolution scope and may not correctly account type redirections.

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,519 @@
...
+                // A ParamDef should always identify its declaring method; failure indicates invalid or unreadable
+                // metadata, so skip this attribute without jeopardizing module loading.
+                Logger::Warn("UnsafeAccessorTypeAttributeUpdater: failed to get the declaring method in ",
+                             module_metadata.assemblyName, ", HRESULT=", HResultStr(owner_hr));
+                continue;
+            }
+
+            const void* unsafe_accessor_blob      = nullptr;
+            ULONG       unsafe_accessor_blob_size = 0;
+            if (metadata_import->GetCustomAttributeByName(method, unsafe_accessor_attribute_name.c_str(),
```
iskiselev, line 468: How unsafe_accessor_blob / unsafe_accessor_blob_size  are different from blob_data / blob_size we already resolved?

iskiselev: Why do we need it if we never use them after?

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,519 @@
...
+
+bool TryRewriteUnsafeAccessorTypeName(const std::string&                                       type_name,
+                                      std::unordered_map<WSTRING, AssemblyVersionRedirection>& assembly_redirects,
+                                      std::string&                                             rewritten_type_name,
+                                      WSTRING&                                                 redirected_assembly_name)
+{
+    const auto assembly_separator = FindAssemblySeparator(type_name);
+    if (assembly_separator == std::string::npos)
+    {
+        // A type without an assembly qualifier cannot participate in assembly redirection.
```
iskiselev, line 229: We may add few words to explain, that it either type in same assembly or mscorlib, neither we redirect - or something like that.

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,519 @@
...
+                                                     [](char left_char, char right_char)
+                                                     {
+                                                         return std::tolower(static_cast<unsigned char>(left_char)) ==
+                                                                std::tolower(static_cast<unsigned char>(right_char));
+                                                     });
+}
+
+size_t FindAssemblySeparator(const std::string& type_name)
+{
+    // The first unescaped comma at bracket depth zero starts the outer assembly qualifier. Ignore commas inside generic
```
iskiselev, line 90: Hm, are we doing it deep in the generics? As for generic, we may also need patch all generic arguments:
```
List<string>
System.Collections.Generic.List`1[[System.String, System.Private.CoreLib, Version=10.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], System.Private.CoreLib, Version=10.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e
```
We have version twice there. And assemblies in that version may be different for each type and each generic arguments any level deep.

alexeypukhov: Thank you for noticing this. I forgot to mention in the description that generic nested types are skipped by this solution and will require some more work from me!

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,519 @@
...
+        else if (current == ',' && bracket_depth == 0)
+        {
+            return i;
+        }
+    }
+
+    return std::string::npos;
+}
+
+bool TryParseVersion(const std::string& value, ASSEMBLYMETADATA& version)
```
eftiquar, line 125: This code is correct and does what it is supposed to do. However, it requires readers to reason about separator positions, index arithmetic, digit parsing, and overflow detection simultaneously.
For a dot-separated version string, an iterator/range-based approach using std::find to locate component boundaries and std::from_chars to parse each component feels more idiomatic than manually tracking indices, accumulating digits, and performing overflow checks inline. This is for future readers who stumble upon this. 
Also a comment illustrating typical version string will help.

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,519 @@
...
+        }
+
+        for (ULONG i = 0; i < attribute_count; i++)
+        {
+            mdToken     owner       = mdTokenNil;
+            mdToken     constructor = mdTokenNil;
+            const void* blob_data   = nullptr;
+            ULONG       blob_size   = 0;
+            const auto  attribute_hr =
+                metadata_import->GetCustomAttributeProps(attributes[i], &owner, &constructor, &blob_data, &blob_size);
```
eftiquar, line 432: This method is primarily concerned with rewriting UnsafeAccessorTypeAttribute metadata, but much of the code is occupied by the mechanics of metadata enumeration (HCORENUM, batching, EnumCustomAttributes, GetCustomAttributeProps, cleanup, etc.). Consider introducing an RAII-backed custom attribute range/iterator abstraction and passing the UnsafeAccessorTypeAttribute constructor token into the enumeration. That would allow the metadata layer to perform the filtering and lifecycle management while the updater focuses on the rewrite logic. The resulting code would be both more idiomatic C++ and more aligned with the domain intent.


Existing clr_helpers.h already has enumeration model that we can use for reference, adding custom attribute enumerator will be nice addition. 

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/90911c2744228ea00753aa43b1cf83634ff30a6c/src/OpenTelemetry.AutoInstrumentation.Native/clr_helpers.h#L48-L179


Enumerator<mdCustomAttribute> EnumCustomAttributes(
    const ComPtr<IMetaDataImport2>& metadata_import,
    mdToken tkParent = mdTokenNil,
    mdToken tkType   = mdTokenNil);

1. The last tkType argument is the custom-attribute constructor token, which will filter only the desired UnsafeAccessorTypeAttribute  attributes. 

2. Enumerator<mdCustomAttribute> will own and close the HCORENUM;

Please let me know if you have any follow up questions.

eftiquar (changes requested):

docs/assembly-conflict-resolution.md
```diff
@@ -352,12 +362,16 @@ architectures.
...
+The native profiler processes assembly references and `UnsafeAccessorType`
+attribute values in module-load order. CoreLib attributes are considered during
+CoreLib loading; in each later module, `AssemblyRef` entries are considered
+before attribute values. If an earlier reference has been
+redirected to the instrumentation's version and a later reference requests a
+**higher** version than the instrumentation ships, the profiler cannot safely
+revise the earlier metadata. This is logged as an error. An attribute in
+`System.Private.CoreLib` can be that first redirection because CoreLib loads
+before application modules. In practice this is rare because the
+instrumentation ships recent versions of its dependencies.
```
iskiselev, line 374: In fact, for most .NET application it is safe situation - it usually means that TPA already have higher-version assembly, and it will win in managed resolver - or even when custom resolver is skipped.

iskiselev: BTW, can we get info about TPA list from native code? We may not need to apply any redirection at all if target assembly is the assembly we will select.

alexeypukhov: 1. Good point. The documentation phrased the profiler’s ordering limitation too broadly. The profiler cannot retroactively change metadata that was already rewritten, but normal runtime resolution may still satisfy the later request when an appropriate version is available in the TPA list. I will improve that wording.
2. I am not aware of a profiler API that exposes the TPA list. A quick search suggests it may be available through the active `hostfxr` context, but integrating that safely would require additional cross-platform hosting work.
3. Since this would benefit the general native redirection policy for both `AssemblyRef` and `UnsafeAccessorTypeAttribute`, it goes beyond the current improvement and I suggest tracking it separately. Both redirection paths in this PR use the same version map, so a future TPA-aware policy can be applied centrally to both.

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,811 @@
...
+
+std::vector<mdToken> FindAttributeConstructors(const ComPtr<IMetaDataImport2>& metadata_import,
+                                               const WSTRING&                  attribute_name)
+{
+    std::vector<mdToken> constructors;
+    const auto           add_constructors = [&](const auto& methods)
+    {
+        for (const auto method : methods)
+        {
+            if (IsConstructor(metadata_import, method))
```
iskiselev, line 434: Ideally will be to validate that it has strictly 1 parameter and first parameter named System.String, though it is check against very edge situations currently and potential new ctor in future.

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,811 @@
...
+    }
+    return constructors;
+}
+
+bool HasCustomAttribute(const ComPtr<IMetaDataImport2>& metadata_import, mdToken owner, const WCHAR* attribute_name)
+{
+    // GetCustomAttributeByName requires blob outputs even though existence is all the caller needs.
+    const void* unused_blob      = nullptr;
+    ULONG       unused_blob_size = 0;
+    return metadata_import->GetCustomAttributeByName(owner, attribute_name, &unused_blob, &unused_blob_size) == S_OK;
```
iskiselev, line 467: Can we pass `nullptr` intead of not used `&unused_blob`, and `&unused_blob_size`?

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,811 @@
...
+        if (FAILED(owner_hr))
+        {
+            // A ParamDef should always identify its declaring method; failure indicates invalid or unreadable
+            // metadata, so skip this attribute without jeopardizing module loading.
+            Logger::Warn("UnsafeAccessorTypeAttributeUpdater: failed to get the declaring method in ",
+                         module_metadata.assemblyName, ", HRESULT=", HResultStr(owner_hr));
+            continue;
+        }
+
+        if (!HasCustomAttribute(metadata_import, method, unsafe_accessor_attribute_name.c_str()))
```
iskiselev, line 760: Strictly speaking that check may be not required - we can just update all UnsafeAccessorTypeAttribute even without validation that owner marked with UnsafeAccessorAttribute - but that attributes are not processed by runtime.
No need to change.

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,811 @@
...
+                redirect->second.ulRedirectionCount++;
+            }
+        }
+        updated_count++;
+    }
+
+    if (applicable_count > 0)
+    {
+        // Keep attributes that do not target the current redirection map visible in debug logs.
+        Logger::Debug("UnsafeAccessorTypeAttributeUpdater: found ", applicable_count,
```
iskiselev, line 801: I'd always write only this message. Probably with either debug/info level based on `updated_count > 0`

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,811 @@
...
+// Reflection type names come from module metadata. Bound recursive generic parsing so malformed metadata cannot
+// exhaust the profiler's native stack; 64 is well above practical generic nesting
+constexpr size_t max_generic_nesting_depth = 64;
+
+bool FindAssemblyReferences(std::string_view                      type_name,
+                            Span                                  type,
+                            size_t                                generic_nesting_depth,
+                            std::vector<ParsedAssemblyReference>& references);
+
+bool FindGenericArguments(const std::string_view                type_name,
```
eftiquar, line 278: The syntax of generic types embeds backtick escaped count specifying the number of arguments.
We can leverage that to make this code readable - though current code is correct, it is hard to follow. 

> The backtick is an unambiguous marker that the current type is generic, and the digits following it give the exact number of arguments. A grammar-aware flow could be:
>
> ```text
> ParseTypeSpec
>   → parse type name
>   → if `N is present:
>       position cursor at '['
>       ParseGenericArguments(cursor, N)
>   → parse array/pointer/byref suffixes
>   → parse assembly qualification
> ```
>
> `ParseGenericArguments` would consume the opening bracket, parse exactly `N` arguments by recursively calling `ParseTypeSpec`, consume separators, then consume the matching closing bracket and leave the cursor advanced.
>
> This would make the implementation easier to reason about and distinguish generic argument brackets from array suffixes. It would also allow malformed argument counts to be rejected explicitly. The `TypeName` is ordinary UTF-8 text inside the compressed `SerString`; only the string length is compressed, so the generic arity can be parsed directly.
>
> I’m suggesting this primarily for readability and robustness rather than performance. The current recursive approach is close in spirit, but using the arity as the grammar guide would reduce the cognitive load of following the parser.

eftiquar (changes requested):

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
iskiselev: Unfortunatly backtick escaped count is not required per ECMA-335 / CLR. It is recommended for CLS-compliance. We can't create alogrithm based on backtick presence.

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,811 @@
...
+    // A null SerString (FF) is not a type name and cannot be redirected.
+    if (blob == nullptr || blob_size < custom_attribute_prolog_size + 1 + named_argument_count_size ||
+        blob[0] != 0x01 || blob[1] != 0x00 || blob[custom_attribute_prolog_size] == 0xFF)
+    {
+        return false;
+    }
+
+    ULONG      string_length  = 0;
+    size_t     length_bytes   = 0;
+    const auto remaining_size = static_cast<size_t>(blob_size) - custom_attribute_prolog_size;
```
iskiselev, line 651: We know, that at least named_argument_count_size will be also not there.
```suggestion
    const auto remaining_size = static_cast<size_t>(blob_size) - custom_attribute_prolog_size - named_argument_count_size;
```

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,811 @@
...
+    if (!TryReadPackedLength(blob + custom_attribute_prolog_size, remaining_size, string_length, length_bytes))
```
iskiselev, line 652: I'd probably call `CorSigUncompressData` directly - we do minimal processing in `TryReadPackedLength`, but first is known to everyone - and econd requires to look inside.

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,811 @@
...
+
+        const auto version_comparison = redirect->second.CompareToAssemblyVersion(reference.parsed_version);
+        if (version_comparison >= 0)
+        {
+            continue;
+        }
+
+        // Never lower a requested version. Before any redirect is committed, let this higher request raise the shared
+        // target; afterward, changing it could disagree with metadata already rewritten.
+        if (redirect->second.ulRedirectionCount == 0)
```
iskiselev, line 537: We have very simialr code in cor profiler around line 430. Can we extract some common helper?

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,811 @@
...
+        if (redirect != assembly_redirects.end())
+        {
+            resolved_references.push_back({&reference, redirect});
+        }
+    }
+
+    // First establish the final target for every assembly. Delay committing a raised target until every reference in
+    // this attribute has been considered, so repeated references use the highest requested version consistently.
+    std::vector<RedirectIterator> raised_redirects;
+    for (const auto& resolved_reference : resolved_references)
```
iskiselev, line 520: Do we need such complexity?
Original attribute most probably use the same version all the time for the same assembly. Even if it is not true, TPA most probably still resolve only one version.
And we can play with that consistency game wider - why are we not requiring to process all attributes / assemblyrefs together so that we select target consistently?
I'd suggest to simplify it. But looks like nothing wrong with current alorithm.

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
```diff
@@ -0,0 +1,811 @@
...
+namespace
+{
+
+const WSTRING unsafe_accessor_type_attribute_name = WStr("System.Runtime.CompilerServices.UnsafeAccessorTypeAttribute");
+const WSTRING unsafe_accessor_attribute_name      = WStr("System.Runtime.CompilerServices.UnsafeAccessorAttribute");
+
+// A custom-attribute SerString prefixes its UTF-8 bytes with an ECMA-335 compressed unsigned integer. Use the CLR's
+// codec for that standard 1/2/4-byte representation, but guard its reader because it examines the first byte before
+// validating the supplied length.
+bool TryReadPackedLength(const BYTE* data, size_t size, ULONG& length, size_t& bytes_read)
```
iskiselev, line 35: We mix in one file / class attribute finding, rewrite, string processing - probably it can be split.
Parsing of assemblies and their version can be moved to separate type/file. May be some other functions - like modyfying type string. Probably something else.

src/OpenTelemetry.AutoInstrumentation.Native/unsafe_accessor_type_attribute_updater.cpp
eftiquar: I looked into the backtick - even if it is optional the suggestion for algorithmic improvement still stands. The code correctness is not the issue - it is readability -  

> Conceptual sketch - that mirrors the grammar of type format

ParseTypeSpec
  → ParseTypeName
  → if '[':
       ParseArraySuffix
       or
       ParseGenericArguments
  → Parse pointer/byref suffixes
  → Parse assembly qualification


```
ParseTypeSpec(cursor)
{
    ParseTypeName(cursor);

    auto arity = TryParseBacktickArity(cursor); // optional
    if (Peek(cursor) == '[')
    {
        if (LooksLikeArraySuffix(cursor))
            ParseArraySuffix(cursor);
        else
            ParseGenericArguments(cursor, arity);
    }

    ParsePointerOrByRefSuffixes(cursor);
    ParseAssemblyQualification(cursor);
}

```
> ParseGenericArguments has two modes:


```
ParseGenericArguments(cursor, optional<size_t> expectedCount)
{
    Consume('[');
    size_t actualCount = 0;
    do
    {
        ParseTypeArgument(cursor); // recursively calls ParseTypeSpec
        actualCount++;
    }

    while (Consume(','));
    Consume(']');
    return !expectedCount || actualCount == *expectedCount;
}
```

PR #4940, opened 2026-03-19
Stop promoting resource attributes from appSettings to environment variables

matt-hensley:
## Why

Follow-up to #4904

When running on IIS, multiple services can share the same Application Pool and thus process. The values for `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES` from web.config/app.config into process environment variables during startup of the first service in the App Pool. This caused all services in the same app pool to have the same resource attributes.

## What

Stops promoting `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES` from appSettings to environment variables. Instead, values from web.config/app.config on .NET Framework are used to build the Resource. 

Precedence ordering, from lowest to highest:
  1. `OTEL_RESOURCE_ATTRIBUTES` from env var
  2. `OTEL_SERVICE_NAME` from  env var
  3. `OTEL_RESOURCE_ATTRIBUTES` from web.config/app.config
  4. `OTEL_SERVICE_NAME` from web.config/app.config

## Checklist


test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/EnvironmentInitializerTests.cs
```diff
@@ -75,4 +77,26 @@ public void Noop_WhenSettingIsNonOTel()
             Environment.SetEnvironmentVariable(NonOtelVariableName, null);
         }
     }
+
+    [Theory]
+    [InlineData(OtelServiceVariableName)]
+    [InlineData(OtelResourceAttributesVariableName)]
+    public void Noop_WhenSettingResources(string variableName)
+    {
+        try
```
martincostello, line 75: Might be useful to copy over [EnvironmentVariableScope](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/test/Shared/EnvironmentVariableScope.cs) to make the set and undo pattern less verbose.

matt-hensley: Originally had that included here but decided to split it out to keep the diff readable.

test/OpenTelemetry.AutoInstrumentation.Tests/Configurations/SettingsTests.cs
```diff
@@ -570,4 +576,449 @@ internal void ResourceSettings_ResourceDetectors_SupportedValues(string resource
...
+
+    [Fact]
+    internal void ResourceSettings_LoadEnvVar_AppSettingsSourceTakesPriorityOverEnvVarSource()
+    {
+        const string envVarValue = "service-name-from-env-var";
+        const string appSettingsValue = "service-name-from-app-settings";
+
+        try
+        {
+            Environment.SetEnvironmentVariable(ConfigurationKeys.ServiceName, envVarValue);
```
martincostello, line 606: Use the environment variable scope class here?

iskiselev:
I think this change needs a broader SIG-level decision on app.config/web.config semantics.

This PR introduces a backward-incompatible behavior change for a narrow edge case: non-IIS multi-AppDomain .NET Framework applications configured through app.config. Previously, appSettings were promoted into process-level environment variables, so values from the first initialized AppDomain could become visible to later AppDomains. With this change, resource configuration becomes AppDomain-local. I think that scenario is narrow enough that we may still decide it does not require a major version bump, but that should be an explicit SIG decision.

There is also a backward-incompatible IIS case if anyone relied on the first application in a shared AppPool configuring later applications. I would classify that as unintended behavior / a bug, so changing it seems reasonable without a major version bump, but again we should be explicit about it.

My main concern is that the current implementation leaves us in a mixed semantic model:
- `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES` from app.config/web.config become AppDomain-local and are merged specially.
- Other `OTEL_*` appSettings are still promoted to process-level environment variables and can leak globally across applications/AppDomains.

If we accept the compatibility break, I think the cleaner model is to stop promoting all managed `OTEL_*` appSettings into environment variables. `AppSettingsConfigurationSource` should already allow our managed `Settings` configuration system to read app.config/web.config as a fallback source. Profiler/bootstrap settings and AppDomain load strategy are exceptions and should remain process/env-level, which would be consistent with YAML configuration limitations.

There are a few design questions we should settle before merging:
- How should we decide whether YAML configuration is enabled, and which YAML file path to use, if those values are provided in app.config/web.config? Today this likely still needs env-level handling.
- YAML `${ENV_VAR}` substitution currently means process environment variables only. We can discuss extending that later, but I would not couple it to this PR.
- Do we really want attribute-level merging for `OTEL_RESOURCE_ATTRIBUTES`, or should app.config/web.config simply be a fallback/override source like other keys? I am not against merging, but it should be an intentional SIG decision.
- What precedence do we want: environment variables over app.config/web.config, or local app.config/web.config over environment variables? The former is more backward compatible. The latter makes sense for IIS/shared AppPool locality, but is another compatibility change.

Finally, this overlaps with #5055. If the plan is to deprecate app.config/web.config configuration once YAML matures, we should decide whether this PR is still worth doing. Improving this behavior now and then deprecating it in 2.0 may send a confusing signal. If we proceed with this PR as the chosen direction, then #5055 may need to be closed or reframed.

docs/config.md
```diff
@@ -13,7 +13,11 @@ with environment variables taking precedence over `App.config` or `Web.config` f
 
     For an application running on .NET Framework, you can use a web configuration
     file  (`web.config`) or an application configuration file (`app.config`) to
-    configure the `OTEL_*` settings.
+    configure the `OTEL_*` settings. For resource configuration, `OTEL_RESOURCE_ATTRIBUTES`
+    and `OTEL_SERVICE_NAME` can come from environment variables or from these
```
iskiselev, line 17: Why is it specific only to that 2 keys?

docs/config.md
```diff
@@ -85,11 +89,22 @@ for more details.
 
 ### Resource attributes
 
+On .NET Framework, `OTEL_RESOURCE_ATTRIBUTES` and `OTEL_SERVICE_NAME` may be
```
iskiselev, line 92: Should we document that `OTEL_RESOURCE_ATTRIBUTES` will be merged?

iskiselev (changes requested):

docs/config.md
matt-hensley: Those are the two env vars called out as supported resource attributes in the doc:

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/v1.16.0-beta.1/docs/config.md#resource-attributes

nrcventura:
> Finally, this overlaps with https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/5055. If the plan is to deprecate app.config/web.config configuration once YAML matures, we should decide whether this PR is still worth doing. Improving this behavior now and then deprecating it in 2.0 may send a confusing signal. If we proceed with this PR as the chosen direction, then https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/5055 may need to be closed or reframed.

That is an important callout. My preference is to align on the standard declarative configuration option, and deprecating this app.config/web.config workaround. Even if we deprecate the app.config/web.config option and standardize on the YAML option, we are still left with a less than ideal option for those hosting multiple applications in a single app pool.

I'm leaning towards saying that processes with multiple app domains will be treated as single applications. One of my big reasons for leaning in that direction is that it minimizes the confusion that can occur for settings that apply to the native profiler (which is only loaded once) and the managed code (which can be loaded multiple times). It also leaves the option open to allowing an app.config/web.config setting that can be used to point to individual YAML files in the future.

I'm also not convinced that processes with multiple app domains should be treated as separate applications. In the single app pool case, if any of the web apps assigned to that single app pool crashes, the whole process will crash. If you monitor each of those web apps independently you lose the visibility into understanding why web app 1 stopped reported when web app 2 crashed. That runtime coupling is important when you are trying to understand the health of your system. By separating it out, you will lose the visibility into that runtime coupling. If these really are separate apps, and you do not want a crash in one app to affect the other, separate app pools need to be used.

> How should we decide whether YAML configuration is enabled, and which YAML file path to use, if those values are provided in app.config/web.config? Today this likely still needs env-level handling.

Ideally we just move to a model where configuration is managed at the process level, which is basically what we have today (even with the current app.config/web.config support).

> YAML ${ENV_VAR} substitution currently means process environment variables only. We can discuss extending that later, but I would not couple it to this PR.

Ideally this can remain a process level concept. If we decide in the future to support a YAML file per app domain, this process level environment variable substitution can still be useful, but can also be a source of confusion. I agree that this should be handled as part of this PR.

> Do we really want attribute-level merging for OTEL_RESOURCE_ATTRIBUTES, or should app.config/web.config simply be a fallback/override source like other keys? I am not against merging, but it should be an intentional SIG decision.

Merging does not seem like the most flexible/understandable behavior to me. I think that it is much easier to explain and understand a configuration precedence order. Supporting merging can lead to awkward scenario where one of the resource attributes defined in the environment variable is not applicable to one of the app domains. To address that in the merging behavior scenario, you would need to remove that attribute from the environment variable and then push it down to relevant configs for each app domain. That would increase the number of "apps" that need to be validated. It's also much easier to reason about what setting should be applied when everything is defined in one place.

If we restrict things to just process-level  configuration, the question about merging is no longer relevant.

> What precedence do we want: environment variables over app.config/web.config, or local app.config/web.config over environment variables? The former is more backward compatible. The latter makes sense for IIS/shared AppPool locality, but is another compatibility change.

I think that local app.config/web.config having a higher precedence makes the most sense. However, I think that we should just be using the YAML configuration instead and keep configuration at the process level.

iskiselev:
I think we should split this into a few separate design questions.

**1. Should we support per-AppDomain managed configuration?**

I agree that there are strong reasons to treat a multi-AppDomain process as one application, especially for native profiler/bootstrap settings. The native profiler is process-level, so trying to make every setting AppDomain-local would create confusing boundaries.

At the same time, today we already initialize the managed agent once per AppDomain. Even with environment-variable configuration, different AppDomains can observe different values if the parent AppDomain changes process environment variables before creating a secondary AppDomain. To make multi-AppDomain applications always behave as one application, we would likely need to snapshot/move configuration into the profiler or another process-level owner. That seems like a larger and more complex design.

For IIS specifically, I think the most reasonable default is to treat each ASP.NET application as an independent managed application, even if multiple apps share an AppPool. The crash/runtime coupling is real, but AppPool sharing is also often a hosting/deployment detail. If every app in the pool gets the same `service.name`, users lose the ability to understand which web application is producing telemetry.

My preferred model would be:
- profiler/bootstrap/AppDomain loading settings remain process-level;
- managed SDK/instrumentation settings can be AppDomain-local;
- IIS uses app-local configuration by default, with an opt-out mechanism;
- non-IIS multi-AppDomain applications keep env-based / global YAML resolution by default, with an opt-in path for per-domain configuration if we decide to support one.

**2. Should we support app.config/web.config as an alternative to environment variables? If yes, what precedence should it have?**

I strongly value the ability to configure ASP.NET applications through `Web.config`. For web-hosted .NET Framework applications, `Web.config` feels like the closest analogue to environment variables: it is the app-local configuration surface users already have. Setting environment variables for .NET Framework AppPool scenarios is not very ergonomic, and often requires machine/server-level configuration.

I would support removing `Web.config`/`App.config` configuration only if we were also planning to remove environment-variable configuration entirely, which I do not expect will ever happen.

That said, the current implementation model is the part I strongly think we should change. Today, app.config/web.config settings are softly merged into process-level environment variables based on whichever AppDomain initializes first. That leaks configuration between separate logical applications. I think we should fix that either by making managed app.config/web.config settings AppDomain-local, or by removing/deprecating app.config/web.config support as suggested in #5055. Keeping the current first-loaded-app-wins behavior seems like the worst option.

If we keep app.config/web.config support, I would prefer appSettings first and environment variables second for AppDomain-local managed settings. That lets machine/AppPool env provide defaults while the app-local config overrides local values. I recognize that appSettings-first is a compatibility change. Question 6 below is where I think we should decide whether the compatibility surface is small enough for a minor release and how much of the old behavior we should preserve outside IIS/per-domain mode.

App.config/web.config is one possible design for per-domain configuration. Per-domain YAML, YAML templates, or some other vessel may also be viable either as alternatives or additional options.

**3. Should we support per-AppDomain YAML configuration?**

Even if we eventually deprecate broad app.config/web.config based configuration, I think it may still be valuable to allow app.config/web.config to point to the YAML config file. That gives IIS apps a local way to choose their YAML config without requiring AppPool/machine-level environment changes.

**4. Should YAML interpolation support app.config/web.config values?**

YAML `${ENV_VAR}` substitution can remain process-env based by default. But if we support per-AppDomain YAML, a follow-up could add an app-config resolver for YAML templating, so a shared YAML template can pull selected values from each app's Web.config/App.config.

With that model, we could emulate separate per-application configuration through YAML while still moving toward standard declarative config.

**5. Should resource attributes be merged or selected by normal source precedence?**

I do not have a strong opinion here, but I would suggest not merging for now. A normal source precedence model, where one source overrides the other, is simpler to explain, understand, implement, and document. If we later get user feedback asking for resource-attribute merging, we can add it deliberately.

**6. Can we do this in a minor release, or does it require a major release?**

I think we may be able to treat this as a minor-release bug fix if the behavior is:
- opt-out for IIS;
- opt-in for non-IIS multi-AppDomain applications.

The backward-incompatible surface would be limited. The main affected cases would be:
- IIS applications that configure only some apps through Web.config and intentionally rely on a specific first-app-load order to configure the whole AppPool;
- Windows Services or other .NET Framework apps that rely on app.config values being promoted to process environment variables and then reused by later AppDomains.

The IIS case looks like unintended configuration leakage between separate logical applications, so I think it is reasonable to classify that as a bug fix.

We can probably reduce the compatibility risk even further by preserving the old app.config-to-env promotion only when per-domain configuration is not enabled and we are in the primary AppDomain. That would keep the old behavior for most non-IIS single-domain and primary-domain scenarios, while allowing IIS/per-domain configuration to avoid leaking app-local settings globally.

nrcventura:
In the SIG meeting today we discussed the 6 proposed questions. This is a summary of that discusion.

1. Should we support per-AppDomain managed configuration? Yes. This is the most compatible choice based on the current behavior.
2. Should we support app.config/web.config as an alternative to environment variables? If yes, what precedence should it have? Yes, but we did not discuss the precedence.
3. Should we support per-AppDomain YAML configuration? Yes, but not necessary right away. It's also possible that the current implementation partially supports this.
4. Should YAML interpolation support app.config/web.config values? Seems like a reasonable approach but not necessary at this point in time.
5. Should resource attributes be merged or selected by normal source precedence? They should not be merged.
6. Can we do this in a minor release, or does it require a major release? Based on the answers to the other questions, this can be done as a minor release.

Other related discussion points:

- We will need documentation to help clarify configuration behavior in mult-app domain scenarios. For example, not all settings can be controlled via app/web config.
- Continuous profiling is functionality that crosses the app domain boundary and is another potential source of confusion
- To prevent confusion we highly recommend that each logical application runs in a separate process. The functionality we are discussing here should really only apply to cases where either it is not possible to split the logical apps into their own processes, or each app domain is part of the same logical app.

Issue #3714, opened 2024-10-16
NetFX dependency redirections fails in some cases


RassK:
### What

Runtime dependency redirection mechanism is failing in some cases. 

Example affected library: System.Diagnostics.DiagnosticSource

Affected version: OpenTelemetry.AutoInstrumentation v1.8.0 (ships with System.Diagnostics.DiagnosticSource v8.0.0)

App 1 (NetFxConsole):
.NET FX 4.8 Console app (very simple, creates one manual Activity + makes a HttpClient call)
ref: System.Diagnostics.DiagnosticSource v8.0.1
result: works

App 2 (SimpleFrameworkApp):
ASP.NET 4.8 (template)
ref: System.Diagnostics.DiagnosticSource v8.0.1
result: 
Could not load file or assembly 'System.Diagnostics.DiagnosticSource, Version=8.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified.

Native logs attached:
[otel-dotnet-auto-5888-NetFxConsole-Native.log](https://github.com/user-attachments/files/17398621/otel-dotnet-auto-5888-NetFxConsole-Native.log)
[otel-dotnet-auto-7856-w3wp-Native.log](https://github.com/user-attachments/files/17398622/otel-dotnet-auto-7856-w3wp-Native.log)

RassK:
Let's make sure we don't release this before a proper fix:
https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3710

This is a very good test example at the moment.

pjanotti:
> Not sure if runtime_information_.is_desktop() doesn't correctly identify IIS apps?
It is based on https://learn.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/icorprofilerinfo3-getruntimeinformation-method so unless that is broken it should correctly identify as "desktop" meaning .NET Fx.

> Does anyone know how important is the initial skiplist for performance at all?
I'm not aware of any measurement of that. Anyway, we need to ensure first that we are correct we then can take care of optimizing if that becomes an issue. This should affect start up time and should be done only once per app domain IIRC.

nrcventura:
[...]

PR #4970, opened 2026-04-03
Fix/netfx entrypoint assembly redirection

### LINKED ISSUE #4905: Assembly version redirect fails for entry point on .NET Framework
Assembly version redirects requires assembly either be installed on GAC or our assembly resolver to be installed.
We install assembly resolver on first called method in customer assembly.
If the same method (or anything inclined in it) uses types from redirected assemblies - it will be to late, as assembly should be resolved before methods executed.
We may try to find some safe point in mscorlib that is called right before customer code - similar to LoaderHook, but in .Net Framework.

aviralgarg05:
## Why

On .NET Framework we currently register the managed loader from the first instrumented method body we see in an AppDomain. That is too late for assembly redirects in one important case: if the entry point itself, or something it touches during JIT, needs a redirected assembly, resolution happens before our loader has installed the resolver.

That is what causes the failure described in #4905.

Fixes #4905

## What

This change updates the .NET Framework startup path so the loader is installed from `System.AppDomain.nExecuteAssembly(...)` instead of relying only on the first instrumented user method in the target AppDomain.

In practice that means:

- the profiler still injects the existing loader type into `mscorlib`
- `System.AppDomain.nExecuteAssembly(RuntimeAssembly, string[])` is rewritten to call the loader before the managed entry point runs
- the existing per-method prolog injection path remains in place for the normal NetFx startup fallback behavior

That gets the resolver installed before the application's entry point executes, which closes the timing gap for assembly redirects without cloning or replacing user methods.

## Tests

Automated / verification done for this change:

- `git diff --check`
- native formatting was run with the repo's formatter
- `Shellcheck` passed on the fork
- `SdkVersionCheck` passed on the fork
- `CodeQL` passed on the fork
- `format native` passed on Linux and Windows; the macOS formatter job could not start on the fork because of runner billing limits
- the updated `ci` run passed all code-sensitive Windows jobs, including `test-build-managed (net462, windows-2022)`, `test-build-managed (net462, windows-2025)`, and both Windows `test-nuget-packages` jobs

Limitations:

- I could not run local `dotnet` build/test commands in this environment because `dotnet` is not installed here
- one `windows-2022` IIS container job failed because the GitHub runner could not connect to the Docker daemon pipe; the same job on `windows-2025` passed, which points to runner instability rather than a code issue

## Checklist


RassK:
Sounds related also to https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/3714

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -58,6 +58,75 @@ namespace trace
 
 CorProfiler* profiler = nullptr;
 
+#ifdef _WIN32
+namespace
+{
+HRESULT ModifyAppDomainExecuteAssembly(ICorProfilerInfo* info,
```
nrcventura, line 67: Can comments be added to this method to make it easier for others to understand what is being done when reading the code?

nrcventura:
The "What" section of the PR description does not match the implementation. Can you update the PR description so that it matches? For example, this code is not injecting a new method or cloning a method.

nrcventura:
I think this just needs a changelog update.

nrcventura (approved):

iskiselev:
LGTM, it will be nice to add some test for it. My idea is next:
- have some .Net Framework app that use old version of any Microsoft.Extension. something (that we use) from Main. After it write to consolve AssebmlyVersion
- Load it and validate that application not crashed and loaded expected assembly version

We have customization for IIS test that does not register assemblies in GAC, here we need something similar for non-IIS app. Stating point to look may be: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/test/IntegrationTests/AspNetTests.cs#L59 (`Gac.UseLocal`).
That test should fail without the fix.

iskiselev (approved):

iskiselev:
Have this PR been tested on any real .NET Framework application affected by the issue?
Per my test nExecuteAssembly is not called at all for normal .NET executable start. Here is list of methods that are run for me:
```
System.AppDomain.SetupDomain()
System.AppDomainSetup.SetupDefaults()
System.String.LastIndexOfAny()
System.String.Substring()
System.String.InternalSubString()
System.Buffer.Memmove()
System.String.Concat()
System.String.FillStringChecked()
System.AppDomain.set_PartialTrustVisibleAssemblies()
System.AppDomain.GetNativeHandle()
System.AppDomain.SetupFusionStore()
System.AppDomainSetup.VerifyDirList()
System.AppDomainSetup.set_DeveloperPath()
System.AppDomainSetup.SetupFusionContext()
System.AppDomainSetup.UpdateContextPropertyIfNeeded()
System.String.Equals()
System.AppDomainSetup.UpdateBooleanContextPropertyIfNeeded()
System.AppDomainSetup.UpdateByteArrayContextPropertyIfNeeded()
System.AppDomain.InitializeCompatibilityFlags()
System.CompatibilitySwitches.InitializeSwitches()
System.CompatibilitySwitches.IsCompatibilitySwitchSet()
System.Threading.Thread.GetDomain()
System.AppDomain.IsCompatibilitySwitchSet()
System.AppDomain.InitializeDomainSecurity()
System.AppDomainSetup.InternalGetApplicationTrust()
System.AppDomain.SetupDomainSecurity()
System.Security.PermissionSet..ctor()
System.AppDomain.TurnOnBindingRedirects()

**LOAD of ENTRY POINT ASSEMBLY**

System.Object.Finalize()
System.Runtime.ConstrainedExecution.CriticalFinalizerObject.Finalize()
System.Runtime.InteropServices.SafeHandle.Finalize()
Microsoft.Win32.SafeHandles.SafePEFileHandle.ReleaseHandle()
System.IntPtr.IsNull()
System.IntPtr..ctor()
System.IntPtr.op_Equality()
Microsoft.Win32.SafeHandles.SafeHandleZeroOrMinusOneIsInvalid.get_IsInvalid()
System.Runtime.InteropServices.SafeHandle.Dispose()
System.Runtime.InteropServices.SafeHandle.Dispose()
Microsoft.Win32.SafeHandles.SafePEFileHandle..ctor()
System.Security.Policy.PEFileEvidenceFactory.CreateSecurityIdentity()
System.Security.Policy.Evidence..ctor()
System.Collections.Generic.Dictionary`2..ctor()
System.Collections.Generic.EqualityComparer`1.get_Default()
System.Collections.Generic.EqualityComparer`1..cctor()
System.Collections.Generic.EqualityComparer`1.CreateComparer()
System.RuntimeType.IsAssignableFrom()
System.RuntimeType.get_UnderlyingSystemType()
System.RuntimeType.get_IsGenericType()
System.RuntimeType.get_IsEnum()
System.RuntimeType.GetBaseType()
System.Type.get_IsInterface()
System.RuntimeType..cctor()
System.Security.Policy.Evidence.get_RuntimeEvidenceTypes()
System.AppDomain.get_IsLegacyCasPolicyEnabled()
System.Collections.Generic.Dictionary`2.Insert()
System.Collections.Generic.Dictionary`2.Initialize()
System.Collections.HashHelpers.GetPrime()
System.Collections.HashHelpers..cctor()
System.String.UseRandomizedHashing()
System.Object..ctor()
System.Collections.Generic.ObjectEqualityComparer`1.GetHashCode()
System.RuntimeType.GetHashCode()
System.Collections.HashHelpers.ExpandPrime()
System.Collections.Generic.Dictionary`2.Resize()
System.Security.Policy.Evidence.QueryHostForPossibleEvidenceTypes()
System.AppDomain.get_DomainManager()
System.Threading.ReaderWriterLock.Finalize()
System.Security.Policy.AssemblyEvidenceFactory.UpgradeSecurityIdentity()
System.Security.Policy.Evidence.set_Target()
EvidenceLockHolder.Dispose()
System.AppDomain.get_HostSecurityManager()
System.Security.HostSecurityManager.get_Flags()
System.Security.HostSecurityManager.ProvideAssemblyEvidence()
System.AppDomain.SetTargetFrameworkName()
```

That behavior may be different on different OS/Arch/Framework version

PR #4987, opened 2026-04-10
Add netfx entrypoint assembly redirection coverage

### LINKED ISSUE #4982: Tests for netfx entrypoint assembly redirection
LGTM, it will be nice to add some test for it. My idea is next:
- have some .Net Framework app that use old version of any Microsoft.Extension. something (that we use) from Main. After it write to consolve AssebmlyVersion
- Load it and validate that application not crashed and loaded expected assembly version

We have customization for IIS test that does not register assemblies in GAC, here we need something similar for non-IIS app. Stating point to look may be: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/test/IntegrationTests/AspNetTests.cs#L59 (`Gac.UseLocal`).
That test should fail without the fix.

_Originally posted by @iskiselev in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/4970#issuecomment-4219267358_

aviralgarg05:
## Why

The recent .NET Framework fix for entrypoint assembly redirection changed the runtime behavior, but the existing test coverage did not exercise the failure mode that triggered the issue.

This PR adds coverage for the case where a .NET Framework entrypoint depends on a `Microsoft.Extensions.*` assembly that is also shipped by auto-instrumentation. That makes the test validate the scenario more directly and helps prevent regressions in the startup path.

Fixes #4982

## What

- update the existing assembly redirection integration test to validate the `.NET Framework` path against `Microsoft.Extensions.Logging.Abstractions`
- make the `TestApplication.AssemblyRedirection` app load that assembly from the entrypoint on `net462`
- keep the change scoped to the existing test application and test suite instead of introducing a new app or broader build changes

## Tests

Automated:
- built `TestApplication.AssemblyRedirection` for `net10.0` with `LibraryVersion=10.0.2`
- built `TestApplication.AssemblyRedirection` for `net9.0` with `LibraryVersion=9.0.0`
- built `TestApplication.AssemblyRedirection` for `net9.0` with `LibraryVersion=10.0.0`
- ran `git diff --check`

Manual:
- verified the `.NET Framework` test data now targets `Microsoft.Extensions.Logging.Abstractions`
- verified the `net462` test application forces the entrypoint dependency before the existing assembly/version checks run
- verified the expected `10.0.225.61305` file version matches the local `Microsoft.Extensions.Logging.Abstractions` `10.0.2` package contents

Notes:
- full `net462` execution was not possible in the local macOS environment because .NET Framework reference packs/runtime are not available there
- full `IntegrationTests` execution is currently blocked locally by unrelated SDK/source-generator issues in this checkout

## Checklist


test/IntegrationTests/AssemblyRedirectionTests.cs
```diff
@@ -37,14 +35,14 @@ public class AssemblyRedirectionTests(ITestOutputHelper output) : TestHelper("As
...
+    [InlineData("10.0.2", "System.Diagnostics.DiagnosticSource", "10.0.0.0", "10.0.526.15411", true)]
+    [InlineData("10.0.2", "System.Diagnostics.DiagnosticSource", "10.0.0.0", "10.0.526.15411", false)]
     // Case 3: Higher version is not possible for DiagnosticSource on .NET 10, the instrumentation tool is already using the highest possible version
 #elif NETFRAMEWORK
-    // Case 1: Lower version should be redirected (native profiler mandatory)
-    [InlineData("6.0.0", AssemblyName, "10.0.0.2", "10.0.225.61305")]
-    // Case 2: Equal version, should NOT be redirected (native profiler mandatory)
-    [InlineData("10.0.2", AssemblyName, "10.0.0.2", "10.0.225.61305")]
+    // Case 1: Lower version should be redirected before the entrypoint runs (native profiler mandatory)
+    [InlineData("6.0.0", "Microsoft.Extensions.Logging.Abstractions", "10.0.0.2", "10.0.225.61305")]
```
iskiselev: @alexeypukhov , I know you have plans to extend this test and make assembly configurable - will this change interfere with you plans?

iskiselev:
I don't think it properly test assembly resolver for .NET Framework scenario.
By default, we install all assemblies in GAC - https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/24279187987/job/70898737911?pr=4987
After it, when we run test, our assembly resolver is not working really - assembly resolved from GAC. To properly validate it, we need run it without installing assemblies in GAC, for example with docker - see how `Gac.UseLocal` is used in another test as example.

iskiselev (changes requested):

test/IntegrationTests/AssemblyRedirectionTests.cs
alexeypukhov: No. For the current state of the test you can use any dependency that the test application shares with the instrumentation tool

test/IntegrationTests/AssemblyRedirectionTests.cs
```diff
@@ -56,13 +64,28 @@ public void SubmitsTraces(
     {
 #if NETFRAMEWORK
         Assert.True(enableNativeProfiler, "Native profiler is required for assembly redirection on .NET Framework");
-        var excludedNames = string.Empty;
+
+        using var collector = new MockSpansCollector(Output, host: "*");
```
alexeypukhov: maybe at this point it makes sense to just split the test into two methods - one for all .NET versions and another for .NET Framework? It may improve the visibility of each test

test/test-applications/integrations/TestApplication.AssemblyRedirection/Program.cs
```diff
@@ -33,6 +36,15 @@
 Console.WriteLine($"  Excluded Assemblies (empty means 'check all'): [{string.Join(",", excludedNames)}]");
 Console.WriteLine();
 
+#if NETFRAMEWORK
+// Force an entrypoint dependency that also exists in the auto-instrumentation payload.
+var entrypointAssembly = NullLogger.Instance.GetType().Assembly;
```
alexeypukhov: Just neat picking but the entrypoint assembly sounds like this program assembly. I assume you are loading the dependency assembly which is `Microsoft.Extensions.Logging.Abstractions` for .NET Framework. In that case in the current terms of this application it's called "expected assembly". I personally don't mind any names, I just think we should make them consistent.

test/test-applications/integrations/TestApplication.AssemblyRedirection/Program.cs
```diff
@@ -33,6 +36,15 @@
...
+Console.WriteLine("Check 0: Entrypoint Assembly");
```
alexeypukhov: technically you get all this output at Check 2 below as long as you guarantee that the expected assembly is loaded.
For .NET it is done by starting an activity below which as a result serves two purposes: a) load expected assembly, b) validate we can send spans.
To keep things consistent I would either 1) just access NullLogger for .NET Framework here and wait for the output at Check 2, or 2) output the expected library for .NET as the Check 0 too.

docker/ubuntu1604.dockerfile
```diff
@@ -20,7 +20,7 @@ RUN apt-add-repository "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-5.
     update-alternatives --config clang++
 
 # Install newer g++
-RUN add-apt-repository ppa:ubuntu-toolchain-r/test -y && \
+RUN for i in 1 2 3; do add-apt-repository ppa:ubuntu-toolchain-r/test -y && break || sleep 10; done && \
```
iskiselev: Looks like it is not really related to the discussed change. Is it attempt to solve test instability? It may be better have as separate PR.

test/IntegrationTests/AssemblyRedirectionTests.cs
```diff
@@ -84,5 +76,90 @@ public void SubmitsTraces(
...
+
+        var arguments = new StringBuilder($"run --rm --name {containerName} --network {networkName}");
+        foreach (var environmentVariable in environmentVariables)
+        {
+            arguments.Append($" -e {environmentVariable.Key}={environmentVariable.Value}");
+        }
+
+        arguments.Append($" {imageName}");
+
+        using var process = Process.Start(new ProcessStartInfo("docker", arguments.ToString())
```
iskiselev: Can we use `ContainerBuilder` / `IContainer` API instead of direct working with `docker` executable?
We probably even cen generelize `IISContainerTestHelper.StartContainerAsync` to make universal helper for running test application in container and use it here (and in future with `WcfIISTests`).

test/test-applications/integrations/TestApplication.AssemblyRedirection/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+COPY bin/OpenTelemetry.DotNet.Auto.psm1 .
+ENV OTEL_DOTNET_AUTO_INSTALL_DIR=C:\opentelemetry `
+    OTEL_DOTNET_AUTO_LOG_DIRECTORY=C:\app\logs `
+    OTEL_LOG_LEVEL=debug
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
+RUN Import-Module .\OpenTelemetry.DotNet.Auto.psm1 -Verbose; `
+    Install-OpenTelemetryCore -LocalPath .\tracer.zip -RegisterAssembliesInGAC $false
+
+WORKDIR /app
+COPY bin/${libraryVersion}/${platform}/${configuration}/net462/ .
```
iskiselev, line 18: It may be possible to simplify test preparation, if docker will not hold application, but application will be host-compiled and mounted inside image as shared folder. It is optional and can be done separately.

aviralgarg05: Will do this separately

test/IntegrationTests/AssemblyRedirectionTests.cs
```diff
@@ -84,5 +72,53 @@ public void SubmitsTraces(
 
         // Assert
         collector.AssertExpectations();
+        return Task.CompletedTask;
+    }
+#endif
+
+#if NETFRAMEWORK
+    private static string GetNetFxDockerImageName(string libraryVersion)
```
alexeypukhov, line 80: I guess you need to move this private method below the test method to satisfy the code styling guide - https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/actions/runs/24710867822/job/72276497566?pr=4987

lachmatt:
SIG: moving to draft.

Kielek:
Cloing. dead PR

PR #4999, opened 2026-04-15
Bump test/infra vulnerable packages

Kielek:
## Why &  What

Bump test/infra vulnerable packages

## Tests

CI

## Checklist


iskiselev (approved):

iskiselev:
We probably can define `<CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>` on root level and have safe set of assemblies be defined in one place at root, not requiring us to add transitive package reference every time we need to change version.
We will still be able to override versions downlevel on needed basis.

PR #5038, opened 2026-04-30
Add native (RTL) stack walk fallback for Windows x64

eftiquar:
**Reviewer Guide -Please read - ** 
https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/5038#issuecomment-4725291675

## Why

When DoStackSnapshot fails (thread's instruction pointer is in native code), the profiler silently skips those threads. This is common for threads blocked in P/Invoke, OS waits, or unmanaged callbacks - resulting in missing stacks in continuous profiling data. This is especially impactful for .NET Framework applications where threads frequently transition to native code.

### Fixes 
Adds an RTL-based native stack walk fallback (Windows x64 only) that activates when DoStackSnapshot returns failure. This applies to both the DotNet and NetFx (.NET Framework) stack capture strategies.

 
## What

**NetFx x64(OS thread suspension, no runtime shield):**
1.	Suspend target + canary threads via ThreadGuard
2.	RTL frame-0 probe (proves RTL/CLR/loader CSes safe)
3.	Canary DSS probe (proves process-wide DSS health - required because NetFx lacks SuspendRuntime)
4.	Branch on probe verdict:
•	Managed frame-0: ctx is the seed (probe left it untouched) - issue seeded DSS directly
•	Native frame-0: emit composed frame0, walk natively from unwound ctx (frame-1) until managed boundary, then seed DSS from there

**CLR (.NET Core, runtime-suspension shield):**
1.	SuspendRuntime already certifies DSS health - no canary needed
2.	Seedless DSS attempted first (fast path under runtime suspension)
3.	On seedless failure: suspend target via ThreadGuard, RTL frame-0 probe, same managed/native dispatch as NetFx
x64 Fallback 
**NetFx x86 No RTL machinery exists on x86. The path reduces to:
•	Suspend target + canary
•	STL/heap gate + canary DSS probe (via StackWalkGuard)
•	Seedless DSS (only safe option without unwind APIs)

**Cooperative Timeout and Abandon Protocol**
The StackWalkGuard worker publishes a ProbeStage breadcrumb (atomic<ProbeStage>) at each hazardous call boundary. On timeout the orchestrator:
1.	Reads the stage to produce a precise timeout result (e.g., LoaderLockTimeout)
2.	Sets abandon_ flag
3.	Returns immediately - no logging or allocation while target is suspended
4.	ThreadGuard destructor resumes the target
5.	Caller logs the failure reason after resume
The worker, when it eventually unblocks, observes abandon_ and discards staged outputs without committing them to caller memory.
Key Safety Properties
•	Zero allocation on the suspending thread: ProbeRequest is a trivially-copyable POD; Schedule() is a memcpy under mutex
•	MSVC C2712 compliance: SEH (__try/__except) functions contain only POD locals; STL objects live exclusively in WorkerLoop
•	Staged output commit: Worker writes to staged_ctx/staged_frame; orchestrator publishes to caller memory under mutex only on success AND !abandon_
•	SEH on all RTL/loader calls: Corrupt pData, partially-unloaded modules, and bad stack pointers produce a clean false return, not a process crash

## Tests
We assert native-only frames as evidence of RTL walk. Each runtime has characteristic native APIs on the suspended thread's stack (e.g., ntdll!NtWaitFor*, KERNELBASE!WaitFor* for .NET Core; clr! internals for NetFx) that DoStackSnapshot never surfaces. Updated expected traces to require these signatures - their presence is conclusive proof the native walker executed and the managed/native frame interleaving is correct.


Kielek:
> Thread enumeration has had a memory leak hiding in plain sight - that leak predates the netfx support that we added - and was part of .Net continuous profiling machinery that enumerated threads for snapshotting.
> A COM interface which is used fir enumerating threads was not cleaned up after enumeration.
> Fixed now by storing it in ComPtr smart pointer.

Can we make this fix separate from this PR?

Kielek:
I think that this PR is missing some integration tests.
our current ContinuousProfiling test apps technically covers some native frames, but I do not see any changes here.
Am I wrongly understand this?

eftiquar:
> I think that this PR is missing some integration tests. our current ContinuousProfiling test apps technically covers some native frames, but I do not see any changes here. Am I wrongly understand this?

The integration tests exclusively look at stacks captured by DoStackSnapshot (DSS) API which is workhorse of managed stack capturing. This PR is fallback for when  DSS fails. And it is hard to write integration test that predictably fails DSS to activate the native walk.  
 ContinuousProfiling  integration tests do not and cannot cover precise native frames - they detect native frame and emit an opaque"Unknown_native_call" frame. 
I wanted to get the PR out to get feedback,  as for testing - I am exploring a few options.

eftiquar:
> > Thread enumeration has had a memory leak hiding in plain sight - that leak predates the netfx support that we added - and was part of .Net continuous profiling machinery that enumerated threads for snapshotting.
> > A COM interface which is used fir enumerating threads was not cleaned up after enumeration.
> > Fixed now by storing it in ComPtr smart pointer.
> 
> Can we make this fix separate from this PR?

Saw that coming :)

src/OpenTelemetry.AutoInstrumentation.Native/dot_net_stack_capture_strategy.h
```diff
@@ -4,51 +4,55 @@
...
+                }
 
-                if (FAILED(frameHr)) {
-                    //trace::Logger::Debug("DoStackSnapshot failed for thread ", tid,
-                    //    " HRESULT=", trace::HResultStr(frameHr));
-                    if (SUCCEEDED(captureResult)) {
-                        captureResult = frameHr; // Remember first error
-                    }
+                if (FAILED(frameHr) && SUCCEEDED(captureResult)) {
+                    captureResult = frameHr;
```
gleocadie: Out of curiosity, why not having a IStackWalker instance:
- one pointing to DoStackSnapShot
- the other one pointing to Rtl-based stack walker 
?
It then would be passed to the class.

RTL stackwalker always works since the JIT lays the code as expected.

In that case, for x86_64, you would just use RTL-based stackwalker, get the callstack and for x86, it will go

gleocadie (approved):
Overall code looks good

zacharycmontoya:
> Overall code looks good

@gleocadie Thanks for taking time to review this 🙏🏼

src/OpenTelemetry.AutoInstrumentation.Native/dot_net_stack_capture_strategy.h
eftiquar: Thank you for reviewing this. I am actually converging to same conclusion - some segments will be walked by DoStackSnapshot(DSS) and some by RTL. Even if RTL can walk managed calls, we can delegate to DSS with seed, as DSS has richer knowhow about transition points. Only when DSS draws blank do we fallback on RTL.  I am working on that.

src/OpenTelemetry.AutoInstrumentation.Native/CMakeLists.txt
```diff
@@ -225,7 +226,9 @@ add_library("OpenTelemetry.AutoInstrumentation.Native.static" STATIC
         rejit_preprocessor.cpp
         rejit_work_offloader.cpp
         environment_variables_util.cpp
+        stack_capturer.cpp
         stub_generator.cpp
+        suspension_guards.cpp
         method_rewriter.cpp
         tracer_tokens.cpp
         lib/coreclr/src/pal/prebuilt/idl/corprof_i.cpp
```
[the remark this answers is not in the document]
eftiquar: Windows CMake builds are intentionally out of scope for this project, and this is a deliberate, explicit policy — not an oversight. - 
if (CMAKE_SYSTEM_NAME MATCHES "Windows")
    message(FATAL_ERROR "Windows builds are not supported using CMAKE. Please use Visual Studio")

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -692,13 +693,25 @@ trace::WSTRING* NamingHelper::Lookup(const FunctionIdentifier& function_identifi
...
-    const auto old_value = function_name_cache_.Put(function_identifier, answer);
-    delete old_value;
+    if (stackWalker_ && function_identifier.IsNative())
+    {
+        if (auto hr = stackWalker_->ResolveNativeSymbolName(function_identifier.native_ip, *owned); hr != S_OK)
+        {
+            // owned's destructor deletes the string
+            return GetOrCreateUnknownNativeSentinel();
+        }
+    }
```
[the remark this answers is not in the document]
eftiquar: Now adding failed IPs to cache.

eftiquar:
> I think that this PR is missing some integration tests. our current ContinuousProfiling test apps technically covers some native frames, but I do not see any changes here. Am I wrongly understand this?

Updated the integration tests to cover the native elements.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -692,13 +693,28 @@ trace::WSTRING* NamingHelper::Lookup(const FunctionIdentifier& function_identifi
...
-    const auto old_value = function_name_cache_.Put(function_identifier, answer);
-    delete old_value;
+    if (stackWalker_ && function_identifier.IsNative())
+    {
+        if (auto hr = stackWalker_->ResolveNativeSymbolName(function_identifier.native_ip, *owned); hr != S_OK)
+        {
+            // Cache a per-key copy of the unknown sentinel so repeated lookups
+            // of the same unresolved IP are O(1) and avoid re-entering the resolver.
+            owned->assign(WStr("Unknown_Native_Function(unknown)"));
+            trace::WSTRING* raw = owned.release();
```
iskiselev: why  are we assign `owned` just to release it one line later? That's was hard to read.

Looks like it is done for symmetry with out-of-if branch.
Probably we should have here only `owned->assign(WStr("Unknown_Native_Function(unknown)"));` - everything else can be served by common code out of if.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.h
```diff
@@ -46,21 +45,33 @@ struct FunctionIdentifier
     mdToken  function_token;
     ModuleID module_id;
     bool     is_valid;
+    UINT_PTR native_ip;
+
     static FunctionIdentifier Managed(mdToken token, ModuleID mod, bool valid)
```
iskiselev, line 50: Do we need `valid` here? as we already have `ManagedInvalid`

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.h
```diff
@@ -46,21 +45,33 @@ struct FunctionIdentifier
...
     }
-
     static FunctionIdentifier ManagedInvalid()
     {
-        return {0, 0, false};
+        return {0, 0, false, 0};
+    }
+    static FunctionIdentifier UnknownNativeSentinel()
+    {
+        return {0, 0, true, static_cast<UINT_PTR>(~static_cast<UINT_PTR>(0))};
```
iskiselev: why do we have double `static_cast` - is negate changing type?

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -714,6 +730,20 @@ FunctionIdentifier NamingHelper::LookupManagedFunction(const FunctionID function
     return resolvedIdentifier;
 }
 
+trace::WSTRING* NamingHelper::GetOrCreateUnknownNativeSentinel()
```
iskiselev: Looks like it is unused?

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -780,20 +804,28 @@ static void CaptureFunctionIdentifiersForThreads(
...
+        auto frameProcessor = [&threadStacksBuffer, prof](continuous_profiler::CapturedFrame* frame) -> HRESULT
         {
-            auto                  thread = snapshot_context->threadId;
+            auto                  thread = frame->threadId;
             DoStackSnapshotParams doStackSnapshotParams{prof, &threadStacksBuffer[thread]};
-            FrameCallback(snapshot_context->functionId, snapshot_context->instructionPointer,
-                          snapshot_context->frameInfo, snapshot_context->contextSize, snapshot_context->context,
-                          &doStackSnapshotParams);
+
+            if (frame->isUnmanagedFrame && frame->functionId == 0 && frame->instructionPointer != 0)
```
iskiselev, line 799: Probably it can be refactored to process both branches in `HandleStackFrame` - it will be easier to understand two branches doing nearly same things.

iskiselev: Are we walking normal and allocation sample differently? in allocation we do not try to resolve native methods - but probably we should make it use similar logic?

src/OpenTelemetry.AutoInstrumentation.Native/clr_runtime_capture.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+
+HRESULT ClrRuntimeCapture::CaptureStack(ThreadID managedThreadId, StackSnapshotCallbackContext* clientData)
+{
+    if (clientData == nullptr)
+    {
+        return E_INVALIDARG;
+    }
+
+    HRESULT hr = profilerApi_->DoStackSnapshotUnseeded(managedThreadId, clientData);
```
iskiselev, line 61: So, we are trying to do unseeded first. If CLR knows how to do it, we will skip potential unmanaged frames on top, don't we?

src/OpenTelemetry.AutoInstrumentation.Native/safe_native_walk_service.cpp
```diff
@@ -0,0 +1,171 @@
...
+    NativeWalkResult walkResult = WalkNativeUntilManaged(ctx, clientData);
+
+    if (FAILED(walkResult.hr) && walkResult.nativeFrameCount == 0)
+        return walkResult.hr;
+
+    // Stage 2: DSS seeded from boundary - managed frames with CLR accuracy
+    HRESULT hr = walkResult.hr;
+    if (walkResult.hasSeed)
+    {
+        hr = profilerApi_->DoStackSnapshot(managedThreadId, StackSnapshotCallbackDefault, COR_PRF_SNAPSHOT_DEFAULT,
```
iskiselev: Ideally we should call `WalkNativeUntilManaged` if callback called with `funcId == 0`

src/OpenTelemetry.AutoInstrumentation.Native/clr_runtime_capture.cpp
eftiquar: No, if CLR reports unmanaged frames, the handler will mark those as native frames - the handler does not care if a native frame is originating from DSS or RTL unwind.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
eftiquar: We will do it in next PR. There is scope for further consolidation.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
eftiquar: Needs to be dropped.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.h
eftiquar: UnknownNativeSentinel  itself will be removed.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.h
eftiquar: ManagedInvalid is a helper to produce a sentinel FunctionIdentifier that specifies unresolved managed function. It is Existing code not new change.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
eftiquar: Yes there is duplication here,  will fix

src/OpenTelemetry.AutoInstrumentation.Native/safe_native_walk_service.cpp
eftiquar: Will keep it simple for now. I experimented with that and it did not yield any new frames. Will require more research. This PR is getting more complex than I anticipated.

gleocadie (approved):
Nice PR
Thanks learnt a few things there and there :)

eftiquar:
> Nice PR Thanks learnt a few things there and there :)

Thanks! Glad you found it useful. Appreciate you taking the time to review it.

I learned a few things while working on it too. One surprise was discovering that STL iterators have a global mutex in debug mode - I managed to hit a deadlock during testing. That's one of the reasons the prober ended up being STL-free. 

I also picked up a new term: ergonomics. Copilot used it when I suggested that the native walks should take a ThreadGuard, making it explicit that the thread is already suspended and enforcing the contract at the API boundary. Apparently that's "good ergonomics". 😄

There's something oddly poetic about discussing thread "handles" and API ergonomics in the same sentence.

src/OpenTelemetry.AutoInstrumentation.Native/OpenTelemetry.AutoInstrumentation.Native.vcxproj
```diff
@@ -187,15 +187,16 @@
     <ClInclude Include="assembly_redirection_net.h" />
     <ClInclude Include="assembly_redirection_netfx.h" />
     <ClInclude Include="calltarget_tokens.h" />
+    <ClInclude Include="captured_frame.h" />
```
iskiselev:
```suggestion
```

You added dead file in vcxproj.

test/IntegrationTests/ContinuousProfilerTests.cs
```diff
@@ -212,4 +268,89 @@ private static IEnumerable<string> GetFrameNames(Stack stack, ProfilesDictionary
...
+                continue;
+            }
+
+            var frames = GetFrameNames(dictionary.StackTable[stackIndex], dictionary).ToList();
+
+            // .NET Core ThreadPool worker threads park in NtWaitForWorkViaWorkerFactory with no managed frames.
+            // DSS yields nothing for them; their presence proves the RTL native walk is working.
+            // RtlSetThreadSubProcessTag appears as a trampoline; security/AV DLLs may also appear.
+            bool hasWorkerFactory = frames.Any(f => f.Contains("NtWaitForWorkViaWorkerFactory", StringComparison.OrdinalIgnoreCase));
+            bool hasThreadStart   = frames.Any(f => f.Contains("RtlUserThreadStart", StringComparison.OrdinalIgnoreCase)
```
iskiselev:
```suggestion
            bool hasThreadStart = frames.Any(f => f.Contains("RtlUserThreadStart", StringComparison.OrdinalIgnoreCase)
```

src/OpenTelemetry.AutoInstrumentation.Native/rtl_stack_walk.h
```diff
@@ -0,0 +1,106 @@
...
+#define OTEL_RTL_STACK_WALK_H_
+
+#if defined(_WIN32) && defined(_M_AMD64)
+
+#include <atomic>
+#include <corhlpr.h>
+#include <corprof.h>
+#include <unordered_map>
+#include "string_utils.h"
+#include "thread_suspend.h"
```
iskiselev: How is it used?

src/OpenTelemetry.AutoInstrumentation.Native/rtl_stack_walk.h
```diff
@@ -0,0 +1,106 @@
...
+
+struct NativeWalkContext
+{
+    StackSnapshotCallbackContext* clientParams  = nullptr;
+    std::atomic<bool>*                                stopRequested = nullptr;
+    IProfilerApi*                 profilerApi   = nullptr;
+    DWORD                                             maxDepth      = 512;
+};
+
+HRESULT WalkNativeStack(void* suspendedThread,
```
iskiselev: I don't see implementation / usage for it and next function.

src/OpenTelemetry.AutoInstrumentation.Native/rtl_stack_walk.h
```diff
@@ -0,0 +1,106 @@
+// Copyright The OpenTelemetry Authors
```
iskiselev: file has misleading name. It has native name resolution implementation mostly.

src/OpenTelemetry.AutoInstrumentation.Native/clr_runtime_capture.cpp
```diff
@@ -0,0 +1,108 @@
...
+    if (SUCCEEDED(hr))
+    {
+        return hr;
+    }
+
+    trace::Logger::Debug("[ClrRuntimeCapture] Seedless DSS failed. ManagedID=", managedThreadId,
+                         ", HRESULT=", trace::HResultStr(hr), " - attempting native walk fallback");
+
+#if defined(_WIN32) && defined(_M_AMD64)
```
iskiselev: I'd suggest to have opt out mechanic here, similar to mechanic we have for opting out from native resolution.

We call supend thread internally. Previously, if thread was somewhere in middle of native pinvoke it SuspendRuntime has zero effect (I believe it just marked that thread should be blocked before return to managed) - now it will be actually supended. I hope, we have not introduced any issues here, but I'll worry much less if there will be a way to opt out that function - even with the cost of less snapshot gathered (as it was before).

iskiselev:
There are many files added that are call-stack specific.  Previously we have not used any subfolders in OpenTelemetry.AutoInstrumentation.Native - but it may benefit that PR as it will collect stack collection logic to same folder. It can be done post PR merge.

src/OpenTelemetry.AutoInstrumentation.Native/clr_runtime_capture.cpp
eftiquar: Native walk is equally hazardous on .Net and netfx. While we have canary that gives safety shield for DSS walk, we can work on making native walk "canary" - read below for details - 
The native walk is easier to make robust - DSS is harder to probe because the walk is blindfolded - that is why we use canary as witness for DSS safety on target thread. After looking at concerns on native walk, I think we can fortify for native walk - native walk is iterative unwinding of stack frames using windows API - which is fully under our control compare with DSS - we cannot do DSS cooperatively it is black box- but native walk is in our control and we will yield if a call deadlocks - by waiting for iteration0 of native walk - if at any state the walk deadlocks , we bail out and abandon stack capture. I am working on that strategy. This is native counterpart of DSS canary.

test/IntegrationTests/ContinuousProfilerTests.cs
eftiquar: I am still working on getting main feature working - I have not looked at formatting error logs ..please focus on functional aspects first - dead files or missing space is  left out for the final commit because things are in flux - this explains why the header file names and content have mismatch.

src/OpenTelemetry.AutoInstrumentation.Native/rtl_stack_walk.h
eftiquar: thx - will fix it - the changes have been rapid and I have not looked at cosmetic or nomenclature aspect yet.

src/OpenTelemetry.AutoInstrumentation.Native/rtl_stack_walk.h
eftiquar: It is deprecated - will remove

src/OpenTelemetry.AutoInstrumentation.Native/rtl_stack_walk.h
eftiquar: not used . will be dropped, thx for pointing - please finish functional review first I am yet to do final cosmetic pass

src/OpenTelemetry.AutoInstrumentation.Native/OpenTelemetry.AutoInstrumentation.Native.vcxproj
eftiquar: dropped now

src/OpenTelemetry.AutoInstrumentation.Native/stack_walk_guard.cpp
```diff
@@ -0,0 +1,576 @@
...
+        //   2) Single source of truth: certifying the heap CS + STL
+        //      iterator/debug-list CS once per round (here) eliminates
+        //      duplication across CanaryDss (all archs) and RtlFrame0
+        //      (x64 only) paths.
+        // Cooperative: if the orchestrator abandoned us while we were
+        // blocked on the heap CS, bail before dispatching the probe.
+        SetStage(ProbeStage::HeapOrStlGate);
+        {
+            std::vector<int> v{1, 2, 3};
+        }
```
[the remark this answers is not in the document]
eftiquar: 1. The debug iter scenario is never going to hit in production. 
2. But in developer setups it will surface if debug iter preprocessor is defined .
3. The d-tor of vector will call _Orphan_all_locked_v3. That call will block if our target thread has held the STL lock and was suspended by stack capturer. If we hit that scenario - the probe worker will block at location shown below. 

>	ConsoleApplication1.exe!std::_Container_base12::_Orphan_all_locked_v3() Line 1255	C++
 	ConsoleApplication1.exe!std::_Container_base12::_Orphan_all() Line 1412	C++
 	ConsoleApplication1.exe!std::vector<int,std::allocator<int>>::_Tidy() Line 2088	C++
 	ConsoleApplication1.exe!std::vector<int,std::allocator<int>>::~vector<int,std::allocator<int>>() Line 830	C++

src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver.cpp
```diff
@@ -0,0 +1,336 @@
...
+        return 0;
+    }
+}
+
+UINT_PTR NativeSymbolResolver::GetImageBaseForIp(UINT_PTR ip)
+{
+    DWORD64              imageBase    = 0;
+    UNWIND_HISTORY_TABLE historyTable = {};
+    RtlLookupFunctionEntry(static_cast<DWORD64>(ip), &imageBase, &historyTable);
+
```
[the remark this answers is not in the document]
eftiquar: Added SEH cushion - this was right call.

src/OpenTelemetry.AutoInstrumentation.Native/netfx_runtime_capture.cpp
```diff
@@ -0,0 +1,235 @@
...
+    {
+        return E_INVALIDARG;
+    }
+
+    const CanarySnapshot canary = SnapshotCanary();
+    if (!canary.IsValid())
+    {
+        trace::Logger::Debug("[NetFxRuntimeCapture] No canary available; skipping. ManagedID=", managedThreadId);
+        return E_FAIL;
+    }
```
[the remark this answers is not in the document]
eftiquar: This is by design - canary failure is transient and we can skip current batch as we are continuously profiling, we can revisit. Added appropriate comments.

src/OpenTelemetry.AutoInstrumentation.Native/stack_walk_guard.cpp
```diff
@@ -0,0 +1,576 @@
...
+        bool     classified     = false;
+        bool     isSystemModule = false;
+        if (imageBase != 0)
+        {
+            WCHAR modulePath[MAX_PATH] = {};
+            DWORD pathLen              = 0;
+            __try
+            {
+                pathLen = GetModuleFileNameW(reinterpret_cast<HMODULE>(imageBase), modulePath, MAX_PATH);
+            }
```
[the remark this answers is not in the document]
eftiquar: force truncate if path length equals MAX_PATH . Fixed.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -692,13 +693,25 @@ trace::WSTRING* NamingHelper::Lookup(const FunctionIdentifier& function_identifi
...
-    delete old_value;
+    if (stackWalker_ && function_identifier.IsNative())
+    {
+        if (auto hr = stackWalker_->ResolveNativeSymbolName(function_identifier.native_ip, *owned); hr != S_OK)
+        {
+            // Cache a per-key copy of the unknown sentinel so repeated lookups
+            // of the same unresolved IP are O(1) and avoid re-entering the resolver.
+            owned->assign(WStr("Unknown_Native_Function(unknown)"));
+        }
+    }
```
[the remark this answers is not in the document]
eftiquar: There is no known scenario where we will fail transiently on symbol resolution - we do not download PDB or perform use external service for resolution.

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -221,16 +218,6 @@ public static void Initialize()
...
-        if (OpAmpSettings.Value.OpAmpClientEnabled)
-        {
-            var resources = ResourceHelper.AggregateResources(_tracerProvider, _meterProvider, LoggerProvider);
-
-            OpAmpHelper.EnableOpAmpClient(resources, OpAmpSettings.Value, _pluginManager);
-        }
-
-        // Notify plugins all initialization is done
-        _pluginManager.Initialized();
     }
```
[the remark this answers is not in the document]
eftiquar: Auto merge conflict led to this .. fixed it

src/OpenTelemetry.AutoInstrumentation.Native/interop.cpp
```diff
@@ -51,7 +51,8 @@ EXTERN_C VOID STDAPICALLTYPE ConfigureContinuousProfiler(bool         threadSamp
                                                          unsigned int threadSamplingInterval,
                                                          bool         allocationSamplingEnabled,
                                                          unsigned int maxMemorySamplesPerMinute,
-                                                         unsigned int selectedThreadSamplingInterval)
+                                                         unsigned int selectedThreadSamplingInterval,
+                                                         bool         nativeSymbolResolutionEnabled)
 {
     return trace::profiler->ConfigureContinuousProfiler(threadSamplingEnabled, threadSamplingInterval,
                                                         allocationSamplingEnabled, maxMemorySamplesPerMinute,
```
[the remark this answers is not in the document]
eftiquar: Auto merge conflict led to this .. fixed it

src/OpenTelemetry.AutoInstrumentation/Instrumentation.cs
```diff
@@ -596,8 +584,6 @@ private static void OnExit(object? sender, EventArgs e)
 
         try
         {
-            OpAmpHelper.StopOpAmpClientIfRunning(_pluginManager);
-
             LazyInstrumentationLoader?.Dispose();
             _sampleExporter?.Dispose();
```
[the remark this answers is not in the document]
eftiquar: Auto merge conflict led to this .. fixed it

eftiquar:
Reviewer Guide: Native Stack Walk (RTL Frame-0 Probe)
Where to focus
The core innovation is in stack_walk_guard.cpp (RunRtlFrame0Checks) and the caller-site pattern in netfx_runtime_capture.cpp / clr_runtime_capture.cpp. Everything else flows naturally from these two decisions.
Key design decision: Frame-0 is production AND hazard detection in one pass
The frame-0 probe is not a dry run. Each hazardous API call simultaneously:
•	Proves the suspended target does not hold the corresponding critical section (deadlock safety)
•	Produces a real output (RUNTIME_FUNCTION, managed/native classification, unwound CONTEXT)
If it completes without timeout, frames 1..N are safe by induction - target is still suspended, cannot acquire any CS it didn't already hold. This collapses a naive "probe pass + work pass" into one.
Key design decision: RAII carries full functional bandwidth
We use destructor-driven control flow for both suspension lifecycle and failure surfacing:

try {
    ThreadGuard target(osThreadId);     // suspends on construction
    // ... probes ...
    if (probe fails) throw ProbeResult;  // <-- granular enum
}
catch (ProbeResult result) {
    // ThreadGuard already destroyed by stack unwinding -> target resumed
    // Safe to log/allocate here - no thread is suspended
    Logger::Debug("Reason=", ProbeResultName(result));
}

This gives us:
•	Guaranteed resume on any exit path (normal return, throw, early bail)
•	Zero allocation while suspended (throw is a scalar enum, no string formatting)
•	Precise failure taxonomy surfaced cleanly at the catch site (RtlLookupTimeout, LoaderLockTimeout, ClrIpMapTimeout, etc.)
What follows naturally (lower review priority)
•	SafeNativeWalkService - stateless dispatcher: managed seed -> seeded DSS, native frame-0 -> walk + seed DSS at boundary
•	NativeSymbolResolver - SRW-cached module classification, export-table RVA walk
•	StackCaptureImpl - thin orchestrator, shutdown is implicit via member destruction order (no Stop() command)
•	Canary bookkeeping in NetFx - standard map-under-mutex, thread lifecycle hooks
Things to scrutinize
1.	The cooperative abandon protocol: AwaitProbeResult reads active_stage_ (atomic) to classify which CS blocked, sets abandon_, returns immediately. Worker observes abandon and discards staged outputs. No logging while target is suspended.
2.	SEH + C2712 compliance: RunRtlFrame0Checks uses __try/__except throughout - all locals are POD. STL gate lives in WorkerLoop (separate function) specifically because MSVC forbids SEH + non-trivial destructors in the same function.
3.	Staged output commit: Worker writes to staged_ctx / staged_frame; orchestrator publishes to caller memory under mutex only on success && !abandon_.

src/OpenTelemetry.AutoInstrumentation.Native/netfx_runtime_capture.cpp
```diff
@@ -0,0 +1,230 @@
...
+    if (!canary.IsValid())
+    {
+        // Canary not ready - expected only during startup. In continuous
+        // profiling mode, failing the batch is acceptable; the next cycle
+        // will capture stacks once the canary is elected.
+        // TODO: permanently disable sampling if canary-not-ready persists
+        // for more than 10 consecutive cycles.
+        trace::Logger::Debug("[NetFxRuntimeCapture] No canary available; skipping. ManagedID=", managedThreadId);
+        return E_FAIL;
+    }
```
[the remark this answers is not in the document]
eftiquar: Not really - we have logged telemetry as to why capture "failed". The semantics are if canary is not ready capture fails. For a single PR this is max bandwidth that we can occupy.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -692,13 +693,25 @@ trace::WSTRING* NamingHelper::Lookup(const FunctionIdentifier& function_identifi
...
+    else
+    {
+        this->GetFunctionName(function_identifier, *owned);
+    }
 
-    return answer;
+    trace::WSTRING* raw = owned.release();
+    owned.reset(function_name_cache_.Put(function_identifier, raw));
+    return raw;
```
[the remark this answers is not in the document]
eftiquar: There is no use after free the Put API returns null or least recently used entry - never raw because raw is constructed in this scope and was never pushed to cache.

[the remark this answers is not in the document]
eftiquar: The only situation when this can happen - if the cache size was 1 - a degenerate case.

src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver.cpp
```diff
@@ -0,0 +1,351 @@
...
+            return true;
+        }
+    }
+
+    // Phase 2: cache miss. Do the I/O OUTSIDE the lock so a stuck loader
+    // CS on the target thread stalls only this single caller (typically
+    // the StackWalkGuard worker, which the sampler drives via abandon)
+    // and never blocks other readers.
+    WCHAR modulePath[MAX_PATH] = {};
+    if (GetModuleFileNameW(reinterpret_cast<HMODULE>(imageBase), modulePath, MAX_PATH) == 0)
```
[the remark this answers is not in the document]
eftiquar: The frame zero probe wraps GetModuleFileNameW call insie SEH because we have suspended the thread whose context we are reading - if the loader lock is held by that thread we block in GetModuleFileNameW , and the orchestrator times out - as a side effect it throws exception - thereby setting of chain of events - the try block unwinds, RAII semantics kick in - the thread guard d-tor will resume thread - which will cause the module reference held by the suspended thread to be removed - and our probe thread is in uncharted territory - this warrants SEH cushion. 
The continuation path has no such hazard - the thread remains suspended - and the thread itself is active reference on the module. If the module were to be unloaded we will be in bigger trouble regardless of our SEH. 
Hope this settles the objection.

src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver.cpp
```diff
@@ -0,0 +1,351 @@
...
+    {
+        return false;
+    }
+    // GetModuleFileNameW may not NUL-terminate when the path fills or
+    // exceeds the buffer. Unconditionally force a terminator.
+    modulePath[MAX_PATH - 1] = L'\0';
```
nrcventura: I don't remember if we need to handle using different character prefixes for windows and linux for these character constants. Do we need to use the `Wstr` macro here?

eftiquar: Yes indeed - I will fix that. Thank you for pointing this out.

eftiquar: Fixed now - using "WStr " as platform agnostic macro for wide char literals.

nrcventura (approved):

Kielek:
@eftiquar,
I asked the codex for the review. The only one P2 issue:

# PR feedback: avoid re-entering the loader during the frame-0 probe

## Issue

`StackWalkGuard::RunRtlFrame0Checks` already calls `GetModuleFileNameW` under its SEH and timeout protocol while the target thread is suspended. Immediately after that, it calls `NativeSymbolResolver::IsSystemModule(imageBase)`.

On a cold resolver cache, `IsSystemModule` goes through `FetchClassification`, which calls `GetModuleFileNameW` again. That second loader call happens outside the probe's local SEH/abandon protocol and while the target thread is still suspended.

That reintroduces the exact hazard the frame-0 probe is trying to contain:

- possible loader-lock stall while the sampled thread is suspended
- no probe-stage timeout classification for the second loader call
- no local SEH cushion around the duplicate loader call
- a path where classification can fault or block after the guarded loader probe already succeeded

## Fix

I added `NativeSymbolResolver::IsSystemModuleFromPath(UINT_PTR imageBase, const WCHAR* modulePath)`.

This method classifies and caches the module using the `modulePath` that `RunRtlFrame0Checks` already obtained under SEH. The frame-0 probe now calls this path instead of `IsSystemModule(imageBase)`, so the probe does not re-enter `GetModuleFileNameW` for the same frame.

The regular `IsSystemModule(imageBase)` path remains available for callers that do not already have a module path.

## Patch

```diff
diff --git a/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver.cpp b/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver.cpp
index 2074ce51c..94acc2e6b 100644
--- a/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver.cpp
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver.cpp
@@ -160,6 +160,54 @@ std::optional<bool> NativeSymbolResolver::IsSystemModule(UINT_PTR imageBase) con
     return isSystem;
 }
 
+std::optional<bool> NativeSymbolResolver::IsSystemModuleFromPath(UINT_PTR imageBase, const WCHAR* modulePath) const
+{
+    if (imageBase == 0 || modulePath == nullptr || modulePath[0] == WStr('\0'))
+    {
+        return std::nullopt;
+    }
+
+    {
+        SrwSharedLock lk(moduleCacheLock_);
+        auto          it = moduleCache_.find(imageBase);
+        if (it != moduleCache_.end())
+        {
+            return it->second.isSystem;
+        }
+    }
+
+    ModuleInfo   info;
+    const WCHAR* slash = wcsrchr(modulePath, WStr('\\'));
+    info.baseName      = slash ? (slash + 1) : modulePath;
+
+    info.isSystem = false;
+    if (!sysDir_.empty())
+    {
+        std::wstring pathLower(modulePath);
+        std::transform(pathLower.begin(), pathLower.end(), pathLower.begin(), ::towlower);
+        info.isSystem = (pathLower.find(sysDir_) == 0);
+    }
+
+    {
+        SrwExclusiveLock lk(moduleCacheLock_);
+        if (auto it = moduleCache_.find(imageBase); it != moduleCache_.end())
+        {
+            return it->second.isSystem;
+        }
+
+        if (moduleCache_.size() >= kMaxModuleCacheSize)
+        {
+            trace::Logger::Debug("[NativeSymbolResolver] Module cache full, clearing ", moduleCache_.size(),
+                                 " entries");
+            moduleCache_.clear();
+        }
+
+        const bool isSystem = info.isSystem;
+        moduleCache_.emplace(imageBase, std::move(info));
+        return isSystem;
+    }
+}
+
 // ----------------------------------------------------------------------------
 // Private helpers
 // ----------------------------------------------------------------------------
diff --git a/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver_impl.h b/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver_impl.h
index 533e15551..c4e8457de 100644
--- a/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver_impl.h
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver_impl.h
@@ -57,6 +57,11 @@ public:
     // SRW-protected cache.
     std::optional<bool> IsSystemModule(UINT_PTR imageBase) const override;
 
+    // Classifies a module using a path already obtained by the caller. This
+    // avoids re-entering the loader when the stack-walk probe has already
+    // called GetModuleFileNameW under its own SEH/timeout protocol.
+    std::optional<bool> IsSystemModuleFromPath(UINT_PTR imageBase, const WCHAR* modulePath) const;
+
 private:
     // Cached per-module metadata - looked up once per unique imageBase.
     struct ModuleInfo
diff --git a/src/OpenTelemetry.AutoInstrumentation.Native/stack_walk_guard.cpp b/src/OpenTelemetry.AutoInstrumentation.Native/stack_walk_guard.cpp
index 8713cdc8a..6144d3dae 100644
--- a/src/OpenTelemetry.AutoInstrumentation.Native/stack_walk_guard.cpp
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/stack_walk_guard.cpp
@@ -515,7 +515,7 @@ bool StackWalkGuard::RunRtlFrame0Checks(ProbeRequest& req) noexcept
 
             if (pathLen != 0)
             {
-                auto sysModule = NativeSymbolResolver::Instance().IsSystemModule(imageBase);
+                auto sysModule = NativeSymbolResolver::Instance().IsSystemModuleFromPath(imageBase, modulePath);
                 isSystemModule = sysModule.value_or(false);
                 classified     = sysModule.has_value();
             }
```

eftiquar:
> @eftiquar, I asked the codex for the review. The only one P2 issue:
> 
> # PR feedback: avoid re-entering the loader during the frame-0 probe
> ## Issue
> `StackWalkGuard::RunRtlFrame0Checks` already calls `GetModuleFileNameW` under its SEH and timeout protocol while the target thread is suspended. Immediately after that, it calls `NativeSymbolResolver::IsSystemModule(imageBase)`.
> 
> On a cold resolver cache, `IsSystemModule` goes through `FetchClassification`, which calls `GetModuleFileNameW` again. That second loader call happens outside the probe's local SEH/abandon protocol and while the target thread is still suspended.
> 
> That reintroduces the exact hazard the frame-0 probe is trying to contain:
> 
> * possible loader-lock stall while the sampled thread is suspended
> * no probe-stage timeout classification for the second loader call
> * no local SEH cushion around the duplicate loader call
> * a path where classification can fault or block after the guarded loader probe already succeeded
> 
> ## Fix
> I added `NativeSymbolResolver::IsSystemModuleFromPath(UINT_PTR imageBase, const WCHAR* modulePath)`.
> 
> This method classifies and caches the module using the `modulePath` that `RunRtlFrame0Checks` already obtained under SEH. The frame-0 probe now calls this path instead of `IsSystemModule(imageBase)`, so the probe does not re-enter `GetModuleFileNameW` for the same frame.
> 
> The regular `IsSystemModule(imageBase)` path remains available for callers that do not already have a module path.
> 
> ## Patch
> ```diff
> diff --git a/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver.cpp b/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver.cpp
> index 2074ce51c..94acc2e6b 100644
> --- a/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver.cpp
> +++ b/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver.cpp
> @@ -160,6 +160,54 @@ std::optional<bool> NativeSymbolResolver::IsSystemModule(UINT_PTR imageBase) con
>      return isSystem;
>  }
>  
> +std::optional<bool> NativeSymbolResolver::IsSystemModuleFromPath(UINT_PTR imageBase, const WCHAR* modulePath) const
> +{
> +    if (imageBase == 0 || modulePath == nullptr || modulePath[0] == WStr('\0'))
> +    {
> +        return std::nullopt;
> +    }
> +
> +    {
> +        SrwSharedLock lk(moduleCacheLock_);
> +        auto          it = moduleCache_.find(imageBase);
> +        if (it != moduleCache_.end())
> +        {
> +            return it->second.isSystem;
> +        }
> +    }
> +
> +    ModuleInfo   info;
> +    const WCHAR* slash = wcsrchr(modulePath, WStr('\\'));
> +    info.baseName      = slash ? (slash + 1) : modulePath;
> +
> +    info.isSystem = false;
> +    if (!sysDir_.empty())
> +    {
> +        std::wstring pathLower(modulePath);
> +        std::transform(pathLower.begin(), pathLower.end(), pathLower.begin(), ::towlower);
> +        info.isSystem = (pathLower.find(sysDir_) == 0);
> +    }
> +
> +    {
> +        SrwExclusiveLock lk(moduleCacheLock_);
> +        if (auto it = moduleCache_.find(imageBase); it != moduleCache_.end())
> +        {
> +            return it->second.isSystem;
> +        }
> +
> +        if (moduleCache_.size() >= kMaxModuleCacheSize)
> +        {
> +            trace::Logger::Debug("[NativeSymbolResolver] Module cache full, clearing ", moduleCache_.size(),
> +                                 " entries");
> +            moduleCache_.clear();
> +        }
> +
> +        const bool isSystem = info.isSystem;
> +        moduleCache_.emplace(imageBase, std::move(info));
> +        return isSystem;
> +    }
> +}
> +
>  // ----------------------------------------------------------------------------
>  // Private helpers
>  // ----------------------------------------------------------------------------
> diff --git a/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver_impl.h b/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver_impl.h
> index 533e15551..c4e8457de 100644
> --- a/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver_impl.h
> +++ b/src/OpenTelemetry.AutoInstrumentation.Native/native_symbol_resolver_impl.h
> @@ -57,6 +57,11 @@ public:
>      // SRW-protected cache.
>      std::optional<bool> IsSystemModule(UINT_PTR imageBase) const override;
>  
> +    // Classifies a module using a path already obtained by the caller. This
> +    // avoids re-entering the loader when the stack-walk probe has already
> +    // called GetModuleFileNameW under its own SEH/timeout protocol.
> +    std::optional<bool> IsSystemModuleFromPath(UINT_PTR imageBase, const WCHAR* modulePath) const;
> +
>  private:
>      // Cached per-module metadata - looked up once per unique imageBase.
>      struct ModuleInfo
> diff --git a/src/OpenTelemetry.AutoInstrumentation.Native/stack_walk_guard.cpp b/src/OpenTelemetry.AutoInstrumentation.Native/stack_walk_guard.cpp
> index 8713cdc8a..6144d3dae 100644
> --- a/src/OpenTelemetry.AutoInstrumentation.Native/stack_walk_guard.cpp
> +++ b/src/OpenTelemetry.AutoInstrumentation.Native/stack_walk_guard.cpp
> @@ -515,7 +515,7 @@ bool StackWalkGuard::RunRtlFrame0Checks(ProbeRequest& req) noexcept
>  
>              if (pathLen != 0)
>              {
> -                auto sysModule = NativeSymbolResolver::Instance().IsSystemModule(imageBase);
> +                auto sysModule = NativeSymbolResolver::Instance().IsSystemModuleFromPath(imageBase, modulePath);
>                  isSystemModule = sysModule.value_or(false);
>                  classified     = sysModule.has_value();
>              }
> ```

Thank you @Kielek 
I have fixed the reported issue - it was a subtle one.

PR #5111, opened 2026-05-26
Support .NET 11 + drop .NET 8 and .NET 9

Kielek:
## Why

Fixes #

## Checklist


src/OpenTelemetry.AutoInstrumentation.Assemblies/Directory.Packages.props
```diff
@@ -47,56 +47,26 @@
 
   <ItemGroup Label="Transient dependencies auto-generated by TransientDependenciesGenerator for all .NET (Core)" Condition=" '$(TargetFrameworkIdentifier)' == '.NETCoreApp' ">
     <!-- TODO: should be discussed or a ticket should be filed: left included only to match the libraries previously present in output -->
```
alexeypukhov, line 50: very minor but the context of this TODO was for the packages `System.Private.ServiceModel` and below

.github/workflows/build-nuget-packages.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     name: Build NuGet packages
     # In principle this job needs the build outputs, however that would cause the upstream reusable workflow
     # to run twice. Instead, we let the caller workflow handle the dependencies.
-    runs-on: windows-2022
+    runs-on: windows-2025-vs2026
```
martincostello: You should be able to just use `windows-2025` now: [Windows 2025 Visual Studio 2026 image migration](https://github.blog/changelog/2026-05-14-github-actions-upcoming-image-migrations/#windows-2025-visual-studio-2026-image-migration)

Kielek: c88816ee3df34f0fe0301b8c91df98f2b65aeacd - fixed, thanks

src/OpenTelemetry.AutoInstrumentation.Native/clr_helpers.cpp
```diff
@@ -649,6 +653,99 @@ ULONG TypeSignature::GetSignature(PCCOR_SIGNATURE& data) const
...
+    mdToken type_token = mdTokenNil;
+    uint32_t token_length = 0;
+    const auto remaining_length = static_cast<uint32_t>(end - current);
+    if (FAILED(CorSigUncompressToken(current, remaining_length, &type_token, &token_length)) || token_length == 0)
+    {
+        return false;
+    }
+    current += token_length;
+
+    const auto type_name = GetTypeInfo(metadata_import, type_token).name;
```
eftiquar: What if GetTypeInfo fails - need to handle empty name and log and exit,

src/OpenTelemetry.AutoInstrumentation.Native/clr_helpers.cpp
```diff
@@ -649,6 +653,99 @@ ULONG TypeSignature::GetSignature(PCCOR_SIGNATURE& data) const
...
+    if (is_generic)
+    {
+        unsigned generic_argument_count;
+        if (!ParseNumber(current, end, &generic_argument_count) || generic_argument_count != 1 || current >= end)
```
eftiquar: We parse the generic argument count before verifying the generic type is Task<T> or ValueTask<T>. Can we move the is_task / is_value_task check ahead of ParseNumber() so unsupported generic types are rejected immediately? This would avoid unnecessary signature parsing and make the control flow a bit clearer.

eftiquar (commented):
Still reviewing

src/OpenTelemetry.AutoInstrumentation.Native/clr_helpers.cpp
```diff
@@ -649,6 +653,99 @@ ULONG TypeSignature::GetSignature(PCCOR_SIGNATURE& data) const
...
+        const bool is_task = element_type == ELEMENT_TYPE_CLASS && type_name == WStr("System.Threading.Tasks.Task`1");
+        const bool is_value_task =
+            element_type == ELEMENT_TYPE_VALUETYPE && type_name == WStr("System.Threading.Tasks.ValueTask`1");
+        if (!is_task && !is_value_task)
+        {
+            return false;
+        }
+
+        PCCOR_SIGNATURE result_start = current;
+        if (!ParseType(current, end) || current != end)
```
eftiquar: ParseType rejects ELEMENT_TYPE_ARRAY, this will cause Task<int[,]> to be rejected. By tying this to ParseType  we are importing limitations into our acync signature builder. I will look for an alternative that does not need ParseType services at all.

eftiquar: Once you have established - 
Task`1
ValueTask`1
generic_argument_count == 1

if (generic_argument_count != 1)
{
    return false;
}
we dont care what the type is - 
so - 
*result_type = {0,
                static_cast<ULONG>(end - current),
                current};
return true;

eftiquar (changes requested):

src/OpenTelemetry.AutoInstrumentation.Native/clr_helpers.cpp
eftiquar: Here is overall summary of my review feedback - 

1. Please do not use ParseType 
The PR makes valid runtime-async methods unsupported. For example, Task<int[,]> is rejected because ParseType does not support ELEMENT_TYPE_ARRAY. The runtime-async path should only decode the outer Task<>/ValueTask<> wrapper, then preserve the remaining bounded bytes as the complete T signature.
ParseNumber is only needed to decode the wrapper’s generic arity and can be replaced with bounded CorSigUncompressData. ParseType is unnecessary.

2. Please signature normalization out of the IL rewriter
The loader/function-resolution layer should produce both forms:
declared signature   = original metadata signature
effective signature  = signature after runtime-async transformation
The IL rewriter should receive the effective signature and use it for locals, stack balancing, and generated end-method tokens. It should not independently determine or transform runtime-async return semantics.
Other consumers should retain the declared signature:
- instrumentation matching
- textual display/signature formatting
- diagnostics and logging
- error messages
This avoids mutating or overloading one signature representation and keeps runtime-async policy in the function-loading machinery, where the method metadata and instrumentation definition are already resolved.

PR #5127, opened 2026-05-27
Adds public plugins API

RassK:
## Why

* Removes all reflection and provides public and statically typed API
* Ensures proper public endpoints tracking via .publicApi
* Helps the ensure all public methods are tested (follow up needed to test missing extensions)
* Helps to build more complex plugins (needed for opamp)

## What

Adds public statically typed plugins api.

## Tests

* Plugin updated
* All plugin integration tests remained the same to ensure no changes to current expectations

## Checklist


docs/plugins.md
```diff
@@ -1,143 +1,295 @@
...
+methods are called on every configured plugin.
 
 ```csharp
+using OpenTelemetry.AutoInstrumentation.PluginApi;
 
-public class MyPlugin 
+public class MyPlugin : IPlugin
 {
-    // To configure plugin, before OTel SDK configuration is called.
     public void Initializing()
```
iskiselev, line 36: Do we need `Initializing` - class will have `.ctor` - are we planning any workflow that may split plugin constructing and call `Initializing` on it?

docs/plugins.md
```diff
@@ -1,143 +1,295 @@
...
-    
-    // To configure any metrics options used by OpenTelemetry .NET Automatic Instrumentation
-    public void ConfigureMetricsOptions(OpenTelemetry.NameSpace.OptionType options)
+}
+```
+
+`BeforeConfigureTracerProvider`, `AfterConfigureTracerProvider`,
+`BeforeConfigureMeterProvider`, `AfterConfigureMeterProvider`, and
+`ConfigureResource` return the builder that automatic instrumentation continues
+to use. Only the first configured plugin implementing `ITelemetryPlugin` is used
```
iskiselev, line 115: Are we pushing builder from first plugin result as input for next? With chain we may give opportunity for every plugin to apply modifications.
Alternatively we may pass 2 arguments - inital builder and current builder. But do we expect plugin to replcae builder or just modify it?
If we do not expect complete builder replacement, it may be better to make it `void`.

docs/plugins.md
```diff
@@ -1,143 +1,295 @@
...
+{
+    public void Initializing()
+    {
+    }
+
+    public void Initialized()
+    {
+    }
+
+    public SelectiveSamplerConfiguration? GetFirstSelectiveSamplingConfiguration()
```
iskiselev, line 239: I'd probably either give each plugin result of previous to be able to observe settings / modify them. But in that case it may require two passes - one to modify, second to observe.

It may be good to provide some API for plugin to understand that callback will not be called or result ignored.

docs/plugins.md
```diff
@@ -185,7 +337,9 @@ public class MyOpAmpPlugin
 
 ## Requirements
 
-* The plugin must use the same version of the `OpenTelemetry` as the
-OpenTelemetry .NET Automatic Instrumentation.
-* The plugin must use the same options versions as the
-OpenTelemetry .NET Automatic Instrumentation (found in the table above).
+* The plugin must use the same `OpenTelemetry.AutoInstrumentation.PluginApi`
```
iskiselev, line 340: Are we planning to release `OpenTelemetry.AutoInstrumentation.PluginApi` on different schedule for stable API or in fact it usually will be version per release?

src/OpenTelemetry.AutoInstrumentation.PluginApi/OpenTelemetry.AutoInstrumentation.PluginApi.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\CommonExcludedAssets.props" />
+
+  <PropertyGroup>
+    <Description>Provides extension points for an OpenTelemetry.AutoInstrumentation plugin</Description>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="OpenTelemetry" />
```
iskiselev, line 10: I'm worry that design may be very fragile for Options configuration.
Previosly plugin referenced Auto-Insrumentation and trasitively referenced all our dependencies, so Options there were usually same that we used.
Now, Pulgin need reference API package + instrumentation library. So, it may be that plugin will have same API, but newer instrumentation library - which will result in hard to identify error.

RassK: I may need to check somehow before loading the plugin, that the auto version and plugin api version are matching.
There is a hard link between versions anyway but currently nothing is safeguarding. See > https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/v1.16.0-beta.1/docs/plugins.md#requirements

docs/plugins.md
RassK: I think it will be quite locked down to auto releases and plugin owners must also distribute a whole package (plugins + auto).

docs/plugins.md
RassK: Currently it's 1:1 match what we have, I would collect improvements to a different issues and do one step at a time.

docs/plugins.md
RassK: There may be a point in the future if dependency injection is used. SDK is using DI a lot. Currently it is 1:1 match what we have.

docs/plugins.md
RassK: before callbacks are before anything is setup by the auto and after callbacks are called when the chain is finalized. I don't think that replacing a builder completely in the final step actually breaks something, so I would leave that option open just in case plugin decides to ignore the whole previous chain.

There may be issue to retain the order in which plugins are loaded - as the iteration order could change a lot. If there are multiple - I don't know anybody using multiple plugins atm.

src/OpenTelemetry.AutoInstrumentation.PluginApi/Telemetry/ITelemetryPlugin.cs
ysolomchenko, line 1: `ITelemetryPlugin` is a large public interface.

A plugin that only needs `ConfigureResource` or `TracerProviderInitialized` is still required to implement the entire set of telemetry hooks. For a public API, this increases boilerplate and makes future evolution harder, since adding a new method to the interface would be a breaking change.

My suggestion is to split `ITelemetryPlugin` into smaller capability-based interfaces, such as separate interfaces for trace hooks, metric hooks, resource customization, and provider initialization callbacks. 

Alternatively, provide an optional abstract base class with no-op implementations.

RassK: I had the same concern. I was thinking to offer optional base class (but then this needs to be for every interface) with virtual methods (abstract is not helping you here). Generating an interface for each method doesn't sound also a good path forward.

RassK: Any more opinions if those are necessary? - since plugin are for limited users and each distro release requires attention anyway.

Kielek: What's about base clad with virtual methods with empty implementations?

docs/plugins.md
```diff
@@ -1,143 +1,295 @@
...
+> this behavior.
+> Example:
+> `OpenTelemetry.Instrumentation.Http.HttpClientTraceInstrumentationOptions.EnrichWithHttpWebRequest`
+> is conditionally set by this project.
+
+## OpAMP
+
+Implement `IOpAmpPlugin` to customize the OpAMP client and observe its
+lifecycle.
+OpAMP methods are called on every configured plugin implementing `IOpAmpPlugin`.
```
zacharycmontoya, line 183: I imagine that the OpAmpPlugin would be an _example_ plugin that is provided by this auto instrumentation project. What do you see users customizing at the moment? Can we remove this for now and potentially add it later if needed?

src/OpenTelemetry.AutoInstrumentation/Plugins/PluginManager.ContinuousProfiler.cs
```diff
@@ -1,27 +1,22 @@
...
-            var mi = plugin.Type.GetMethod("GetContinuousProfilerConfiguration");
-            if (mi?.Invoke(plugin.Instance, null) is Tuple<bool, uint, bool, uint, TimeSpan, TimeSpan, object>
-                continuousProfilerConfiguration)
+            if (plugin.Instance is IContinuousProfilerPlugin profilerPlugin)
             {
-                return (continuousProfilerConfiguration.Item1, continuousProfilerConfiguration.Item2,
-                    continuousProfilerConfiguration.Item3, continuousProfilerConfiguration.Item4,
-                    continuousProfilerConfiguration.Item5, continuousProfilerConfiguration.Item6,
-                    continuousProfilerConfiguration.Item7);
+                return profilerPlugin.GetFirstContinuousProfilerConfiguration();
```
zacharycmontoya, line 16: Rather than returning on the first `IContinuousProfilerPlugin`, it would be helpful to continue iterating and log other types that were found later but not loaded since a plugin was already found

zacharycmontoya: If the plugins are only providing options, should we replace the `IContinuousProfilerPlugin` model with declarative config?

src/OpenTelemetry.AutoInstrumentation/Plugins/PluginManager.cs
```diff
@@ -69,85 +106,140 @@ public void Initialized()
...
-    private void CallPlugins(string methodName, (Type Type, object Value) arg)
+    private TReturn CallPlugins<TPlugin, TReturn>(Func<TPlugin, TReturn> action)
     {
         foreach (var plugin in _plugins)
         {
-            var mi = plugin.Type.GetMethod(methodName, [arg.Type]);
-            mi?.Invoke(plugin.Instance, [arg.Value]);
+            if (plugin.Instance is TPlugin tplugin)
+            {
+                return action(tplugin);
```
zacharycmontoya: This returns after the first plugin, whereas the Action one calls all plugins. Was that intended?

docs/plugins.md
zacharycmontoya: It seems inconsistent for some methods to return a builder and some to be void. Can we standardize them all to one or the other?

docs/plugins.md
```diff
@@ -1,143 +1,295 @@
...
+    public void TracerProviderInitialized(TracerProvider tracerProvider)
     {
-        // My custom logic here
-        // Find supported options below
+        // Called after TracerProviderBuilder.Build().
     }
 
-    // To configure metrics SDK before Auto Instrumentation configured SDK
-    public OpenTelemetry.Metrics.MeterProviderBuilder BeforeConfigureMeterProvider(OpenTelemetry.Metrics.MeterProviderBuilder builder)
+    public MeterProviderBuilder BeforeConfigureMeterProvider(MeterProviderBuilder builder)
```
zacharycmontoya, line 87: It seems like you're missing a Before/After LogsProvider

docs/plugins.md
RassK: That's the root cause that needs a better auto <-> plugin interop layer. The API has to be reworked completely anyway.

Currently effective config handling is full custom, nothing in upstream but vendor can provide one with custom spec. Prob some other parts will follow, since internal spec gets developed faster than upstream.

src/OpenTelemetry.AutoInstrumentation/Plugins/PluginManager.ContinuousProfiler.cs
RassK: Can be updated later, since this is currently almost 1:1 (besides signature. there is no point to use current one) logic that we already have.

src/OpenTelemetry.AutoInstrumentation/Plugins/PluginManager.cs
RassK: yes, no change to current released logic.

docs/plugins.md
RassK: true, but current logic also doesn't have it.

src/OpenTelemetry.AutoInstrumentation.PluginApi/Telemetry/ITelemetryPlugin.cs
RassK: I don't mind adding them if there is actual need. I don't have any issues also adding a method to our distro if one pops up (they rarely do) - and it also brings some attention to the plugin if something is missing.

RassK:
Important note: This PR is not changing any internal logic besides adding a public statically typed api with all public parts verified (see .publicApi). Of course there was no point to mirror tuple<T1, ...> signatures and those are replaced with simple custom types.

It does ship one extra dll, which is this `PublicApi` - to indicate exactly which parts are safe to use in plugins. This api is not causing any side-effects, like force loading any other dll (see module load tests, only public api added).

src/OpenTelemetry.AutoInstrumentation/Plugins/PluginManager.cs
zacharycmontoya: Are you sure this maintains the previous logic? Looking at `BeforeConfigureTracerProviderBuilder`, this would previously invoke ConfigureBuilder which would loop over all plugins before returning, whereas now this method is called and returns after the first plugin invocation.

zacharycmontoya (approved):
Overall I think this does a good job of keeping the _existing_ behavior but with strong typing rather than reflection. I've resolved most of my previous comments but one still stands where I think there's a behavioral difference. Other than that, LGTM

src/OpenTelemetry.AutoInstrumentation/Plugins/PluginManager.cs
RassK: Thanks for catching this, you are correct.

Now I remember, this is erroneous also in original implementation. The plugin api is forcing you to return something (if null check is turned on). The old imp expects to return something but does not check or enforce it, therefore could cause some null return issues.

Please recheck if this little divert from original logic is acceptable.

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/5127/commits/acea4389dbe114b9f330c0c7eb0e4e470bdcb2a8

zacharycmontoya: LGTM, thanks!

nrcventura:
I have been trying to think about the plugin API for a while, and while I feel like the interface based approach can work and makes plugin development easier, we may want to consider some other options as well. I'll try to mention a few here so that we have some record of it when/if we search for information about the design decision in the future.

1. Interfaces (the approach in this PR)
2. Static API class for managing plugin behavior
3. Existing plugin style

For the interface approach:
Pros:
- Makes it easy to see what functionality is available
- Simple migration from the current behavior
Cons:
- Plugins will now implement an interface, and define all methods on the interface, even the ones that are not being used
- Adding a new method to the plugin API either requires a new interface or leveraging default interface implementations (I don't remember if this is applicable to .net framework or not)

The static API implementation would look more like the following. There would be a public plugin api assembly, that contains a public static class that can be used to register custom plugin logic. Something like the following just to provide an idea.

```
public static class PluginApi
{
    void ConfigureTracerProvider(Action<TracerProvider> configurer);
}
```

The plugin API can potentially detect if auto instrumentation is present or not. If it is not, the plugin API can function as a no-op. If auto instrumentation is present, it can ensure that the `Action` passed to the API is forwarded to auto configuration so that it can be used at the right time.

Pros:
- Makes it easy to see what functionality is available
- Only the APIs that you want to use need to be called
Cons:
- This approach would need to be proved out. I don't know if the timing is quite right, because we might need to run some of the configuration before the app's code runs.
- Migrating from the current plugin style to this approach is more complicated

The existing plugin style.

Pros:
- Only need to define what you use
- If a method is defined, but it is unknown by auto instrumentation, it doesn't break
Cons:
- Harder to see what all of the options are
- If a method is defined by it is unknown by auto instrumentation, it silently is ignored

Using an abstract class instead of an interface for option 1 (interfaces) may improve that option slightly so that users don't have to define implementations for the things they don't use. Default interface implementations may help with that as well (if it's possible).

Creating and using a source analyzer may help with the current reflection based implementation. It could at least help with identifying unrecognized method names.

The main idea with option 2 is that it's trying to reduce the auto instrumentation specific types that are necessary for the plugin so that you only have to define what you want to use. There are other ways to handle it, but the example is attempting to explain a public API that allows you to "build" an opaque configuration options object using, and then auto instrumentation will use that opaque configuration options object. This approach may still need a plugin class that is much simpler (possibly just a `void Run()` method) that allows the opaque configuration object to be created at the right time.

If we think the interface based approach is good enough, then I think the current PR is fine. We can probably improve upon it with abstract classes or default interface implementations, which can probably be done incrementally so long as we do not declare the plugin API stable.

Kielek:
I pushed 2 commits, if we decide to go this path.
It will be great to made some decision on the SIG.

RassK:
The reason why I don't see interfaces (without default imp) causing a big headache:
- Interfaces come by "topics", if you are not interested in the part, nothing changes
- Extension points are very rarely added to the same "topic".
- Updating the plugin requires some work anyway (syncing versions, making sure whole distro builds correctly)
- Missing implementations can be added as quickly as updating the nuget package.
- Missing implementation generates attention for the "topic" (remember that the state is experimental, some important changes might need some extra attention to fix or update).

Default implementations are easy to provide if needed. I wouldn't use one for Splunk specifically, since most extension points are needed for our distro. In case of changes, often we would need to use it anyway. There is a micro performance cost for using defaults but it seems mostly irrelevant, no high throughput functions.

Kielek (approved):
LGTM, I think that now it is good time to merge it and potentially include some further adjustment before next release.

I agree with @nrcventura concerns, but if we consider fact that changes in new package will be not considered as a breaking changes it should be fine to merge it.

nrcventura (approved):

PR #5195, opened 2026-06-17
SqlClient .NET Framework trace context propagation

Kielek:
# General

## Why

In Splunk we have received request to extend support known from https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/Instrumentation.SqlClient-1.15.2/src/OpenTelemetry.Instrumentation.SqlClient

## What

SqlClient .NET Framework trace context propagation
Intentionally I have omitted possibility to configure this by yaml file.
We do not have such support for .NET for now. It can be done on follow up when needed.

## Tests

New tests added, see implementation details section

## Checklist

# Implementation details

> [!NOTE]
> Following description is AI generated. Still I find it useful to understand code design, especially flow diagrams.

Adds .NET Framework SqlClient trace context propagation by instrumenting `SqlCommand.WriteBeginExecuteEvent` with CallTarget instrumentation.

The .NET SqlClient instrumentation can propagate trace context because DiagnosticSource payloads expose the live `DbCommand`. On .NET Framework, SqlClient uses EventSource payloads that only contain scalar values, so the listener cannot access the live command, connection, or transaction. This change fills that gap with profiler instrumentation at the point where the SqlClient span activity already exists and the live `SqlCommand` instance is still available.

The feature remains controlled by the same environment variable as the .NET implementation:

```text
OTEL_DOTNET_EXPERIMENTAL_SQLCLIENT_ENABLE_TRACE_CONTEXT_PROPAGATION
```

No separate YAML option is added, so the option surface stays consistent between .NET and .NET Framework.

## Flow

```mermaid
sequenceDiagram
    participant App as Application
    participant Cmd as SqlCommand
    participant Event as SqlClient EventSource
    participant OTel as OTel SqlClient instrumentation
    participant CT as AutoInstrumentation CallTarget
    participant Prop as SqlClientTraceContextPropagator
    participant SQL as SQL Server

    App->>Cmd: Execute*()
    Cmd->>Event: WriteBeginExecuteEvent()
    Event->>OTel: BeginExecute event
    OTel->>OTel: Create SqlClient Activity
    OTel->>OTel: Set Activity.Current

    Event-->>Cmd: return
    CT->>Prop: OnMethodEnd(command)

    Prop->>Prop: Check feature flag
    Prop->>Prop: Check CommandType.Text
    Prop->>Prop: Check open connection
    Prop->>Prop: Check Activity.Current source is SqlClient

    Prop->>Cmd: connection.CreateCommand()
    Prop->>Prop: Copy original transaction
    Prop->>Prop: Build traceparent from SqlClient Activity
    Prop->>SQL: SET CONTEXT_INFO @opentelemetry_traceparent
    SQL-->>Prop: context stored on same session

    Prop-->>Cmd: return
    Cmd->>SQL: Execute original user command
    SQL-->>Cmd: result with CONTEXT_INFO available

    Cmd->>Event: WriteEndExecuteEvent()
    Event->>OTel: EndExecute event
    OTel->>OTel: Stop SqlClient Activity
```

## Why This Flow Is Correct

The propagation happens after the SqlClient instrumentation has created the SqlClient `Activity`, so the propagated traceparent comes from the actual SqlClient span, not from a parent ASP.NET, request, or custom span.

The injected `SET CONTEXT_INFO` command uses the same open connection and copies the original transaction. That matters because SQL Server `CONTEXT_INFO` is scoped to the current session, and using a separate connection would set context on the wrong session.

The injected command is wrapped in instrumentation suppression and temporarily clears `Activity.Current`, preventing recursive SqlClient spans and avoiding pollution of telemetry.

The original application command is not modified. Propagation is best-effort: failures are logged at debug level and do not block the user command.

## Testing

Tests cover `System.Data.SqlClient` and `Microsoft.Data.SqlClient` on .NET Framework by executing `SELECT CONTEXT_INFO()` and verifying that the SQL Server session received a traceparent matching the emitted SqlClient span.

## Test Flow

The propagation tests verify the behavior from both sides of the boundary:

- The test application reads `CONTEXT_INFO()` from SQL Server, proving the value was written to the SQL Server session that executed the user command.
- The mock OTLP collector receives the SqlClient span, proving the propagated value belongs to the actual SqlClient activity created by instrumentation.

```mermaid
sequenceDiagram
    participant Test as Integration test
    participant Collector as Mock OTLP collector
    participant App as Test application
    participant Instr as AutoInstrumentation
    participant SQL as SQL Server

    Test->>Test: Enable OTEL_DOTNET_EXPERIMENTAL_SQLCLIENT_ENABLE_TRACE_CONTEXT_PROPAGATION
    Test->>Collector: Start and configure exporter endpoint
    Test->>App: Run with SELECT CONTEXT_INFO()

    App->>Instr: Execute SqlCommand
    Instr->>Instr: Create SqlClient Activity
    Instr->>SQL: SET CONTEXT_INFO traceparent
    App->>SQL: SELECT CONTEXT_INFO()
    SQL-->>App: traceparent bytes

    App-->>Test: Print CommandResult=<traceparent>
    Instr-->>Collector: Export SqlClient span

    Test->>Test: Parse CommandResult
    Test->>Collector: Find one SqlClient span
    Test->>Test: Compare CONTEXT_INFO trace id/span id with SqlClient span
```

The test runs with and without an explicit SQL transaction. This is important because the injected `SET CONTEXT_INFO` command copies the original command transaction; without that, propagation could fail for commands executed inside a transaction even though non-transactional commands work.

The test also asserts that exactly one SqlClient span is collected for the command. That confirms the injected `SET CONTEXT_INFO` command is suppressed and does not create a second visible span.


RassK (approved):

nrcventura (approved):
I do not think that connection level context is the right way to handle this, especially when connections can be reused, but this at least matches the existing functionality that we are trying to port over.

Kielek:
@nrcventura, I fully understand your concerns, but both implementations follow semantic convention (at the time of writing Experimental/Development state) https://github.com/open-telemetry/semantic-conventions/blob/v1.42.0/docs/db/sql-server.md#set-context_info

PR #5206, opened 2026-06-22
feat(kafka): add messaging.kafka.cluster.id span attribute

### LINKED ISSUE #5278: feat(kafka): add messaging.kafka.cluster.id span attribute
## Problem

Confluent.Kafka producer and consumer spans do not include `messaging.kafka.cluster.id`. When the same topic name exists across multiple clusters or environments, there is no way to distinguish which cluster a span came from.

## Expected behavior

`messaging.kafka.cluster.id` ([semconv](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/kafka.md)) should be emitted on Confluent.Kafka producer and consumer spans. It is a **Recommended** attribute in the messaging semantic conventions, so it is populated **by default** (no configuration flag) for Confluent.Kafka ≥ 2.0.0.

## Context

Part of a cross-language effort to add this attribute consistently across OTel instrumentations. See [opentelemetry-java-instrumentation#18977](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/18977) for the cross-language tracking issue.

shashank-reddy-nr:
## Summary

Adds `messaging.kafka.cluster.id` ([semantic-conventions #3819](https://github.com/open-telemetry/semantic-conventions/pull/3819), merged) to Confluent.Kafka producer and consumer spans.

`messaging.kafka.cluster.id` is a **Recommended** attribute in the messaging semantic conventions, so it is emitted **by default** — there is no configuration flag. It is populated for Confluent.Kafka ≥ 2.0.0; older versions are unaffected.

### How it works

- The cluster ID is resolved **once per unique `bootstrap.servers`** via a background `AdminClient.DescribeClusterAsync()` call and cached with a TTL. The produce/consume hot path only reads the cached value — no per-span broker calls.
- When a fully-constructed producer/consumer is available, `DependentAdminClientBuilder` reuses its existing `rdkafka` handle, so no additional client handles are allocated.
- SASL/SSL client configuration is forwarded so authenticated clusters are supported.
- If the cluster ID has not been resolved yet, the attribute is omitted for that span (no blocking on the hot path).

## Testing

- Unit tests for `BootstrapServersCache` and `KafkaClusterIdCache`.
- Integration test `SubmitsTraces` asserts `messaging.kafka.cluster.id` is present and non-empty on producer and consumer spans for Confluent.Kafka ≥ 2.0.0.

Fixes #5278

Assisted-by: Claude Sonnet 4.6

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation/Instrumentations/Kafka/KafkaInstrumentation.cs
+++ b/src/OpenTelemetry.AutoInstrumentation/Instrumentations/Kafka/KafkaInstrumentation.cs
@@ -31,6 +31,15 @@ internal static class KafkaInstrumentation
             {
                 activity.SetTag(MessagingAttributes.Keys.Kafka.ConsumerGroupId, groupId);
             }
+
+            if (BootstrapServersCache.TryGet(consumer, out var bootstrapServers))
+            {
+                var clusterId = KafkaClusterIdCache.GetClusterIdAndScheduleRefresh(bootstrapServers);
+                if (clusterId is not null)
+                {
+                    activity.SetTag(MessagingAttributes.Keys.Kafka.ClusterId, clusterId);
+                }
+            }
         }
 
         return activity;
@@ -62,7 +71,8 @@ internal static class KafkaInstrumentation
     public static Activity? StartProducerActivity<TTopicPartition, TMessage, TClient>(
         TTopicPartition partition,
         TMessage message,
-        TClient producer)
+        TClient producer,
+        object? rawProducer = null)
     where TTopicPartition : ITopicPartition
     where TMessage : IKafkaMessage
     where TClient : INamedClient
@@ -81,6 +91,16 @@ internal static class KafkaInstrumentation
                 producer);
 
             activity.SetTag(MessagingAttributes.Keys.Kafka.IsTombstone, message.Value is null);
+
+            if (rawProducer is not null &&
+                BootstrapServersCache.TryGet(rawProducer, out var bootstrapServers))
+            {
+                var clusterId = KafkaClusterIdCache.GetClusterIdAndScheduleRefresh(bootstrapServers);
+                if (clusterId is not null)
+                {
+                    activity.SetTag(MessagingAttributes.Keys.Kafka.ClusterId, clusterId);
+                }
+            }
         }
 
         return activity;
--- a/.cspell/other.txt
+++ b/.cspell/other.txt
@@ -60,6 +60,7 @@ proto
 protobuf
 protos
 RABBITMQ
+SASL
 Serilog
 Sonoma
 spdlog
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,13 @@ release.
 
 ### Added
 
+- Add `messaging.kafka.cluster.id` span attribute to Confluent.Kafka producer
+  and consumer spans (Confluent.Kafka ≥2.0.0). The cluster UUID is fetched
+  once per unique bootstrap-servers string via a background
+  `AdminClient.DescribeClusterAsync()` call and cached; no overhead on the
+  produce/consume hot path. SASL and SSL client configuration is forwarded to
+  the `AdminClient` so authenticated clusters are supported. If the cluster ID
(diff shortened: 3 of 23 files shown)


Kielek:
I think that you could use Assited-by instead of Co-authored commit messages.
Without this we cannot proceed with this PR.

What is more, some meaningful PR description.

Kielek:
It is not compliant with semantic convention (yet).
Could you please work towards fixing https://github.com/open-telemetry/semantic-conventions/issues/32 first?

Thanks @lachmatt for the link.

shashank-reddy-nr:
Hi @Kielek, sure! I will let you know once I've added it to the semantic-conventions repo.

Kielek (changes requested):
Blocked by https://github.com/open-telemetry/semantic-conventions/pull/3819.
Adding request changes to avoid unintentional merge.

shashank-reddy-nr:
Hi @Kielek Can you please review this PR? As semantic-convention PR got merged?

martincostello:
> As semantic-convention PR got merged?

We shouldn't ship until there's a stable release containing the change (v1.44) - same as https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4739#issuecomment-4968703567

src/OpenTelemetry.AutoInstrumentation/Instrumentations/Kafka/KafkaInstrumentation.cs
Kielek, line 16: As semantic convention is already merged, I would consider migrating our implementation to the new version (separate PR). Then, as a follow up we can consider this PR.

Kielek (commented):
I do not fully reviewed this PR.
Based on the short review, the separate method to fetch cluster it should be deeply reviewed. I do not recall any similar mechanism for similar solution.

@zacharycmontoya, could you please advice here what is the best/battled tested solution for  this?

src/OpenTelemetry.AutoInstrumentation/Instrumentations/Kafka/KafkaInstrumentation.cs
shashank-reddy-nr: Sure @Kielek I will raise a new PR to include new sem conv version, post that we can come to this PR.

shashank-reddy-nr: Thanks for the flag — I don't think a version bump is needed here.

`messaging.kafka.cluster.id` is additive (not a rename/replacement) and still `Development` stability in v1.44.0, so there's no settled convention to migrate onto yet. Also, this `ActivitySource` doesn't carry a `TelemetrySchemaUrl` at all — it's tagged with the product version, not a semconv version (`KafkaInstrumentation.cs:16`) — so there's nothing here that goes "stale."

The only semconv-version references in this area are two doc-link comments in `MessagingAttributes.cs` (lines 6, 19), both still on v1.23.0. Bumping those — and any real convention migration, gated behind `OTEL_SEMCONV_STABILITY_OPT_IN` — feels like separate work from adding this one attribute. Happy to send a follow-up for those comment links if useful, but don't think it should block this PR.

shashank-reddy-nr: Hi @Kielek  Semantic-convs is released. Now, I Migrated existing implementation using #5453: five renamed attributes, the span name order, and `error.type`. Can you please review it and get it merged?

The red checks are not from this PR. `main` has been failing since #5443 - the BuildTasks targets is still on `1.18.0-beta.1` while `Directory.Packages.props` is on `1.18.0-beta.2`. #5454 is the one-line fix.

shashank-reddy-nr: Hi @Kielek sorry to bother you. Can you please move this forward when you have time?

zacharycmontoya: Let's focus on merging #5453 and then we can merge this on top of that PR

PR #5213, opened 2026-06-23
[infra] Pin CMake installer checksum in ubuntu dockerfile

ysolomchenko:
## Why

Fixes # N/A

Found by Codex analysis

## What

Reverted the independent verification for the CMake installer. Removed downloading the SHA-256 manifest from the same GitHub release and pinned the hardcoded SHA-256 directly in the Dockerfile

## Checklist

### DIFF
--- a/docker/ubuntu1604.dockerfile
+++ b/docker/ubuntu1604.dockerfile
@@ -35,12 +35,11 @@ RUN add-apt-repository ppa:ubuntu-toolchain-r/test -y && \
 
 # Install cmake directly from GitHub releases (Kitware Xenial apt repo no longer serves cmake)
 RUN CMAKE_INSTALLER="cmake-${CMAKE_VERSION}-linux-x86_64.sh" && \
-    CMAKE_SHA256_FILE="cmake-${CMAKE_VERSION}-SHA-256.txt" && \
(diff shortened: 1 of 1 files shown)


docker/ubuntu1604.dockerfile
```diff
@@ -35,12 +35,11 @@ RUN add-apt-repository ppa:ubuntu-toolchain-r/test -y && \
 
 # Install cmake directly from GitHub releases (Kitware Xenial apt repo no longer serves cmake)
 RUN CMAKE_INSTALLER="cmake-${CMAKE_VERSION}-linux-x86_64.sh" && \
-    CMAKE_SHA256_FILE="cmake-${CMAKE_VERSION}-SHA-256.txt" && \
+    CMAKE_INSTALLER_SHA256="f582e02696ceee81818dc3378531804b2213ed41c2a8bc566253d16d894cefab" && \
```
Kielek, line 38: It will probably fail when new cmake will be fetched. Do you see any possibility to auto-update it?

ysolomchenko: That's actually the issue i trying to fix. The problem was that the Dockerfile downloaded both the CMake installer and SHA-256 checksum, Re-enabling automatic SHA-256 updates doesn't improve security, it leaves the same attack vector in place.

Kielek (approved):

PR #5247, opened 2026-07-02
chore work 

eftiquar:
draft .. deleting


nrcventura:
My main concern with this approach is based on my experience with subscribing to the event pipe from managed code within the same process that is being monitored (by using an EventListener to listen to the same event pipe source used in this PR). When subscribing to the event pipe in that manner, I have seen numerous errors across multiple versions of .net. The errors are usually reported as memory corruption errors and can affect the exit code of the application. I'm hoping that subscribing to the event pipe from native code does not suffer from the same problems (and that the problems I've seen stem from the in-process native to managed layer translation). As a result, I am very nervous about building another feature on top of that functionality.

eftiquar:
> My main concern with this approach is based on my experience with subscribing to the event pipe from managed code within the same process that is being monitored (by using an EventListener to listen to the same event pipe source used in this PR). When subscribing to the event pipe in that manner, I have seen numerous errors across multiple versions of .net. The errors are usually reported as memory corruption errors and can affect the exit code of the application. I'm hoping that subscribing to the event pipe from native code does not suffer from the same problems (and that the problems I've seen stem from the in-process native to managed layer translation). As a result, I am very nervous about building another feature on top of that functionality.

Thank you for raising this concern — EventPipe in-process subscription can indeed be fragile, and your experience with memory corruption via EventListener is well-documented in the .NET community.
However, there's a **critical architectural distinction** between the approach you encountered and what we're implementing here: we are native only implementation. Additionally, memory allocation profiling already uses EventPipe session and has been part of Otel Profiler. We will make this feature opt-in as further mitigation.   

Your Scenario (Managed EventListener)
•	Managed C# code using System.Diagnostics.Tracing.EventListener
•	Every event crosses the native→managed boundary with marshaling
•	Known stability issues across .NET versions (as you've experienced)
Our Implementation (Native Profiling API)
•	Uses the official CLR Profiling API (ICorProfilerInfo12::EventPipeStartSession)
•	Events are delivered entirely in native code via EventPipeEventDelivered callbacks
•	No managed/native boundary crossing per event

Evidence of stability - 
Our Otel Profiler already uses native event pipe session to track memory allocation and publishes call stacks for memory profiling.  This PR tunes the pipe to emit contention events. 

// This has been running in production without memory corruption
COR_PRF_EVENTPIPE_PROVIDER_CONFIG sessionConfig[] = {
    {WStr("Microsoft-Windows-DotNETRuntime"), 
     kClrGcKeyword,  // <-- Already using EventPipe for AllocationTick
     COR_PRF_EVENTPIPE_VERBOSE, 
     nullptr}
};


Mitigation Strategy - 
We will gate this using configuration - make it opt -in. 

Bottom line: The memory corruption you experienced with EventListener is a managed-layer problem. Our native profiling API approach is architecturally immune to that failure mode.

PR #5268, opened 2026-07-09
Prevent stack buffer overflow when instrumenting methods with many variables

martincostello:
## Why

Fixes stack buffer overflow in the case of attempting to instrument a method with an extreme (700+) number of local variables.

Discussed with @Kielek over Slack and we agreed that this doesn't meet the bar for a security issue and is just a standard bug fix for the purposes of robustness.

The only likely valid use case that would hit this bug would be trying to instrument some sort of extremely large machine-generated code.

## What

`CallTargetTokens::ModifyLocalSig`, `TracerTokens::WriteBeginMethod` (fast path) and the related token emitters copied variable-length metadata signatures into fixed 500-byte stack buffers with no size check. A method whose local-variable or argument signatures exceed the buffer overflowed the stack during instrumentation.

Add bounds checks before every write into those buffers, and make the rewrite callback honour the resulting HRESULT so a failed signature build aborts the rewrite (leaving the method uninstrumented) instead of committing an inconsistent method body.

## Tests

Manually verified with Claude's assistance.

## Checklist


Kielek:
@zacharycmontoya, I think that you have starting looking on same/similar issue, so any feedback is more than welcomed.

nrcventura (approved):

src/OpenTelemetry.AutoInstrumentation.Native/calltarget_tokens.cpp
eftiquar, line 336: if (offset != signatureLength)
{
    Logger::Warn("GetCallTargetDefaultValueMethodSpec: signature size mismatch.");
    return mdMethodSpecNil;
}

eftiquar (commented):
Still reviewing.

src/OpenTelemetry.AutoInstrumentation.Native/tracer_tokens.cpp
eftiquar, line 326: Same as above - to make it robust please consider adding validation on computed signature length. 
The buffer checks are valid, thank you for addressing those. Please also add this check at each site where we are computing signature length. 

if (offset != signatureLength)
{
    Logger::Warn("GetCallTargetDefaultValueMethodSpec: signature size mismatch.");
    return mdMethodSpecNil;
}

src/OpenTelemetry.AutoInstrumentation.Native/method_rewriter.cpp
```diff
@@ -175,9 +175,19 @@ HRESULT TracerMethodRewriter::Rewrite(RejitHandlerModule* moduleHandler, RejitHa
     mdToken  exceptionToken        = mdTokenNil;
     mdToken  callTargetReturnToken = mdTokenNil;
     ILInstr* firstInstruction      = nullptr;
-    tracerTokens->ModifyLocalSigAndInitialize(&reWriterWrapper, caller, &callTargetStateIndex, &exceptionIndex,
-                                              &callTargetReturnIndex, &returnValueIndex, &callTargetStateToken,
-                                              &exceptionToken, &callTargetReturnToken, &firstInstruction);
+    hr = tracerTokens->ModifyLocalSigAndInitialize(&reWriterWrapper, caller, &callTargetStateIndex, &exceptionIndex,
+                                                   &callTargetReturnIndex, &returnValueIndex, &callTargetStateToken,
+                                                   &exceptionToken, &callTargetReturnToken, &firstInstruction);
+    if (FAILED(hr))
```
eftiquar, line 181: Consider adding "hr" to the logs, will be useful signal for diagnosis.

eftiquar (approved):
Suggested minor changes. Approving.

zacharycmontoya (approved):
LGTM. Thanks for the contribution and thank you @eftiquar for recommending the additional hardening

PR #5269, opened 2026-07-09
Bound ExtractPublicKeyToken to matched token

martincostello:
## Why

Fix out-of-bounds write.

## What

The copy loop was driven by the caller-supplied length rather than the 16-hex-character (8-byte) regex match, allowing an out-of-bounds write past the destination buffer when length > 8. Clamp the loop to the bytes actually available in the match.

## Tests

Added new unit test.

## Checklist


src/OpenTelemetry.AutoInstrumentation.Native/regex_utils.cpp
```diff
@@ -148,9 +148,11 @@ bool ExtractPublicKeyToken(const WSTRING& str, unsigned char* data, const int le
 
     if (std::regex_search(str, match, re) && match.size() == 2)
     {
-        for (int i = 0; i < length; i++)
+        const auto token           = match.str(1);
```
nrcventura: I think that it would be helpful to have some comments that explain how the token is being extracted. There are enough moving parts between windows and linux and their character widths, as well as sizes of ints and longs on those platforms to keep straight.

martincostello: Comments added.

src/OpenTelemetry.AutoInstrumentation.Native/regex_utils.cpp
```diff
@@ -142,15 +142,34 @@ bool ExtractPublicKeyToken(const WSTRING& str, unsigned char* data, const int le
...
+    // its low 8 bits are meaningful and are stored through the `unsigned char` cast. The loop is
+    // bounded by both `length` (the size of the caller's `data` buffer) and `available_bytes` (the
+    // number of bytes the match can supply), so it never reads past the match nor writes past the
+    // destination buffer.
+
 #ifdef _WIN32
     static std::wregex re(WStr("PublicKeyToken=([a-fA-F0-9]{16})"));
     std::wsmatch       match;
 
     if (std::regex_search(str, match, re) && match.size() == 2)
```
eftiquar, line 218: ExtractVersion already uses a similar TryParseVersionMatch template pattern, so this change makes ExtractPublicKeyToken consistent with the existing style in the file.
We can follow same approach - 

namespace {
// A public key token is 8 bytes, represented as 16 hex characters.
constexpr int kPublicKeyTokenBytes = 8;

// Converts a single hex character to its 0-15 value. Returns false if invalid.
// Using auto allows generic character handling without template brackets.
constexpr bool TryParseHexDigit(auto ch, unsigned char& value) {
    if (ch >= '0' && ch <= '9') {
        value = static_cast<unsigned char>(ch - '0');
        return true;
    }
    if (ch >= 'a' && ch <= 'f') {
        value = static_cast<unsigned char>(ch - 'a' + 10);
        return true;
    }
    if (ch >= 'A' && ch <= 'F') {
        value = static_cast<unsigned char>(ch - 'A' + 10);
        return true;
    }
    return false;
}

// Parses a hex token capture group into bytes, writing exactly 8 bytes.
// Using auto references the sub_match type transparently.
bool TryParsePublicKeyToken(const auto& sub_match, unsigned char* data, const int length) {
    // Check buffer length constraint first
    if (length < kPublicKeyTokenBytes) {
        return false;
    }

    if (!sub_match.matched || sub_match.length() < static_cast<size_t>(kPublicKeyTokenBytes * 2)) {
        return false;
    }

    // Direct pointer tracking avoids type extraction acrobatics completely
    auto current_char = sub_match.first;

    for (int i = 0; i < kPublicKeyTokenBytes; i++) {
        unsigned char high;
        unsigned char low;
        
        if (!TryParseHexDigit(*current_char++, high) || !TryParseHexDigit(*current_char++, low)) {
            return false;
        }
        
        data[i] = static_cast<unsigned char>((high << 4) | low);
    }
    return true;
}
} // namespace

bool ExtractPublicKeyToken(const WSTRING& str, unsigned char* data, const int length) {
    if (str.empty() || data == nullptr || length < kPublicKeyTokenBytes) {
        return false;
    }

#ifdef _WIN32
    static std::wregex re(WStr("PublicKeyToken=([a-fA-F0-9]{16})"));
    std::wsmatch match;
    if (std::regex_search(str, match, re) && match.size() == 2) {
        // Pass the specific capture group sub_match directly
        return TryParsePublicKeyToken(match[1], data, length);
    }
#else
    // On Linux/MacOS, convert to narrow string for regex
    std::string narrow_str = ToString(str);
    static std::regex re("PublicKeyToken=([a-fA-F0-9]{16})");
    std::smatch match;
    if (std::regex_search(narrow_str, match, re) && match.size() == 2) {
        // Pass the specific capture group sub_match directly
        return TryParsePublicKeyToken(match[1], data, length);
    }
#endif

    return false;
}

eftiquar: Reasoning - 
•	Buffer overflow removed: bytes_to_write is clamped to min(length, available_bytes), so it never writes past data.
•	No substr allocations: it indexes characters directly instead of creating temporary strings each iteration.
•	No stringstream: hex parsing is done with a small helper, avoiding uninitialized x and silent parse failures.
•	Explicit failure: invalid hex returns false instead of writing garbage.
•	Shared logic: both _WIN32 and non-Windows branches call the same ParseHexToken, removing duplicated loops.

eftiquar (commented):
Proposed an approach that is consistent with existing version parser - TryParseVersionMatch, 
Core issue of buffer overrun is addressed and made the approach consistent with "ExtractVersion".

eftiquar (approved):

nrcventura (approved):

PR #5270, opened 2026-07-09
Fix null dereference

martincostello:
## Why

Fix null dereference if profiler is null.

## What

The exported entry point dereferenced the static `profiler_info` without a null check, unlike its sibling exports, crashing if it was invoked before profiler initialisation (or with profiling disabled).

## Tests

Added a new test.

## Checklist


nrcventura (approved):

zacharycmontoya (approved):

eftiquar (approved):
A debug level log will be useful -
if (profiler_info == nullptr)
{
trace::Logger::Debug("ContinuousProfilerSetNativeContext skipped: profiler_info is null.");
return;
}

nrcventura (approved):

nrcventura (approved):

PR #5272, opened 2026-07-09
Fix OOB read in the method signature parser

martincostello:
## Why

Fix out-of-bounds read when parsing method signatures.

## What

`ParseRetType`, `ParseParamOrLocal` and `ParseType` dereferenced `*pbCur` before checking `pbCur < pbEnd`, reading one byte past a truncated signature blob via `FunctionMethodSignature::TryParse`. Reorder the bounds checks before each dereference.

## Tests

Added test.

## Checklist


test/OpenTelemetry.AutoInstrumentation.Native.Tests/clr_helper_test.cpp
```diff
@@ -265,4 +268,119 @@ TEST_F(CLRHelperTest, DoesNotFindDoubleNestedTypeDefsByName)
         EXPECT_FALSE(found) << "Failed type is : " << def << std::endl;
         EXPECT_EQ(typeDef, mdTypeDefNil) << "Failed type is : " << def << std::endl;
     }
-}
\ No newline at end of file
+}
+
+#ifdef _WIN32
```
nrcventura, line 273: Why are these tests windows only? Is it necessary to call these windows apis in these tests? It just seems like we just need to configure a buffer with "invalid" content in order to trigger the error conditions that this PR is addressing.

martincostello: I questioned this with Claude locally, and its response was that the tests are only run on Windows already anyway. Initially it didn't even generate these guard clauses, it added these after I questioned it.

As it stands today, the tests only run on Windows, but if they were expanded to other OSs then these new tests would just be skipped. Getting the gtest suite to run on a non-Windows OS is already non-trivial (I assume, otherwise it would already be cross-platform?), so it's not making anything worse.

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/350f9d8ebfeab8b16cc0a5281114426f2ca175fe/build/Build.Steps.Linux.cs#L78-L88

The TODO above was committed 5 years ago.

My guess at why the tests are written like this is also so that they leverage the `/GS` support by MSVC, rather than needing to complicate them with additional code to support other OSs were there isn't currently any validation anyway.

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/350f9d8ebfeab8b16cc0a5281114426f2ca175fe/src/OpenTelemetry.AutoInstrumentation.Native/OpenTelemetry.AutoInstrumentation.Native.DLL.vcxproj#L165

eftiquar (approved):

nrcventura (approved):

PR #5312, opened 2026-07-29
Profiler runtime configuration

ysolomchenko:
## Why

It change is needed for remote configuration for opamp
https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#servertoagentremote_config

## What

- ConfigureContinuousProfiler remains the separate, one-time native resource initialization method.
- Added six runtime APIs:
  - `SetContinuousProfilerEnabled(bool)`
  - `SetContinuousProfilerSamplingInterval(uint)`
  - `SetContinuousProfilerAllocationSamplingEnabled(bool)`
  - `SetContinuousProfilerMaxMemorySamplesPerMinute(uint)`
  - `SetContinuousProfilerSnapshotsEnabled(bool)`
  - `SetContinuousProfilerSnapshotSamplingInterval(uint)`
- CPU, allocation, and snapshot sampling can be enabled, disabled, and re-enabled at runtime. Configured intervals and limits are retained while sampling is disabled.
- CPU and snapshot sampling share one sampling thread. Runtime changes do not create duplicate threads, and disabling one mode keeps the thread running while the other remains enabled.
- CPU and snapshot configuration changes are published through the atomic sampling_configuration_dirty_ fast-path and observed at sampling-cycle boundaries. Configuration changes do not wake the stop condition variable; stopping the final sampling mode or shutting down wakes the thread immediately.
- A single mutex protects the effective CPU/snapshot configuration, stop-request state, and sampling-thread lifecycle. A single stop condition variable is used only by WaitForStop() and is notified only when stopping the sampling service.
- Shutdown() requests sampling-service stop once. SamplingThreadMain observes only the stop request and does not depend on shutdown state.
- Allocation sampling remains a separately managed service. Its sampling limit can be updated at runtime without restarting the active EventPipe session.
- CPU sample batches remain stored in FIFO order together with the sampling interval used when they were captured.
- Runtime setters do not initialize native resources or managed export pipelines. Each sampling pipeline must be prepared during initial configuration before it can be enabled later at runtime.

## Tests

Ci

## Checklist


eftiquar:
The current implementation appears to enforce strict local versioning semantics for runtime CPU sampling configuration, but I do not think that guarantee is meaningful for OpAMP-driven config.

OpAMP is a remote control plane. Between the user/agent pushing config and that config reaching the native profiler, there are multiple asynchronous hops outside this code’s control. Because of that, barriers and version checks that try to prevent “stale” local captures do not provide meaningful end-to-end correctness.

Each exported CPU sample batch already carries the sampling interval used for that batch. A batch captured with the previous interval is still valid data; it is not an illegal capture. The exporter can interpret it correctly from the batch metadata.

If we accept eventual consistency, the sampler can snapshot config once per cycle. At worst, one sample cycle may use the previous config after a runtime update. That is acceptable for OpAMP and greatly simplifies the implementation.

Since config changes are expected to be rare and the sampler loop is the fast path, I suggest optimizing for the common case. Keep a local config snapshot in SamplingThreadMain, use an atomic dirty flag to signal that config changed, and only acquire the config mutex on the rare reload path.

To summarize - 
CPU Sampling config is applied eventually. The sampler observes the latest config at cycle boundaries. In-flight or just-scheduled samples may complete under the previous config, and that is valid because each batch carries its own sampling interval metadata. Disable/shutdown should still wake the sampler promptly and avoid starting new expensive captures once observed. 
If this is acceptable I can provide detailed review feedback based on this contract.

ysolomchenko:
> The current implementation appears to enforce strict local versioning semantics for runtime CPU sampling configuration, but I do not think that guarantee is meaningful for OpAMP-driven config.
> 
> OpAMP is a remote control plane. Between the user/agent pushing config and that config reaching the native profiler, there are multiple asynchronous hops outside this code’s control. Because of that, barriers and version checks that try to prevent “stale” local captures do not provide meaningful end-to-end correctness.
> 
> Each exported CPU sample batch already carries the sampling interval used for that batch. A batch captured with the previous interval is still valid data; it is not an illegal capture. The exporter can interpret it correctly from the batch metadata.
> 
> If we accept eventual consistency, the sampler can snapshot config once per cycle. At worst, one sample cycle may use the previous config after a runtime update. That is acceptable for OpAMP and greatly simplifies the implementation.
> 
> Since config changes are expected to be rare and the sampler loop is the fast path, I suggest optimizing for the common case. Keep a local config snapshot in SamplingThreadMain, use an atomic dirty flag to signal that config changed, and only acquire the config mutex on the rare reload path.
> 
> To summarize - CPU Sampling config is applied eventually. The sampler observes the latest config at cycle boundaries. In-flight or just-scheduled samples may complete under the previous config, and that is valid because each batch carries its own sampling interval metadata. Disable/shutdown should still wake the sampler promptly and avoid starting new expensive captures once observed. If this is acceptable I can provide detailed review feedback based on this contract.

I think it’s okay. I updated the profiler in this commit: [db1c17f](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/5312/commits/db1c17fd959349178317d8747b63fe12a9dd5e23)

eftiquar:
A) Synchronization needs to be revisited 

Please use one mutex to protect:
   effective config
   stop_requested
   thread lifecycle state 

Please use one stop CV:
  waited on by WaitForStop()
  notified only for stop/shutdown

Please use sampling_configuration_dirty_ as an atomic fast-path hint for config refresh.
Remove config-change CV notification and strict reload/versioning behavior.
Remove thread_sampling_thread_mutex_; avoid mutex proliferation and lock-order complexity.

B) Any reason we are not covering allocation sample settings ? 

C) 
Thread Lifecycle - 
You added runtime ability to start/stop the thread sampling service, not just change intervals. This is great feature. 
My suggestion - Shutdown may not leak into SamplingThreadMain; shutdown simply stops the dependent sampling service.
SamplingThreadMain observes only “stop requested,” not “shutdown requested.”
The Shutdown will request stop - that is sufficient to stop sampler thread - else we cause duplicate signals - stop requested and shutdown requested. This will simplify the code. Existing way of shut down and stop signals are unnecessary duplications.

Let me know your thoughts.

eftiquar:
The profiling service  was designed for a “start once, seed once” model, while OpAMP requires dynamic configuration refresh and on-demand enablement and disablement.
@ysolomchenko's  PR has been critical in exposing the resulting lifecycle and concurrency scenarios, including interactions between OpAMP, AppDomain initialization, process-wide native state, and configuration changes during active profiling.

I propose splitting the work into two PRs:

1. Evolve the native profiling service into a dynamically configurable service and define its lifecycle, precedence, shutdown, and mid-capture semantics.
2. Connect the managed OpAMP control-plane integration and exporter pipeline to that native contract.

This builds on the current PR rather than replacing its contribution: the work here revealed the architectural requirements and provides the scenarios the revised native contract must support.

eftiquar:
Thank you for the discussion around the runtime-configuration changes. This PR has been instrumental in exposing the broader process-wide lifecycle and AppDomain ownership semantics that dynamic configuration requires.

I opened #5360 to discuss the native runtime-service contract and the proposed horizontal delivery split independently of this implementation review. The issue includes the current design draft.

The intent is to establish and test the native lifecycle contract first, and then connect the managed OpAMP control plane and exporter pipeline to it.

PR #5349, opened 2026-08-11
Npgsql trace context propagation

### LINKED ISSUE #5338: Propagate Npgsql trace context to PostgreSQL through application_name
# Feature Request

Description assisted by Codex/GPT5.6Sol

## TLDR

Propagate context by application_name as it is done by js and collector-contrib.
The solution is not recommended by the Npgsql team, but it is what we have right now.

SQLCommenter is a worse idea, as it kills benefits from cached commands.

## Requested framework or library

- Framework or library name: [Npgsql](https://www.npgsql.org/)
- Library type: PostgreSQL ADO.NET data provider
- Library version: Npgsql 6.0 and later, matching the
  [currently documented support range](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/10cc9a24196699515538901a98d993595377fe7c/docs/config.md#L159)

## Problem

The existing Npgsql automatic instrumentation records client-side activities by
subscribing to the `Npgsql` activity source
([current registration](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/10cc9a24196699515538901a98d993595377fe7c/src/OpenTelemetry.AutoInstrumentation/Configurations/EnvironmentConfigurationTracerHelper.cs#L37)).
It does not propagate the client span context to the PostgreSQL session. As a
result, telemetry observed from the server cannot be correlated reliably with
the Npgsql client span.

This affects the OpenTelemetry Collector Contrib PostgreSQL receiver in
particular. Its query-sample feature reads `pg_stat_activity` and emits
`db.server.query_sample` log records. The receiver already supports trace
correlation when the complete value of `application_name` is a valid W3C
`traceparent`:

- the receiver extracts `traceparent` from `application_name`
  ([implementation](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3456d18100676b2a160809040d84faaace0caeed/receiver/postgresqlreceiver/client.go#L989-L1033));
- the extracted trace and span identifiers are written to the OTLP log record;
  see the
  [generated log builder](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3456d18100676b2a160809040d84faaace0caeed/receiver/postgresqlreceiver/internal/metadata/generated_logs.go#L21-L54);
- this behavior was introduced by
  [Collector Contrib commit `8bd91eec220a75e8c4b30a39f16780f6ff168586`](https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/8bd91eec220a75e8c4b30a39f16780f6ff168586).

The Collector-side support is available without a custom receiver, but query
sampling itself must be enabled and its monitoring user must have `pg_monitor`
([receiver configuration](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3456d18100676b2a160809040d84faaace0caeed/receiver/postgresqlreceiver/README.md#L75-L99)).
The receiver emits correlated query-sample **logs**, not PostgreSQL server
spans, and its log support is currently marked Development
([receiver status](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3456d18100676b2a160809040d84faaace0caeed/receiver/postgresqlreceiver/README.md#L1-L12)).

The related Npgsql request
[#6255](https://github.com/npgsql/npgsql/issues/6255) proposes SQLCommenter
propagation. SQLCommenter is allowed by the OpenTelemetry database semantic
conventions, but that section is Development, the feature should be opt-in,
and comments should normally be appended to the query
([semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/e9c74f02b130ed8b7122f271475fa781545dd1e3/docs/db/database-spans.md#L538-L568)).
The current PostgreSQL receiver does not extract context from SQL comments, so
SQLCommenter alone would not solve interoperability with that receiver.

## Proposed solution

Add an opt-in Npgsql context-propagation mode that places the current Npgsql
client span's W3C `traceparent` in the PostgreSQL session's
`application_name` immediately before executing the corresponding command.
This should be the recommended compatibility solution while the PostgreSQL
server and Npgsql client do not provide an end-to-end native propagation
mechanism that is also consumed by server-side telemetry.

The capability should preferably be implemented in Npgsql itself, with this
distribution enabling or configuring that support. Npgsql owns physical
connection scheduling, prepared commands, pooling, multiplexing, retries, and
the exact point at which the client span is created, so it can preserve the
required ordering more reliably than an external activity listener. Work
should be coordinated with
[npgsql/npgsql#6255](https://github.com/npgsql/npgsql/issues/6255).

The initial behavior should be:

1. Keep propagation disabled by default because it changes
   `application_name` and can add an extra command/round trip for every user
   command. The exact configuration key and configuration-source precedence
   should be agreed by the .NET Auto Instrumentation SIG.
2. When enabled and a valid client span context exists, execute the equivalent
   of:

   ```sql
   SET application_name =
     '00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01';
   ```

   on the same physical connection, ordered immediately before the user
   command. The value must be the complete W3C `traceparent`, with no prefix,
   suffix, or additional metadata, because this is what the receiver parses.
3. Do not put `tracestate` or baggage in `application_name`. The receiver only
   consumes `traceparent`, and PostgreSQL restricts `application_name` to fewer
   than 64 characters. A version `00` W3C `traceparent` is 55 characters and
   fits that limit
   ([PostgreSQL documentation](https://www.postgresql.org/docs/current/runtime-config-logging.html#GUC-APPLICATION-NAME)).
4. Treat the propagation command as internal instrumentation: it must not
   produce an Npgsql client span, must not invoke user command callbacks as if
   it were application SQL, and must not recurse into propagation.
5. Never fail the application command because propagation failed. Report the
   failure through instrumentation diagnostics and continue with the original
   command.
6. Prevent context leakage through pooled connections. Before an uncorrelated
   command, replace a previous propagated value with the connection's configured
   application name or an empty value. Leaving the most recent traceparent in
   place until the next command allows the polling receiver to observe a
   completed query, but it must never be attributed to a subsequent command
   that has no valid context.
7. Preserve strict pairing for concurrent commands. Multiplexing must either be
   supported with correct per-command ordering or rejected/documented for this
   mode; one command must never run with another command's trace context.
8. Support simple and prepared commands without rewriting the user query text.
   Support synchronous and asynchronous execution paths.
9. Document that enabling this mode takes ownership of `application_name` for
   trace correlation and therefore conflicts with applications that use that
   field as a stable human-readable client name.

Node.js PostgreSQL instrumentation provides implementation precedent. Its
opt-in `enableTraceContextPropagation` mode sends
`SET application_name = '<traceparent>'` before the user query and suppresses
its internal command
([implementation](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/a405e8596bdb06c6669294de00b0f60dbd74e339/packages/instrumentation-pg/src/instrumentation.ts#L470-L503),
[configuration and overhead warning](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/a405e8596bdb06c6669294de00b0f60dbd74e339/packages/instrumentation-pg/src/types.ts#L81-L96)).
The proposed Npgsql behavior should also address stale context explicitly, even
if another implementation does not yet do so.

### Preferred long-term path

`application_name` is a compatibility carrier, not the desired permanent
transport. When PostgreSQL and Npgsql both expose a native, per-operation trace
context mechanism and the server-side telemetry path can consume it, that
native mechanism should take precedence:

1. use native client/server propagation when both ends support it;
2. do not also issue `SET application_name` in that case; and
3. retain `application_name` only as an explicit fallback for older deployments
   and for Collector versions that depend on it.

Native support should avoid query mutation, avoid taking over the user-visible
application name, work with prepared commands and multiplexing, and avoid the
extra SQL round trip. The transition and fallback rules should be agreed with
the Npgsql, PostgreSQL, Collector, and OpenTelemetry maintainers before changing
the default behavior.

### Acceptance criteria

- The feature is opt-in and disabled behavior remains unchanged.
- A valid Npgsql client span causes the matching
  `db.server.query_sample` OTLP log to contain the same trace ID and the client
  span ID when collected by the unmodified Collector Contrib PostgreSQL
  receiver.
- Simple, parameterized, and prepared commands are covered in synchronous and
  asynchronous paths.
- Internal propagation commands do not create exported spans or appear as the
  application's database operation.
- Pool reuse cannot leak a previous traceparent into an uncorrelated command.
- Concurrent or multiplexed commands cannot exchange contexts; unsupported
  combinations fail configuration validation or are documented clearly.
- Propagation failure does not change the user command's result.
- Integration coverage exercises PostgreSQL and the Collector receiver without
  introducing fixed multi-second sleeps into the test suite.
- Configuration, performance cost, `application_name` ownership, Collector
  prerequisites, and the future native migration path are documented.

### SQLCommenter

SQLCommenter is the generic mechanism currently described by the OpenTelemetry
database semantic conventions and is implemented by other language
instrumentations. It remains a reasonable optional solution for database logs
or receivers that explicitly parse it.

It is not the recommended first solution for this request because the current
Collector Contrib PostgreSQL receiver does not parse SQL comments. It would
still emit the usual query-sample attributes, but the OTLP log would not receive
the propagated trace and span identifiers. SQLCommenter also mutates query text,
has difficult prepared-statement lifetime semantics, and an appended comment
can be absent from a truncated `pg_stat_activity.query` value.

Adding SQLCommenter parsing to the Collector receiver could make it interoperable
in the future, but it would require a separate Collector feature and would not
remove the prepared-command and truncation concerns. It is therefore outside
the initial scope of this issue.

### Application-level configuration

Applications can execute `SET application_name` themselves, but this requires
source changes, must be repeated for every operation, and is easy to implement
incorrectly with pooling and concurrency. It does not satisfy the automatic
instrumentation use case.

### Wait only for native support

Native server/client propagation is the preferred end state, but waiting for it
leaves currently deployed Npgsql applications unable to correlate their client
spans with query samples from the existing Collector receiver. An explicitly
opt-in `application_name` compatibility mode provides value now and can be
retired gradually after native support is broadly available.

## Additional context

The intended correlation is:

```text
Npgsql client span
  -> W3C traceparent in PostgreSQL application_name
  -> pg_stat_activity
  -> Collector Contrib PostgreSQL receiver
  -> db.server.query_sample log with TraceId and SpanId
```

Relevant references:

- [Npgsql issue #6255: Support OpenTelemetry context propagation for database traces](https://github.com/npgsql/npgsql/issues/6255)
- [Collector implementation commit](https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/8bd91eec220a75e8c4b30a39f16780f6ff168586)
- [Collector `application_name` extraction](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3456d18100676b2a160809040d84faaace0caeed/receiver/postgresqlreceiver/client.go#L989-L1033)
- [OpenTelemetry database context-propagation conventions](https://github.com/open-telemetry/semantic-conventions/blob/e9c74f02b130ed8b7122f271475fa781545dd1e3/docs/db/database-spans.md#L538-L568)
- [Node.js PostgreSQL `application_name` precedent](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/0a1f4648ceb11d9a2a8b4bccab4b8a2a28546b09)
- [PostgreSQL `application_name` documentation](https://www.postgresql.org/docs/current/runtime-config-logging.html#GUC-APPLICATION-NAME)
- [PostgreSQL `pg_stat_activity` documentation](https://www.postgresql.org/docs/current/monitoring-stats.html#MONITORING-PG-STAT-ACTIVITY-VIEW)

**Tip:** React with 👍 to help prioritize this issue, as described in
[GitHub's reaction guidance](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/).
Please use comments to provide useful context instead of `+1` or `me too`; see
the OpenTelemetry community's
[issue participation guidance](https://opentelemetry.io/community/end-user/issue-participation/).

Kielek:
## Note

PR and two detailed tests scenarios Assisted by Codex GPT 5.6 Sol

## Why

Npgsql client spans currently stop at the client boundary. PostgreSQL does not
receive their trace context, so server-side query samples cannot be correlated
with the client span.

The OpenTelemetry Collector Contrib PostgreSQL receiver already recognizes a
complete W3C `traceparent` stored in `application_name` and assigns the extracted
trace and span identifiers to `db.server.query_sample` log records
([receiver implementation](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3456d18100676b2a160809040d84faaace0caeed/receiver/postgresqlreceiver/client.go#L989-L1033),
[generated log builder](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3456d18100676b2a160809040d84faaace0caeed/receiver/postgresqlreceiver/internal/metadata/generated_logs.go#L21-L54)).
This change provides an opt-in compatibility path for that existing receiver
without rewriting user SQL or invalidating prepared statements.

Native, per-operation support in PostgreSQL and Npgsql remains the preferred
long-term solution. When it becomes available end to end, it should take
precedence over this `application_name` compatibility mechanism. See
[npgsql/npgsql#6255](https://github.com/npgsql/npgsql/issues/6255).

Fixes #5338

## What

- Adds the disabled-by-default
  `OTEL_DOTNET_AUTO_NPGSQL_CONTEXT_PROPAGATION` option and its file-based
  configuration equivalent.
- Adds bytecode hooks at the Npgsql activity lifecycle point immediately before
  the database request:
  - `TraceCommandStart` for Npgsql 6;
  - `TraceCommandEnrich` for Npgsql 7-8 on .NET Framework and Npgsql 7-10 on
    .NET;
  - `TraceCopyStart` for Npgsql 10.x on .NET. This shared COPY lifecycle is
    invoked by the text, binary, and raw COPY implementations
    ([pinned Npgsql source](https://github.com/npgsql/npgsql/blob/d3768398c17877b3a916c3c4d87e8e11698991fc/src/Npgsql/Internal/NpgsqlConnector.cs#L2964-L2980)).
- Executes `SET application_name = '<traceparent>'` on the same physical
  connector before a traced request. The internal propagation command is not
  exported as an application span.
- Tracks propagated state per physical connector and executes
  `RESET application_name` before the next untraced command or COPY request,
  preventing a pooled connector from attributing that request to stale context.
- Detects multiplexing, logs one warning, and safely skips propagation. A
  connection-scoped value cannot preserve per-command pairing when multiple
  logical commands are written to one physical connector
  ([pinned multiplexing loop](https://github.com/npgsql/npgsql/blob/d3768398c17877b3a916c3c4d87e8e11698991fc/src/Npgsql/MultiplexingDataSource.cs#L176-L199)).
- Keeps propagation failures isolated from application commands and documents
  the additional round trip, ownership of `application_name`, Collector
  prerequisites, stale-context behavior, and multiplexing limitation.

### Supported database request matrix

`Supported` means the request passes through an instrumented Npgsql activity
lifecycle and receives the activity's W3C `traceparent`. `Reset only` means no
context is injected; if this connector previously carried propagated context,
the configured `application_name` is restored before the request.

| Database request or mode | .NET Framework, Npgsql 6-8.x | .NET, Npgsql 6-8.x | .NET, Npgsql 9.x | .NET, Npgsql 10.x | Coverage notes |
|---|---|---|---|---|---|
| `ExecuteScalar` sync/async | Supported | Supported | Supported | Supported | Sync scalar and async prepared scalar are exercised. |
| `ExecuteReader` sync/async | Supported | Supported | Supported | Supported | Parameterized async reader is exercised; both paths use the same command lifecycle. |
| `ExecuteNonQuery` sync/async | Supported | Supported | Supported | Supported | Async non-query is exercised and read back from the session. |
| Prepared command execution | Supported | Supported | Supported | Supported | Preparation does not rewrite SQL; prepared async execution is exercised. |
| Stored-procedure command execution | Supported | Supported | Supported | Supported | Uses the common command lifecycle; not separately exercised. |
| SQL command inside an explicit transaction | Supported | Supported | Supported | Supported | .NET Framework uses synchronous `BeginTransaction`; .NET uses `BeginTransactionAsync`. |
| `NpgsqlBatch` execution | Supported | Supported | Supported | Supported | A multi-command batch is exercised. |
| Text COPY import/export | Not supported | Not supported | Not supported | Supported | Npgsql 10 introduced the shared COPY activity lifecycle; text export is exercised. |
| Binary COPY import/export | Not supported | Not supported | Not supported | Supported | Uses the same Npgsql 10 `TraceCopyStart` lifecycle; not separately exercised. |
| Raw binary COPY | Not supported | Not supported | Not supported | Supported | Uses the same Npgsql 10 `TraceCopyStart` lifecycle; not separately exercised. |
| Filtered or suppressed command/COPY activity | Reset only | Reset only | Reset only | Reset only | Tests verify that the previous traceparent is not visible to the untraced request. |
| Multiplexed command | Safely skipped | Safely skipped | Safely skipped | Safely skipped | Concurrent tests verify the configured application name is preserved and requests do not hang. |
| Connection open/close, keepalive, prepare/unprepare, and transaction-control protocol operations | Not independently propagated | Not independently propagated | Not independently propagated | Not independently propagated | These operations do not use the instrumented command/COPY activity lifecycle. Executed commands inside a transaction are supported. |

Npgsql 9 and later do not support .NET Framework. Consequently, the .NET
Framework profiler registers only Npgsql 6-8 command definitions and does not
contain the Npgsql 10 COPY integration.

### Propagation flow

```mermaid
flowchart TD
    A["Npgsql database request starts"] --> B{"Propagation enabled?"}
    B -- "No" --> N["Keep existing Npgsql behavior"]
    B -- "Yes" --> C{"Multiplexed connector?"}
    C -- "Yes" --> W["Warn once and skip propagation"]
    C -- "No" --> D{"Instrumented lifecycle?"}
    D -- "Npgsql 6 command" --> E["TraceCommandStart"]
    D -- "Npgsql 7-10 command" --> F["TraceCommandEnrich"]
    D -- "Npgsql 10.x COPY on .NET" --> G["TraceCopyStart"]
    D -- "Other protocol operation" --> U["No independent propagation"]
    E --> H{"Valid W3C Activity ID?"}
    F --> H
    G --> H
    H -- "Yes" --> I["SET application_name = traceparent"]
    I --> R["Execute application request"]
    H -- "No, previous context exists" --> J["RESET application_name"]
    J --> R
    H -- "No previous context" --> R
```

## Tests

New tests added + manual verification with contrib added. See separate commit.

## Checklist


Kielek:
# Npgsql context propagation validation with Collector Contrib

These instructions reproduce a standalone validation of
`OTEL_DOTNET_AUTO_NPGSQL_CONTEXT_PROPAGATION`. They run the repository's Npgsql
test application as a normal process; no xUnit or Nuke test target is used for
the application runs.

The validation was performed on 2026-08-10 from repository revision
`8611f301fd5edd5f8c408f0f5efcd6a82979ba8b`.

## Pinned components

- OpenTelemetry Collector Contrib
  [`v0.158.0`](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.158.0),
  source commit
  [`821a9d9c2c1623c4a0ceba5d47b57c48879c3f84`](https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/821a9d9c2c1623c4a0ceba5d47b57c48879c3f84),
  image digest
  `otel/opentelemetry-collector-contrib@sha256:c5918f78992ee73b0d6f0e599423ac5ec52dd5d9726733114d6eca53d5a32ed5`.
- PostgreSQL 18.4, image
  `postgres:18.4@sha256:8ff36f3c66371cba71d20ceedccfc3de9669a68737607888c4ef0af93abe8e39`.
- Npgsql versions `8.0.9`, `9.0.4`, and `10.0.3`. These are the generated
  [Npgsql test versions](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/8611f301fd5edd5f8c408f0f5efcd6a82979ba8b/test/IntegrationTests/LibraryVersions.g.cs#L269-L284).

Collector Contrib `v0.158.0` documents that query sampling is disabled by
default, requires `pg_monitor`, and maps a valid W3C traceparent in
`application_name` to the emitted record's trace and span IDs. See the pinned
[receiver documentation](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/821a9d9c2c1623c4a0ceba5d47b57c48879c3f84/receiver/postgresqlreceiver/README.md#L110-L141),
[event metadata](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/821a9d9c2c1623c4a0ceba5d47b57c48879c3f84/receiver/postgresqlreceiver/metadata.yaml#L291-L303),
and [query-sampling SQL](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/821a9d9c2c1623c4a0ceba5d47b57c48879c3f84/receiver/postgresqlreceiver/templates/querySampleTemplate.tmpl#L1-L67).

## Prerequisites

- Windows x64 and Docker Desktop using Linux containers.
- .NET SDKs 8, 9, and 10.
- .NET Framework 4.6.2 targeting pack and a runtime capable of running the
  `net462` application.
- PowerShell 7.
- Repository dependencies already restored, or network access for restore.

The following host ports were used because the usual ports overlapped local
reserved ranges:

| Service | Host | Container |
|---|---:|---:|
| PostgreSQL | 15432 | 5432 |
| OTLP/gRPC | 14317 | 4317 |
| OTLP/HTTP | 14318 | 4318 |
| Collector health | 14133 | 13133 |

## 1. Build the current auto-instrumentation distribution

Run this first so that `bin/tracer-home` contains the current managed loader and
native profiler:

```powershell
Set-Location C:\dev\opentelemetry-dotnet-instrumentation
.\build.cmd BuildTracer --configuration Release
```

## 2. Add a pollable standalone scenario

The PostgreSQL receiver polls `pg_stat_activity`; very short queries are not
guaranteed to overlap a collection. The standalone application therefore has a
`--collector-context-propagation-scenario` switch which runs a 750 ms regular
query and, on Npgsql 10+, a 750 ms COPY operation. The scenario is in
`test/test-applications/integrations/TestApplication.Npgsql/Program.cs`.

The regular query is:

```sql
SELECT current_setting('application_name') FROM pg_sleep(0.75)
```

For Npgsql 10+, the COPY operation is:

```sql
COPY (SELECT current_setting('application_name') FROM pg_sleep(0.75)) TO STDOUT
```

Both return the server-visible `application_name` to standard output, which
provides an application-side check in addition to Collector output.

## 3. Configure Collector Contrib

Create `test-artifacts/npgsql-collector-validation/collector-config.yaml`:

```yaml
extensions:
  health_check:
    endpoint: 0.0.0.0:13133

receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
      http:
        endpoint: 0.0.0.0:4318
  postgresql:
    endpoint: postgres:5432
    transport: tcp
    username: postgres
    password: otel-validation
    databases:
      - postgres
    collection_interval: 100ms
    initial_delay: 100ms
    tls:
      insecure: true
    events:
      db.server.query_sample:
        enabled: true
    query_sample_collection:
      max_rows_per_query: 1000

processors:
  batch:
    timeout: 100ms

exporters:
  file/client_spans:
    path: /results/client-spans.json
  file/query_samples:
    path: /results/query-samples.json
  debug/query_samples:
    verbosity: basic

service:
  extensions:
    - health_check
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
      exporters: [file/client_spans]
    logs:
      receivers: [postgresql]
      processors: [batch]
      exporters: [file/query_samples, debug/query_samples]
```

The receiver's supported query-sampling configuration is shown in its pinned
[example](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/821a9d9c2c1623c4a0ceba5d47b57c48879c3f84/receiver/postgresqlreceiver/README.md#L243-L260).

## 4. Start PostgreSQL and Collector Contrib

```powershell
$root = (Get-Location).Path
$results = Join-Path $root 'test-artifacts\npgsql-collector-validation'
$config = Join-Path $results 'collector-config.yaml'

New-Item -ItemType Directory -Force $results | Out-Null
Remove-Item -ErrorAction SilentlyContinue `
  (Join-Path $results 'client-spans.json'), `
  (Join-Path $results 'query-samples.json')

docker network create otel-npgsql-context-validation

docker run -d `
  --name otel-npgsql-postgres-validation `
  --network otel-npgsql-context-validation `
  --network-alias postgres `
  -e POSTGRES_HOST_AUTH_METHOD=trust `
  -p 15432:5432 `
  postgres:18.4@sha256:8ff36f3c66371cba71d20ceedccfc3de9669a68737607888c4ef0af93abe8e39

docker run -d `
  --name otel-npgsql-collector-validation `
  --network otel-npgsql-context-validation `
  -p 14317:4317 `
  -p 14318:4318 `
  -p 14133:13133 `
  --mount "type=bind,source=$config,target=/etc/otelcol-contrib/config.yaml,readonly" `
  --mount "type=bind,source=$results,target=/results" `
  otel/opentelemetry-collector-contrib@sha256:c5918f78992ee73b0d6f0e599423ac5ec52dd5d9726733114d6eca53d5a32ed5
```

The local PostgreSQL container used trust authentication. Collector `v0.158.0`
still validates that `password` is non-empty, so the configuration contains a
non-empty value even though PostgreSQL does not check it in this local setup.

Wait for PostgreSQL and Collector readiness:

```powershell
docker exec otel-npgsql-postgres-validation pg_isready -U postgres
Invoke-WebRequest -UseBasicParsing http://127.0.0.1:14133/
docker logs otel-npgsql-collector-validation --tail 40
```

## 5. Set auto-instrumentation environment variables

```powershell
$tracerHome = Join-Path $root 'bin\tracer-home'
$nativeProfiler = Join-Path $tracerHome 'win-x64\OpenTelemetry.AutoInstrumentation.Native.dll'

$env:CORECLR_ENABLE_PROFILING = '1'
$env:CORECLR_PROFILER = '{918728DD-259F-4A6A-AC2B-B85E1B658318}'
$env:CORECLR_PROFILER_PATH = $nativeProfiler
$env:COR_ENABLE_PROFILING = '1'
$env:COR_PROFILER = '{918728DD-259F-4A6A-AC2B-B85E1B658318}'
$env:COR_PROFILER_PATH = $nativeProfiler
$env:DOTNET_STARTUP_HOOKS = Join-Path $tracerHome 'net\OpenTelemetry.AutoInstrumentation.StartupHook.dll'
$env:OTEL_DOTNET_AUTO_HOME = $tracerHome
$env:OTEL_DOTNET_AUTO_NPGSQL_CONTEXT_PROPAGATION = 'true'
$env:OTEL_DOTNET_AUTO_TRACES_ENABLED_INSTRUMENTATIONS = 'NPGSQL'
$env:OTEL_TRACES_EXPORTER = 'otlp'
$env:OTEL_METRICS_EXPORTER = 'none'
$env:OTEL_LOGS_EXPORTER = 'none'
$env:OTEL_EXPORTER_OTLP_PROTOCOL = 'http/protobuf'
$env:OTEL_EXPORTER_OTLP_ENDPOINT = 'http://127.0.0.1:14318'
$env:OTEL_BSP_SCHEDULE_DELAY = '100'
```

## 6. Build and run every framework/library combination

The effective matrix is:

| Npgsql | net462 | net8.0 | net9.0 | net10.0 |
|---|:---:|:---:|:---:|:---:|
| 8.0.9 | yes | yes | yes | yes |
| 9.0.4 | no | yes | yes | yes |
| 10.0.3 | no | yes | yes | yes |

Npgsql 9+ is not built for .NET Framework. Npgsql 10 COPY is exercised only
where Npgsql exposes the tracing lifecycle used by the bytecode integration.

```powershell
$project = Join-Path $root 'test\test-applications\integrations\TestApplication.Npgsql\TestApplication.Npgsql.csproj'
$applicationRoot = Join-Path $root 'test\test-applications\integrations\bin\TestApplication.Npgsql'
$logRoot = Join-Path $results 'profiler-logs-final'
New-Item -ItemType Directory -Force $logRoot | Out-Null

$matrix = @(
  @{ Framework = 'net462'; Version = '8.0.9' },
  @{ Framework = 'net8.0'; Version = '8.0.9' },
  @{ Framework = 'net9.0'; Version = '8.0.9' },
  @{ Framework = 'net10.0'; Version = '8.0.9' },
  @{ Framework = 'net8.0'; Version = '9.0.4' },
  @{ Framework = 'net9.0'; Version = '9.0.4' },
  @{ Framework = 'net10.0'; Version = '9.0.4' },
  @{ Framework = 'net8.0'; Version = '10.0.3' },
  @{ Framework = 'net9.0'; Version = '10.0.3' },
  @{ Framework = 'net10.0'; Version = '10.0.3' }
)

foreach ($case in $matrix) {
  $framework = $case.Framework
  $version = $case.Version
  $serviceSuffix = "$($framework.Replace('.', '-'))-v$($version.Replace('.', '-'))"

  dotnet build $project `
    --configuration Release `
    --framework $framework `
    -p:TargetFrameworks=$framework `
    -p:Platform=x64 `
    -p:LibraryVersion=$version
  if ($LASTEXITCODE -ne 0) { throw "Build failed for $framework / $version" }

  $env:OTEL_SERVICE_NAME = "npgsql-validation-final-$serviceSuffix"
  $env:OTEL_RESOURCE_ATTRIBUTES = "validation.framework=$framework,validation.npgsql.version=$version"
  $env:OTEL_DOTNET_AUTO_LOG_DIRECTORY = Join-Path $logRoot $serviceSuffix
  New-Item -ItemType Directory -Force $env:OTEL_DOTNET_AUTO_LOG_DIRECTORY | Out-Null

  $output = Join-Path $applicationRoot "$version\x64\Release\$framework"
  if ($framework -eq 'net462') {
    & (Join-Path $output 'TestApplication.Npgsql.exe') `
      --postgres 15432 --collector-context-propagation-scenario
  }
  else {
    dotnet (Join-Path $output 'TestApplication.Npgsql.dll') `
      --postgres 15432 --collector-context-propagation-scenario
  }
  if ($LASTEXITCODE -ne 0) { throw "Run failed for $framework / $version" }
}
```

`-p:TargetFrameworks=$framework` is intentional. Without it, restore evaluates
all project target frameworks and fails when an Npgsql 9 or 10 package is paired
with `net462`, even though the requested build framework is .NET 8+.

## 7. Flush and inspect the output

Stop Collector gracefully so the file exporters flush their final batches:

```powershell
docker stop otel-npgsql-collector-validation
```

The two newline-delimited OTLP JSON files are:

- `test-artifacts/npgsql-collector-validation/client-spans.json`
- `test-artifacts/npgsql-collector-validation/query-samples.json`

For every application operation, verify all of the following:

1. Exactly one client span exists under the expected `service.name`, framework,
   and Npgsql version resource attributes.
2. At least one receiver log has `eventName = db.server.query_sample` and the
   expected normalized `db.query.text`.
3. The receiver log's `traceId` and `spanId` exactly equal the client span's
   identifiers.
4. `postgresql.application_name` is exactly
   `00-{traceId}-{spanId}-01`.
5. The application's `CollectorApplicationName` or
   `CollectorCopyApplicationName` output contains the same value.

Receiver sample counts are deliberately not exact assertions. At a 100 ms
polling interval, a 750 ms operation normally appears several times, and timing
can add or remove an active or recently completed idle sample. The receiver SQL
which produces this behavior is pinned
[here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/821a9d9c2c1623c4a0ceba5d47b57c48879c3f84/receiver/postgresqlreceiver/templates/querySampleTemplate.tmpl#L47-L67).

## 8. Cleanup

```powershell
docker rm -f otel-npgsql-collector-validation
docker rm -f otel-npgsql-postgres-validation
docker network rm otel-npgsql-context-validation
```

Do not delete `test-artifacts/npgsql-collector-validation` until the raw OTLP
JSON and profiler logs are no longer needed.

Kielek:
# Npgsql context propagation validation results

Validation date: 2026-08-10  
Repository revision: `8611f301fd5edd5f8c408f0f5efcd6a82979ba8b`  
Execution method: standalone `TestApplication.Npgsql` processes, not tests

## Outcome

The supported regular-command and Npgsql 10 COPY paths are compatible with the
OpenTelemetry Collector Contrib PostgreSQL receiver in every tested
framework/library combination:

- 10 of 10 standalone application combinations built and exited successfully.
- 13 of 13 operation contexts were present in both the application's exported
  Npgsql client spans and the receiver's `db.server.query_sample` logs.
- Every matched receiver record had the exact client TraceId and SpanId and an
  exact `00-{traceId}-{spanId}-01` value in
  `postgresql.application_name`.
- No `CallTargetInvokerException`, propagation failure, or missing callback was
  found in the final profiler logs.

The overall result is a **qualified pass**, not an unconditional correctness
claim. A separate transaction-control probe found that `COMMIT` can be emitted
with the preceding query's trace context because `application_name` remains set
until another instrumented command resets or replaces it. Details are in
[Observed receiver edge cases](#observed-receiver-edge-cases).

## Environment

| Component | Version or identity |
|---|---|
| Repository | `8611f301fd5edd5f8c408f0f5efcd6a82979ba8b` |
| Collector Contrib | [`v0.158.0`](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.158.0) |
| Collector source | [`821a9d9c2c1623c4a0ceba5d47b57c48879c3f84`](https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/821a9d9c2c1623c4a0ceba5d47b57c48879c3f84) |
| Collector image | `otel/opentelemetry-collector-contrib@sha256:c5918f78992ee73b0d6f0e599423ac5ec52dd5d9726733114d6eca53d5a32ed5` |
| Collector image ID | `sha256:837606a793453fd0c2eef9a6d4ee47ecc970d228ede7bc0c15d32ea9324c9e80` |
| PostgreSQL image | `postgres:18.4@sha256:8ff36f3c66371cba71d20ceedccfc3de9669a68737607888c4ef0af93abe8e39` |
| PostgreSQL image ID | `sha256:a1892662ab56773f96746107054fcfec3d7cd32d39a34fabec4a57f66bd15524` |
| PostgreSQL server | `postgres (PostgreSQL) 18.4 (Debian 18.4-1.pgdg13+1)` |
| .NET Framework runtime | `.NET Framework 4.8.9337.0` |
| .NET 8 runtime | `.NET 8.0.29` |
| .NET 9 runtime | `.NET 9.0.18` |
| .NET 10 runtime | `.NET 10.0.10` |

Collector Contrib `v0.158.0` was the latest published release when validation
was performed. Its pinned PostgreSQL receiver documentation explicitly defines
the [`application_name` traceparent correlation](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/821a9d9c2c1623c4a0ceba5d47b57c48879c3f84/receiver/postgresqlreceiver/README.md#L113-L141).

## Build result

The required command was run before any standalone matrix build:

```text
.\build.cmd BuildTracer --configuration Release
```

Result: succeeded in 1 minute 27 seconds. The current managed distribution and
Windows x64/x86 native profiler were written to `bin/tracer-home`.

Each of the 10 application combinations was then built independently in Release
x64 configuration. All 10 builds completed with zero warnings and zero errors.

## Matrix results

`Receiver samples` is `expected-query samples / all samples carrying that
context`. Multiple samples are normal because the receiver polls every 100 ms.

| Runtime | Npgsql | Operation | Client spans | Receiver samples | TraceId | SpanId | Result |
|---|---:|---|---:|---:|---|---|---|
| net462 | 8.0.9 | command | 1 | 10 / 11 | `8e014fce12bb350ede7838c5e09fa7a1` | `84121d4925429c4e` | pass |
| net8.0 | 8.0.9 | command | 1 | 9 / 9 | `9b048e3a64bca4c739a926d12184bd21` | `42724803be0e1a94` | pass |
| net9.0 | 8.0.9 | command | 1 | 9 / 9 | `ed24205cfd0b6868a4e3551d20fbe22f` | `bcfa810edc49b83f` | pass |
| net10.0 | 8.0.9 | command | 1 | 9 / 9 | `0a98ad128fdf3c11fcf47efe53b169ec` | `ecf6d194d1fcbdc3` | pass |
| net8.0 | 9.0.4 | command | 1 | 8 / 8 | `dabd14f2c81de5033e95ac900c0eb6ba` | `1a758e6a7376e9de` | pass |
| net9.0 | 9.0.4 | command | 1 | 9 / 9 | `6bb43cd48c8ea80fd1c969260ca3c748` | `030f0bc1474000e9` | pass |
| net10.0 | 9.0.4 | command | 1 | 9 / 9 | `d7b6e764f9476b349df8950f9e6b3fe5` | `a37ce498ad2d16b1` | pass |
| net8.0 | 10.0.3 | command | 1 | 8 / 8 | `d6769c82805aa139b232df3a9ec7af36` | `b341f673d7826dd5` | pass |
| net8.0 | 10.0.3 | COPY | 1 | 8 / 8 | `8862be02b6cb3b2cb6c9dc00bfb12eb0` | `01d646d5db6b259e` | pass |
| net9.0 | 10.0.3 | command | 1 | 8 / 8 | `2537651d43e13119570984f6ac009d47` | `c4abb5237b374209` | pass |
| net9.0 | 10.0.3 | COPY | 1 | 7 / 7 | `da64fb8b0fb576d76ccb1b6ac5972e88` | `454245f8d1a919bd` | pass |
| net10.0 | 10.0.3 | command | 1 | 8 / 8 | `27987ec39d6a358047a4fe1a22d42445` | `5c62f508095b802d` | pass |
| net10.0 | 10.0.3 | COPY | 1 | 8 / 8 | `2dada4a029512ccaa0ab7ad38c8c147c` | `8c1ee5d0028b8e87` | pass |

The expected normalized receiver query texts were:

```text
SELECT current_setting ( ? ) FROM pg_sleep ( ? )
COPY ( SELECT current_setting ( ? ) FROM pg_sleep ( ? ) ) TO STDOUT
```

Samples included active rows and, depending on polling timing, a recently
completed idle row. This matches the receiver's pinned
[`pg_stat_activity` query](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/821a9d9c2c1623c4a0ceba5d47b57c48879c3f84/receiver/postgresqlreceiver/templates/querySampleTemplate.tmpl#L12-L24)
and its state/time filter
[`WHERE` clause](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/821a9d9c2c1623c4a0ceba5d47b57c48879c3f84/receiver/postgresqlreceiver/templates/querySampleTemplate.tmpl#L47-L67).

## What was verified

For each row in the table:

1. The client trace exporter contained one Npgsql client span for the unique
   service/framework/library resource tuple.
2. The receiver log exporter contained at least one
   `eventName = db.server.query_sample` record for the expected normalized SQL.
3. The log record's OTLP `traceId` and `spanId` exactly matched the client span.
4. The log attribute `postgresql.application_name` exactly matched the W3C
   value `00-{traceId}-{spanId}-01`.
5. The application itself read the same value from PostgreSQL.

This is the same correlation behavior specified in Collector Contrib's pinned
[receiver README](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/821a9d9c2c1623c4a0ceba5d47b57c48879c3f84/receiver/postgresqlreceiver/README.md#L130-L138).

## Observed receiver edge cases

### Propagation statement can be sampled

The net462/Npgsql 8.0.9 command context had 10 correct query records and one
additional idle record:

```text
db.query.text = SET application_name = ?
postgresql.application_name = 00-8e014fce12bb350ede7838c5e09fa7a1-84121d4925429c4e-01
traceId = 8e014fce12bb350ede7838c5e09fa7a1
spanId = 84121d4925429c4e
```

This is expected with polling: `SET application_name` is itself a PostgreSQL
statement and can briefly be the backend's last completed query. The desired
application query was also sampled correctly 10 times. Consumers should not
assume that every receiver record carrying a client context represents the
client span's application SQL; they should compare `db.query.text` as well.

### Transaction-control statement can inherit the preceding command context

A standalone net8.0/Npgsql 10 smoke run using the existing
`--context-propagation-scenario` produced this receiver record:

```text
db.query.text = COMMIT
postgresql.state = idle
postgresql.application_name = 00-eff772e610676c0034ce9a8e4ed1dbd6-4a0befdabdae651b-01
traceId = eff772e610676c0034ce9a8e4ed1dbd6
spanId = 4a0befdabdae651b
```

The matching client span was for:

```sql
SELECT current_setting('application_name') /* transaction */
```

There was no matching client `COMMIT` span. The receiver therefore correlated
the transaction-control statement to the preceding application query. This can
happen because the current implementation leaves `application_name` in place
until the next traced command replaces it or the next untraced command resets
it, while `COMMIT` is sent outside the instrumented command tracing hook.

This is a real correctness gap for transaction-control query samples. It does
not invalidate the regular query or COPY matches in the matrix, but it must be
resolved or explicitly accepted before claiming that every receiver query
sample is correctly correlated.

## Collector diagnostics

Collector started as `service.version = 0.158.0`, reported `Everything is
ready`, exported query-sample logs throughout the run, and shut down cleanly
with exit code 0.

No Collector errors were logged. One unrelated warning stated that the
`receiver.postgresql.separateSchemaAttr` feature gate was not enabled. The
PostgreSQL receiver also identifies itself as a Development component, which is
consistent with its generated event stability.

## Raw local evidence

The raw evidence remains under
`test-artifacts/npgsql-collector-validation/`:

| Artifact | Size |
|---|---:|
| `client-spans.json` | 73,413 bytes |
| `query-samples.json` | 384,215 bytes |
| `profiler-logs-final/` | per-process managed/native profiler logs |
| `collector-config.yaml` | 1,085 bytes |

These files are ignored test artifacts and are intentionally not proposed as
repository changes. Reproduction commands and the exact verification criteria
are in `npgsql-collector-validation-instructions.md`.

test/test-applications/integrations/TestApplication.Npgsql/Program.cs
```diff
@@ -16,14 +21,373 @@ public static async Task Main(string[] args)
...
+            WriteValue("FailedTransactionBackendProcessId", failedTransactionBackendProcessId);
+            WriteValue("FailedTransactionUntracedBackendProcessId", result.BackendProcessId);
+        }
+    }
+
+    private static async Task RunCollectorContextPropagationScenarioAsync(NpgsqlConnection connection)
+    {
+        using (var command = new NpgsqlCommand(
+                   "SELECT current_setting('application_name') FROM pg_sleep(0.75)",
+                   connection))
```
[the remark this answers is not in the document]
Kielek: false positive IMO.

RassK:
Can you add any performance deltas (with propagation enabled and disabled)? I see injected code contains locks and not sure if this causes any pain to db performance. This can change the npgsql execution model entirely if injected into a hot path / working thread.

Kielek:
> Can you add any performance deltas (with propagation enabled and disabled)? I see injected code contains locks and not sure if this causes any pain to db performance. This can change the npgsql execution model entirely if injected into a hot path / working thread.

I measured this with Npgsql 10.0.3, .NET 10, and PostgreSQL 17.8, using seven independent process runs per mode. For a sequential SELECT 1, enabling propagation increased mean latency from 416.8 µs to 1,193.2 µs (+186%) and reduced throughput by 65.1%. With 16 independent connectors, throughput decreased by 61.9%. For a query with 5 ms of server work, the mean increase was 14.7%.
To isolate the lock, I also built a controlled variant that retained both per-connector locks and all propagation bookkeeping but suppressed the SET/RESET execution. That path added 13.8 µs sequentially and 9.4 µs with 16 connectors, approximately 1–3%. The extra PostgreSQL roundtrips account for an estimated 98–99% of the total overhead.
The lock is per physical connector rather than global, and Npgsql already prevents concurrent user actions on a non-multiplexed connector. However, the roundtrip concern is real: an enabled operation performs synchronous SET, application query, and RESET exchanges. The feature is opt-in and disabled by default, but this cost should be explicitly documented. Native client/server propagation support remains the preferred long-term solution.

docs/config.md
```diff
@@ -289,6 +293,40 @@ was only available for stored procedures. With this setting enabled, it is also
 available for raw queries. This changes the behavior of events emitted by the
 [`SqlEventSource`](https://github.com/dotnet/SqlClient/blob/v1.0.19239.1/src/Microsoft.Data.SqlClient/netfx/src/Microsoft/Data/SqlClient/SqlCommand.cs#L6369),
 which might impact other parts of the application if this mechanism is used.
+\[3\]: The OpenTelemetry Collector Contrib PostgreSQL receiver exposes the
```
martincostello, line 296: IMHO this footnote is quite hard to read when rendered.

docs/config.md
martincostello, line 1: Does the performance penalty of enabling this option need to be documented somewhere?

src/OpenTelemetry.AutoInstrumentation/Instrumentations/Npgsql/NpgsqlTraceContextPropagator.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        for (var i = 3; i < value.Length; i++)
+        {
+            if (i is 35 or 52)
+            {
+                continue;
+            }
+
+            var character = value[i];
+            if (character is not (>= '0' and <= '9') and not (>= 'a' and <= 'f'))
```
martincostello, line 233: Could consider using `!char.IsAsciiHexDigitLower()` for .NET 8+.

martincostello (approved):
A few minor comments but otherwise looks OK to me.

In a follow-up, if there already isn't something similar in place, adding some integration tests with the collector to avoid the need for manual testing (at least for a basic smoke test) would be good.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/Npgsql/NpgsqlConstants.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    public const string IntegrationName = nameof(TracerInstrumentation.Npgsql);
+    public const string AssemblyName = "Npgsql";
+    public const string CommandTypeName = "Npgsql.NpgsqlCommand";
+    public const string ConnectorTypeName = "Npgsql.Internal.NpgsqlConnector";
+    public const string TraceCommandStartMethodName = "TraceCommandStart";
+    public const string TraceCommandEnrichMethodName = "TraceCommandEnrich";
+    public const string EndUserActionMethodName = "EndUserAction";
+    public const string MinVersion = "6.0.0";
+    public const string TraceCommandStartMaxVersion = "6.*.*";
```
lachmatt: We might need to adjust the ranges - current ranges will leave some of the versions uninstrumented.
Consider the following:
Npgsql versions | Connector-taking method
-- | --
6.0.0–6.0.11 | TraceCommandStart
6.0.12–6.0.13 | TraceCommandEnrich
7.0.0–7.0.7 | TraceCommandStart
7.0.8+ | TraceCommandEnrich
8.0.0–8.0.3 | TraceCommandStart
8.0.4+ | TraceCommandEnrich

nrcventura (approved):

docs/config.md
Kielek: Agree, but it requires redesignign documentation IMO. I will keep as is in this PR.

docs/config.md
Kielek: I think that current wording is sufficient, but if you see option to improvement, I open in this PR/follow up.

src/OpenTelemetry.AutoInstrumentation/Instrumentations/Npgsql/NpgsqlTraceContextPropagator.cs
Kielek: da5647177452da20835f9e3dd3a0c500f412a391

src/OpenTelemetry.AutoInstrumentation/Instrumentations/Npgsql/NpgsqlConstants.cs
Kielek: Fixed

Issue #5360, opened 2026-08-13
Design process-wide lifecycle and dynamic configuration contract for continuous profiling


eftiquar:
# Feature Request

## Context

The OpAMP runtime-configuration work in #5312 has exposed a broader architectural requirement in the continuous-profiler service.

The service was originally designed around a start-once, seed-once lifecycle. Dynamic configuration introduces additional requirements:

- enabling and disabling profiling features at runtime;
- changing CPU, selective-thread, and allocation-sampling settings atomically;
- preserving process-wide configuration across .NET Framework AppDomain initialization and unload;
- defining precedence between startup Seed and ControlPlane configuration;
- preventing multiple AppDomain-local exporters from competing for process-wide native samples;
- recovering the managed exporter and canary after their owning AppDomain unloads;
- preserving compatibility between different managed and native package versions.

These are process-wide service semantics and should be established independently of the OpAMP transport and managed exporter implementation.

## Proposed direction

Define a protocol-neutral native runtime-service contract that provides:

- one-time, source-neutral bootstrap;
- complete atomic runtime-configuration application;
- `ControlPlane > Seed > None` authority precedence;
- last-known-good behavior for invalid or stale updates;
- lazy producer activation and safe quiescence;
- explicit capture, disable, and terminal-shutdown boundaries;
- opaque managed-host ownership tokens;
- AppDomain-unload revocation and demand-driven host recovery;

Managed code would remain responsible for interpreting and merging OpAMP configuration, constructing the exporter pipeline and canary, and invoking this native contract.

Although the proposed contract is comprehensive, much of that complexity already exists in the runtime environment and is currently implicit. This is particularly true on .NET Framework, where process-wide native profiling state must coexist with managed components whose AppDomains can arrive and depart independently. Exporter ownership, AppDomain unload, canary recruitment and invalidation, competing initialization attempts, and recovery after owner loss all require defined semantics. The purpose of this design is to make those lifecycle rules explicit, deterministic, and independently testable before dynamic OpAMP integration depends on them.
## Proposed delivery

Split delivery horizontally:

1. Implement and test the native runtime-service lifecycle and interop contract.
2. Connect the managed OpAMP control plane and exporter pipeline to that contract.

This allows native concurrency, ownership, authority, and lifecycle semantics to be reviewed and tested independently before adding managed protocol integration.

## Design draft

The attached document is a working design proposal intended to support discussion and refinement. 

[...]

PR #5377, opened 2026-08-20
[Native] Dynamic continuous profiler

ysolomchenko:
## Why

Towards #5360

## What

Native includes:
- atomic V1 configuration/state ABI;
- dynamic enable/update/disable/re-enable capabilities;
- shared CPU/selective worker modes;
- a dedicated EventPipe lifecycle allocation profiler;
- actual interval metadata in V2 batches while preserving the legacy ABI;
- safe bootstrap and terminal shutdown;
- protection against partial stacks, buffer overflows, and callback races.

## Tests

Ci

## Checklist

### DIFF
--- a/src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
@@ -1043,9 +1043,11 @@ static bool CaptureFunctionIdentifiersForThreads(
         const HRESULT                            hr = stackWalker->CaptureStacks(selectedThreads, &request);
         if (FAILED(hr))
         {
-            trace::Logger::Debug("Stack capture failed. HRESULT=0x", std::setfill('0'), std::setw(8), std::hex, hr);
-            threadStacksBuffer.clear();
-            return false;
+            // CaptureStacks reports an aggregate failure when any requested thread cannot be captured.
+            // Preserve stacks captured for other threads and let the non-empty check below discard cycles
+            // that produced no usable stacks.
+            trace::Logger::Debug("Some thread stacks could not be captured. HRESULT=0x", std::setfill('0'),
+                                 std::setw(8), std::hex, hr);
         }
 
         return true;
--- a/src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
+++ b/src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
@@ -1444,39 +1444,61 @@ void ContinuousProfiler::SamplingThreadMain() noexcept
         while (true)
         {
             RuntimeSamplerConfiguration cycleConfiguration;
+            std::uint64_t               cycleGeneration;
             {
                 std::unique_lock<std::mutex> lock(sampling_state_mutex_);
-                if (shutdown_requested_.load(std::memory_order_acquire) || thread_sampling_stop_requested_)
+                if (shutdown_requested_.load(std::memory_order_acquire))
                 {
-                    return;
+                    break;
                 }
 
-                cycleConfiguration = configuration_;
-                if (!cycleConfiguration.HasThreadSampling())
+                if (!periodic_thread_sampling_interval_.has_value() && !selective_thread_sampling_interval_.has_value())
                 {
-                    return;
+                    const auto quiescentGeneration                = thread_configuration_generation_;
+                    acknowledged_thread_configuration_generation_ = quiescentGeneration;
+                    sampling_state_cv_.notify_all();
+                    sampling_state_cv_.wait(lock,
+                                            [this, quiescentGeneration]
+                                            {
+                                                return shutdown_requested_.load(std::memory_order_acquire) ||
+                                                       thread_configuration_generation_ != quiescentGeneration;
+                                            });
+                    continue;
                 }
 
-                if (observedGeneration != configuration_generation_)
+                cycleConfiguration = {periodic_thread_sampling_interval_, selective_thread_sampling_interval_,
+                                      std::nullopt};
+                cycleGeneration    = thread_configuration_generation_;
+
+                if (observedGeneration != cycleGeneration)
                 {
-                    observedGeneration = configuration_generation_;
+                    observedGeneration = cycleGeneration;
                     iteration          = 0;
                 }
 
                 const unsigned int sleepTime = GetSleepTime(cycleConfiguration);
                 if (sleepTime == 0)
                 {
-                    trace::Logger::Warn("Unexpected sampling interval configured, exiting sampling thread.");
-                    return;
+                    trace::Logger::Warn("Unexpected sampling interval configured, quiescing sampling thread.");
+                    sampling_state_cv_.wait(lock,
+                                            [this, cycleGeneration] {
+                                                return shutdown_requested_.load(std::memory_order_acquire) ||
+                                                       thread_configuration_generation_ != cycleGeneration;
+                                            });
+                    continue;
                 }
 
                 if (sampling_state_cv_.wait_for(lock, std::chrono::milliseconds(sleepTime),
-                                                [this] {
+                                                [this, cycleGeneration] {
                                                     return shutdown_requested_.load(std::memory_order_acquire) ||
-                                                           thread_sampling_stop_requested_;
+                                                           thread_configuration_generation_ != cycleGeneration;
                                                 }))
                 {
-                    return;
+                    if (shutdown_requested_.load(std::memory_order_acquire))
+                    {
+                        break;
+                    }
+                    continue;
                 }
             }
 
@@ -1501,7 +1523,7 @@ void ContinuousProfiler::SamplingThreadMain() noexcept
 
                 if (IsShutdownRequested())
                 {
-                    return;
+                    break;
                 }
 
                 const auto now = std::chrono::steady_clock::now();
@@ -1558,6 +1580,12 @@ void ContinuousProfiler::SamplingThreadMain() noexcept
         {
         }
     }
+
+    {
+        std::lock_guard<std::mutex> lock(sampling_state_mutex_);
+        thread_sampling_worker_exited_ = true;
+    }
+    sampling_state_cv_.notify_all();
 }
 
 void ContinuousProfiler::SetGlobalInfo7(ICorProfilerInfo7* cor_profiler_info7)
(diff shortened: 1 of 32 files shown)


src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -208,7 +233,8 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
     tracer_integration_preprocessor = std::make_unique<TracerRejitPreprocessor>(rejit_handler, work_offloader);
 
     DWORD event_mask = COR_PRF_DISABLE_TRANSPARENCY_CHECKS_UNDER_FULL_TRUST | COR_PRF_MONITOR_MODULE_LOADS |
-                       COR_PRF_MONITOR_ASSEMBLY_LOADS | COR_PRF_MONITOR_APPDOMAIN_LOADS | COR_PRF_ENABLE_REJIT;
+                       COR_PRF_MONITOR_ASSEMBLY_LOADS | COR_PRF_MONITOR_APPDOMAIN_LOADS | COR_PRF_ENABLE_REJIT |
+                       COR_PRF_MONITOR_THREADS | COR_PRF_ENABLE_STACK_SNAPSHOT;
```
eftiquar: In original implementation,  continuous-profiler event masks and native machinery are initialized lazily from ConfigureContinuousProfiler, which is invoked only when the exporter plugin is installed and sampling is enabled. Now we moved setting of COR_PRF_MONITOR_THREADS, COR_PRF_ENABLE_STACK_SNAPSHOT, and runtime_sampler_controller_.Prepare() into unconditional profiler initialization. This is undesirable because the sampling may never be enabled, we do not know yet. If it is not, the application subscribes to these events in CLR withut any real value. Both event-mask configuration and native bootstrap should remain behind the first valid configuration that enables a sampling feature

eftiquar (changes requested):

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
eftiquar: We desire zero footprint of cont profiling until it is explicitly enabled.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -1266,30 +1614,210 @@ IStackWalker* ContinuousProfiler::GetStackWalker() const
...
+    for (const auto& threadStateEntry : managed_tid_to_state_)
+    {
+        delete threadStateEntry.second;
+    }
+    managed_tid_to_state_.clear();
+}
+
+bool ContinuousProfiler::ApplyConfiguration(const RuntimeSamplerConfiguration& candidate)
+{
+    std::unique_lock<std::mutex> transitionLock(configuration_transition_mutex_);
```
eftiquar: Thread joins and all the machinery is making the code complex and unreadable - 
We explicitly chose a different model in our design [#53360](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/5360)
 
>> Quiescent worker: A previously created CPU/selective sampling worker retained
>> in a non-producing wait state. Re-enabling wakes it; terminal shutdown joins it.

eftiquar (changes requested):

eftiquar (commented):
With lazy one-time construction and configuration-driven quiescence, many of these states and their associated synchronization, error paths, ABI results, and tests disappear. I believe restarting from the smaller lifecycle contract would produce a substantially more reviewable and maintainable change. Currently there are too many moving parts, the size of the change is not itself the concern; the concern is that much of it supports lifecycle states outside the required contract. In particular, preparation while disabled, repeated worker construction and joining, transition rollback, and duplicated controller/service state substantially expand both the implementation and its concurrency state space. 
The clearest signal is that a function named ApplyConfiguration requires staged threads, ownership transfer, joining, rollback, and several locks. That is not inherent complexity in applying configuration—it is complexity created by the chosen lifecycle model.

eftiquar:
Before we proceed I have few suggestions based on my recce of how this will unfold - each suggestion is backed by the respective problem it addresses 

1. Ordered control-plane configuration
Every OpAMP profiler configuration must carry (schemaVersion, sourceVersion). Initially, sourceVersion is a Unix-epoch timestamp in nanoseconds, assigned by the OpAMP backend. Native uses it as a pseudo-monotonic stale-write guard.
This is not ideal but solves the issue of stale writes subject to time skew. The ideal solution is for OpAMP backend to stamp is config push with monotonically increasing version stamp. 

2. Durable restoration of the last committed control-plane configuration
Managed code must persist the last successfully committed OpAMP configuration together with its schema version, source version, and OpAMP hash. After process restart, it must restore this snapshot before depending on server synchronization, which may be delayed or unavailable.
The restored configuration serves as the restart bootstrap, but it retains control-plane authority—it must not be reclassified as seed.
3. Complete configuration replacement
Every OpAMP update must contain the complete intended profiler configuration. Native validates and atomically replaces the entire snapshot. There is no patch merging, expected-generation exchange, conflict-resolution merge, or retry loop. Stale, conflicting, or invalid submissions preserve the last-good configuration.

Let me know your thoughts.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -1180,109 +1191,291 @@ void CorProfiler::InternalAddInstrumentation(WCHAR* id, CallTargetDefinition* it
...
+        }
+        catch (...)
+        {
+        }
+    }
+#endif
 }
 
-void CorProfiler::ConfigureContinuousProfilerInternal(const ContinuousProfilerParams& params)
+bool CorProfiler::TryEnterContinuousProfilerCallback(std::shared_lock<std::shared_mutex>& callbackLock) noexcept
```
eftiquar: TryEnterContinuousProfilerCallback is a concrete example of complexity introduced by the current lifetime model. It uses an atomic flag, a shared mutex, a mutable RAII-lock output parameter, and a second flag check to protect a raw callback pointer that may be destroyed. Under the agreed start-once/quiesce-resume contract, the callback facade remains stable and only sampling admission changes, so this lifetime gate should disappear rather than be incrementally refined.

eftiquar (commented):
Increasing complexity and drift from the proposed design.

ysolomchenko:
> Before we proceed I have few suggestions based on my recce of how this will unfold - each suggestion is backed by the respective problem it addresses
> 
> 1. Ordered control-plane configuration
>    Every OpAMP profiler configuration must carry (schemaVersion, sourceVersion). Initially, sourceVersion is a Unix-epoch timestamp in nanoseconds, assigned by the OpAMP backend. Native uses it as a pseudo-monotonic stale-write guard.
>    This is not ideal but solves the issue of stale writes subject to time skew. The ideal solution is for OpAMP backend to stamp is config push with monotonically increasing version stamp.
> 2. Durable restoration of the last committed control-plane configuration
>    Managed code must persist the last successfully committed OpAMP configuration together with its schema version, source version, and OpAMP hash. After process restart, it must restore this snapshot before depending on server synchronization, which may be delayed or unavailable.
>    The restored configuration serves as the restart bootstrap, but it retains control-plane authority—it must not be reclassified as seed.
> 3. Complete configuration replacement
>    Every OpAMP update must contain the complete intended profiler configuration. Native validates and atomically replaces the entire snapshot. There is no patch merging, expected-generation exchange, conflict-resolution merge, or retry loop. Stale, conflicting, or invalid submissions preserve the last-good configuration.
> 
> Let me know your thoughts.

Thanks for the suggestions. I have a few concerns, mainly around points 1 and 2.

1. Ordered control-plane configuration

I'm not sure we need an additional (schemaVersion, sourceVersion) ordering mechanism on the profiler side.

OpAMP already provides the control-plane transport and sequencing between the client and server. If there is a real scenario where an older Remote Config can be delivered after a newer one and still needs to be rejected by the agent, I think this should ideally be handled by the OpAMP layer rather than introducing another pseudo-monotonic ordering mechanism in the profiler.

Using a backend-generated Unix timestamp also feels problematic because it introduces clock-skew concerns without giving us true ordering guarantees. Before adding this complexity, I think we should identify a concrete case that cannot already be handled by the existing OpAMP protocol/client.

2. Durable restoration of the last committed control-plane configuration

I don't think we should persist Remote Config on disk.

As far as I understand, there is no requirement in the OpAMP specification that the agent must preserve Remote Config across a process restart. In fact, our current model is the opposite:

`process restart -> startup/initial profiler configuration -> OpAMP reconnects -> Remote Config is sent again -> profiler is reconfigured`

After a restart, the OpAMP client starts a new connection/session, and the server can detect the new state, request/report full state if needed, and resend the appropriate Remote Config.

Persisting the control-plane configuration locally introduces additional questions around stale state, file lifetime, permissions, corruption, multiple processes, upgrades, etc., without a clear protocol requirement for it. I would prefer not to introduce durable state unless we have a concrete requirement from the OpAMP specification/server behavior that makes it necessary.

3. Complete configuration replacement

This part makes sense to me.

Having every Remote Config contain the complete intended profiler configuration and atomically replacing the current snapshot keeps the model simple. If validation or application fails, we should keep the last successfully applied configuration rather than partially applying or merging changes.

---

So overall, I agree with the complete-replacement/last-good behavior, but I think the ordering and especially disk persistence need stronger justification before we add that complexity.

eftiquar:
>> I think we should identify a concrete case that cannot already be handled by the existing OpAMP protocol/client.

What happens if stale config arrives after fresh config is committed - what mechanism is used to detect stale arrival. 
I agree that native should not invent a pseudo-monotonic timestamp; however, the managed control plane needs either a documented OpAMP guarantee that later delivery is authoritative or a server-issued monotonic revision embedded in the configuration. Can we identify which guarantee the current OpAMP implementation provides?
The problematic scenario is semantic staleness rather than network reordering: B may be the newer configuration, yet an older A may be delivered afterward. 

>>Persistence across restarts. 

Is it acceptable that we run with seed config if restart happens after OpAmp config is applied and OpAmp server is unavailable. This needs to be spelled out -sampling will fallback to seeded config in this scenario.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.h
```diff
@@ -32,58 +34,50 @@ class ContinuousProfiler;
...
-{
-    bool         threadSamplingEnabled;
-    unsigned int threadSamplingInterval;
-    bool         allocationSamplingEnabled;
-    unsigned int maxMemorySamplesPerMinute;
-    unsigned int selectedThreadsSamplingInterval;
-};
-
-class CorProfiler : public CorProfilerBase
+class CorProfiler : public CorProfilerBase, private continuous_profiler::IRuntimeSamplerLifecycle
```
eftiquar, line 37: This is concerning - our design goal is - zero footprint for sampler until it is enabled.
Moreover the cor_profiler is already too crowded - my recommendation - create a facade for sampling service and hide all the sampling service lifecycle and config management behind the facade. The facade accepts config, manages lifecycle and cor_profiler bootstraps the facade only if sampling is enabled. Only create facade if at least one service is enabled.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -954,6 +964,7 @@ HRESULT STDMETHODCALLTYPE CorProfiler::ProfilerDetachSucceeded()
         return S_OK;
     }
 
+    runtime_sampler_controller_.Shutdown();
```
eftiquar, line 967: ShutDown may reach EventPipeStopSession - the CLR prohibits calling any methods from detach that will enter profiler callbacks. This is theoretical hazard. 
But practical problem is we set COR_PRF_DISABLE_TRANSPARENCY_CHECKS_UNDER_FULL_TRUST which belongs to immutable profiler flag class, that means the profiler detach is not supported for our profiler - also we rewrite IL and call SetILFunctionBody, thus our profiler code is injected in the process address space and we cannot undo that safely. CLR has in built guard rail for this - any profiler that does any of these operations is not detachable. 
This makes our life easy - we do not need to handle detach, only shut down. It simplifies our cleanup.

eftiquar (commented):
Still reviewing

src/OpenTelemetry.AutoInstrumentation.Native/runtime_sampler_controller.cpp
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    shutdown_requested_ = true;
+    if (bootstrap_attempted_)
+    {
+        lifecycle_.ShutdownSampling();
+    }
+}
+
+RuntimeSamplerApplyResult RuntimeSamplerController::ApplyConfigurationLocked(
```
eftiquar, line 53: The config application is done synchronously under lock - this induces changes in sampling core - I would have a facade that accepts the config from managed and then lazy pull from sampler thread will propagate the config. As much as possible we will aim to preserve the core sampling algorithm.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -101,61 +113,113 @@ static std::mutex name_cache_lock = std::mutex();
 
 static std::shared_mutex profiling_lock = std::shared_mutex();
 
-static ICorProfilerInfo7* profiler_info; // After feature sets settle down, perhaps this should be refactored and have
-                                         // a single static instance of ThreadSampler
+// After feature sets settle down, perhaps this should be refactored and have a single static instance of ThreadSampler.
+static std::atomic<ICorProfilerInfo7*> profiler_info{nullptr};
+static std::atomic_bool                profiler_callbacks_enabled{false};
```
eftiquar, line 118: This lock and associated machinery can be simplified and isolated from ContinuousProfiler.

eftiquar: The invariant becomes:
- The facade is published once.
- Capture objects are created before CLR sampling callbacks are enabled.
- Once created, those objects remain at stable addresses until process termination.
- Ordinary disablement only quiesces producers; it does not destroy callback targets.
- Terminal shutdown closes admission, aborts capture, stops EventPipe, and joins the worker.
- Profiler detachment is unsupported, so no live object withdrawal is required.

eftiquar (commented):
The sampling machinery needs to be kept intact as much as possible, right now we have introduced changes that make it hard to reason about config and sampler seam.

PR #5397, opened 2026-08-28
Ibmmq xms instrumentation

Aryan16138:
## Why

There is currently no IBM MQ support in this distribution. XMS.NET (`IBMXMSDotnetClient`) is IBM's
messaging API for .NET, and applications using it get no messaging spans today, so a queue hop shows
up as a gap in the trace. This adds producer and consumer instrumentation for it, including context
propagation so a trace continues across the queue.

Fixes #

## What

New bytecode instrumentation for IBM MQ via XMS.NET, enabled with the `Xms` instrumentation id.
It targets the `amqmxmsstd` assembly (namespace `IBM.XMS`) for versions 9.0.0 through 10.x.

Instrumented methods:

* `XmsMessageProducerImpl.Send`, all four overloads, produces a Producer span
* `XmsMessageConsumerImpl.Receive()`, `Receive(long)` and `ReceiveNoWait()`, produce Consumer spans
* `XmsProviderMessageListener.OnMessage(ProviderMessage)`, produces a Consumer span for async
  delivery through a `MessageListener`

Spans follow the messaging semantic conventions and carry `messaging.system` (`ibmmq`),
`messaging.operation`, `messaging.destination.name` and `messaging.message.id`. Span names are
`{destination} publish`, `{destination} receive` and `{destination} deliver`. The producer message id
is read after `Send` returns, because JMS assigns `JMSMessageID` during the send rather than before.

W3C trace context is injected into and extracted from XMS string message properties, so the consumer
span becomes a child of the producer span. Property names are passed through a sanitiser first,
because JMS property names have to be valid Java identifiers. `traceparent` and `tracestate` are
already valid so they are unchanged, but a propagator using something like `X-B3-TraceId` would
otherwise fail to set the property.

This also includes one change outside the new instrumentation. `DuckType.CreateMethods` built its
fallback candidate list from a raw `GetMethods` call on the target type, which for an interface does
not include members inherited from base interfaces. Duck typing against `IBM.XMS.IMessage` needs
`SetStringProperty` and `GetStringProperty`, both of which are declared on a base interface, so those
members were invisible. The flattened interface list is now used when the target type is an
interface, and the previous behaviour is kept for classes and structs. Happy to split this into its
own PR if you would prefer to review it separately.

Synchronous `Receive` spans are created after the call returns, so they do not measure the blocking
wait. Transacted `Commit`/`Rollback` and `IQueueBrowser` are not instrumented.

## Tests

Added `TestApplication.Xms` and `XmsTests.SubmitsTraces`. The test application exercises all three
paths in a single run, a `Send`, a synchronous `Receive`, and async delivery through a named
`MessageListener` target. The test asserts the three spans, their messaging attributes, and that the
consumer span shares the producer's trace id, which is the assertion that actually covers the
property get and set path end to end.

The broker comes from the IBM MQ developer image through Testcontainers, reusing the `DEV.QUEUE.*`
queues, the `DEV.APP.SVRCONN` channel and the `app` user that the image already creates.

 Verified the behaviour by hand against a live queue manager, running an uninstrumented
XMS application under the profiler and exporting over OTLP. Producer and consumer span counts matched
one to one over about 5000 messages, and traces contained the expected producer and consumer pair with
the consumer's parent id equal to the producer's span id.

.NET Framework is untested. The instrumentation itself is not framework specific, but XMS dropped its
`netstandard2.0` library at 9.4.0, so .NET Framework would need XMS 9.2.x or 9.3.x and that
combination is not covered here.

## Checklist


Aryan16138:
Closing this PR. Clean Changes are in a new PR: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/5404

PR #5400, opened 2026-09-01
Make the native continuous profiler dynamically configurable

eftiquar:
## Why

The native continuous profiler is process-wide, while managed startup may run more than once, particularly across .NET Framework AppDomains. The existing start-once, seed-once shape cannot safely support runtime enablement, disablement, and full configuration refresh without allowing later startup calls to overwrite process-wide state.

This PR implements the native half of the horizontal split described in #5360. Managed OpAMP sequencing and exporter integration will consume this contract in a follow-up PR.

## How 
This redesign was human-directed and emerged from the review of the earlier implementation. I defined the quiescent lifecycle and eventual-consistency model, configuration authority boundaries, dependency-DAG ordering for activation and shutdown, and the simplified full-snapshot native contract. Codex assisted with implementation, testing, and iterative review; I reviewed and directed the resulting changes

## What

- Add a process-wide runtime sampler facade that owns configuration and lifecycle.
- Accept complete V1 configuration snapshots with explicit Seed or ControlPlane authority. The first valid Seed wins; ControlPlane supersedes Seed; later Seed attempts cannot overwrite active state.
- Keep `ConfigureContinuousProfiler` as the Seed-compatible adapter.
- Create sampling infrastructure only on first enablement. Later disablement quiesces the thread sampler without destroying it, and re-enablement reuses the parked worker.
- Publish thread configuration for the sampler loop to pull at a sampling boundary; update the allocation cap atomically.
- Close allocation admission before stopping EventPipe.
- Define capture boundaries explicitly: ordinary disable completes and publishes the admitted cohort before quiescing, while terminal shutdown aborts at a frame boundary and discards unpublished partial work.
- Preserve the existing core sampling algorithm behind the facade and expose V1 apply/query entry points plus the V2 thread-sample reader.

Managed integration remains responsible for supplying complete configuration snapshots and rejecting stale OpAMP documents before calling the native API.

## Tests

- Release x64 native static-library build
- Release x64 profiler DLL build and export-table verification
- Release x64 native test build
- Native unit tests: 111/111 passed
- `scripts/format-native.sh` passed in a clean detached worktree
- Mutation-checked lazy-quiesce regression: an eager-abort implementation fails the admitted-cohort test
- Non-Windows source and CMake audit; platform builds remain covered by CI

The solution-wide `dotnet format --verify-no-changes` command could not complete locally because the checkout does not contain the locally built `bin/nuget-artifacts` feed. This PR does not change managed C# source; CI remains authoritative for that gate.

## Checklist


src/OpenTelemetry.AutoInstrumentation.Native/runtime_sampler_configuration.h
```diff
@@ -0,0 +1,105 @@
...
+    bool AnyFeatureEnabled() const noexcept;
+
+    bool operator==(const RuntimeSamplerConfigurationV1& other) const noexcept;
+    bool operator!=(const RuntimeSamplerConfigurationV1& other) const noexcept;
+};
+
+struct RuntimeSamplerStateV1
+{
+    // Committed controller state. Asynchronous workers may still be converging to this
+    // configuration, and disabled infrastructure may remain allocated or await cleanup.
```
ysolomchenko, line 73: `Applied` currently describes only the committed controller state; it does not tell the managed consumer when the sampling worker has crossed the safe boundary. A cohort admitted under the previous configuration may therefore publish after a disabling `Apply` returns, so managed code cannot safely perform `disable -> final drain -> stop consumer` without racing a late batch.

src/OpenTelemetry.AutoInstrumentation.Native/runtime_sampler_service.cpp
```diff
@@ -0,0 +1,373 @@
...
+        return;
+    }
+    sampler_->ThreadDestroyed(threadId);
+    if (stackWalker_ != nullptr)
+    {
+        stackWalker_->OnThreadDestroyed(threadId);
+    }
+}
+
+void RuntimeSamplerService::OnThreadNameChanged(const ThreadID threadId, const ULONG cchName, WCHAR name[]) noexcept
```
ysolomchenko, line 383: This `noexcept` wrapper calls code that can allocate, including `unordered_map::operator[]`, `new ThreadState`, and `wstring::append`. `OnAllocationTick` similarly constructs vectors and encoded buffers. If any of these operations throws, the exception reaches a `noexcept` function and invokes `std::terminate`, taking down the instrumented application.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -1264,32 +1391,126 @@ IStackWalker* ContinuousProfiler::GetStackWalker() const
...
-    thread_sampling_thread_ = std::make_unique<std::thread>(SamplingThreadMain, this);
+    if (thread_sampling_thread_ != nullptr)
+    {
+        return true;
+    }
+
+    try
+    {
+        thread_sampling_thread_ = std::make_unique<std::thread>(SamplingThreadMain, this);
+        return true;
```
ysolomchenko, line 1519: This returns `true` as soon as the OS thread is created, while the worker calls `InitializeCurrentThread()` later and ignores its HRESULT. `RuntimeSamplerService` can therefore commit the configuration and return `Applied` without knowing whether the worker is valid for profiler API calls. Subsequent applies also return success because `thread_sampling_thread_` is already non-null.

src/OpenTelemetry.AutoInstrumentation.Native/runtime_sampler_service.cpp
eftiquar: Yes that is known issue that is being worked on.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
eftiquar: Yes this needs to be fixed - thank you for reviewing it.

src/OpenTelemetry.AutoInstrumentation.Native/runtime_sampler_configuration.h
eftiquar: Yes this is the managed - native protocol that needs to be finalized -the config "apply" semantics are asynchronous so I will rename the API appropriately and then a mechanism to signal the consumer is being designed - I am actively working on that part. 
The challenge is to ensure no native to managed calls only managed to native interop is allowed. I have few options and I am evaluating those.

eftiquar: I thought about this further. Applied acknowledges that the controller committed the configuration; it is not a quiescence or drain barrier.
Runtime disablement is different from terminal shutdown. Once the native producer observes the disabled configuration, it stops admitting new sampling work and becomes quiescent after completing and publishing any cohort already in flight. The managed exporter is not terminated: it continues its existing timed read loop, drains any final batch, and subsequently returns empty-handed.
A later enablement resumes production, and the existing consumer consumes the new samples without any thread teardown or reconstruction.
This is intentional for the current iteration as we transition from one-time initialization to dynamic lifecycle semantics. If profiling is never enabled, sampling infrastructure is not created. After the first enablement, we retain the low-cost infrastructure—sleeping or parked threads and the required CLR flags—across disable/re-enable cycles. We can optimize the residual empty polling in a subsequent refinement, but it is not required for correctness.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -1350,16 +1589,24 @@ AllocationSubSampler::AllocationSubSampler(uint32_t targetPerCycle_, uint32_t se
...
           std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()))
     , sampleLock()
     , rand(std::default_random_engine((unsigned int)(nextCycleStartMillis.count())))
 {
 }
+
+void AllocationSubSampler::SetTargetPerCycle(const uint32_t target) noexcept
+{
+    targetPerCycle.store(target, std::memory_order_relaxed);
+}
```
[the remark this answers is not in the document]
eftiquar: I will address this.

src/OpenTelemetry.AutoInstrumentation.Native/cor_profiler.cpp
```diff
@@ -1240,57 +1234,106 @@ void CorProfiler::ConfigureContinuousProfiler(bool         threadSamplingEnabled
...
-        Logger::Info("ContinuousProfiler::StartThreadSampling");
-        this->continuousProfiler->StartThreadSampling();
+        GetContinuousProfilerStateV1(actualState);
+        return runtime_sampler_shutdown_requested_.load(std::memory_order_acquire)
+                   ? continuous_profiler::RuntimeSamplerApplyResult::ShuttingDown
+                   : continuous_profiler::RuntimeSamplerApplyResult::ActivationFailed;
     }
 
-    if (params.allocationSamplingEnabled)
+    const auto outcome = service->ApplyConfigurationV1(authority, configuration);
```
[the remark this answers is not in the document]
eftiquar: I will address this. This is blocker.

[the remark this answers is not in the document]
eftiquar: RuntimeSamplerService now owns the sole lifecycle/configuration gate. Apply holds that gate for the complete transaction, while shutdown marks the service terminal under the same gate and releases it before performing blocking teardown. Therefore, an already-admitted apply completes before shutdown drains it; an apply ordered after shutdown returns ShuttingDown. The previous outer service-acquisition mutex and shutdown handoff have been removed.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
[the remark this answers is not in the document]
eftiquar: SetTargetPerCycle now resets the counters, startup pacing deadline, and cycle boundary under sampleLock, then publishes the new target. This prevents both inherited long suppression after increasing the target and bursts after lowering it. Regression coverage was added in AllocationSamplerHigherTargetDoesNotInheritOldSpacing and AllocationSamplerLowerTargetUsesNewSpacingWithoutBursting

src/OpenTelemetry.AutoInstrumentation.Native/runtime_sampler_service.cpp
eftiquar: The CLR-facing callback boundaries remain noexcept, but their complete bodies—including OnAllocationTick—are now enclosed in try/catch (...). Failures are logged through an exception-contained helper and cannot escape through the noexcept boundary into the host application.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
eftiquar: he worker now reports its InitializeCurrentThread() HRESULT through a startup promise before entering the sampling loop. StartThreadSampling() waits for that handshake;

eftiquar:
@codex review

eftiquar:
@codex review

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
 
 ### Added
 
+- Added a native continuous-profiler runtime configuration contract with
```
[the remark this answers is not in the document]
eftiquar: The reviewed commit [b023157d](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/commit/b023157d4d8488e4be06419b8978365a2c052a21) already contains Assisted-by: GPT-5.6. The changelog entry originated in [9ff51cb5](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/commit/9ff51cb50058a8d97e76f8168860dd7de7b06eac), which contains Assisted-by: OpenAI Codex. All Codex-assisted implementation commits carry the required trailer; the trailer-free commits are mechanical merges from main. Resolving this as already addressed.

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
```diff
@@ -1165,57 +1298,78 @@ static std::chrono::steady_clock::time_point GetNextRefreshTime(
...
 
-static bool ShouldTrackIterations(const ContinuousProfiler* const prof)
+static bool ShouldTrackIterations(const ThreadSamplingConfiguration& configuration)
 {
-    return prof->selectedThreadsSamplingInterval.has_value() && prof->threadSamplingInterval.has_value();
+    return configuration.BothEnabled();
 }
 
-static void SamplingThreadMain(ContinuousProfiler* prof)
+static void SamplingThreadMain(ContinuousProfiler*   prof,
```
lachmatt: should we make this function `noexcept` and add a last-resort catch around the complete worker body?

eftiquar:
@codex review

src/OpenTelemetry.AutoInstrumentation.Native/continuous_profiler.cpp
eftiquar: Yes that is needed for both sampler main as  well as stack walk guard - I have done those changes.

eftiquar:
@codex review

src/OpenTelemetry.AutoInstrumentation.Native/CMakeLists.txt
```diff
@@ -219,6 +219,8 @@ add_library("OpenTelemetry.AutoInstrumentation.Native.static" STATIC
         rejit_handler.cpp
         rejit_preprocessor.cpp
         rejit_work_offloader.cpp
+        runtime_sampler_configuration.cpp
+        runtime_sampler_service.cpp
```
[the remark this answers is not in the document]
eftiquar: Fixed in ba23aa5c2ef3c5fe7e655356242490521aa857e5

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ This component adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.h
...
+  process-wide startup precedence and lazy CPU, selective-thread, and allocation
+  sampler lifecycle management.
```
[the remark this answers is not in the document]
eftiquar: Fixed - abe313e9

src/OpenTelemetry.AutoInstrumentation.Native/stack_walk_guard.cpp
```diff
@@ -185,12 +204,80 @@ StackWalkGuard::ProbeResult StackWalkGuard::AwaitProbeResult()
...
+        std::lock_guard<std::mutex> lk(mutex_);
+        if (state_ == State::Stopping)
+        {
+            return;
+        }
+
+        if (FAILED(initializeResult))
+        {
+            result_ = ProbeResult::Failed;
+            state_  = State::Stopping;
```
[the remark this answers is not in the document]
eftiquar: fixed in abe313e9

CHANGELOG.md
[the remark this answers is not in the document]
eftiquar: fixed in abe313e9

eftiquar:
@ysolomchenko Thanks for the detailed review. I separated the feedback into changes that belong in this PR and lifecycle/host-selection work intentionally deferred to the next phase described in #5360.

Addressed in this PR:

- `StackWalkGuard` now exposes and invokes `InitializeCurrentThread` before its worker accepts requests. Worker entry points are `noexcept` and contain exception handling so failures fail closed at the sampling-service boundary instead of escaping into the process.
- `SetNativeContext` now uses a lock-free Compare and Swap (CAS)  admission gate with an RAII admission token. This is inspired by the thread suspend resume mechanism used by stack sampling. Shutdown closes admission and waits on the cold-path condition variable until all calls admitted before shutdown have returned. This avoids adding a mutex to the hot path and closes the race where a caller had already loaded `profiler_info`.
- The stack-walk request now snapshots the caller-provided `CONTEXT` before scheduling, so the worker does not access caller-owned context after the scheduling call returns.
- Added deterministic coverage for the context snapshot and shutdown-admission behavior.

Deferred / intentional lifecycle boundary:

- Disabling `Apply` is intentionally lazy quiescing. `Applied` acknowledges that the complete configuration snapshot was accepted and committed; it is not a producer-publication barrier. The managed consumer therefore remains alive and continues polling while the native worker reaches quiescence. This preserves the existing consumer lifecycle from `master` and avoids blocking a managed thread during AppDomain unload. A host-aware wait/CV contract can be added with managed host selection in a follow-up.
- Process-wide managed-host election, reader ownership tokens, destructive-read authorization, and owner revocation are deferred. This PR retains the existing lock-protected pool of consumers and does not introduce the managed host-selection protocol.
- Managed-host readiness, including .NET Framework canary readiness and readiness revocation on AppDomain unload, is also deferred to the managed integration/host-selection phase. This PR establishes the native full-snapshot and lifecycle boundary; it does not implement the OpAMP-to-managed protocol.

Intentional shutdown semantics:

- Shutdown is terminal and eager. Once shutdown is observed, an in-progress DSS is aborted at a frame boundary and the unpublished cohort is discarded. We do not publish captured stacks after the shutdown signal. Preserving completed stacks would require a shutdown drain and could make teardown appear hung, which is particularly problematic for IIS or other services being restarted. This is separate from ordinary disablement, which remains lazy and eventually quiesces.

eftiquar:
Something to chew on - 
Me -to Codex -  to be honest I did not ask you explicitly to use RAII - I merely suggested ref count that is my genuine contributaion - you were sharp enough to pick RAII for that - so I am asking was your thinking influenced by ambient code and its mechanisms - this is an attempt to peer into your thinking process :) 

Codex - " Yes. Your ref-count/CAS suggestion was the key design input. I mapped it to a scoped operation:
- admission succeeds → increment count;
- scope exits → decrement count;
- shutdown closes future admission and waits for count zero.
The existing RuntimeGuard and ScopedThreadSuspend made RAII the natural C++ idiom for expressing that lifetime safely. So RAII was my synthesis from your ref-count idea plus the repository’s established patterns—not an explicit instruction you gave."

eftiquar:
@codex review

src/OpenTelemetry.AutoInstrumentation.Native/profiler_api_adapter.cpp
```diff
@@ -13,6 +13,7 @@ namespace ProfilerStackCapture
 struct ProfilerApiAdapter::Impl
 {
     ICorProfilerInfo2*  profilerInfo;
+    ICorProfilerInfo4*  profilerInfo4  = nullptr;
```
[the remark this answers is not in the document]
eftiquar: Fixed.

docs/internal/continuous-profiler.md
```diff
@@ -16,6 +16,76 @@ events:
...
+Interop callers must initialize
+`RuntimeSamplerConfigurationV1.structureSize` to
+`sizeof(RuntimeSamplerConfigurationV1)` (16 bytes) and
+`RuntimeSamplerStateV1.structureSize` to `sizeof(RuntimeSamplerStateV1)` (24
+bytes). Each apply call makes one synchronous activation attempt. Native code
+does not schedule retries, impose a retry budget, or permanently disable a
+failed feature; retry cadence and stale-request suppression remain managed
+control-plane responsibilities. A failed attempt returns `ActivationFailed`,
+preserves the last committed state, and can be retried by a later explicit
+apply.
```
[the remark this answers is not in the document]
eftiquar: Revised the doc.

eftiquar:
@codex review

eftiquar:
@codex review

eftiquar:
@codex review

eftiquar:
@codex review

eftiquar:
@codex review

ysolomchenko (approved):

CHANGELOG.md
Kielek, line 12: I do not think that we need this message at all. it is not user-facing change. Developers/distribution providers can analyze commits.

src/OpenTelemetry.AutoInstrumentation.Native/shutdown_admission.cpp
```diff
@@ -0,0 +1,88 @@
...
+// Cold-path wait primitives. Sampler callbacks and publications use only the CAS admission in the normal case; the
+// mutex is acquired only to publish and wait for the terminal drained predicate.
+std::mutex              shutdown_mutex;
+std::condition_variable shutdown_cv;
+bool                    shutdown_admissions_drained = false;
+
+bool TryEnterShutdownAdmission() noexcept
+{
+    auto state = shutdown_admission_gate.load(std::memory_order_acquire);
+    for (;;)
```
Kielek, line 30: It is strange from C# developer perspective, shouldn't it be while(true)?

Just asking, not necessary, changes are needed.

src/OpenTelemetry.AutoInstrumentation.Native/runtime_sampler_configuration.h
```diff
@@ -0,0 +1,106 @@
...
+
+enum class RuntimeSamplerStateQueryResult : int32_t
+{
+    Succeeded         = 0,
+    InvalidArgument   = 1,
+    UnsupportedLayout = 2
+};
+
+#pragma pack(push, 4)
+struct RuntimeSamplerConfigurationV1
```
Kielek, line 47: I would fully remove a version suffix. We always release native part together with managed code. ABI compatibility is not required between releases. This comment should be applied across whole PR.

src/OpenTelemetry.AutoInstrumentation.Native/shutdown_admission.cpp
eftiquar: Consider it culturally received style - the loop in question is compare and swap (CAS) - a lock free construct to atomically check for conditions in the loop body. I'd say for (;;) has acquired a cultural semantic meaning in C/C++ beyond its literal syntax. Not a language rule, but a strong stylistic signal that the loop itself is intentionally unconditional and the exit conditions live entirely inside the body.
You are onto something here - I once wrote for(;;) in C# code and was forced to retreat by the reviewers

nrcventura (approved):

lachmatt (approved):

Kielek (approved):

PR #5404, opened 2026-09-02
Add IBM MQ (XMS.NET) traces instrumentation

Aryan16138:
## Why

There is currently no IBM MQ support in this distribution. XMS.NET (`IBMXMSDotnetClient`) is IBM's
messaging API for .NET, and applications using it get no messaging spans today, so a queue hop shows
up as a gap in the trace. This adds producer and consumer instrumentation for it, including context
propagation so a trace continues across the queue.

Fixes #
Issue: **https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/5419**
## What

New bytecode instrumentation for IBM MQ via XMS.NET, enabled with the `Xms` instrumentation id.
It targets the `amqmxmsstd` assembly (namespace `IBM.XMS`) for versions 9.0.0 through 10.x.

Instrumented methods:

* `XmsMessageProducerImpl.Send`, all four overloads, produces a Producer span
* `XmsMessageConsumerImpl.Receive()`, `Receive(long)` and `ReceiveNoWait()`, produce Consumer spans
* `XmsProviderMessageListener.OnMessage(ProviderMessage)`, produces a Consumer span for async
  delivery through a `MessageListener`

Spans carry `messaging.system` (`ibmmq`), `messaging.operation`, `messaging.destination.name` and
`messaging.message.id`. Span names are `{destination} publish`, `{destination} receive` and
`{destination} deliver`. The producer message id is read after `Send` returns, because JMS assigns
`JMSMessageID` during the send rather than before.

W3C trace context is injected into and extracted from XMS string message properties. Consumer spans
are separate trace roots, each carrying exactly one link back to the producer span, rather than being
children of it. This matches how the existing Kafka and RabbitMqLegacy instrumentations in this
distribution handle the same problem, and follows the messaging semantic conventions, which treat
links as the default correlation mechanism between messaging spans and allow a parent-child
relationship only as an option for single-message scenarios. Property names are passed through a
sanitiser first, because JMS property names have to be valid Java identifiers. `traceparent` and
`tracestate` are already valid so they pass through unchanged, but a propagator using something like
`X-B3-TraceId` would otherwise fail to set the property.

The three `Receive` overloads share one `ConsumerReceiveIntegration` class with three stacked
`InstrumentMethod` attributes, because its two `OnMethodBegin` overloads differ by a concrete `long`
parameter and reflection can still tell them apart. The four `Send` overloads stay on four separate
classes instead, because four generic `OnMethodBegin` overloads on one class make CallTarget throw
`AmbiguousMatchException` on every instrumented call.

TThis also adds an experimental `messaging.ibmmq.queue_manager.id` attribute, the IBM MQ queue
manager identifier, behind `OTEL_DOTNET_AUTO_XMS_EXPERIMENTAL_SPAN_ATTRIBUTES` and off by
default. Queue manager names are not globally unique, so two hosts each running a `QM1` is
ordinary rather than an edge case, and the identifier is what lets a span be attributed to a
specific queue manager. It is read from connection properties the XMS client has already
resolved and cached at `MQCONN`, so collecting it is a dictionary lookup rather than an `MQINQ`
round trip, and it is re-read per operation because IBM refreshes those properties after a
client reconnect.

Both this attribute and the `ibmmq` member of the `messaging.system` enum that these spans use
are proposed in open-telemetry/semantic-conventions#4081.

There is one change outside the new instrumentation. `DuckType.CreateMethods` built its
candidate list from a raw `GetMethods` call, which for an interface omits members inherited from
base interfaces. Duck typing against `IBM.XMS.IMessage` needs `SetStringProperty` and
`GetStringProperty`, both declared on a base interface, so those members were invisible. It now
flattens the interface list when the target type is an interface; behaviour for classes and
structs is unchanged.

Synchronous `Receive` spans are created after the call returns, so they do not measure the blocking
wait. Transacted `Commit`/`Rollback` and `IQueueBrowser` are not instrumented.

## Tests

Added `TestApplication.Xms` and `XmsTests.SubmitsTraces`. The test application exercises all three
paths in a single run: a `Send`, a synchronous `Receive`, and async delivery through a named
`MessageListener`. The test asserts four spans, two producer and two consumer, their messaging
attributes, and that each consumer span is a trace root carrying exactly one link whose trace id and
span id both match the corresponding producer span, for both the synchronous receive path and the
async `MessageListener` path.

The broker comes from the IBM MQ developer image through Testcontainers, reusing the `DEV.QUEUE.*`
queues, the `DEV.APP.SVRCONN` channel and the `app` user the image already creates.

I also verified the behaviour by hand against a live queue manager, running an uninstrumented XMS
application under the profiler and exporting over OTLP. Producer and consumer span counts matched one
to one over about 5000 messages, and across 10001 exported spans, `ParentSpanId` was empty on every
one, with each consumer span carrying a link to its producer span.

.NET Framework is untested. XMS dropped its `netstandard2.0` library at 9.4.0, so .NET Framework
would need XMS 9.2.x or 9.3.x, and that combination is not covered here.

## Checklist


Kielek:
Hi @Aryan16138 for your contribution.
Before we start review, it will be great to understand your needs.

There are 2 possible ways to handle any package instrumentation
1. Implement native instrumentation directly within the library. Some time ago [OpenSearch](https://github.com/opensearch-project/opensearch-net/pull/1011) decided to do this. It is the best option if you are able to influence maintainers or you are the library maintainer. Then, everything you need is to use `OTEL_DOTNET_AUTO_TRACES_ADDITIONAL_SOURCES` env. var. IT is the most recommended way, as it allow to use instrumentation even without byte-code dark magic and the autoinstrumentation itself.
2. The second option, which can be less performant, is [configuration based instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/102e59153f751f2920490a2929f7f9d64d72d789/docs/nocode-instrumentation.md). For sure it is not so powerful as bytecode instrumentation written in the code, but it should allow to handle most cases. Please consider this as an option.

-----------

SIG notes: need to understand end-user needs and ask about possibility for native instrumentation

Aryan16138:
> Hi @Aryan16138 for your contribution. Before we start review, it will be great to understand your needs.
> 
> There are 2 possible ways to handle any package instrumentation
> 
> 1. Implement native instrumentation directly within the library. Some time ago [OpenSearch](https://github.com/opensearch-project/opensearch-net/pull/1011) decided to do this. It is the best option if you are able to influence maintainers or you are the library maintainer. Then, everything you need is to use `OTEL_DOTNET_AUTO_TRACES_ADDITIONAL_SOURCES` env. var. IT is the most recommended way, as it allow to use instrumentation even without byte-code dark magic and the autoinstrumentation itself.
> 2. The second option, which can be less performant, is [configuration based instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/102e59153f751f2920490a2929f7f9d64d72d789/docs/nocode-instrumentation.md). For sure it is not so powerful as bytecode instrumentation written in the code, but it should allow to handle most cases. Please consider this as an option.
> 
> SIG notes: need to understand end-user needs and ask about possibility for native instrumentation



Thanks for asking before review. Context first, Let's break down the problem we are trying to solve, and then move on to consider the provided options.

### What is missing today

Nothing instruments XMS.NET, IBM's messaging API for .NET. Neither this distribution nor
opentelemetry-dotnet-contrib covers it. A .NET application that sends or receives over an IBM MQ
queue produces no spans for those operations.

So a request that crosses a queue arrives as two unrelated traces. Nothing says which queue it
used, or how long it waited there.

### Why it cannot be done inside the application

Adding an `ActivitySource` means changing the application. For these systems that is not
available:
- They are long-lived business applications. A change needs review, regression testing, and a
  redeploy of something that currently works.
- The team operating the application often does not own its source.
- Some are maintained under support terms that do not permit modification.

This rules out every in-application option, which is why auto-instrumentation is the only route
that reaches them.

### What this PR adds

- Producer spans for the four `Send` overloads.
- Consumer spans for `Receive()`, `Receive(long)` and `ReceiveNoWait()`.
- A consumer span for async delivery through a registered `MessageListener`.
- W3C context injected into and extracted from XMS string message properties, so the trace
  continues across the queue.

### Consistency with the existing instrumentations

The consumer span is a trace root with a link to the producer span. That matches Kafka and
RabbitMqLegacy, and matches the messaging conventions, which use links as the default correlation
mechanism.

Span naming, the `messaging.*` attribute set, the class layout and the experimental attribute gate
all follow those two.

### Queue manager identifier

The PR adds `messaging.ibmmq.queue_manager.id`, gated behind
`OTEL_DOTNET_AUTO_XMS_EXPERIMENTAL_SPAN_ATTRIBUTES` and off by default.

Queue manager names are not unique. Two hosts each running a `QM1` is normal. The identifier is
what makes a span attributable to one specific queue manager.

It is read from connection properties the XMS client already caches at `MQCONN`, so it costs a
dictionary lookup rather than an `MQINQ` call.

### Semantic conventions

The attribute, and the `ibmmq` value of `messaging.system`, are both proposed in
open-telemetry/semantic-conventions#4081.

### The same attribute in Java

This is about the convention rather than this PR. The Java agent has an equivalent change in
open-telemetry/opentelemetry-java-instrumentation#19896, so the proposed convention has two
independent implementations behind it.

### Option 1, native instrumentation in the library

Not available here. opensearch-net#1011 worked because opensearch-net is an Apache-2.0 repository,
so a contributor could open a pull request against the client itself.

XMS.NET ships as a compiled NuGet package, `IBMXMSDotnetClient`, with no public source. Nobody
outside IBM can add an `ActivitySource` to it, and `OTEL_DOTNET_AUTO_TRACES_ADDITIONAL_SOURCES`
needs a source name that does not exist for this library.

### Option 2, configuration based instrumentation

Attributes would work. `NoCodeExpressionContext` exposes `ReturnValue`, so the message id JMS
assigns during `Send` is readable.

Propagation would not:

- Injecting `traceparent` needs a call to `SetStringProperty`. CEL reads properties via
  `GetProperty` with `BindingFlags.Public | BindingFlags.Instance`. There is no method invocation
  and no way to write back into an argument.
- `NoCodeSpan` has `name`, `name_source`, `kind`, `attributes` and `status`. There is no links
  field, so an extracted context cannot be attached to the span.

Propagation is the point of this instrumentation. Without it, the spans on either side of a queue
cannot be joined, which leaves the original problem in place. That is also why Kafka and RabbitMQ
here are written in C# rather than as configuration.

The PR is a single commit on current `main`, with integration tests covering all three paths
against the IBM MQ developer image. I have also checked the behaviour against a live queue
manager.

Kielek:
Could you please resolve conflicts and double check if you are able to build code locally? CI seems to be failing due to your chnages.

Aryan16138:
@Kielek I am able to build locally

PR #5452, opened 2026-09-13
feat: Add opt-in RabbitMQ vhost/cluster_name span attributes

### LINKED ISSUE #5451: Add opt-in messaging.rabbitmq.vhost.name and messaging.rabbitmq.cluster.name span attributes
# Feature Request

- Framework or library name : `RabbitMQ.Client`
- Library type: message broker client
- Library version: 5.x / 6.x (the `rabbitmq-legacy` bytecode instrumentation)

RabbitMQ spans produced by the `RabbitMqLegacy` instrumentation carry no information about which
virtual host or which cluster a message was published/consumed on. Both are basic facts about a
RabbitMQ connection, and without them it's hard to distinguish traffic across vhosts (a common
multi-tenancy boundary in RabbitMQ) or to tell which cluster a trace touched when several clusters
are monitored from the same backend.

Neither attribute is part of the messaging semantic conventions yet
(open-telemetry/semantic-conventions#3997 is tracking that), so they can't be added as stable,
always-on attributes today.

Add two new **opt-in, experimental** span attributes to the RabbitMQ instrumentation:

- `messaging.rabbitmq.vhost.name` — the virtual host of the connection.
- `messaging.rabbitmq.cluster.name` — the broker's `cluster_name` server property.

Both disabled by default, gated behind a single new configuration property, following the existing
pattern used for other opt-in attributes in this instrumentation (e.g.
`OTEL_DOTNET_AUTO_ORACLEMDA_SET_DBSTATEMENT_FOR_TEXT`, `OTEL_DOTNET_AUTO_GRAPHQL_SET_DOCUMENT`):

- `OTEL_DOTNET_AUTO_RABBITMQ_CAPTURE_VHOST_AND_CLUSTER_NAME` (default `false`)

Both attributes share the same rationale for being opt-in (the semantic convention isn't stable
yet), so one flag covers both rather than introducing two nearly-identical ones.

**Additional context**

abhishuraina:
## Why

RabbitMQ spans produced carry no information about which virtual host or which cluster a message was published/consumed on. Both are basic facts about a RabbitMQ connection, and without them it's hard to distinguish traffic across vhosts (a common multi-tenancy boundary in RabbitMQ) or to tell which cluster a trace touched when several clusters are monitored from the same backend.

Neither attribute is part of the messaging semantic conventions yet (open-telemetry/semantic-conventions#3997), so they're added as opt-in, disabled by default rather than committing to a stable shape ahead of that discussion.

Fixes #5451 

## What

- Adds messaging.rabbitmq.vhost.name and messaging.rabbitmq.cluster.name to publish, receive, and deliver spans, gated behind a new configuration property (default false), following the existing pattern used for other opt-in attributes in this instrumentation (OTEL_DOTNET_AUTO_ORACLEMDA_SET_DBSTATEMENT_FOR_TEXT, OTEL_DOTNET_AUTO_GRAPHQL_SET_DOCUMENT):

  - OTEL_DOTNET_AUTO_RABBITMQ_CAPTURE_VHOST_AND_CLUSTER_NAME

  Both attributes share the same rationale for being opt-in (the semantic convention isn't stable yet), so one flag covers both.

## Tests

- Added SubmitsVhostAndClusterNameWhenOptedIn to RabbitMqTests.cs: enables the new flag and asserts both attributes on publish, receive, and deliver spans (skipped for 7.x+, which uses the client's own native instrumentation, not this bytecode path).
- Extended the existing default-path test's span validators with an explicit absence assertion, to lock in the opt-in-by-default contract (nothing previously asserted either way).
- Manually verified end-to-end against a live 3-node RabbitMQ cluster (not just the test suite): attributes absent by default and present when opted in, on a genuinely separate producer process and consumer process (not just a single combined app), confirming both the attribute values and correct cross-process trace linking. Also specifically reproduced and re-verified the AutorecoveringModel fix against a real automatic-recovery consumer before and after the change.

## Checklist


Kielek:
SIG notes:
1. We will not proceed with this until semantic convention includes these attributes.
2. Bytecode instrumentation implemented here supports only legacy 5 and 6 version. All new versions are natively supported by Activity(Source) and we only _listen to_ it. Before bringing it to the legacy version, you should ensure that it is supported by the latest supported version. We can consider brining it to the legacy version when it is done first.

abhishuraina:
I've created a PR to support for the latest version https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/2049

PR #4950, opened 2026-03-25
ADO.NET and SQLite bytecode instrumentation

Kielek:
## Why

Fixes #1126

## What

Brings instrumentation for generic ADO.NET traces instrumentation.
Brings also support for SQLite instrumentation for [Microsoft.Data.Sqlite](https://www.nuget.org/packages/Microsoft.Data.Sqlite) 8.0.0+.

Generic ADO.NET instrumentation is disabled for known cases, covered by our other traces instrumentations.
Generic ADO.NET instrumentation is working only with direct implementors of DbCommand. What is more it cannot utilize `new` keyword to override behavior.

This implementation covers also Devart.Data.Oracle 9+. For now, one specific method is included into generic ADO.NET instrumentation. It can be separated to separate switch as it was done for SQLite.

SQLite is included here, to have a proof that it can work independently from ADO.NET.

### Potential extenions

There is a chance that other libraries also needs some tweaks. In such cases we can extend generic ADO.NET or decided to extract it as separate functionalities. Both cases are illustrated in this PR

For sure, there is a lack of support for https://www.nuget.org/packages/System.Data.SQLite for SQLite. Not a blocker IMO.

## Tests

Added tests based on the Stub implemented DbCommands..
Added tests for SQLite.
Separate tests for Devart.Data.Oracle under https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/commit/013718fe2fe2e1cd5719cede3d8f7f14c1026625. It is working fine for 9/10/11. ~~Still checking licesnse issues for 9.*.~~

## Checklist

- [x] `CHANGELOG.md` is updated.
- [x] Documentation is updated.
- [x] New features are covered by tests.

## Notes

Thanks for Maryna Dimova from Devart team for providing license to be able to cover Devart.Data.Oracle case.

PR #4791, opened 2026-01-19
MongoDB traces - update semantic conventions to v1.39.0

aviralgarg05:
# MongoDB - update semantic convention to v1.39.0 (#4786)

## Why

OpenTelemetry Semantic Conventions have standardized database attribute names to ensure consistency across different language implementations. Specifically, the attribute `db.system` has been transitioned to `db.system.name` in the stable semantic conventions (finalized in v1.30.0 and kept in v1.39.0). This PR aligns the .NET AutoInstrumentation with these standards.

## What

- **Code Change**: Updated `DbSystem` constant from `"db.system"` to `"db.system.name"` in [src/OpenTelemetry.AutoInstrumentation/Instrumentations/DatabaseAttributes.cs](cci:7://file:///Users/aviralgarg/Everything/opentelemetry-dotnet-instrumentation/src/OpenTelemetry.AutoInstrumentation/Instrumentations/DatabaseAttributes.cs:0:0-0:0).
- **Test Alignment**: Updated [test/IntegrationTests/MongoDBTests.cs](cci:7://file:///Users/aviralgarg/Everything/opentelemetry-dotnet-instrumentation/test/IntegrationTests/MongoDBTests.cs:0:0-0:0) to expect the new `db.system.name` attribute in emitted spans.
- **Changelog**: Added an entry to [CHANGELOG.md](cci:7://file:///Users/aviralgarg/Everything/opentelemetry-dotnet-instrumentation/CHANGELOG.md:0:0-0:0) under the "Unreleased" section.

## Tests

- **Static Analysis**: Verified that [MongoDBInstrumentation.cs](cci:7://file:///Users/aviralgarg/Everything/opentelemetry-dotnet-instrumentation/src/OpenTelemetry.AutoInstrumentation/Instrumentations/MongoDB/MongoDBInstrumentation.cs:0:0-0:0) uses the `DatabaseAttributes.Keys.DbSystem` constant, ensuring the change propagates automatically.
- **Mock Collector Verification**: Updated the integration test assertions to validate that the spans now correctly contain `db.system.name`.
- **Global Search**: Performed a codebase-wide `grep` to ensure no other hardcoded strings of `"db.system"` were missed in the `src` or `test` directories.

## Checklist

- [x] [CHANGELOG.md](cci:7://file:///Users/aviralgarg/Everything/opentelemetry-dotnet-instrumentation/CHANGELOG.md:0:0-0:0) is updated.
- [x] New features are covered by tests.

PR #5453, opened 2026-09-14
Kafka traces - update semantic conventions to v1.44.0

shashank-reddy-nr:
## Why

The Kafka instrumentation still emits the [v1.23.0](https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/messaging/messaging-spans.md) messaging conventions. Five attributes keep names that have since been renamed, the span name uses the pre-v1.24.0 `{destination} {operation}` order rather than `{messaging.operation.name} {destination}`, and `error.type` is not emitted at all.

This is the split requested in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/5206#discussion_r3768913003 and, for the sibling instrumentation, in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4739#issuecomment-5060107586: migrate to the current conventions first, without adding new recommended attributes, so compliance with v1.44.0 can be reviewed on its own.

## What

Migrates the Kafka instrumentation to [v1.44.0](https://github.com/open-telemetry/semantic-conventions/blob/v1.44.0/docs/messaging/kafka.md).

Renames:

| v1.23.0 | v1.44.0 |
| --- | --- |
| `messaging.operation` | `messaging.operation.name` |
| `messaging.client_id` | `messaging.client.id` |
| `messaging.kafka.consumer.group` | `messaging.consumer.group.name` |
| `messaging.kafka.destination.partition` (int) | `messaging.destination.partition.id` (string) |
| `messaging.kafka.message.offset` | `messaging.kafka.offset` |

Also:

- Producer operation name changes from `publish` to `send`.
- Span name changes from `{destination} {operation}` to `{operation} {destination}`.
- Adds `messaging.operation.type`, `Conditionally Required` on the send, receive and process spans.
- Adds `error.type`, `Conditionally Required` when the operation has failed, on the produce and consume paths that surface an exception. No `Recommended` attribute is added — `messaging.kafka.cluster.id` stays out of scope, per the split above.
- The `ActivitySource` now declares `TelemetrySchemaUrl`, as the ADO.NET (v1.40.0) and MongoDB (v1.39.0) instrumentations already do.

`error.type` reports the canonical class name rather than `Type.FullName`. Confluent.Kafka surfaces produce failures as `ProduceException<TKey, TValue>`, and the `FullName` of a constructed generic type carries assembly-qualified type arguments that shift with the runtime version — neither predictable nor low cardinality:

```
Confluent.Kafka.ProduceException`2[[System.String, System.Private.CoreLib, Version=10.0.0.0, ...],[...]]
```

The RabbitMQ instrumentation is untouched — it still emits v1.23.0, so the shared constants it uses are left in place and it declares no schema URL.

Follows #4791 (MongoDB → v1.39.0, which also added `error.type` as part of its migration) and #4950 (ADO.NET → v1.40.0): a direct rename, no `OTEL_SEMCONV_STABILITY_OPT_IN` gate, recorded in `CHANGELOG.md` under `Changed`. These trace instrumentations are documented as Experimental "due to lack of stable semantic convention". Happy to gate this behind `OTEL_SEMCONV_STABILITY_OPT_IN` instead if you would prefer that for messaging, since it is still Development.

## Tests

- Six unit tests added in `KafkaInstrumentationTests` covering the producer span, the consumer span, delivery-result attributes, the emitted schema URL, `error.type`, and the canonical-name handling for generic exceptions. Each was written first and confirmed failing against the old behaviour.
- `KafkaTests` integration expectations updated to the new names, the string partition id, the new span-name order, `messaging.operation.type`, and the schema URL (via the `schemaUrl` parameter added in #4791).
- Managed suite: 757 passed, 2 skipped, 0 failed. `dotnet format <sln> --verify-no-changes`: clean. Builds clean for net462/net8.0/net10.0.
- Also ran the built distribution against Apache Kafka 3.8.1 with a Confluent.Kafka 2.6.1 application (`ProduceAsync`, sync produce with and without a delivery handler, a tombstone, a forced `Local_UnknownPartition` failure, and consumes) and inspected the console exporter: 13 spans named `send <topic>` / `receive <topic>`, all attributes on the new names, none on the old ones, schema URL `1.44.0`, and `error.type: Confluent.Kafka.ProduceException` on the failed produce.

Known gap, left for a follow-up: the delivery-report path reports failures by wrapping a Kafka `Error` in a synthetic `Exception`, so it carries no meaningful exception type and is not given an `error.type`. Reporting the Kafka error code there (as `OpenTelemetry.Instrumentation.ConfluentKafka` does) needs new duck-typed surface for `Error.Code`.

## Checklist

- [x] ~~Documentation is updated.~~ No documented attribute lists reference these names.

Disclosure per the [OTel GenAI policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md): AI-assisted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Kielek:
SIG decision, we SHOULD follow

> [!IMPORTANT]
>
> Existing messaging instrumentations that are using
> [v1.24.0 of this document](https://github.com/open-telemetry/semantic-conventions/blob/v1.24.0/docs/messaging/messaging-spans.md)
> (or prior):
>
> * SHOULD NOT change the version of the messaging conventions that they emit by default
>   until the messaging semantic conventions are marked stable.
>   Conventions include, but are not limited to, attributes,
>   metric and span names, span kind and unit of measure.
> * SHOULD introduce an environment variable `OTEL_SEMCONV_STABILITY_OPT_IN`
>   in the existing major version as a comma-separated list of category-specific values
>   (e.g., http, databases, messaging). The list of values includes:
>   * `messaging` - emit the new, stable messaging conventions,
>     and stop emitting the old experimental messaging conventions
>     that the instrumentation emitted previously.
>   * `messaging/dup` - emit both the old and the stable messaging conventions,
>     allowing for a seamless transition.
>   * The default behavior (in the absence of one of these values) is to continue
>     emitting whatever version of the old experimental messaging conventions
>     the instrumentation was emitting previously.
>   * Note: `messaging/dup` has higher precedence than `messaging` in case both values are present
> * SHOULD maintain (security patching at a minimum) the existing major version
>   for at least six months after it starts emitting both sets of conventions.
> * SHOULD drop the environment variable in the next major version.
> * SHOULD emit the new, stable values for span name, span kind and similar "single"
> valued concepts when `messaging/dup` is present in the list.

Please remember that we should also support file based configuration. You should add there temporary configuration, to be able to configure this feature.

PR #5438, opened 2026-09-10
Update native code to upstream tag v2.22.0 (include v2.14.0 to v2.21.0)

Kielek:
# Sync native instrumentation changes from DataDog v2.14.0 through v2.22.0

🎉 Welcome in 2023!

While review, please check tables below. It contains most important decisions.

The general flow

- manually select candidates for synchornisation
- Ask GPT 5.6 sol for bringing changes
- manual review of each decision commit. Often override decision to bring changes at least partially

Intentionally skipped native linux infrastracture in this PR.

## Why

OpenTelemetry .NET Automatic Instrumentation was originally derived from
DataDog's .NET tracer and still shares substantial native profiler, CallTarget,
IL rewriting, and instrumentation infrastructure with it. Periodically syncing
applicable upstream changes reduces divergence and makes later synchronization
work smaller and safer.

This continues the synchronization performed in
https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/5411,
which covered DataDog changes through v2.13.0.

## What

This PR reviews all 50 selected DataDog commits between v2.13.0 and v2.22.0. It
ports shared changes that have useful equivalents in this repository, records
equivalent behavior that is already present, and documents why product-specific
or obsolete changes were excluded. The applied commits remain in the same
chronological order as their DataDog counterparts and retain the full upstream
commit link in each commit message.

The resulting series contains:

- 4 fully ported upstream commits;
- 12 partially ported upstream commits, limited to reusable infrastructure; and
- 34 commits skipped because they are DataDog-specific, debugger-specific,
  obsolete, already implemented, or have no equivalent component here.

The ported changes include:

- native and managed support for instrumenting interface methods, including the
  additive RabbitMQ interface definitions;
- CallTarget async-continuation infrastructure for `OnAsyncMethodEnd` methods
  that return a `Task`; its test suite is intentionally deferred to a separate
  PR;
- the standalone `CallTargetBubbleUpException` contract and native filters that
  allow direct marker exceptions to escape synchronous begin/end callbacks;
- CallTarget token cleanup, generic-type safety, and IL-rewriter correctness
  fixes;
- copying custom attributes to generated DuckTyping reverse proxies;
- profiler lifecycle guards, assembly and loader-callsite exclusions, and
  improved IL-dump diagnostics;
- portable native numeric conversion and whitespace handling;
- selected CppCheck findings and mechanical C++ modernization; and
- reduced exported-symbol visibility on Unix.

[...]

PR #5456, opened 2026-09-14
Support runtime async bytecode instrumentation

### LINKED ISSUE #5447: .NET 11 Runtime Async causes invalid CallTarget-rewritten IL
## Summary

Issue description is AI-Assited. Codex GPT5.6 Sol/Max

Enabling [.NET 11 Runtime Async V2](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-11/runtime#runtime-async) for the `TestApplication.NoCode` test application causes the auto-instrumented process to terminate with `System.InvalidProgramException` when the first rewritten runtime-async method executes.

The failure is reproducible from [PR #5111: Support .NET 11 + drop .NET 8 and .NET 9](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/5111), at head commit [`7992cbe4e98712a17326733e14f0766adc08db87`](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/commit/7992cbe4e98712a17326733e14f0766adc08db87).

The application builds and runs normally without the profiler. The crash occurs only after the native CallTarget rewriter modifies a runtime-async method.

## Environment

- PR: `open-telemetry/opentelemetry-dotnet-instrumentation#5111`
- PR source branch: `Kielek:net11`
- Tested commit: `7992cbe4e98712a17326733e14f0766adc08db87`
- .NET SDK: `11.0.100-rc.1.26425.128`
- PowerShell Desktop: `5.1.26100.9168`
- Test target framework: `net11.0`
- Containers: disabled

## Reproduction

Run the following commands from PowerShell on Windows.

### 1. Check out PR #5111

Check out locally #5111

### 2. Enable Runtime Async for the .NET 11 NoCode application

If this change is not already present in the branch, add the following property group near the beginning of `test/test-applications/integrations/TestApplication.NoCode/TestApplication.NoCode.csproj`:

```xml
<PropertyGroup Condition="'$(TargetFramework)' == 'net11.0'">
  <Features>$(Features);runtime-async=on</Features>
</PropertyGroup>
```

The condition keeps the preview feature isolated to .NET 11 and leaves the project's other target frameworks unchanged.

### 3. Verify the uninstrumented control

```powershell
dotnet build test/test-applications/integrations/TestApplication.NoCode/TestApplication.NoCode.csproj `
  --framework net11.0 `
  --configuration Release

dotnet run `
  --project test/test-applications/integrations/TestApplication.NoCode/TestApplication.NoCode.csproj `
  --framework net11.0 `
  --configuration Release `
  --no-build
```

Expected and observed: the project builds with no errors and the uninstrumented application exits successfully.

### 4. Build the profiler and run the minimal failing test

```powershell
dotnet tool restore
./build.cmd BuildTracer

$env:OTEL_LOG_LEVEL = 'debug'
$env:OTEL_DOTNET_AUTO_DUMP_ILREWRITE_ENABLED = 'true'

./build.cmd `
  --target ManagedTests `
  --test-project IntegrationTests `
  --test-name NoCodeTests.SubmitsTraces `
  --test-target-framework net11.0 `
  --containers none
```

## Expected result

`NoCodeTests.SubmitsTraces` passes and receives the expected spans for the NoCode application's synchronous and asynchronous methods.

## Actual result

The instrumented application terminates with CLR exception exit code `-532462766` (`0xE0434352`):

```text
Unhandled exception. System.InvalidProgramException:
Common Language Runtime detected an invalid program.
   at TestApplication.NoCode.NoCodeTestingClass.TestMethodAsync()
   at TestApplication.NoCode.Program.Main(String[] args)
```

The first confirmed failing target is:

```text
TestApplication.NoCode.NoCodeTestingClass.TestMethodAsync()
```

Its source body is simply:

```csharp
public async Task TestMethodAsync()
{
    await Task.Delay(TimeSpan.FromMilliseconds(100)).ConfigureAwait(false);
}
```

## Rewritten IL evidence

With `OTEL_DOTNET_AUTO_DUMP_ILREWRITE_ENABLED=true`, the native profiler log under `test-artifacts/profiler-logs/` shows that `GetReJITParameters` is called and that the method is rewritten.

The relevant tail of the original runtime-async IL is:

```text
call      System.Threading.Tasks.Task.Delay(...)
callvirt  System.Threading.Tasks.Task.ConfigureAwait(...)
call      System.Runtime.CompilerServices.AsyncHelpers.Await(...)
ret
```

The stack is empty at this `ret`, as required for a runtime-async method declared to return `Task`.

The rewriter changes that tail to the equivalent of:

```text
call      System.Runtime.CompilerServices.AsyncHelpers.Await(...)
stloc.0   // inserted; local 0 has the metadata return type System.Threading.Tasks.Task
leave.s   END_METHOD
...
ldloc.0
ret
```

The inserted `stloc.0` requires a `Task` value on the evaluation stack, but runtime-async IL intentionally has no value there. The resulting IL is invalid.

## Root cause

Runtime-async methods keep a `Task`/`ValueTask` return type in metadata but use a different IL return convention. The [runtime-async specification](https://github.com/dotnet/runtime/blob/main/docs/design/specs/runtime-async.md) defines the following:

- `Task` and `ValueTask`: the evaluation stack is empty at `ret`.
- `Task<T>` and `ValueTask<T>`: the evaluation stack contains `T`, not the task-like wrapper.

The CallTarget rewriter instead derives its behavior exclusively from the metadata method signature:

1. [`method_rewriter.cpp`](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/7992cbe4e98712a17326733e14f0766adc08db87/src/OpenTelemetry.AutoInstrumentation.Native/method_rewriter.cpp#L113-L120) reads the declared return type and classifies `Task` as non-void.
2. `ModifyLocalSigAndInitialize` therefore creates a return-value local of type `Task`.
3. [`method_rewriter.cpp`](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/7992cbe4e98712a17326733e14f0766adc08db87/src/OpenTelemetry.AutoInstrumentation.Native/method_rewriter.cpp#L570-L586) inserts `StLocal(returnValueIndex)` before every original `ret` for any method classified as non-void.
4. The generated end-method call also uses the declared task-like type and expects the normal CallTarget continuation path.

This assumption is valid for ordinary synchronous IL and compiler-generated async state-machine entry methods, but not for methods marked with the new runtime-async method implementation flag.

The same defect should affect all four runtime-async return shapes:

| Declared metadata return | Runtime-async IL value at `ret` | Current rewriter expects |
|---|---|---|
| `Task` | no value | `Task` |
| `ValueTask` | no value | `ValueTask` |
| `Task<T>` | `T` | `Task<T>` |
| `ValueTask<T>` | `T` | `ValueTask<T>` |

The `TestApplication.NoCode` application already exercises all four shapes, as well as a generic `Task<T>` method.

## Why this is not the earlier Runtime Async ReJIT publication issue

The symptom differs from [dotnet/runtime#128944](https://github.com/dotnet/runtime/issues/128944) and its fix in [dotnet/runtime#131681](https://github.com/dotnet/runtime/pull/131681). In this reproduction:

- `GetReJITParameters` is invoked.
- The native profiler logs the modified method body.
- The rewritten body is selected and then rejected as an invalid program.

The failure is therefore in the OpenTelemetry CallTarget rewrite semantics, not in failure to publish the ReJIT version.

### Defensive compatibility fix

Detect the runtime-async method implementation flag and skip CallTarget rewriting for those methods. This prevents host crashes, but telemetry from affected bytecode instrumentations is lost. It is suitable only as a temporary safety measure.

### Full Runtime Async support

Teach the CallTarget pipeline about the runtime-async calling convention:

1. Detect the runtime-async method implementation flag (`MethodImplOptions.Async`, metadata value `0x2000`).
2. Derive the logical IL return type:
   - `void` for `Task` and `ValueTask`;
   - `T` for `Task<T>` and `ValueTask<T>`.
3. Generate return locals and rewrite each `ret` using that logical type.
4. Add a managed completion path that calls the integration's async end callback directly. The existing path expects a task-like wrapper and attaches a continuation, but that wrapper is not present in a runtime-async method body.
5. Verify exception behavior and `Activity.Current`/`ExecutionContext` restoration across actual suspension.

Coverage should include synchronous completion, suspension, exceptions before and after an `await`, cancellation, `Task`, `Task<T>`, `ValueTask`, `ValueTask<T>`, static methods, instance methods, and generic methods.

## Additional local test result

Running the whole `NoCodeTests` class locally produced two failures. `SubmitsTraces` is the deterministic runtime-async `InvalidProgramException` described above. `MalformedConfiguration_ComprehensiveTest` reached normal application exit and exported its expected spans, but the collector rejected them due to an instrumentation-scope version mismatch; that appears separate from Runtime Async and should not be included in the minimal reproduction.

Kielek:
## Why

Fixed #5447

.NET 11 preparation + [Runtime Async](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-11/runtime#runtime-async) replaces compiler-generated async state machines with runtime-managed suspension and resumption.

For runtime-async methods, the return type used by the IL body differs from the declared metadata signature:

- `Task` and `ValueTask` use `void`.
- `Task<T>` and `ValueTask<T>` use `T`.

CallTarget previously assumed that the declared return type also described the IL evaluation stack. It also handled async completion by registering a continuation on the returned task. These assumptions do not hold for runtime-async methods and can result in invalid rewritten IL or missing async completion handling.

Applications can contain a mixture of assemblies compiled with and without Runtime Async. Therefore, the profiler must detect and handle the compilation mode per method rather than per process or target framework.

Related to #5111.

## What

- Detect runtime-async methods through the `MethodImplOptions.Async` metadata flag.
- Preserve both the declared return type and the effective return type used by the method body.
- Resolve the logical result type for `Task`, `Task<T>`, `ValueTask`, and `ValueTask<T>`.
- Support complex result signatures, including generic and multidimensional array types.
- Skip instrumentation safely when a runtime-async return signature cannot be resolved.
- Add `EndRuntimeAsyncMethod` CallTarget entry points that invoke the existing `OnAsyncMethodEnd` callback directly, without requiring a returned task or registering a continuation.
- Keep separate native member references for `EndMethod` and `EndRuntimeAsyncMethod`.
- Add a dedicated `TestApplication.NativeAsync` application and a conventionally compiled companion library to verify mixed runtime-async and conventional-async assemblies in the same process.
- Preserve coverage and compatibility for .NET 8 and .NET 9 without introducing a .NET 11 target framework.

## Tests

- `BuildTracer` in Release configuration, including native x64 and x86 builds.
- Native unit tests on x64 and x86: 80 tests passed on each architecture.
- `NativeAsyncTests` integration test on:
  - .NET 8
  - .NET 9
  - .NET 10 with Runtime Async enabled
- The integration test covers:
  - `Task` and `Task<T>`
  - `ValueTask` and `ValueTask<T>`
  - successful and faulted methods
  - static and instance methods
  - generic results
  - multidimensional array results
  - runtime-async and conventionally compiled assemblies in one process

## Checklist


nrcventura (approved):

src/OpenTelemetry.AutoInstrumentation.Native/calltarget_tokens.h
```diff
@@ -91,7 +91,7 @@ class CallTargetTokens
     mdAssemblyRef GetCorLibAssemblyRef();
     mdToken GetCurrentTypeRef(const TypeInfo* currentType, bool& isValueType);
 
-    HRESULT ModifyLocalSigAndInitialize(void* rewriterWrapperPtr, FunctionInfo* functionInfo,
+    HRESULT ModifyLocalSigAndInitialize(void* rewriterWrapperPtr, TypeSignature* methodReturnValue,
```
iskiselev, line 94:
```suggestion
    HRESULT ModifyLocalSigAndInitialize(void* rewriterWrapperPtr, TypeSignature* methodEffectiveReturnType,
```

As we now have both effective and declared return type, it may be helpfull to state what are we working with.
We also accept not return value, but return type.

src/OpenTelemetry.AutoInstrumentation.Native/clr_helpers.h
```diff
@@ -523,34 +527,38 @@ struct FunctionInfo
     const MethodSignature signature;
     const MethodSignature function_spec_signature;
     const mdToken method_def_id;
+    const DWORD method_impl_flags;
     FunctionMethodSignature method_signature;
+    TypeSignature effective_return_type{};
```
iskiselev, line 532: We probabl should init it from ctors. Design when it is inited from `SetEffectiveReturnType` is fragile - we can't access it before anyone call it first, and there is no way for user code to validate it.
It can be inited to some special value that allow us understand that it is not set (offset/length both 0) - but better make sure that once we constructed it, it is always usable. Only if there are valid hot path which doesn't need EffectiveReturnType we may opt into design with some valid placeholder instead.

iskiselev: Alternatively it can be even derved class. In that case type system will help us judge when we need it - and when we don't need it.
But, really effective type calculation is not too hard and we may want to have it resolved always. We may try to extract TypeInfo calculation to not force it resolve effective type.

src/OpenTelemetry.AutoInstrumentation.Native/clr_helpers.cpp
```diff
@@ -649,6 +657,118 @@ ULONG TypeSignature::GetSignature(PCCOR_SIGNATURE& data) const
...
+    if (!type_info.IsValid() || type_info.name.empty())
+    {
+        Logger::Warn("[TryGetRuntimeAsyncResultType] Failed to resolve return type token: ", type_token);
+        return false;
+    }
+
+    const auto& type_name = type_info.name;
+    if (is_generic)
+    {
+        const bool is_task = element_type == ELEMENT_TYPE_CLASS && type_name == WStr("System.Threading.Tasks.Task`1");
```
iskiselev, line 718: We try to validate parent type here. For it we need `metadata_import`.
But currently per spec method that marked with `MethodImplOptions.Async` should have either Task/ValueTask return (or their generic). I don't think we really need to validate that invariant.
If we skip that validation effective return type calculation will be much easier:
- for non-async do not change return type
- else use void for non-generic return type
- else use 1st generic argument

CLR runtime most probably will reject invalid assemblies loading / using bad methods. Our current check is still non-ideal. as we do not validate if it is real Task/ValueTask or just some same-named class from customer assembly. I do not suggest improve it, as it should be valid to have it in customer assembly it it will be type forwarded.

src/OpenTelemetry.AutoInstrumentation.Native/clr_helpers.cpp
```diff
@@ -809,8 +928,39 @@ bool ParseType(PCCOR_SIGNATURE& pbCur, PCCOR_SIGNATURE pbEnd)
             return false;
 
         case ELEMENT_TYPE_ARRAY:
```
iskiselev, line 930: Looks like we do only partial handling of modreq / modpot also. They can be in surprisingly many places, not only top level return type definition: https://github.com/dotnet/runtime/blob/main/docs/design/specs/Ecma-335-Augments.md#signatures
In practive it is better to track as separate issue, as it is hard to get them in connection with async methods - C# will not give a way for it. Much more iteresting and still legal things may be done with IL :)

src/OpenTelemetry.AutoInstrumentation.Native/rejit_preprocessor.cpp
```diff
@@ -130,6 +130,25 @@ void RejitPreprocessor<RejitRequestDefinition>::ProcessTypeDefForRejit(const Rej
...
         }
 
+        // Keep the declared metadata signature intact for matching and diagnostics. Runtime-async
+        // methods use a different effective return type on the IL evaluation stack: Task/ValueTask
+        // become void, while Task<T>/ValueTask<T> become T.
+        functionInfo.SetEffectiveReturnType(functionInfo.GetDeclaredReturnType());
+        if (functionInfo.IsRuntimeAsync())
+        {
+            TypeSignature runtime_async_result_type;
+            if (!TryGetRuntimeAsyncResultType(functionInfo.GetDeclaredReturnType(), metadataImport,
```
eftiquar, line 140: Could we move runtime-async normalization into the function loader and expose both declared and resultant `SignatureView`s on `FunctionInfo`?

The loader should initialize both once: consumers use declared for matching and diagnostics, while IL rewriting always uses resultant. For regular methods, they are identical. This keeps `ParseType` limitations separate and avoids requiring consumers to normalize or compare signatures. We will need to fix ParseType in separate PR.

eftiquar (changes requested):

src/OpenTelemetry.AutoInstrumentation.Native/clr_helpers.cpp
eftiquar, line 930: Runtime-async normalization only needs to identify the outer Task/ValueTask wrapper and preserve the bounded raw T signature. Please move this normalization into signature loading and expose declared/resultant views. Keep ParseType and type-name handling separate for ordinary matching and diagnostics.

eftiquar (changes requested):

src/OpenTelemetry.AutoInstrumentation.Native/clr_helpers.h
iskiselev: On the other hand, it repeats the design of FunctionMethodSignature::TryParse, which inits members only after `TryParse` was called. I still think that the design is fragile.

src/OpenTelemetry.AutoInstrumentation.Native/clr_helpers.cpp
iskiselev: Do you mean that we don't need to parse entire signature for declariation -> effective return type transformation?
As technically signature is defined by only pointer. We need only shift this pointer to first generic arg (or our special place to represent void).
But in practice we still need to know legth of signature - and we can do it only if we parse it. Without it we will not be able to create transformation, as return type used in local varibale signature.
So, if we are not able to parse signature - we can't transfrom a method.

eftiquar: I want this PR to be focused on handling async runtime ret type - we have discovered limitations in ParseType that predate this PR. 
Runtime-async return signatures have different requirements from regular method returns. Regular returns need granular `ParseType` parsing; broader parser support should be handled separately. For runtime-async methods, coarse parsing is sufficient: identify the `Task`/`ValueTask` wrapper using bounded `CorSigUncompress*` APIs and preserve the raw `T` signature.

`ParseType` is still needed to unpack method arguments for signature matching and IL argument handling, so that existing behavior should remain unchanged. Please bypass it only for the runtime-async return transformation, keeping that path independent of its pre-existing limitations. Those parser limitations can be addressed in a separate PR as they are orthogonal to async runtime return time transformation.

Kielek:
@eftiquar, @iskiselev, I will check your comment when https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/5438 lands on the main first. As I remember there is some overlap with the touched code.

martincostello:
As it doesn't appear to be explicitly called-out anywhere in the PR description or documentation in the diff, Runtime Async is still considered preview for .NET 11 ([docs](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-11/runtime#runtime-async)) (which is why I think it's opt-in):

> Runtime Async is a preview feature.

I think it's worth explicitly calling that out somewhere, so that even if it's tested as seeming to work, there might be things not under OTel's control that could issues due to the preview status of the feature.

Kielek:
> As it doesn't appear to be explicitly called-out anywhere in the PR description or documentation in the diff, Runtime Async is still considered preview for .NET 11 ([docs](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-11/runtime#runtime-async)) (which is why I think it's opt-in):
> 
> > Runtime Async is a preview feature.
> 
> I think it's worth explicitly calling that out somewhere, so that even if it's tested as seeming to work, there might be things not under OTel's control that could issues due to the preview status of the feature.

Last time, when I have check there were a plan to include it by default in .NET11. It is the only reason I decided to touch it. As I remember even dotnet libraries should be compiled with this feature,
If it will be not a case, we can defer this PR about ~1 year.

martincostello:
If it wasn't a preview feature, I would have expected it to have dropped any reference to it being a preview in rc.1's release notes as C# 15 stopped being preview then too. Maybe the docs are just out of date?

https://github.com/dotnet/runtime/issues/109632 looks out-of-date - it hasn't been edited since March, and it states preview for .NET 10 (which didn't happen AFAIK) and stable for .NET 11.

https://github.com/dotnet/runtime/issues/133672 was opened recently that's deferred to .NET 12 regarding some performance issues caused by runtime-async.

The only change I can find in the release notes is that needing to also enable `EnablePreviewFeatures` was removed in preview 3: [Runtime async removes the preview-API opt-in requirement](https://github.com/dotnet/core/blob/6624c24dd992f719f5de97429d4150a7e1a5cd59/release-notes/11.0/preview/preview3/runtime.md#runtime-async-removes-the-preview-api-opt-in-requirement)

> If it will be not a case, we can defer this PR about ~1 year.

IMHO if it stays in preview it's still worth making changes to the code for it to work. But if that's the case, then it should just be documented as preview because it depends on a preview.

I can't seem to find a concrete statement either way beyond that one sentence in the release notes I linked to.

martincostello:
> I can't seem to find a concrete statement either way beyond that one sentence in the release notes I linked to.

[Performance Improvements in .NET 11](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-11/#runtime-async) seems to suggest that it maybe isn't a preview, and .NET 12 is about making it opt-out not opt-in, but it's still not worded concretely either way compared to the release notes.

> Note that there’s no new C# syntax involved, so `LangVersion=preview` isn’t required, nor is `EnablePreviewFeatures`. While this is opt-in at the application layer, most of the in-box shared framework is already built this way for .NET 11. The `async`/`await` performance goal for .NET 11 is parity with .NET 10, and in general runtime async is already as good as or better than the older implementation in many important paths. It isn’t yet fully optimized, though, and there are known cases where it still produces less efficient code. I’d encourage you to experiment in .NET 11 with opting-in your applications and services; just make sure to measure. My hope is that it’ll be on by default starting in .NET 12.

andrewlock:
> > I can't seem to find a concrete statement either way beyond that one sentence in the release notes I linked to.
> 
> [Performance Improvements in .NET 11](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-11/#runtime-async) seems to suggest that it maybe isn't a preview, and .NET 12 is about making it opt-out not opt-in, but it's still not worded concretely either way compared to the release notes.

FWIW, and in case it's not clear, in RC1 there categorically _are_ core libraries being shipped with runtime-async enabled in ASP.NET Core. Which I think means _regardless_ of whether _customers_ set runtime-async enabled, it needs to be supported, otherwise you'll get instrumentation errors if you instrument a runtime-async method. 

Whether that will persist into the .NET 11 GA release I guess we'll find out, but the fact that the application can be _partially_ runtime-async means it needs support sooner rather than later I think

